PR #168, opened 2026-02-22
fix(agent): unregister service endpoints on metal process delete

### LINKED ISSUE #167: [BUG] Endpoints aren't cleared after service deletion
## Bug Description

I delete a service using CLI:

```
go run cmd/cli/main.go delete llama-3.2-3b 
```

Endpoints still show up:

```
% kubectl get endpointslices 
NAME                 ADDRESSTYPE   PORTS     ENDPOINTS        AGE
kubernetes           IPv4          8443      192.168.49.2     26m
llama-3-2-3b-6wv54   IPv4          <unset>   <unset>          21m
llama-3-2-3b-vpg6b   IPv4          8080      192.168.65.254   21m
```

## Steps to Reproduce

1. Fresh minikube setup.
2. Install controller: `make install`
3. Run controller: `make run ARGS="--model-cache-path=/tmp/llmkube-models"`
4. Run metal agent: `go run cmd/metal-agent/main.go -llama-server /opt/homebrew/bin/llama-server -log-level debug`
5. Create a service: `go run cmd/cli/main.go deploy llama-3.2-3b --accelerator metal`
6. Delete the service: `go run cmd/cli/main.go delete llama-3.2-3b`

## Expected Behavior

After the last command (service deletion), endpoints still show up in K8S:

```
% kubectl get endpointslices 
NAME                 ADDRESSTYPE   PORTS     ENDPOINTS        AGE
kubernetes           IPv4          8443      192.168.49.2     15m
llama-3-2-3b-6wv54   IPv4          <unset>   <unset>          9m20s
llama-3-2-3b-vpg6b   IPv4          8080      192.168.65.254   9m17s
```

## Actual Behavior

Endpoints should be removed.

## Environment

- macOS Tahoe 26.2
- Macbook Pro (M4 Pro, Nov 2024)
- minikube v1.38.1 / commit c93a4cb9311efc66b90d33ea03f75f2c4120e9b0
- Tested using a fresh/clean setup a few times.
- Latest `main` / 07630b8f9c48bac9233873c7a37684b26d0e182f

**Cluster Type:**
- [ ] GKE
- [ ] EKS
- [ ] AKS
- [x] minikube
- [ ] kind
- [ ] K3s
- [ ] Other:

**GPU (if applicable):**
- [ ] NVIDIA T4
- [ ] NVIDIA L4
- [ ] NVIDIA V100
- [ ] NVIDIA A100
- [ ] None (CPU only)
- [x] Other: Metal

## Logs
-

## YAML Manifests
-

matiasinsaurralde:
## What

- Add best-effort endpoint cleanup in the Metal agent delete flow by calling `UnregisterEndpoint` both before and after process stop.
- Make `UnregisterEndpoint` idempotent by ignoring Kubernetes `NotFound` errors for Service and Endpoints deletion.
- Improve reliability of Metal service teardown to reduce stale networking artifacts (including lingering EndpointSlices).

## Why

- Deleting a Metal-backed inference service previously stopped the process but could leave Kubernetes networking objects behind.
- Stale endpoint state causes confusing service discovery behavior and makes users think deletion did not fully work.
- Idempotent cleanup is required for reconcile/retry safety, where delete paths may run multiple times.

Fixes #167 

## How

- Updated `deleteProcess` in `pkg/agent/agent.go` to run endpoint cleanup as a best-effort sequence:
  1. Parse `namespace/name` from the internal process key.
  2. Call `registry.UnregisterEndpoint(ctx, namespace, name)` before stopping the process.
  3. Stop the local llama.cpp process.
  4. Call `UnregisterEndpoint` again after stop to handle timing/race windows.
  5. Aggregate errors from all steps so we surface failures without skipping later cleanup attempts.

- Updated `UnregisterEndpoint` in `pkg/agent/registry.go` to be idempotent:
  - Service delete now ignores `apierrors.IsNotFound`.
  - Endpoints delete now ignores `apierrors.IsNotFound`.
  - Other API errors are still returned (RBAC, connectivity, etc.).

- Design choices:
  - **Pre-stop + post-stop cleanup** was chosen over a single call to improve robustness against transient ordering/race issues in endpoint reconciliation.
  - **Idempotent unregister** ensures repeated delete/reconcile paths are safe and do not fail on already-removed resources.
## Checklist

- [ ] Tests added/updated - **Note:** Don't know if we have tests for this in Metal and/or how complex it would be to cover this.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -222,15 +222,20 @@ func (a *MetalAgent) deleteProcess(ctx context.Context, key string) error {
 	a.logger.Infow("stopping inference service", "key", key)
 	namespace, name := parseKey(key)
 
+	var deleteErrors []error
 	if err := a.executor.StopProcess(process.PID); err != nil {
-		return fmt.Errorf("failed to stop process: %w", err)
+		deleteErrors = append(deleteErrors, fmt.Errorf("failed to stop process: %w", err))
 	}
 
 	// Unregister after the process has stopped. UnregisterEndpoint is idempotent
 	// (tolerates 404), so this is safe even if a prior cleanup attempt already
 	// removed the resources.
 	if err := a.registry.UnregisterEndpoint(ctx, namespace, name); err != nil {
-		return fmt.Errorf("failed to unregister endpoint for %s: %w", key, err)
+		deleteErrors = append(deleteErrors, fmt.Errorf("failed to unregister endpoint for %s: %w", key, err))
+	}
+
+	if len(deleteErrors) > 0 {
+		return fmt.Errorf("delete process cleanup errors: %w", errors.Join(deleteErrors...))
 	}
 
 	a.logger.Infow("stopped inference service", "key", key)
--- a/pkg/agent/agent_test.go
+++ b/pkg/agent/agent_test.go
@@ -21,8 +21,10 @@ import (
 	"testing"
 
 	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	inferencev1alpha1 "github.com/defilantech/llmkube/api/v1alpha1"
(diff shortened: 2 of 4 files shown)


pkg/agent/agent.go
```diff
@@ -195,9 +195,26 @@ func (a *MetalAgent) deleteProcess(_ context.Context, key string) error {
 	a.mu.Unlock()
 
 	fmt.Printf("🛑 Stopping inference service: %s\n", key)
+	namespace, name := parseKey(key)
+
+	var cleanupErrors []error
+
```
Defilan: Hey Matias, nice fix! Quick question on the double-unregister approach: could you walk me through the scenario where the pre-stop call is needed? Since UnregisterEndpoint is now idempotent, I'd think a single post-stop call would handle it cleanly. If there's a specific race you observed with the EndpointSlice reconciler during testing, it would be great to capture that in a comment here so future readers understand why we call it twice.

Defilan (changes requested):
Great fix, Matias -- this cleanly addresses the stale endpoint bug from #167 and the idempotent `UnregisterEndpoint` pattern is exactly right for controller/reconcile safety.

One required change: `deleteProcess` currently early-returns when `StopProcess` fails, which skips `UnregisterEndpoint`. Since the process is already removed from the map at that point, a retry will never clean up the endpoints. Please continue to attempt endpoint cleanup even on `StopProcess` failure, aggregating both errors.

Also, as mentioned in the first review, a small test for the idempotent UnregisterEndpoint (calling it on already-deleted resources) would be a welcome addition to lock in the contract.

Nice work on the `errors.Join` / `shutdownErrors` rename -- clean improvements.

Defilan (approved):
This is a clean, well-tested fix for #167. The idempotent `UnregisterEndpoint` pattern is exactly right for controller/reconcile safety, and the error aggregation in `deleteProcess` ensures endpoint cleanup always runs even when `StopProcess` fails. Both new tests lock in the behavioral guarantees nicely. All prior review feedback has been addressed. Really appreciate the thorough work here, Matias — great contribution!

One optional follow-up for a future PR: the `Shutdown` method could also call `UnregisterEndpoint` for each process during graceful shutdown to avoid stale endpoints on agent restart.

PR #274, opened 2026-04-11
fix: inherit runAsUser/runAsGroup from podSecurityContext 

castorinop:
…tainer

The model downloader init container now properly inherits runAsUser and runAsGroup from the InferenceService's podSecurityContext. This fixes the 'Permission denied' error when the main container (running as a non-root user) tries to read model files downloaded by the init container.

Changes:
- initContainerSecurityContext() now accepts InferenceService parameter
- buildCachedStorageConfig() now accepts isvc parameter
- buildEmptyDirStorageConfig() now accepts isvc parameter
- Inherits runAsUser/runAsGroup from podSecurityContext if specified
- Sets RunAsNonRoot=true to require explicit user configuration
- No arbitrary defaults - users must specify their UID

Users must now configure podSecurityContext in their InferenceService:
  podSecurityContext:
    runAsUser: 1000
    runAsGroup: 1000
    fsGroup: 0

Added unit tests for:
- Inheritance of runAsUser/runAsGroup from podSecurityContext
- Verification that no defaults are applied without user config

## What

Fixes the "Permission denied" error when the main container tries to read model files downloaded by the init container.

The `initContainerSecurityContext()` function now properly inherits `runAsUser` and `runAsGroup` from the InferenceService's `podSecurityContext`.

## Why

The model downloader init container was running as root (default) while the main container ran as a non-root user (e.g., UID 1000). This caused permission denied errors when the main container tried to read model files downloaded by the init container, because the files were owned by root.

## How

- Modified `initContainerSecurityContext()` to accept an `*InferenceService` parameter
- Added logic to inherit `runAsUser`/`runAsGroup` from `podSecurityContext` if specified
- Set `RunAsNonRoot: true` to require explicit user configuration
- Removed arbitrary default UID (previously hardcoded to 1000)
- Added unit tests to verify inheritance behavior

## Usage

Users must now configure `podSecurityContext` in their InferenceService:

```yaml
spec:
  podSecurityContext:
    runAsUser: 1000
    runAsGroup: 1000
    fsGroup: 0
```

## Checklist

- [x] Tests added/updated
- [ ] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)
- [ ] All commits are signed off (`git commit -s`) per [DCO](https://developercertificate.org/)

### DIFF
--- a/internal/controller/inferenceservice_controller.go
+++ b/internal/controller/inferenceservice_controller.go
@@ -84,14 +84,30 @@ func inferContainerSecurityContext(isvc *inferencev1alpha1.InferenceService) *co
 	}
 }
 
-func initContainerSecurityContext() *corev1.SecurityContext {
-	return &corev1.SecurityContext{
+// initContainerSecurityContext sets up security context for the model downloader init container.
+// It inherits runAsUser/runAsGroup from podSecurityContext if specified by the user.
+// Users MUST specify runAsUser/runAsGroup in their InferenceService podSecurityContext to avoid
+// permission denied errors on the model volume.
+func initContainerSecurityContext(isvc *inferencev1alpha1.InferenceService) *corev1.SecurityContext {
+	sc := &corev1.SecurityContext{
 		AllowPrivilegeEscalation: boolPtr(false),
 		ReadOnlyRootFilesystem:   boolPtr(false),
 		Capabilities: &corev1.Capabilities{
 			Drop: []corev1.Capability{"ALL"},
 		},
 	}
+
+	// Inherit runAsUser/runAsGroup from podSecurityContext if specified
+	if isvc != nil && isvc.Spec.PodSecurityContext != nil {
+		if isvc.Spec.PodSecurityContext.RunAsUser != nil {
+			sc.RunAsUser = isvc.Spec.PodSecurityContext.RunAsUser
+		}
+		if isvc.Spec.PodSecurityContext.RunAsGroup != nil {
+			sc.RunAsGroup = isvc.Spec.PodSecurityContext.RunAsGroup
+		}
+	}
+
+	return sc
 }
 
 // isLocalModelSource delegates to the shared isLocalSource helper in source.go.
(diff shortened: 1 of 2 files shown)


Defilan (changes requested):
Hey @castorinop, welcome to LLMKube and thank you for this contribution! This is exactly the kind of real-world bug fix we love to see. The init container file ownership mismatch is a genuine pain point, and your approach of inheriting runAsUser/runAsGroup from podSecurityContext is the right pattern. Clean code, good tests, solid PR description. Really appreciate the effort here.

We want to get this merged. There are a few things to address first:

**CI failures (blocking):**

1. **DCO sign-off is missing.** All commits need a `Signed-off-by` line. You can fix this with `git commit --amend -s` and force-push.
2. **Lint (goconst):** The linter flagged `"/models"` as a repeated string. There's an existing constant `DefaultModelCachePath` you can use instead. Check line 2187 in the test file.

**Design concern (blocking):**

3. **`RunAsNonRoot: true` is a breaking change.** Adding this to the init container will cause pod failures for every existing InferenceService that doesn't set `podSecurityContext.runAsUser`. The UID/GID inheritance is the core fix and it's non-breaking on its own. I'd recommend pulling `RunAsNonRoot` out of this PR and we can tackle that as a separate, documented change. That way your fix ships cleanly without breaking anyone's existing deployments.

**Nice to have:**

4. Adding a test for the emptyDir (non-cached) storage path would strengthen coverage since both `buildCachedStorageConfig` and `buildEmptyDirStorageConfig` call `initContainerSecurityContext`.

Looking forward to the next revision!

PR.md
```diff
@@ -0,0 +1,29 @@
+## What
```
Defilan: This file was accidentally committed — please remove it from the PR. You can do:
```bash
git rm PR.md
git commit -s -m "fix: remove accidentally committed PR.md"
```

Defilan (changes requested):
Great revision, @castorinop! You addressed every item from the previous review — DCO sign-off is present, the goconst lint issue is fixed, RunAsNonRoot was correctly removed to avoid breaking existing deployments, and the emptyDir test coverage was added. The core code change is clean and ready.

One small thing to fix before merge: it looks like a `PR.md` file was accidentally committed to the repo root. This file should be removed from the commit — PR descriptions belong on GitHub, not in the tree. Once that file is dropped, this is good to go.

castorinop:
@Defilan 
This fix was developed with the assistance of an AI coding agent. for clarification.

 hope this helps!

Defilan:
Thanks for the transparency, @castorinop! Using AI coding agents is totally fine with us. What matters is that the contribution is correct, well-tested, and solves a real problem, and this PR checks all those boxes. It helps us all iterate faster.

Defilan (approved):
PR.md removed, code is clean. Thanks for the thorough revision, @castorinop -- welcome to LLMKube!

PR #340, opened 2026-04-27
feat(crd): add ParallelSlots support for vllm and fix llamacpp

### LINKED ISSUE #339: fix(controller): parallelSlots=1 silently ignored; document field; extend to vLLM and Ollama
## Background

Originally surfaced from [Discussion #338](https://github.com/defilantech/LLMKube/discussions/338) by @Faylixe (Raspberry Pi 5 k3s cluster running Gemma 4 models, wanting strict per-replica request serialization). When proposing this issue I assumed the feature was missing entirely. @Faylixe then went and read the code and pointed out that `spec.parallelSlots` already exists for the llama.cpp runtime. Big thanks to them for catching that.

This issue is now scoped to (a) the bug they unintentionally helped me find while verifying their report, (b) docs work, and (c) extending the same capability to runtimes that don't have it yet.

## Bug: `parallelSlots: 1` silently does nothing

`internal/controller/runtime_llamacpp_args.go:50-53`:

```go
func appendParallelSlotsArgs(args []string, parallelSlots *int32) []string {
    if parallelSlots != nil && *parallelSlots > 1 {
        return append(args, ""--parallel"", fmt.Sprintf(""%d"", *parallelSlots))
    }
    return args
}
```

The `> 1` guard means `parallelSlots: 1` does NOT inject `--parallel 1`, so the runtime falls back to llama-server's default (4 parallel slots). For users intentionally setting `parallelSlots: 1` to serialize requests on resource-constrained hardware, the field appears to do nothing. They have to fall back to `extraArgs: [""--parallel"", ""1""]`.

The guard should probably be `>= 1` so any explicit non-nil value gets respected. The original intent (don't inject when nil/unset → use llama-server default) is still preserved.

## Documentation gap

`spec.parallelSlots` is documented in the CRD type as a Go comment, but it's not surfaced anywhere a user would look:

- Not in the README
- Not in any docs page about resource-constrained deployments
- Not in any sample InferenceService manifest

@Faylixe ran into the underlying use case and resorted to filing a discussion because the docs didn't lead them to the existing field. That's the bigger win to fix.

## Cross-runtime parity

`parallelSlots` is currently a llama.cpp-only field. Other runtimes have equivalent flags but LLMKube doesn't translate them:

| Runtime | Equivalent flag/env |
|---|---|
| llamacpp | `--parallel <N>` (already implemented, modulo the `> 1` bug) |
| vllm | `--max-num-seqs <N>` |
| ollama | `OLLAMA_NUM_PARALLEL=<N>` env var |
| omlx, generic | (no equivalent; log warning that field is not honored) |

Conflict resolution: when `extraArgs` already contains a `--parallel` (or runtime equivalent), the auto-injection should detect the conflict and skip with a warning rather than double-inject.

## Acceptance criteria

- [ ] Fix the `> 1` guard in `appendParallelSlotsArgs` so `parallelSlots: 1` correctly injects `--parallel 1`
- [ ] Add a unit test covering `parallelSlots: 1` injecting the flag
- [ ] Add documentation for `spec.parallelSlots`:
  - [ ] CRD reference entry (kubebuilder field doc and an example)
  - [ ] Short ""resource-constrained deployments"" section in the docs (homelab, Pi clusters, edge)
- [ ] Translate `parallelSlots` to vLLM `--max-num-seqs` in the vLLM runtime
- [ ] Translate `parallelSlots` to `OLLAMA_NUM_PARALLEL` env var in the Ollama runtime
- [ ] Log a warning for omlx and generic runtimes if `parallelSlots` is set
- [ ] Skip auto-injection (with a warning) when `extraArgs` already contains a conflicting flag

## Future work this unlocks

- **Queue-depth-aware autoscaling**: with `parallelSlots` standardized across runtimes, the operator knows per-replica capacity. HPA can scale on `queue_length / capacity`.
- **Predictable backpressure**: future per-Service queue depth metric for client-side circuit-breaking.

## References

- Discussion: https://github.com/defilantech/LLMKube/discussions/338
- llama.cpp `--parallel` docs: [server README](https://github.com/ggml-org/llama.cpp/blob/master/tools/server/README.md)

Faylixe:
## What

Adding support for `ParallelSlots` parameter in `vllm` runtime.

## Why

As discussed in #339, `spec.parallelSlots` is actually used to control `llamacpp` concurrency. And is not used in other runtime. In addition to that, since we can already submit `--parallel` through `spec.extraArgs` the actual implementation allows to submit twice the arg to the entry point.

## How

This PR is a first draft to address those issues:

- Reusing existing `ParallelSlots` parameter to infer `vllm` CLI args `--max-num-seqs`
- Adding extra check to the existing llama runtime when injecting `--parallel args` to enforce precedence rule.
- Refactor vllm runtime so it is aligned with llama runtime args builder pattern

I added `runtime_vllm_args.go` file that would follow args builder pattern already existing in `llama` runtime. I refactor the entier args building process and  perform a O(n) lookup into `ExtraArgs` in order to ensure parameters don't overlap.

Fixes https://github.com/defilantech/LLMKube/issues/339: parallelSlots: 1 now emits --parallel 1 (previously fell through to llama.cpp's auto-default of 4).

## Checklist

- [x] Tests added/updated


Faylixe:
I still need to add tests and ensure compliance with your repository guidelines so I leave the PR as draft for now, I will add more tomorrow. Meanwhile let me know if you have any recommendations on this.

Faylixe:
Quick update @Defilan, I reworked at bit on it and here it goes:

- Added a log for runtime that does not support parallelSlots (generic, personaflex, and tgi).
- Extracted build args behavior for vLLM runtime to a dedicated `runtime_vllm_args.go` file that use the same pattern than `runtime_llamacpp_args.go` with testable helper function so it normalizes it :)
- Added support for parallelSlots in vLLM runtime.
- Fix guard on parallelSlots support in llama cpp runtime (`> 1` to `>= 1`).

I will now work on test + documentation edit. In the meantime can you tell me if the submitted change suits you ? (I have minor Golang development experience and probably not up to date on the language idoms and recent good practices).

Thanks :)

Defilan:
Thanks @Faylixe, the direction here is exactly right. Mirroring the args-builder pattern from runtime_llamacpp_args.go is the move, and collision detection is the right shape for the problem.

Two things while you're polishing:

1. Run `make build` locally before the next push. A few compile errors snuck in (a couple of `*bool`/`*boolean` type typos, a missing import block in `runtime_vllm_args.go`, and the log name in `runtime_personaplex.go` reads `"runtime.tgi"`). Once those are sorted CI will give you much cleaner feedback when you mark the PR ready.

2. Want to split the unsupported-runtime warnings (generic/personaplex/tgi) into a follow-up PR? Keeps this one focused on the vLLM extension and the llamacpp guard, which makes review easier on both sides. Happy either way, just offering the option.

Excited to land this. Ping me when it's out of draft.

Faylixe:
> Thanks @Faylixe, the direction here is exactly right. Mirroring the args-builder pattern from runtime_llamacpp_args.go is the move, and collision detection is the right shape for the problem.
> 
> Two things while you're polishing:
> 
> 1. Run `make build` locally before the next push. A few compile errors snuck in (a couple of `*bool`/`*boolean` type typos, a missing import block in `runtime_vllm_args.go`, and the log name in `runtime_personaplex.go` reads `"runtime.tgi"`). Once those are sorted CI will give you much cleaner feedback when you mark the PR ready.

Yes absolutely i was planning to install local setup so I can build / lint / test before to switch the PR to ready state :).

> 2. Want to split the unsupported-runtime warnings (generic/personaplex/tgi) into a follow-up PR? Keeps this one focused on the vLLM extension and the llamacpp guard, which makes review easier on both sides. Happy either way, just offering the option.

Sure no problem ! 

> Excited to land this. Ping me when it's out of draft.

Yes I will mark the PR ready as soon as I finished polishing 🎉

Faylixe:
I just mark the PR ready for review, I added following changes:

- Merge recent changes and adapt vllm `KVCustomDtype` handling to the new builder pattern.
- Test suite for llama cpp runtime build args following vllm test model (passing).
- Adapted test suite for vllm runtime so it can works with `InferenceServiceSpec` parameters as well.
- Added test case for vllm runtime `parallelSlots`.
- Fix build and lint errors.

I have `build`, `lint` and `test` target that pass on my machine, I will check the CI feedback to ensure it works as well. In the meantime do not hesitate to tell me if you need more work on that or any other changes you want to see, I will work on the documentation part this afternoon :)

internal/controller/inferenceservice_deployment_test.go
```diff
@@ -966,9 +966,9 @@ var _ = Describe("Context Size Configuration", func() {
 			Expect(args).NotTo(ContainElement("--parallel"))
 		})
 
-		It("should NOT include --parallel flag when parallelSlots is 1", func() {
+		It("should NOT include --parallel flag when parallelSlots is 0", func() {
```
Faylixe: This test was failing but I assume it was due to the fact that previous guarding has the wrong check so I fixed it.

internal/controller/runtime_vllm_test.go
```diff
@@ -71,107 +71,194 @@ func TestVLLMBuildArgs(t *testing.T) {
 
 	cases := []struct {
 		name        string
-		cfg         *inferencev1alpha1.VLLMConfig
+		spec        *inferencev1alpha1.InferenceServiceSpec
```
Faylixe, line 74: I switched from `VLLMConfig` to `InferenceServiceSpec` directly as I need that to work on spec level parameter testing. It is a bit redondant since I have to specify model and runtime for each test case.

pkg/agent/agent_test.go
```diff
@@ -32,6 +32,8 @@ import (
 	inferencev1alpha1 "github.com/defilantech/llmkube/api/v1alpha1"
 )
 
+const Q8_0 = "q8_0"
```
Faylixe, line 35: I got warning from linter saying that I should switch to a constant since this value appear more than 3 times :)

Faylixe:
Just added a documentation file with concurrency exemple (manifest + cli) as well as listing runtime support

Faylixe:
@Defilan sorry I went rogue as I amended all my previous commit to have them signed off 😓.

Defilan:
> @Defilan sorry I went rogue as I amended all my previous commit to have them signed off 😓.

haha, no worries. I got the CI enabled for you. Thanks for all your work!! We're getting close :)

Faylixe:
> > @Defilan sorry I went rogue as I amended all my previous commit to have them signed off 😓.
> 
> haha, no worries. I got the CI enabled for you. Thanks for all your work!! We're getting close :)

Thanks ! I got only the DCO hooks that have failed (I actually have one merge commit that has been done through the Github web page and I don't feel like I can sign it, also one GA bot commit failed)

internal/controller/runtime_vllm.go
```diff
@@ -71,90 +70,48 @@ func (b *VLLMBackend) BuildArgs(isvc *inferencev1alpha1.InferenceService, model
...
-// resolveKVCacheDtype returns the custom vLLM KV cache type when set,
-// otherwise the enum-validated standard value (dereferenced; nil → ""). Lets
-// users opt into vLLM image-specific cache formats (TurboQuant turbo2 from
-// v0.20+, future variants) without expanding the CRD enum on every release,
-// while keeping the standard field discoverable for the common case. Mirrors
-// resolveCacheType on the llama.cpp side.
-func resolveKVCacheDtype(custom string, standard *string) string {
-	if custom != "" {
-		return custom
+	args, err = appendMaxNumSeqsArgs(args, isvc.Spec.ParallelSlots, isvc.Spec.ExtraArgs)
```
Defilan, line 102: Could you move this `appendMaxNumSeqsArgs` call above the `ExtraArgs` append (just before line 102)? The `BuildArgs` doc comment promises ExtraArgs is always last so user flags win, and right now this call is emitting `--max-num-seqs` *after* the user's args. The precedence check covers the conflict case for that specific flag, but for any other user-supplied flags the operator-emitted ones would still come last, which is the opposite of the documented contract. Same idea you already got right on the llamacpp side.

internal/controller/runtime_llamacpp_args.go
```diff
@@ -47,11 +48,17 @@ func appendContextSizeArgs(args []string, contextSize *int32) []string {
 	return args
 }
 
-func appendParallelSlotsArgs(args []string, parallelSlots *int32) []string {
-	if parallelSlots != nil && *parallelSlots > 1 {
-		return append(args, "--parallel", fmt.Sprintf("%d", *parallelSlots))
+func appendParallelSlotsArgs(args []string, parallelSlots *int32, extraArgs []string) ([]string, error) {
+	// NOTE(#339): extra args has precedence.
+	if parallelSlots != nil && *parallelSlots >= 1 {
```
Defilan, line 53: Quick question: was the threshold change from `> 1` to `>= 1` intentional? It's a silent llamacpp behavior change — `parallelSlots=1` now causes `--parallel 1` to be emitted where before it relied on llama.cpp's default of 1. For vLLM the change is meaningful (`--max-num-seqs` default is 256), so `>= 1` makes sense there. For llamacpp it changes the rendered Pod spec without changing observable behavior against today's llama.cpp.

Two clean options:
- Revert this side to `> 1` and keep `>= 1` only for `appendMaxNumSeqsArgs`, or
- Keep `>= 1` for both and call out the llamacpp change in the PR description so it surfaces in the changelog.

Whichever you pick, the renamed Ginkgo test in `inferenceservice_controller_test.go:969` (currently `is 0`, which the CRD's `Minimum=1` rejects, so the case is unreachable) should be replaced with one that exercises the chosen behavior on `parallelSlots=1`.

internal/controller/runtime_llamacpp_args.go
```diff
@@ -47,11 +48,17 @@ func appendContextSizeArgs(args []string, contextSize *int32) []string {
...
+		for _, v := range extraArgs {
+			if v == "--parallel" {
```
Defilan: Tiny edge case: this check matches `--parallel` only as a discrete arg, but llama.cpp also accepts `--parallel=4` as a single token. A user passing the `=`-form would slip past the check and end up with both `--parallel 4` (operator-emitted) and `--parallel=4` (user-supplied) in args.

One-line fix:

```go
if v == "--parallel" || strings.HasPrefix(v, "--parallel=") {
```

(with `"strings"` added to imports). Worth pulling the same logic into a small shared helper since the vllm side has the identical check on `--max-num-seqs`.

internal/controller/runtime_vllm_args.go
```diff
@@ -0,0 +1,152 @@
...
+		return append(args, "--max-num-batched-tokens", fmt.Sprintf("%d", *maxNumBatchedTokens))
+	}
+	return args
+}
+
+func appendMaxNumSeqsArgs(args []string, parallelSlots *int32, extraArgs []string) ([]string, error) {
+	// NOTE(#339): extra args has precedence.
+	if parallelSlots != nil && *parallelSlots >= 1 {
+		for _, v := range extraArgs {
+			if v == "--max-num-seqs" {
```
Defilan: Same `=`-form blind spot as the llamacpp side — `--max-num-seqs=4` won't be caught by the equality check. Recommend `v == "--max-num-seqs" || strings.HasPrefix(v, "--max-num-seqs=")`. If you go with a shared helper for the llamacpp version, this one slots into it cleanly.

internal/controller/inferenceservice_deployment_test.go
Defilan: The renamed test (`is 1` → `is 0`) is testing a `parallelSlots` value that the CRD's `+kubebuilder:validation:Minimum=1` rejects, so this case is unreachable through normal Kube traffic — the only way to hit it is direct Go struct manipulation. The original `is 1` test was guarding the `> 1` threshold; deleting it silently weakens regression coverage for that decision.

Could you replace this with a test that asserts the *new* behavior on `parallelSlots=1`? Something like "should include --parallel 1 when parallelSlots is 1" if you keep `>= 1`, or "should NOT include --parallel flag when parallelSlots is 1" if you revert to `> 1`. The 1-value case is the load-bearing one and deserves a deliberate test.

docs/resource-constrained-deployments.md
```diff
@@ -0,0 +1,54 @@
...
+
+| Runtime     | Behavior                          |
+| ----------- | --------------------------------- |
+| generic     | Not supported                     |
+| llama-cpp   | Set `--parallel` CLI arg          |
+| ollama      | Set `OLLAMA_NUM_PARALLEL` env var |
+| omlx        | Not supported                     |
+| personaflex | Not supported                     |
+| tgi         | Not supported                     |
+| vllm        | Set `--num-max-seqs` CLI arg      |
```
Defilan: Two issues with this runtime table:

1. `ollama` and `omlx` aren't valid `InferenceService.spec.runtime` values. The CRD enum is `llamacpp;personaplex;vllm;tgi;generic` (see `api/v1alpha1/inferenceservice_types.go:39`). `ollama` and `omlx` exist only as `--runtime` flags on the metal-agent binary — they're Apple Silicon agent runtimes, not cluster-side runtimes.
2. `personaflex` is a typo for `personaplex`, and `llama-cpp` should be `llamacpp` to match the enum value.

As written, this table will mislead users into trying `runtime: ollama` in their `InferenceService` spec, which the validating webhook will reject. Two clean fixes:

- Drop the `ollama` and `omlx` rows entirely and fix the typos, or
- Split into two tables: "Cluster runtimes" (the enum values) and "Apple Silicon metal-agent runtimes" (`ollama`, `omlx`).

internal/controller/runtime_llamacpp_test.go
```diff
@@ -0,0 +1,397 @@
...
+				ModelRef:        "test-model",
+				TensorOverrides: []string{"value1", "value2"},
+			},
+			contains: []flagCheck{
+				{"--override-tensor", "value1"},
+				{"--override-tensor", "value2"},
+			},
+		},
+		{
+			name: "cacheTypeV nil does not emits flag",
```
Defilan: Tiny one: this test case has the same `name` as the one at line 286 (`"cacheTypeV nil does not emits flag"`), but it's actually testing `--override-tensor`. `go test` will warn on duplicate subtest names, and the label is wrong. Suggest:

```go
name: "tensorOverrides nil does not emit flag",
```

(Also fixes the `emits` → `emit` grammar — same fix worth applying to the cacheTypeK / cacheTypeV / cacheTypeV cases on lines 269, 277, 286.)

internal/controller/runtime_llamacpp_test.go
```diff
@@ -0,0 +1,397 @@
...
+			name: "parallelSlots set emits flag (without precedence)",
+			spec: &inferencev1alpha1.InferenceServiceSpec{
+				Runtime:       "llama",
+				ModelRef:      "test-model",
+				ParallelSlots: ptrInt32(8192),
+			},
+			contains: []flagCheck{{"--parallel", "8192"}},
+		},
+		{
+			name: "parallelSlots nil does not emit flag (without precedence)",
```
Defilan: Could you add a couple of cases here covering the new precedence path? Specifically:

- `parallelSlots: 4` + `extraArgs: ["--parallel", "8"]` → asserts `--parallel 8` is in args, no duplicate `--parallel 4`.
- `parallelSlots: 4` + `extraArgs: ["--parallel=8"]` → same, once the `=`-form check lands.

Same coverage gap exists on `runtime_vllm_test.go` (substituting `--max-num-seqs`). The whole error-return plumbing on `appendParallelSlotsArgs` and `appendMaxNumSeqsArgs` is currently uncovered, which is also why CI is happily green on the order regression in `runtime_vllm.go:106` — there's no test that exercises operator-emitted flags coming after user flags.

Defilan (commented):
Hey @Faylixe — thanks so much for picking this up! The args-builder alignment between llamacpp and vllm is a great direction (matches where we headed with #359 for cache-type fields), and the precedence-conflict check fixes a real footgun for users who already had `--parallel` in their `extraArgs`. Plus the iteration in your commits — adding the warning log and then taking it back out — tells me you thought through the right design questions along the way. Lots of good engagement on this one.

A few small things to land before merge — none of them are big, mostly polish around an edge case and a couple of test coverage gaps. Inline comments below on the line-specific bits:

1. **Order regression in `runtime_vllm.go`** — `appendMaxNumSeqsArgs` is being appended after `ExtraArgs`, which violates the documented "ExtraArgs last" contract. One-line move.
2. **`parallelSlots=1` behavior change** — the `>1` → `>=1` threshold flip means llamacpp now emits `--parallel 1` where before it didn't. Either revert or call it out in the PR description so it shows up in the changelog.
3. **Precedence check misses the `--flag=value` form** — `v == "--parallel"` doesn't catch `--parallel=4`. Easy fix with `strings.HasPrefix`. Worth pulling into a small shared helper since both runtimes need the same logic.
4. **Renamed Ginkgo test** (`is 1` → `is 0`) tests a `parallelSlots` value the CRD's `Minimum=1` rejects, so the case is unreachable. Could you swap it for one that asserts the new behavior on `parallelSlots=1`?
5. **Test coverage** — the new error-return paths in both runtimes are uncovered. Could you add at least one case per runtime where `parallelSlots` is set and the matching flag is already in `extraArgs`, asserting only the user's flag survives? That would also have caught the order regression in #1.
6. **Docs runtime table** — `ollama`/`omlx` aren't in the `spec.runtime` enum (those are metal-agent flags), and `personaflex`/`llama-cpp` are typos for `personaplex`/`llamacpp`. As written it'll mislead users into `runtime: ollama` which the webhook would reject.
7. **DCO** failing on the merge commit `fd0a19f3` — easiest fix is `git rebase main` for linear history, or `git commit -s --amend` the merge and force-push.

Smaller things, take or leave (none are blockers):

- `vllmLog.Error(nil, err.Error(), …)` is a logr anti-pattern for a configuration warning — `Info` matches the semantics better.
- The `Q8_0` constant extraction in `agent_test.go` is unrelated scope; happy to keep it but a tiny follow-up `refactor(test):` would keep this PR focused.
- Two test cases in `runtime_llamacpp_test.go` share the name `"cacheTypeV nil does not emits flag"` (the second one is actually for `--override-tensor`). Also `does not emits` → `does not emit` while you're in there.
- Couple of grammar fixes in the new doc (`this features is` → `this feature is`; `prevent your service to crash` → `prevent your service from crashing`).

Really nice work overall — this is close. Ping me whenever you've pushed and I'll take another look right away. Thanks again for the thorough writeup and the tests.

internal/controller/runtime_vllm.go
Faylixe: fixed by 7ab0cebf50b09b1f41f430d8636a4fe6540e68f6

internal/controller/runtime_llamacpp_args.go
Faylixe: Yes I was thinking about that but since it is a first contribution I thought that I shall wait to be instructed so :)

I added a new file `runtime_args.go` that would aggregate shared utility for args parsing.

Fixed by 3395e573204c9cfa9eefa74dec1521e20a0c2523 and af7ccca24d22f7a2522c56bcbd7a91e4b2f21a78

internal/controller/runtime_vllm_args.go
Faylixe: Fixed by 3395e573204c9cfa9eefa74dec1521e20a0c2523 and af7ccca24d22f7a2522c56bcbd7a91e4b2f21a78

docs/resource-constrained-deployments.md
Faylixe: Fixed by 6cbd35722603d3d5791c2e75b113ae378008fdea (also fixed typos)

internal/controller/runtime_llamacpp_test.go
Faylixe: Fixed by 52e1d56915162c488c259f02340d19c44e1862f3

internal/controller/runtime_llamacpp_test.go
Faylixe: Fixed by cd97a6dda2d805682a82482b94b8aea62ade1059

I was also thinking that maybe helper functions from `runtime_vllm_test.go` could be moved to a dedicated `runtime_test.go` file since it would be shared by all runtime tests ? Let me know if this is relevant.

internal/controller/runtime_llamacpp_args.go
Faylixe: > Quick question: was the threshold change from `> 1` to `>= 1` intentional? It's a silent llamacpp behavior change — `parallelSlots=1` now causes `--parallel 1` to be emitted where before it relied on llama.cpp's default of 1. For vLLM the change is meaningful (`--max-num-seqs` default is 256), so `>= 1` makes sense there. For llamacpp it changes the rendered Pod spec without changing observable behavior against today's llama.cpp.

Yes it was intentional as it was mentioned in the related issue as a bug (the description of the issue actually mention that `llamacpp` default is 4).

Please let me know which one you prefer (if default is actually 1 then I will revert my change to not break CRD current definition, if it is 4 and the bug is valid I will align that with the current fix) and I will change it tomorrow.

Faylixe:
Thanks ! I fixed almost every blocker except for the parallelSlots guard >= 1 as I need some validation to be sure.

> * `vllmLog.Error(nil, err.Error(), …)` is a logr anti-pattern for a configuration warning — `Info` matches the semantics better.

✅ 

> * The `Q8_0` constant extraction in `agent_test.go` is unrelated scope; happy to keep it but a tiny follow-up `refactor(test):` would keep this PR focused.

Sure, well as I mentioned I did it only because running `make lint` locally got me errors on this matter. But I can totally remove it if you prefer, just let me know.

> * Two test cases in `runtime_llamacpp_test.go` share the name `"cacheTypeV nil does not emits flag"` (the second one is actually for `--override-tensor`). Also `does not emits` → `does not emit` while you're in there.

✅ 

> * Couple of grammar fixes in the new doc (`this features is` → `this feature is`; `prevent your service to crash` → `prevent your service from crashing`).

✅

Defilan:
Yeah, you're right. I framed that wrong. Just checked: `llama-server --help` says `--parallel` defaults to `-1` (auto), and auto picks 4 on every machine I tested. So the existing `> 1` was the bug, exactly like #339 says. Keep your `>= 1`.

Two small follow-ups:

1. Replace the `is 0` test (unreachable, since the CRD's `Minimum=1` rejects it) with one that asserts `--parallel 1` IS in args when `parallelSlots: 1`. That's the case worth pinning.
2. Could you add a line to the PR description so release-please catches it for 0.7.6? Something like *"Fixes #339: `parallelSlots: 1` now emits `--parallel 1` (previously fell through to llama.cpp's auto-default of 4)."*

And yes, please pull the shared fixtures into `runtime_test.go`. That's the right move.

Faylixe:
Hey @Defilan I fixed everything:

- Signed missing commit from Github Actions.
- Rebase from `upstream/main`.
- Remove old parallelSlots test and replace it to check regression if parallelSlots: 1.
- Added fix entry into PR description as requested.

Also ensure `make lint test build` pass locally :).

Quick question: I realized while writing test that the current CRD doc for `ParallelSlots` is now incorrect:

```go
// ParallelSlots sets the number of concurrent request slots for the llama.cpp server (--parallel flag).
// Each slot can process one request independently, enabling concurrent inference.
// Higher values use more memory. If not specified, llama.cpp defaults to 1.
// +kubebuilder:validation:Minimum=1
// +kubebuilder:validation:Maximum=64
// +optional
ParallelSlots *int32 `json:"parallelSlots,omitempty"`
```

Do you want me to fix that now or shall we move that in another PR later (I guess second option as this one is already quite big). Let me know !

Defilan:
Catch the doc fix here, it's one line plus a regen and it's directly tied to the behavior fix in this PR. Leaving wrong docs in the next contributor's path is the kind of thing that creates the next #339.

Suggested wording for the field comment:

```go
// ParallelSlots sets the number of concurrent request slots for the llama.cpp
// server (--parallel flag). Each slot processes one request independently;
// higher values use more KV cache memory. If not specified, the operator
// omits --parallel and llama.cpp picks an auto value (currently 4).
```

Then `make generate manifests chart-crds` to refresh the generated CRD YAML in `config/crd/bases/` and `charts/llmkube/templates/crds/`.

Once that lands and CI is green I'll squash-merge. Thanks for catching it.

Faylixe:
> Catch the doc fix here, it's one line plus a regen and it's directly tied to the behavior fix in this PR. Leaving wrong docs in the next contributor's path is the kind of thing that creates the next #339.
> 
> Suggested wording for the field comment:
> 
> ```go
> // ParallelSlots sets the number of concurrent request slots for the llama.cpp
> // server (--parallel flag). Each slot processes one request independently;
> // higher values use more KV cache memory. If not specified, the operator
> // omits --parallel and llama.cpp picks an auto value (currently 4).
> ```
> 
> Then `make generate manifests chart-crds` to refresh the generated CRD YAML in `config/crd/bases/` and `charts/llmkube/templates/crds/`.
> 
> Once that lands and CI is green I'll squash-merge. Thanks for catching it.

Done !

Defilan (approved):

Issue #186, opened 2026-02-27
Metal agent: runtime memory pressure monitoring and eviction protection


Defilan:
## Background

macOS manages unified memory dynamically and distinguishes between pageable and wired (pinned) memory. Metal GPU buffers used by `llama-server` are wired memory. Under memory pressure, macOS's **wired collector** kernel mechanism can evict GPU buffers, causing severe performance degradation or process crashes — with no warning to the Metal agent.

The agent currently has no visibility into memory pressure and cannot react to it. A process that was healthy at startup can silently degrade as other applications compete for unified memory.

This issue was informed by [research on vllm-metal's memory allocation strategy](https://ranranhaoranzhang.com/blog/2026/llm-inference-memory-allocation-apple-silicon/) which highlights the wired collector problem on Apple Silicon.

## Implementation Plan

This is being implemented in multiple PRs to keep reviews manageable:

### PR A: Memory pressure detection + metrics (Phase 1)
- [x] Extend `MemoryProvider` with `WiredMemory()` and `ProcessRSS()`
- [x] Add `MemoryPressureLevel` type (Normal/Warning/Critical)
- [x] Create `MemoryWatchdog` goroutine (configurable interval, thresholds)
- [x] Add 5 new Prometheus metrics (available, wired, RSS, pressure level, evictions)
- [x] Add CLI flags (`--memory-watchdog-interval`, `--memory-pressure-warning`, `--memory-pressure-critical`, `--eviction-enabled`)
- [x] Unit tests for watchdog, pressure detection, RSS tracking
- **PR:** #216

### PR B: Proactive eviction + status reporting (Phase 2)
- [ ] Eviction logic in the `onPressure` callback (lowest-priority process evicted first)
- [ ] Add `MemoryPressure` condition to InferenceService status
- [ ] Emit Kubernetes events on pressure detection and eviction
- [ ] Add `/memstats` endpoint to health server
- [ ] Unit tests for eviction logic and status updates

## Current State

- Agent polls health every 5 seconds via the K8s watcher, but this only checks for InferenceService changes
- `Healthy` flag is set once at startup and never re-evaluated (tracked in #171)
- No monitoring of system memory pressure or Metal buffer eviction
- No graceful degradation — if macOS kills a `llama-server` process, the agent doesn't know until the next health check (which doesn't exist continuously today)
- No memory-related status information surfaced to Kubernetes

### Memory Pressure Monitoring
[...]

PR #382, opened 2026-05-02
feat(agent): memory-pressure eviction and respawn protection (#186)

Defilan:
## Summary

Phase 2A of #186. Wires the metal-agent's MemoryWatchdog to an eviction selector + respawn block so a runaway model on a Mac Studio actually has a way out other than a hard reboot.

## Behavior

- **Eviction trigger**: only at `MemoryPressureCritical` AND when llmkube's share of `TotalRSS` exceeds **50%** of `TotalMemory`. Pressure caused by a build job or browser does not friendly-fire on inference workloads.
- **Selection**: lowest `InferenceService.Spec.Priority` (low/batch evicted first), tie-break by largest RSS, final tie-break by oldest `StartedAt`. Only `llama-server`-managed processes are eligible; oMLX/Ollama are skipped because their lifecycle is owned by a shared daemon, not by `ManagedProcess`.
- **Stampede control**: at most one eviction per watchdog tick.
- **Respawn block**: after eviction, the key lands in `pressureBlocked`. Subsequent `ensureProcess` calls for that key are no-ops while `lastPressureLevel != Normal`, so the controller's UPDATED event loop cannot silently re-spawn the service we just killed for memory.
- **Recovery**: when the watchdog reports `Normal`, `pressureBlocked` is cleared and the next reconcile respawns the service.
- **Visibility**: a new `MemoryPressure` status condition is patched on every managed InferenceService on level transitions (Reasons: `Normal` / `Warning` / `Critical` / `Evicted`).

## What this is NOT

- **Not** event emission (no `EventRecorder` plumbed into the agent yet) — deferred to PR-2B along with the operator-controlled CRD/CLI flag for opting out, configurable cadence, and recovery metrics.
- **Not** a touch on the controller side: this PR is entirely scoped to `pkg/agent/`.

## Test plan

- [x] Unit tests for selector ordering (priority, RSS, StartedAt, skipped runtimes, unknown priority)
- [x] Unit tests for `shouldEvict` guard at every relevant boundary (disabled, warning-only, below-50%, above-50%, zero TotalMemory)
- [x] Unit tests for the pressure handler: condition patch on transition, no eviction at Warning, friendly-fire guard, Normal clears blocked set
- [x] Unit test for `ensureProcess` back-pressure path (blocked + Critical = no respawn)
- [x] `go vet`, `go build ./...`, `golangci-lint run ./pkg/agent/...` clean for code introduced here (preexisting `health.go:213` G118 and `agent_test.go:639` `q8_0` goconst are not from this PR).
- [ ] Manual smoke on a Mac Studio: spawn 3 services, drive `mem_pressure`, verify the lowest-priority one stops and the others get a `MemoryPressure=True/Critical` condition.

Closes part of #186. PR-2B will follow with events, opt-out flag, and recovery metrics.

### DIFF
--- a/cmd/metal-agent/main.go
+++ b/cmd/metal-agent/main.go
@@ -327,6 +327,7 @@ func main() {
 		ApplePowerEnabled:         cfg.ApplePowerEnabled,
 		ApplePowerInterval:        cfg.ApplePowerInterval,
 		PowermetricsBin:           cfg.PowermetricsBin,
+		EvictionEnabled:           cfg.EvictionEnabled,
 	}
 	if cfg.WatchdogInterval > 0 {
 		agentCfg.WatchdogConfig = &agent.MemoryWatchdogConfig{
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -117,6 +117,17 @@ type MetalAgent struct {
 	mu             sync.RWMutex
 	memoryProvider MemoryProvider
 	memoryFraction float64
+
+	// pressureBlocked records namespacedName keys of processes the agent
+	// evicted under memory pressure. Subsequent ensureProcess calls for these
+	// keys are no-ops while lastPressureLevel != Normal, to prevent a
+	// thrashing respawn loop where the controller's UPDATED event simply
+	// re-spawns the process we just killed for memory.
+	pressureBlocked map[string]bool
+	// lastPressureLevel is the most recent pressure level reported by the
+	// watchdog. Used to gate respawn (above) and to detect transitions for
+	// status condition updates.
+	lastPressureLevel MemoryPressureLevel
 }
 
 // ManagedProcess represents a running inference process (llama-server, oMLX, or Ollama model).
@@ -134,6 +145,12 @@ type ManagedProcess struct {
 	// changed, require respawning the underlying process. Used by ensureProcess
 	// to detect spec drift on UPDATED events and respawn instead of no-oping.
 	SpecHash string
+
+	// Priority is the InferenceService.Spec.Priority enum value
+	// (critical/high/normal/low/batch) captured at spawn time. Used by the
+	// memory-pressure eviction selector to pick the lowest-priority running
+	// process when system memory is critical. Empty defaults to "normal".
+	Priority string
 }
 
 // NewMetalAgent creates a new Metal agent instance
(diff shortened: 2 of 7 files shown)


Defilan:
## Manual smoke test results (M5 Max, 128GB, kind v1.35.0)

Ran the watchdog end-to-end against a real cluster. Three `InferenceService` resources at priorities `batch` / `normal` / `critical`, all referencing the same Qwen3-0.6B-Q8_0 model so RSS stays small enough not to exhaust the laptop.

### Setup found one real bug

The `--eviction-enabled` CLI flag was already declared in `cmd/metal-agent` but never threaded into the agent. The handler hardcoded `evictionEnabled=true` every call. Fixed in 276d562: `MetalAgentConfig.EvictionEnabled` is now consumed by the handler, default false (operators opt in). Added two tests for the gap (the actual eviction path was not previously exercised end-to-end in unit tests).

### What the smoke run validated

Started agent with aggressive thresholds (`--memory-pressure-warning 0.99 --memory-pressure-critical 0.95 --memory-watchdog-interval 5s --eviction-enabled`) so the very first watchdog tick fires `Critical` after the three llama-servers are running.

Watchdog log on first Critical tick:
```
"memory pressure detected","level":"critical","available":"89.9 GB","total":"128.0 GB",
"wired":"8.0 GB","totalRSS":"2.4 GB"
```

`kubectl get isvc … -o jsonpath` immediately after:
```
=== smoke-batch ===
MemoryPressure=True reason=Critical msg=watchdog reported critical memory pressure
=== smoke-normal ===
MemoryPressure=True reason=Critical msg=watchdog reported critical memory pressure
=== smoke-critical ===
MemoryPressure=True reason=Critical msg=watchdog reported critical memory pressure
```

Prometheus metrics:
```
llmkube_metal_agent_managed_processes 3
llmkube_metal_agent_memory_pressure_level 2     # Critical
llmkube_metal_agent_evictions_total 0           # guard correctly held
llmkube_metal_agent_process_rss_bytes{name="smoke-batch"}    8.66e+08
llmkube_metal_agent_process_rss_bytes{name="smoke-critical"} 8.66e+08
llmkube_metal_agent_process_rss_bytes{name="smoke-normal"}   8.66e+08
```

### Confirmed behaviors

- [x] Watchdog goroutine runs and reports level transitions
- [x] `MemoryPressure=True reason=Critical` condition appears on every managed `InferenceService` on the Normal -> Critical transition
- [x] Custom message string flows through (`"watchdog reported critical memory pressure"`)
- [x] Friendly-fire guard correctly prevents eviction at TotalRSS 2.4 GB / 128 GB total = 1.9% (well below 50% threshold) even with `--eviction-enabled` and Critical level
- [x] Process map and metric counters stay correct under sustained Critical pressure
- [x] Per-process RSS gauges populate via `processMemInfoSnapshot`

### Known gap, deferred to PR-2B

Conditions are only patched on level transitions, so a service that spawns AFTER a pressure transition does not get a condition until the next transition. Acceptable for PR-2A (matches K8s controller convention) but worth a follow-up so newly-spawned services see existing pressure on first sight.

### What was NOT exercised

The full eviction path (delete actual `llama-server` PID + set `pressureBlocked` + verify next ensureProcess refuses respawn) requires either ~64 GB of model RSS to satisfy the friendly-fire guard naturally, or synthetic stats injection. Both paths are covered by the unit tests in `pressure_test.go` (`TestHandleMemoryPressure_EvictsLowestPriorityWhenEnabledAndAboveGuard`, `TestEnsureProcess_BlockedUnderCriticalSkipsRespawn`).

PR #394, opened 2026-05-04
feat(crd): add gpuMemoryUtilization and cpuOffloadGB to VLLMConfig

### LINKED ISSUE #307: feat(crd): add gpuMemoryUtilization and cpuOffloadGB to VLLMConfig
## Why

Smoke-testing #306 on 2× RTX 5060 Ti (32 GB total VRAM) surfaced that agentic-coding workloads with FP8 models in the 30B+ range run into VRAM-fit problems that the user can solve with two vLLM flags we currently don't expose through the CRD:

- \`--gpu-memory-utilization\` — vLLM's throttle for what fraction of VRAM it'll claim (default 0.90). On tight setups raising it to 0.95-0.97 buys the headroom needed for profile-run + cudagraph capture.
- \`--cpu-offload-gb\` — spills cold weights to CPU RAM. Gives a clean path for running models that don't fit VRAM at all (e.g., Qwen3.6-35B-A3B-FP8 on 2x 5060 Ti needs ~18 GB/shard > 15.48 GB available → today unrunnable; \`--cpu-offload-gb 4\` per rank makes it fit at a throughput cost).

## Proposed CRD additions

Extends \`VLLMConfig\` (added in #306):

\`\`\`yaml
spec:
  runtime:
    vllm:
      # When set, passes --gpu-memory-utilization to vLLM. Range 0.1-0.99.
      # Default unset (vLLM uses 0.90).
      gpuMemoryUtilization: "0.95"

      # When set, passes --cpu-offload-gb to vLLM. Per-rank, so 4 on TP=2 means
      # 4 GB of CPU RAM per GPU. Use when FP8 model weights don't fit VRAM.
      # Throughput hit is 2-5x on the offloaded path.
      cpuOffloadGB: 4
\`\`\`

## Acceptance

- [ ] Both fields added to VLLMConfig with kubebuilder validation (float 0.1-0.99 for util; int32 min 0 for offload)
- [ ] BuildArgs emits the corresponding flags when set, nothing when nil
- [ ] Table-driven test case for each
- [ ] Sample YAML at \`config/samples/inferenceservice_qwen36_35b_fp8_cpu_offload.yaml\` demonstrating the flagship model running on 2x 5060 Ti with CPU offload

## Context: live smoke test on #306

Qwen3-Coder-30B-A3B-Instruct-FP8 OOM'd during vLLM's profile run with maxModelLen=65536, maxNumBatchedTokens=4096, every other flag from #306 correctly set. GPU 0 used 15.46 GiB / 15.48 GiB usable (13 MiB free). Bumping \`--gpu-memory-utilization\` to 0.95 + dropping maxModelLen to 32768 would likely succeed but is not expressible through the current CRD.

Found during #306 validation.

Faylixe:
## What

Add two new attributes to `VLLMConfig` into the CRD:

- CPUOffloadGB
- GPUMemoryUtilization

## Why

Some vLLM specific args are not exposed into CRD via `VLLMConfig`:

- `--gpu-memory-utilization`
- `--cpu-offload-gb`

Fixes #307

## How

Added matching arg builder function to `runtime_vllm_args.go` and integrate them to the backend `BuildArgs`. Since those two parameters are controlling GPU behavior, I limit them to case where GPU count is > 0 only.

Unit test was added to reflect this behavior with a model routing (with or without GPU).

## Checklist

- [X] Tests added/updated
- [X] `make test` passes locally
- [X] `make lint` passes locally
- [X] All commits are signed off (`git commit -s`) per [DCO](https://developercertificate.org/)
- [X] Documentation updated (if user-facing change)


Faylixe:
Quick thinking that is out of scope and I would probably address in a follow up PR:

- Extract test utilities to a dedicated `runtime_test.go` file (ptr func, containsArgs func, etc).
- resolveGPUCount might be extracted as well as it lives in llama runtime but used in vllm as well.
- Generalize CRD based attributes check VS extraArgs like we did on parallelSlots to prevent conflict.

api/v1alpha1/inferenceservice_types.go
```diff
@@ -560,6 +560,22 @@ type VLLMConfig struct {
...
 	// +optional
 	HFTokenSecretRef *corev1.SecretKeySelector `json:"hfTokenSecretRef,omitempty"`
+
+	// CPUOffloadGb increases the GPU memory size. When set, passes
+	// --cpu-offload-gb to vLLM. Per-rank, so 4 on TP=2 means 4 GB of
+	// CPU RAM per GPU. Use when FP8 model weights don't fit VRAM.
+	// Throughput hit is 2-5x on the offloaded path.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	CPUOffloadGb *int32 `json:"cpuOffloadGB,omitempty"`
```
Defilan: **Style:** Go convention treats `GB` as an initialism. `GPUMemoryUtilization` two lines below already follows that style — renaming this field to `CPUOffloadGB` (the JSON tag `cpuOffloadGB` is already uppercase) makes the two adjacent fields consistent and lines up with `revive`/`stylecheck` defaults.

Follow-up edits needed:
- `make generate` to regenerate `zz_generated.deepcopy.go`
- Call site in `internal/controller/runtime_vllm.go`
- Helper name `appendCPUOffloadGb` → `appendCPUOffloadGB` in `runtime_vllm_args.go`
- Test references in `runtime_vllm_test.go`

internal/controller/runtime_vllm_args.go
```diff
@@ -68,6 +75,13 @@ func appendEnablePrefixCaching(args []string, enablePrefixCaching *bool) []strin
 	return args
 }
 
+func appendGPUMemoryUtilization(args []string, gpuMemoryUtilization *float64) []string {
+	if gpuMemoryUtilization != nil {
+		return append(args, "--gpu-memory-utilization", fmt.Sprintf("%.2f", *gpuMemoryUtilization))
```
Defilan: **Precision loss:** `%.2f` silently rounds. With kubebuilder bounds 0.1–0.99 a user can legitimately write `gpuMemoryUtilization: 0.875` and end up emitting `--gpu-memory-utilization 0.88`. On an 80 GB H100 that's ~800 MB you didn't ask to give up. vLLM accepts arbitrary precision, so we should preserve what the user wrote.

```suggestion
		return append(args, "--gpu-memory-utilization", strconv.FormatFloat(*gpuMemoryUtilization, 'f', -1, 64))
```

…and add `"strconv"` to the imports.

internal/controller/runtime_vllm.go
```diff
@@ -92,9 +93,13 @@ func (b *VLLMBackend) BuildArgs(isvc *inferencev1alpha1.InferenceService, model
 			)
 		}
 		args = appendEnableExpertParallel(args, cfg.EnableExpertParallel)
+
+		if gpuCount > 0 {
```
Defilan: Worth a one-line comment here explaining *why* both new flags are gated on `gpuCount > 0`. `--gpu-memory-utilization` obviously has no meaning on CPU, but `--cpu-offload-gb` reads to a casual reviewer like it should still be valid (the test name `"cpuOffloadGb set does not emit flag if gpu is not enabled"` had me double-take). Documenting intent here saves a future debugging session.

Optional bonus: a `vllmLog.Info(...)` warning when `cfg.CPUOffloadGb != nil && gpuCount == 0`, mirroring the warning pattern in `appendSpeculativeModel`. Silent drops on misconfigured CRDs are a recurring footgun.

internal/controller/runtime_vllm_test.go
```diff
@@ -379,7 +465,12 @@ func TestVLLMBuildArgs(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{Name: "isvc-" + strings.ReplaceAll(tc.name, " ", "-"), Namespace: "default"},
 				Spec:       *tc.spec,
 			}
-			args := backend.BuildArgs(isvc, model, modelPath, port)
+			var args []string
+			if tc.spec.ModelRef == "test-model-gpu" {
```
Defilan, line 469: **Nit (non-blocking):** branching on `tc.spec.ModelRef == "test-model-gpu"` to choose the model fixture works but is fragile — adding a third model fixture later means another string check here. A cleaner pattern is a `model *inferencev1alpha1.Model` field on the test-case struct that defaults to `model` when nil. Happy to take this as a follow-up if you'd rather keep this PR scoped to the CRD additions.

Defilan (changes requested):
Thanks Félix — this is a clean, well-tested implementation of #307. CI is green across all 19 checks, every commit is signed off and GPG-verified, and the Conventional Commits prefixes will land correctly with release-please. The pointer-with-nil-means-unset pattern matches the rest of `VLLMConfig`, the kubebuilder bounds are sensible, and the table tests cover the matrix.

Four small asks before merge:

1. **Go field naming.** `CPUOffloadGb` → `CPUOffloadGB`. Go style treats `GB` as an initialism, and `GPUMemoryUtilization` two lines below already follows that convention. The JSON tag `cpuOffloadGB` is already uppercase, so this just brings the Go identifier in line. Touches the type definition, the regenerated deepcopy, the helper name `appendCPUOffloadGb`, the call site in `runtime_vllm.go`, and the test references.

2. **Float-precision rounding.** `fmt.Sprintf("%.2f", ...)` silently rounds — a user writing `gpuMemoryUtilization: 0.875` gets `--gpu-memory-utilization 0.88` on the wire. With 80 GB-class GPUs that delta matters. Round-trip what the user wrote with `strconv.FormatFloat(*v, 'f', -1, 64)` (and add `"strconv"` to imports).

3. **Document the `gpuCount > 0` gate.** A one-line comment in `runtime_vllm.go` explaining that *both* flags are silently dropped on CPU-only specs would help the next reader — `--cpu-offload-gb` looks like it should still apply when there's no GPU, and the test name `"cpuOffloadGb set does not emit flag if gpu is not enabled"` documents this as a feature without explaining why. Optional bonus: a `vllmLog.Info(...)` warning when `cpuOffloadGB` is set but ignored, mirroring the pattern in `appendSpeculativeModel`. Silent drops on misconfigured CRDs are a recurring footgun.

4. **Missing sample YAML from #307's acceptance list.** The issue explicitly asks for `config/samples/inferenceservice_qwen36_35b_fp8_cpu_offload.yaml` showing the Qwen3.6-35B-A3B-FP8 + 2× 5060 Ti + CPU offload setup. Could you add that, or call it out so I can decide whether to defer?

Non-blocking: the `tc.spec.ModelRef == "test-model-gpu"` sentinel in the test runner is mildly fragile — happy to take that as a follow-up if you'd rather keep this PR scoped.

Once 1–4 are in, I'll merge. Appreciate the work.

internal/controller/runtime_vllm_test.go
Faylixe: Yes I felt it was weird too :).

Well I think I would take it on a follow up PR focused on testing to clean up a bit (extracting common test utilities to a dedicated `runtime_test.go` for instance). So I can clean up testing for good at once :)

internal/controller/runtime_vllm.go
Faylixe: Indeed, I migrate the GPU check into the helper with error emitting pattern + log to reflect that :)

Faylixe:
Hey @Defilan 

I fixed most of it. The only thing I need is the review for the sample inference service (as I don't have any GPU at home I can't really test if the sample is right in term of parameter :). Also didn't fix the test model routing for now. Once this PR is merged I will just right on a test cleanup one :)

Defilan (approved):
Thank you very much @Faylixe ! Approving and merging now. 

Alt Text

PR #407, opened 2026-05-06
refactor(controller): move runtime shared test helpers to a dedicated file

Faylixe:
## What

Cleanup testing and shared helpers.

## Why

As discussed in previous PR reviews, some refactoring and cleanup was needed on the runtime testing part.

## How

- Move shared helper functions for runtime testing to a dedicated file `runtime_test.go` (`containsArgs` and `ptr*`).
- Move duplicated FlagCheck structure to `runtime_test.go`.
- Move duplicated test case inline structure to `runtime_test.go.
- Add model to test case to add fixture variations based on model spec.
- Move `resolveGPUCount` from `runtime_llama.go` to `runtime.go`.
- Add test suite for `resolveGPUCount`.

## Checklist

- [X] Tests added/updated
- [X] `make test` passes locally
- [X] `make lint` passes locally
- [X] All commits are signed off (`git commit -s`) per [DCO](https://developercertificate.org/)
- [X] Documentation updated (if user-facing change)

### DIFF
--- a/internal/controller/runtime_test.go
+++ b/internal/controller/runtime_test.go
@@ -1,0 +1,26 @@
+package controller
+
+// containsArg reports whether args contains the given flag. When value is
+// non-empty, it also requires the immediately following entry to equal value
+// (i.e. `--flag value` as separate slice elements, which is how BuildArgs
+// emits everything).
+func containsArg(args []string, flag, value string) bool {
+	for i, a := range args {
+		if a != flag {
+			continue
+		}
+		if value == "" {
+			return true
+		}
+		if i+1 < len(args) && args[i+1] == value {
+			return true
+		}
+	}
+	return false
+}
+
+// ptrString, ptrBool, ptrInt32 are local helpers so tests read naturally.
+func ptrBool(b bool) *bool          { return &b }
+func ptrFloat64(f float64) *float64 { return &f }
+func ptrInt32(i int32) *int32       { return &i }
+func ptrString(s string) *string    { return &s }
--- a/internal/controller/runtime_test.go
+++ b/internal/controller/runtime_test.go
@@ -1,5 +1,7 @@
 package controller
 
+import inferencev1alpha1 "github.com/defilantech/llmkube/api/v1alpha1"
+
 // containsArg reports whether args contains the given flag. When value is
 // non-empty, it also requires the immediately following entry to equal value
 // (i.e. `--flag value` as separate slice elements, which is how BuildArgs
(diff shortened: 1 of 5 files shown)


internal/controller/runtime_test.go
```diff
@@ -0,0 +1,44 @@
...
+func ptrFloat64(f float64) *float64 { return &f }
+func ptrInt32(i int32) *int32       { return &i }
+func ptrString(s string) *string    { return &s }
+
+type FlagCheck struct {
+	flag  string
+	value string
+}
+
+type RuntimeBuildArgsTestCase struct {
```
Faylixe: @Defilan I extracted as well the BuildArgs test structure. Although it locks what a runtime build args test case is which might in the future preventing some variations ? I am not convinced about that and I might revert to having per runtime inline struct so it remains flexible. Let me know :)

Defilan: Your instinct is right — let's revert the `RuntimeBuildArgsTestCase` extraction and keep the test-case struct inline per runtime. The shared helpers (`containsArg`, `FlagCheck`, the `ptr*` family, and `resolveGPUCount` moved into `runtime.go`) are the genuine wins; those should stay.

I looked at how the closest analog in the space handles this. KServe is a Kubernetes operator with multiple ML serving runtimes (HuggingFace, ONNX, LightGBM, Paddle, PMML, Triton, custom). Their convention at `pkg/apis/serving/v1beta1/predictor_*_test.go`:

- One test file per runtime (same as our split)
- Each test function declares its own inline anonymous struct, tailored to what that test asserts:
  - Validation tests: `struct { spec PredictorSpec; matcher gomega.GomegaMatcher }`
  - Defaulter tests: `struct { spec PredictorSpec; expected PredictorSpec }`
  - Simple presence tests: `struct { spec PredictorSpec }`
- They DO share helpers (Gomega matchers, builders like `proto.String()`)
- They do NOT share a runtime-test-case struct

The reasoning maps cleanly to our situation: today both runtimes happen to assert the same thing (find flags in argv), but probes, image selection, env vars, command shape, and port mapping are all things `RuntimeBackend.BuildArgs` could grow tests around — and those will plausibly want different assertion shapes per runtime. Pre-locking the struct would force struct-gymnastics on the first divergence, when an inline anonymous struct just lets it happen.

So the path forward:

1. **Keep:** `containsArg`, `FlagCheck`, `ptrBool/Int32/String/Float64` in `runtime_test.go` — shared helpers
2. **Keep:** `resolveGPUCount` moved to `runtime.go` and the new `TestResolveGPUCount` — that one's a clear win
3. **Revert:** `RuntimeBuildArgsTestCase` extraction; let each test function (`TestLlamaCppBuildArgs`, `TestVLLMBuildArgs`, future `TestVLLMSwiftBuildArgs`) declare its own inline anonymous struct, threading `model` through naturally

Once that's in I'll squash-merge. Thanks for raising the question — caught a real pre-locking risk before it shipped.

Faylixe: Make sense ! Just revert the change ;)

Defilan (approved):
Thanks for cycling back on this so quickly, Felix — the revert landed exactly how I had in mind. Per-test inline structs in `runtime_llamacpp_test.go` and `runtime_vllm_test.go`, shared helpers (`containsArg`, `FlagCheck`, `ptr*`) and `TestResolveGPUCount` in the new `runtime_test.go`, and `resolveGPUCount` properly hoisted to `runtime.go` since both backends consume it. Nice bonus: threading `model` through each row cleaned up the old `ModelRef == "test-model-gpu"` branch in the vLLM runner — that reads better than what we had before. All checks green, DCO clean across all five commits. Approving — will squash-merge. Appreciate the patience on the back-and-forth here.

PR #474, opened 2026-05-16
feat: add scale sub resource

### LINKED ISSUE #473: [FEATURE] InferenceService should implement the Kubernetes scale subresource
## Feature Description

Scale-to-zero with KEDA: keep InferenceService.spec.replicas: 0 at idle, use a KEDA ScaledObject (HTTP or external trigger) to scale to 1 on demand, and scale back to 0 after a cooldown period. Essentially allowing me to hot swap models on limited hardware.

## Problem Statement

InferenceService does not implement the Kubernetes `/scale` subresource, which prevents external autoscalers (KEDA, HPA) from targeting it. KEDA immediately deletes any ScaledObject that targets a CRD without /scale. 

## Proposed Solution

Add the scale subresource to the InferenceService CRD definition:    

```yaml
subresources:
  scale:
    specReplicasPath: .spec.replicas
    statusReplicasPath: .status.replicas
```

And populate status.replicas in the controller to reflect the running pod count.      

## Alternatives Considered

What alternative solutions or features have you considered?

## Additional Context

This feature would allow me to have multiple model deployments all scaled to zero in my cluster, then use KEDA to scale them up to 1 when requests arrive. After a given amount of time with no activity, the model would go back to zero replicas and I can spin up another.

## Priority

How important is this feature to you?

- [ ] Medium - Nice to have
- [ ] Low - Minor improvement

## Willingness to Contribute

Are you willing to contribute to this feature?

- [x] Yes, I can submit a PR
- [ ] Yes, I can help test
- [ ] No, just requesting

mircea-pavel-anton:
## What

Scale-to-zero support: keep InferenceService.spec.replicas: 0 at idle, use a KEDA ScaledObject (HTTP or external trigger) to scale to 1 on demand, and scale back to 0 after a cooldown period. Essentially allowing me to hot swap models on limited hardware.

## Why

This feature allows InferenceServices to be scaled up/down on-demand via `kubectl scale` or via external scalers such as KEDA. This also allows scale to zero to work.

Fixes #473 

## How

Added scale sub-resource.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,11 @@ spec:
 
 ```bash
 kubectl apply -f model.yaml
+
+# Scale replicas at any time — InferenceService supports the standard Kubernetes scale subresource
+kubectl scale inferenceservice/tinyllama --replicas=3
 ```
+
 </details>
 
 **Full setup guides:** [Minikube Quickstart](docs/minikube-quickstart.md) | [GKE with GPUs](docs/gpu-setup-guide.md) | [Air-Gapped Deployment](docs/air-gapped-quickstart.md) | [OpenShift](#troubleshooting)
@@ -314,7 +318,7 @@ Consistent ~53 tok/s across 3-8B models with automatic layer sharding. See [v0.4
 - Automatic model download from HuggingFace, HTTP, or PVC (S3 planned)
 - Persistent model cache, download once, deploy instantly ([guide](docs/MODEL-CACHE.md))
 - OpenAI-compatible `/v1/chat/completions` API
-- Multi-replica horizontal scaling
+- Multi-replica horizontal scaling with scale subresource support (`kubectl scale`, KEDA)
 - License compliance scanning for GGUF models
 
 **Routing & policy ([ModelRouter](#composition-modelrouter)):**
--- a/README.md
+++ b/README.md
@@ -155,7 +155,6 @@ spec:
 ```bash
 kubectl apply -f model.yaml
 ```
-
 </details>
 
 **Full setup guides:** [Minikube Quickstart](docs/minikube-quickstart.md) | [GKE with GPUs](docs/gpu-setup-guide.md) | [Air-Gapped Deployment](docs/air-gapped-quickstart.md) | [OpenShift](#troubleshooting)
--- a/README.md
+++ b/README.md
@@ -154,9 +154,6 @@ spec:
 
 ```bash
 kubectl apply -f model.yaml
-
-# Scale replicas at any time — InferenceService supports the standard Kubernetes scale subresource
-kubectl scale inferenceservice/tinyllama --replicas=3
 ```
 
 </details>
--- a/api/v1alpha1/inferenceservice_types.go
+++ b/api/v1alpha1/inferenceservice_types.go
@@ -650,7 +650,7 @@ type InferenceServiceStatus struct {
 	// +optional
 	DesiredReplicas int32 `json:"desiredReplicas,omitempty"`
 
-	// Replicas is the current number of running inference pods, exposed via the scale subresource.
+	// Replicas is the current number of running inference pods
 	// +optional
 	Replicas int32 `json:"replicas,omitempty"`
 
--- a/api/v1alpha1/inferenceservice_types.go
+++ b/api/v1alpha1/inferenceservice_types.go
@@ -650,6 +650,10 @@ type InferenceServiceStatus struct {
 	// +optional
 	DesiredReplicas int32 `json:"desiredReplicas,omitempty"`
 
+	// Replicas is the current number of running inference pods, exposed via the scale subresource.
+	// +optional
+	Replicas int32 `json:"replicas,omitempty"`
+
 	// Endpoint is the service URL where inference requests can be sent
 	// +optional
 	Endpoint string `json:"endpoint,omitempty"`
@@ -699,6 +703,7 @@ type InferenceServiceStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas
 // +kubebuilder:resource:shortName=isvc
 // +kubebuilder:printcolumn:name="Phase",type=string,JSONPath=`.status.phase`
 // +kubebuilder:printcolumn:name="Model",type=string,JSONPath=`.spec.modelRef`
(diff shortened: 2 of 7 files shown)


internal/controller/status_builder.go
```diff
@@ -111,6 +111,7 @@ func (r *InferenceServiceReconciler) updateStatusWithSchedulingInfo(
 	isvc.Status.Phase = phase
 	isvc.Status.ModelReady = modelReady
 	isvc.Status.ReadyReplicas = readyReplicas
+	isvc.Status.Replicas = readyReplicas
```
Defilan: `Status.Replicas` is the path `statusReplicasPath` reads for the scale subresource. By convention it should report the *total current* replica count (so HPA computes a correct scaling ratio), not the *ready* count. Reporting `readyReplicas` here can cause over-scaling during rollouts when pods aren't yet Ready. Consider sourcing this from the Deployment's `*Spec.Replicas` instead, keeping `ReadyReplicas` as the separate ready signal. If KEDA-only support is the intent, that's acceptable but should be documented and the README's HPA claim dropped.

api/v1alpha1/inferenceservice_types.go
```diff
@@ -699,6 +703,7 @@ type InferenceServiceStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas
```
Defilan, line 706: For full HPA compatibility through `/scale`, the scale subresource also needs a `selectorpath` pointing at a `status` field with a serialized label selector. Without it, `autoscaling/v2` HPA targeting this CRD fails with `unable to get selector`. KEDA's external scaler doesn't need it, so this is optional if the PR is scoped to KEDA, but the README/PR currently claim HPA support. Either add a `status.selector` field and a `selectorpath`, or narrow the stated scope to KEDA.

config/samples/inferenceservice_scale_subresource.yaml
```diff
@@ -0,0 +1,31 @@
...
+  image: ghcr.io/ggml-org/llama.cpp:server
+  resources:
+    cpu: "2"
+    memory: "4Gi"
+
+---
+apiVersion: keda.sh/v1alpha1
+kind: ScaledObject
+metadata:
+  name: tinyllama-scalable
```
Defilan: The `ScaledObject` has no `namespace` while the `InferenceService` above is in `default`. KEDA requires the `ScaledObject` to live in the same namespace as its target, so set `namespace: default` here for clarity and correctness.

config/samples/inferenceservice_scale_subresource.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: inference.llmkube.dev/v1alpha1
+kind: InferenceService
+metadata:
+  name: tinyllama-scalable
+  namespace: default
+spec:
+  modelRef: tinyllama
+  replicas: 0
```
Defilan, line 29: Worth a comment in this sample noting that the scale subresource (driven by KEDA here) and `spec.autoscaling` are mutually exclusive. Both contend for `spec.replicas`, and using them together makes two controllers fight over the field.

README.md
```diff
@@ -155,6 +155,7 @@ spec:
 ```bash
 kubectl apply -f model.yaml
 ```
+
```
Defilan: This blank line is unrelated to the feature, so please drop it to keep the diff focused.

Defilan (changes requested):
Thanks for this, and thanks for getting the commits signed off. A clean, well-scoped PR: clear What/Why/How, the CRD regen done correctly, and a sample, test, and docs included. CI is green.

One thing I'd like resolved before merge. `status.replicas` currently mirrors `readyReplicas`, but the scale subresource's `statusReplicasPath` is conventionally the *total current* replica count. HPA reads that field to compute its scaling ratio, so reporting "ready" instead of "total" can cause over-scaling during rollouts. The PR and the README mention HPA, but the subresource also lacks a `selectorpath`, which `autoscaling/v2` HPA needs to resolve pods. Two clean ways forward: either (a) point `status.replicas` at the Deployment's replica count and add a `status.selector` plus `selectorpath` for real HPA support, or (b) scope this PR to KEDA, which needs neither, and adjust the README to say KEDA rather than KEDA/HPA, with HPA as a follow-up.

Also strongly suggested: a test that drives the `/scale` subresource directly (and a scale-to-0 case) would be more reassuring than the current field-copy assertion, and a one-line note that the scale subresource and `spec.autoscaling` are mutually exclusive would save users a footgun. Nice contribution overall, and happy to help land it.

README.md
mircea-pavel-anton: fixed in [2e40c88](https://github.com/defilantech/LLMKube/pull/474/commits/2e40c882ae371e86632a0c50b2e3166a5163938a)

config/samples/inferenceservice_scale_subresource.yaml
mircea-pavel-anton: Fixed in [22b3860](https://github.com/defilantech/LLMKube/pull/474/commits/22b386014166bfedebc5d31de88b79205de61f71)
I re-did the whole example, adding what I had deployed in my cluster to be confirmed working.

internal/controller/status_builder.go
mircea-pavel-anton: Something like this perhaps? [d797121](https://github.com/defilantech/LLMKube/pull/474/commits/d79712126467ab3d826923d55349a85b7710c328)

If I understand correctly, this means that KEDA always sees the intended replica count rather than the *ready* replica count. During a rollout with pods not yet `Ready`, it reports 3 (desired) instead of 0 (ready), which prevents any false over-scaling signal.

config/samples/inferenceservice_scale_subresource.yaml
mircea-pavel-anton: Fixed in [22b3860](https://github.com/defilantech/LLMKube/pull/474/commits/22b386014166bfedebc5d31de88b79205de61f71)

api/v1alpha1/inferenceservice_types.go
mircea-pavel-anton: I think that it's best to scope this PR to KEDA. My argument for that would be that, at least as far as I understand, the current `spec.autoscaling` functionality already implements HPA for scaling, so I think this could end up being confusing, having 2 means to the same end. (plus, to be completely honest I only intended/tested KEDA to work here because it allows scale to zero)

Defilan (approved):
Excellent work getting this done so fast! You addressed everything I mentioned. I'll get this merged momentarily. Welcome to LLMKube!

PR #483, opened 2026-05-17
fix: scale llama.cpp on requests_deferred, not requests_processing

### LINKED ISSUE #482: [BUG] llama.cpp default autoscaling metric never triggers a scale-up
## Bug Description

The HPA that LLMKube generates for a llama.cpp `InferenceService` (when `spec.autoscaling` is set with no explicit metrics) uses the default metric `llamacpp:requests_processing` with a target `AverageValue` of `2`. But llama-server runs with a single slot by default (no `--parallel` flag), so `llamacpp:requests_processing` is capped at 1 per pod and can never reach the target of 2. `spec.autoscaling` therefore never scales a llama.cpp service up under load.

## Steps to Reproduce

1. Deploy a llama.cpp `InferenceService` with `spec.autoscaling` set and no explicit `metrics`.
2. The operator creates an HPA on `llamacpp:requests_processing`, target `AverageValue: 2`.
3. Drive sustained concurrent load at the service.
4. `requests_processing` saturates at 1 (one slot); excess requests land in `requests_deferred`.

## Expected Behavior

A llama.cpp service with `spec.autoscaling` and default settings scales up under load.

## Actual Behavior

It never scales up. `requests_processing` is a gauge bounded by the slot count (default 1), so it cannot reach the hardcoded target of 2. Verified on a live cluster.

## Environment

**LLMKube Version:** current `main`
**Cluster Type:** not cluster-specific (reproduced on a CPU-only InferenceService)

## Fix

Change the llama.cpp `DefaultHPAMetric()` to `llamacpp:requests_deferred`: the queue-depth gauge. It is unbounded, directly signals "requests are waiting for capacity," and works at any slot count.

## Additional Context

Found while building the metrics-driven autoscaling tutorial (#478), alongside #480 (the PodMonitor selector bug). With both fixed, native `spec.autoscaling` works end to end.

Defilan:
## What

Changes the llama.cpp backend's default HPA metric from `llamacpp:requests_processing` to `llamacpp:requests_deferred`.

## Why

When `spec.autoscaling` is set with no explicit metrics, the operator generates an HPA on the runtime's `DefaultHPAMetric()` with a target `AverageValue` of `2`. For llama.cpp that metric was `llamacpp:requests_processing`.

`requests_processing` is a gauge bounded by the llama-server slot count, and llama-server runs a single slot by default (no `--parallel`). So the metric is capped at 1 per pod and can never reach the target of 2. With default settings, `spec.autoscaling` never scaled a llama.cpp service up. Verified on a live cluster: under sustained concurrent load the HPA stayed at minReplicas while requests piled into the queue.

## How

`requests_deferred` is the gauge of requests waiting for a slot. It is unbounded, directly signals "this service is out of capacity," and works at any slot count. Switching the default to it makes `spec.autoscaling` autoscale out of the box.

The HPA test that asserted the default metric name is updated.

## Checklist

- [x] Bug reproduced and fix verified on a live cluster
- [x] Tests updated (`constructHPA` default-metric assertion)
- [x] `make build` / `go vet` pass locally

Fixes #482

### DIFF
--- a/api/v1alpha1/inferenceservice_types.go
+++ b/api/v1alpha1/inferenceservice_types.go
@@ -408,7 +408,8 @@ type AutoscalingSpec struct {
 	MaxReplicas int32 `json:"maxReplicas"`
 
 	// Metrics defines the scaling metrics and target values.
-	// If empty, defaults to llamacpp:requests_processing with target average value of 2.
+	// If empty, defaults to the runtime's default metric with a target
+	// average value of 2 (llama.cpp: llamacpp:requests_deferred).
 	// +optional
 	Metrics []MetricSpec `json:"metrics,omitempty"`
 }
--- a/charts/llmkube/templates/crds/inferenceservices.yaml
+++ b/charts/llmkube/templates/crds/inferenceservices.yaml
@@ -98,7 +98,8 @@ spec:
                   metrics:
                     description: |-
                       Metrics defines the scaling metrics and target values.
-                      If empty, defaults to llamacpp:requests_processing with target average value of 2.
+                      If empty, defaults to the runtime's default metric with a target
+                      average value of 2 (llama.cpp: llamacpp:requests_deferred).
                     items:
                       description: MetricSpec defines a single metric for HPA scaling.
                       properties:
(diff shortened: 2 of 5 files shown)


Defilan:
Closing: the premise of this PR was wrong.

llama-server runs 4 slots by default (not 1), so `llamacpp:requests_processing` reaches 4 under load and the existing default HPA target of 2 is reachable. Confirmed by load test: 12 concurrent requests produced 4 processing + 8 deferred. The default metric autoscales fine, and switching to `requests_deferred` is arguably worse since it only scales once requests are already queued. No change needed. See #482.

PR #513, opened 2026-05-21
docs(macos-metal): replace broken port-forward step with host-localhost curl

### LINKED ISSUE #512: [BUG] macOS Metal guide's kubectl port-forward step fails: metal-path Service is selector-less by design
## Bug Description

The macOS Metal quickstart instructs the user to `kubectl port-forward`
the InferenceService's Service, but kubectl refuses with:

```
error: cannot attach to *v1.Service: invalid service 'phi-4-mini':
Service is defined without a selector
```

This isn't a bug in the user's deployment -- it's a fundamental
incompatibility between `kubectl port-forward svc/...` (which requires
a Selector to find target pods) and the LLMKube operator's
metal-path Service shape (which is intentionally selector-less so the
metal-agent's manually-registered Endpoints aren't overwritten by k8s
auto-discovery).

The current doc instruction can never work for a metal-served
InferenceService. It needs to be replaced.

Affected:

- `docs/site/guides/macos-metal.md:170` -- the only metal-path
  port-forward in the docs.
- The cluster-side GPU + CPU port-forwards in `docs/gpu-setup-guide.md`
  are fine (those paths use Deployments with selectors).

## Steps to Reproduce

1. Deploy a metal-served InferenceService (any model on Apple Silicon
   via metal-agent -- the macos-metal guide is the canonical path).
2. Wait for `kubectl get inferenceservice phi-4-mini` to show
   `PHASE=Ready`.
3. Run the next instruction from the guide:

   ```bash
   kubectl port-forward svc/phi-4-mini 8080:8080 &
   ```

4. Observe the error above.

Reproduced live against an unrelated metal InferenceService
(`qwen36-35b-carnice-mtp`, Ready, Endpoints populated by metal-agent
at host IP) -- same error, byte-for-byte.

## Expected Behavior

The guide's flow lets the user verify their deployment works. Either
the port-forward succeeds, or the guide shows the user a command that
actually works for metal-served services.

## Actual Behavior

`kubectl port-forward` errors out with no path forward inside the
guide. A reader who doesn't already know that the Service is
selector-less by design has no obvious way to continue.

## Root Cause

The metal-path operator skips the Deployment entirely (llama-server
runs natively on the host, supervised by metal-agent). To make the
Service still resolve, the metal-agent registers `Endpoints` (or, on
modern k8s, EndpointSlice -- see #377) pointing at the host's IP +
port. This works for in-cluster traffic via ClusterIP, but
`kubectl port-forward svc/X` is hard-wired to require a Selector and
will not consult the Endpoints object.

Code references:

- `internal/controller/inferenceservice_controller.go:214` -- "No
  Deployment for metal" branch.
- `internal/controller/service_builder.go:64` -- sets Selector for
  the normal (Deployment-backed) path; the metal branch
  intentionally elides it via reconcileService's metal stub.
- `internal/controller/inferenceservice_controller.go:324` --
  `metalReadyEndpoints` reads the Endpoints object the metal-agent
  populates.
- Recent fix landing the Endpoints watch: #486 (fixed in PR #488).

## Proposed Fix (docs)

Replace the port-forward step in `docs/site/guides/macos-metal.md`
with a direct host-side curl (llama-server is already on the
host you're following the guide on), and mention the two ways to
expose this Service to other machines.

Current snippet (line 170-176):

```bash
kubectl port-forward svc/phi-4-mini 8080:8080 &
curl -sS http://localhost:8080/v1/chat/completions \
  -H 'content-type: application/json' \
  -d '{"model":"phi-4-mini","messages":[{"role":"user","content":"hi"}]}'
```

Suggested replacement:

```bash
# llama-server runs natively on this host via metal-agent;
# no port-forward needed when you are on the same machine.
curl -sS http://localhost:8080/v1/chat/completions \
  -H 'content-type: application/json' \
  -d '{"model":"phi-4-mini","messages":[{"role":"user","content":"hi"}]}'
```

Then a follow-up paragraph for "from another machine":

```markdown
### Reaching the service from elsewhere

The InferenceService Service on the metal path is selector-less by
design (the metal-agent registers the host as the Endpoint), so
`kubectl port-forward svc/...` will not work. Two supported ways
to expose the service to other clients:

1. **Set `spec.endpoint.type: NodePort` on the InferenceService.**
   Then `curl http://<host-ip>:<node-port>/v1/...` from anywhere
   that can route to the host.
2. **Hit the host directly.** metal-agent binds llama-server to
   `0.0.0.0` and registers `<host-ip>:8080` as the Endpoint, so
   `curl http://<host-ip>:8080/v1/...` works straight from any
   client on the same network.
```

## Proposed Fix (operator, v0.2)

Optional and lower priority: emit a clearer error when a user runs
`port-forward` against a metal-path Service, e.g. via an annotation
or a NOTES.md the chart surfaces post-install. Anyone hitting this
without context will lose time on it.

A cleaner architectural fix would be to make the metal-agent inject
itself as a Pod that owns the actual llama-server process -- that
restores the standard selector->Pod model and port-forward "just
works." That's a structural change worth considering for v0.2 but
not justified by this defect alone.

## Additional Context

- This is the second metal-tutorial doc defect filed in the same
  session. The first was #511 (broken HuggingFace link).
  Worth thinking about whether the macOS Metal guide deserves an
  end-to-end "freshly install on a clean Mac" CI suite, or at
  minimum a manual quarterly walkthrough.
- The `Documentation Validation` CI job appears to be schema/lint
  oriented; it does not exercise the guide commands.

Defilan:
## What

Replaces the macOS Metal quickstart's `kubectl port-forward svc/...`
instruction (which never works against a metal-served InferenceService)
with a direct `curl http://localhost:8080/...` from the same Mac, plus
a short "Reaching the service from elsewhere" subsection covering
NodePort + direct host-IP access for off-host clients.

Single-file diff: `docs/site/guides/macos-metal.md`, +21 / -2 lines.

## Why

Fixes #512

The metal path operator deliberately creates a selector-less Service
(metal-agent runs `llama-server` natively on the host and registers
the host IP as the Endpoint manually). `kubectl port-forward svc/...`
is hard-wired to require a Selector and refuses to consult the
Endpoints object, so the guide's step returns:

```
error: cannot attach to *v1.Service: invalid service 'phi-4-mini':
Service is defined without a selector
```

Every reader of the macOS Metal quickstart hits this at Step 5, and
the diagnostic is uninformative without architectural context. See
issue #512 for the full root-cause writeup + alternatives considered.

## How

- Reframed the leading sentence: the original "Query the model from
  anywhere in the cluster" was misleading (the metal path's
  ClusterIP service is reachable from in-cluster, but the
  port-forward call is bound to your laptop, so the natural reading
  is wrong twice over).
- New copy explicitly names the design (selector-less Service) so a
  reader hitting the same error via another doc path can grep and
  land here.
- Two supported off-host exposure paths documented: `NodePort`
  endpoint type, and direct `<host-ip>:<port>` access. Both work
  today; nothing needed from the operator.

## Checklist

- [x] Tests added/updated -- N/A; docs-only change.
- [x] `make test` passes locally -- N/A; no code touched. CI's
      `Documentation Validation` job exercises the doc lint surface.
- [x] `make lint` passes locally -- N/A; same reason.
- [x] Commit messages follow conventional commits.
- [x] All commits are signed off (`git commit -s`) per DCO.
- [x] Documentation updated -- this PR *is* the documentation update.

### DIFF
--- a/docs/site/guides/macos-metal.md
+++ b/docs/site/guides/macos-metal.md
@@ -164,14 +164,33 @@ The agent's log should show:
 "msg":"started inference service","name":"phi-4-mini","pid":<llama-server-pid>
 ```
 
-Query the model from anywhere in the cluster:
+Query the model from this Mac (`llama-server` is already running
+natively on this host via metal-agent, so no port-forward is needed
+when you are on the same machine):
 
 ```bash
-kubectl port-forward svc/phi-4-mini 8080:8080 &
 curl -sS http://localhost:8080/v1/chat/completions \
   -H 'content-type: application/json' \
   -d '{"model":"phi-4-mini","messages":[{"role":"user","content":"hi"}]}'
 ```
+
+### Reaching the service from elsewhere
+
+The InferenceService Service on the metal path is selector-less by
+design: the metal-agent registers the host as the Endpoint, so the
+Service has no Pods to target. That means
+`kubectl port-forward svc/phi-4-mini ...` returns
+`error: cannot attach to *v1.Service: ... Service is defined without
+a selector` and cannot be used here. Two supported ways to expose
+this Service to clients on other machines:
+
+1. **Set `spec.endpoint.type: NodePort` on the InferenceService.**
+   Then `curl http://<host-ip>:<node-port>/v1/...` from anywhere that
+   can route to the host.
+2. **Hit the host directly.** metal-agent binds `llama-server` to
+   `0.0.0.0` and registers `<host-ip>:<port>` as the Endpoint, so
+   `curl http://<host-ip>:8080/v1/...` works from any client on the
+   same network.
 
 ## Memory budgets
 


Defilan:
Verified on an M5 Max running this exact setup
(`InferenceService/qwen36-35b-carnice-mtp` served by metal-agent +
llama-server natively):

**The new command (host-localhost curl) returns a clean 200:**

```sh
$ curl -sS http://localhost:8080/v1/chat/completions \
    -H 'content-type: application/json' \
    -d '{"model":"qwen36-35b-carnice-mtp","messages":[{"role":"user","content":"Reply with: HELLO"}],"max_tokens":200,"temperature":0}' \
    | jq '{model, usage, tps: .timings.predicted_per_second}'
{
  "model": "Carnice-Qwen3.6-MoE-35B-A3B-APEX-MTP-I-Balanced.gguf",
  "usage": { "completion_tokens": ..., "prompt_tokens": 16, "total_tokens": ... },
  "tps": ~102
}
```

**The original port-forward command fails byte-for-byte as #512 documents:**

```sh
$ kubectl port-forward svc/qwen36-35b-carnice-mtp -n default 18080:8080
error: cannot attach to *v1.Service: invalid service 'qwen36-35b-carnice-mtp': Service is defined without a selector
```

Same metal-path Service shape any user following the guide would hit
(selector-less by design; `metal-agent` registers the host as the
Endpoint). The doc change in this PR is what actually works on this
hardware today.

Defilan:
Force-pushing not needed; added a follow-up commit (`8626b18`)
after the first real-tester walkthrough exposed a second wrong
assumption in the original snippet.

What changed:

- The original `curl localhost:8080` only works if metal-agent was
  started with `--llama-server-port 8080`. Without that flag, the
  agent allocates llama-server an ephemeral port and registers it
  on the Service's Endpoints. The tester's `kubectl get endpoints`
  showed `192.168.65.254:63344` -- right host, different port, so
  the doc command got `Failed to connect to localhost port 8080`.
- New section *Find the endpoint* reads the actual IP:port from
  `kubectl get endpoints` first; *Query the model* uses
  `PORT=$(kubectl get endpoints ...)` to derive it. Now portable
  across any agent flag set.
- Added a single-line curl alongside the multi-line form to
  protect against paste mediums that strip trailing backslashes
  (another tester failure mode in the same walkthrough).
- *Reaching the service from elsewhere* reordered: "hit the host
  directly" is now option #1 (it works for any LAN client with no
  extra config) and includes a concrete example reusing the address
  kubectl already printed.

Verified live against the running Carnice InferenceService on the
same M5 Max from the first verification comment: HTTP 200, model
name matches, ~119 tok/s.

PR #537, opened 2026-05-25
[WIP] fix(metal-agent): reconcile pre-existing InferenceService status on startup

### LINKED ISSUE #506: [BUG] metal-agent does not reconcile pre-existing zero-replica InferenceServices on startup
## Bug Description

After a metal-agent restart, InferenceServices that already had `spec.replicas=0` set before the restart keep their stale pre-restart status (often `status.readyReplicas=1` and `status.phase=Progressing`). The new metal-agent process only emits status patches when it observes a state *change*; pre-existing zero-replica state does not trigger a startup-time reconcile, so the cluster shows an InferenceService that looks deployed when no process is running.

## Steps to Reproduce

1. Have an InferenceService with `spec.replicas: 1` served by metal-agent, healthy on Apple Silicon (a `llama-server` child process is running, `status.readyReplicas: 1`, `status.phase: Ready`).

2. Scale down without stopping the agent first:
   ```
   kubectl patch inferenceservice <name> --type=merge -p '{"spec":{"replicas":0}}'
   ```
   The running metal-agent observes the spec change, calls Stop on the executor, patches status to `Stopped` with `readyReplicas: 0`. This works correctly (it's exactly what PR #498 fixed).

3. Stop the metal-agent before the in-cluster status reflects step 2 *or* simulate the pre-existing-state case some other way (e.g. apply an InferenceService with `replicas: 0` to a cluster the metal-agent has never run against, then manually pre-populate the status to `readyReplicas: 1, phase: Progressing` via `kubectl patch --subresource=status`).

4. Start the metal-agent (fresh process, fresh Kubernetes client).

5. Observe `kubectl get inferenceservice <name>` for several minutes.

## Expected Behavior

On startup, the metal-agent should sweep all watched InferenceServices and reconcile each against ground truth:

- If `spec.replicas == 0` AND `status.readyReplicas != 0` → patch the status (`readyReplicas: 0`, `phase: Stopped`).
- If `spec.replicas > 0` AND no `llama-server` process is running for it → spawn it (this case already works).
- If a `llama-server` child process is supervising a service that no longer exists in the cluster → reap it (orphaned-process cleanup).

The startup reconcile should converge each service's `status` to match observed reality, then the existing event-driven loop takes over.

## Actual Behavior

The metal-agent's initial pass only registers the watcher and emits status updates on subsequent change events. Pre-existing zero-replica state from a previous agent run keeps its stale status indefinitely (until a user manually edits the spec, an unrelated event nudges the controller, or someone deletes the resource).

In `kubectl get`, this surfaces as a confusing `REPLICAS: 1` column on an InferenceService whose `spec.replicas` is `0` and whose backing `llama-server` process does not exist:

```
NAME              PHASE         MODEL          REPLICAS   ENDPOINT
minimax-opencode  Progressing   minimax-m2-7   1          http://minimax-opencode.default.svc.cluster.local:8080/...
```

(The `REPLICAS: 1` reads from `status.readyReplicas`, which is stale; the kubectl listing has no easy way to surface the `spec.replicas: 0` mismatch.)

## Environment

**LLMKube Version:** HEAD of `main` (post-#501, post-#504; `feat/foreman-scheduler` had the same behavior before merge)

**Cluster Type:**
- [x] kind  (`kind-llmkube-local` on the M5 Max for local dev)
- Also expected on `microk8s` / production clusters; same code path

**GPU:** Apple Silicon (M5 Max, Metal backend). Bug is in the metal-agent path; the in-cluster operator path is fine (the operator is event-driven on Pods / Deployments which it owns).

## Additional Context

Discovered while swapping out the MiniMax M2.7 InferenceService for the Carnice-Qwen3.6-MoE-35B-A3B-APEX-MTP setup on the M5 Max. Sequence that exposed it:

1. The previous metal-agent was supervising `minimax-opencode` (real `llama-server` running, ~75 GB RSS).
2. `kubectl patch ... replicas: 0` to scale down. (metal-agent stopped llama-server; status updated correctly.)
3. SIGTERM the metal-agent (preparing to bring it back up with the new runtime binary).
4. Apply new Model + InferenceService for the Carnice setup.
5. Restart metal-agent. New agent observes the Carnice service, starts it. Correctly handles the new resource.
6. `kubectl get inferenceservices` shows the OLD `minimax-opencode` with `REPLICAS: 1`. The minimax llama-server is NOT running; the metal-agent just never patched the stale status from the previous agent's lifetime.

Related fixes that adjacently address this code path:
- #486 / #488 fixed metal-path operator status reporting for `InferenceService.spec.replicas=0` (the in-cluster operator side).
- #498 fixed the in-flight scale-down case (metal-agent observes the patch and stops).

The gap left is the **startup-time sweep**: convergence against ground truth on agent boot, independent of observed change events.

## Severity

- [x] **Medium** - Confusing status surface; user-visible "this is deployed" claim when nothing is running. No active service impact (the actual functional state is correct), but operationally misleading and blocks clean dashboards / `kubectl get` triage.

## Proposed fix

In `cmd/metal-agent/main.go` or `pkg/agent/agent.go`, after the watcher initializes but before entering the event loop, perform a one-shot reconcile pass:

```go
// On startup, sweep current InferenceServices to converge their status
// against observed reality. Required because the watcher only fires on
// CHANGES; pre-existing zero-replica state would otherwise keep stale
// readyReplicas / phase from the previous agent's lifetime.
func (a *MetalAgent) reconcileOnStartup(ctx context.Context) error {
    var list inferencev1alpha1.InferenceServiceList
    if err := a.kc.List(ctx, &list, client.InNamespace(a.namespace)); err != nil {
        return err
    }
    for i := range list.Items {
        svc := &list.Items[i]
        if err := a.reconcileOne(ctx, svc); err != nil {
            // log + continue; one bad service shouldn't block the rest
        }
    }
    return nil
}
```

`reconcileOne` is mostly the same logic the event loop uses; refactoring the existing handler to be callable both from the watcher and from the startup sweep is the bulk of the change.

Probably an afternoon of work + tests for the orphaned-process case.

Defilan:
> **🤖 Foreman-authored PR.** Branch produced by a Foreman pipeline on 2026-05-25: Qwen 3.6 35B-A3B coder + gate Job + M5-lite reviewer. Marked `[WIP]` for human review.

## What

Adds a `reconcileOnStartup` sweep to the metal-agent that lists all watched `InferenceService` resources at boot and calls `ensureProcess` on each, so stale `replicas=0` status from a previous metal-agent shutdown is converged against ground truth on restart.

## Why

Fixes #506

If the metal-agent shuts down while an `InferenceService` has `replicas=0` in status (because the model was being unloaded), the next boot won't trigger a reconcile until the watcher emits an event. The status drifts from reality until something else nudges it.

## How

- `pkg/agent/agent.go`: new `reconcileOnStartup` method (+31 lines). Runs synchronously in `Start()` before watcher goroutines start, so no concurrency issues with `ensureProcess`'s starting mutex.
- `pkg/agent/agent_test.go`: two new regression tests (+95 lines).
  - `TestReconcileOnStartup_CorrectsStaleStatus`: verifies stale `replicas=0` is corrected on boot.
  - `TestReconcileOnStartup_SkipsNonZeroReplicas`: verifies errors on individual services don't abort the sweep.

## M5-lite reviewer summary

**Verdict: APPROVE.** riskLevel: low. testsAdded: 2.

Quoted issue ask: *"On startup, the metal-agent should sweep all watched InferenceServices and reconcile each against ground truth."*

Findings (all `low` severity):
- **scope**: diff exactly matches the issue ask; commit message correctly references `Fixes #506`.
- **minimalism**: only 2 files touched, +31/-0 in `agent.go`, +95/-0 in `agent_test.go`. No generated files modified. Style matches surrounding code.
- **tests**: both new tests assert real behavior (stale status correction, error isolation) rather than tautological constants. All existing tests pass.
- **side-effects**: reviewer verified `reconcileOnStartup` runs synchronously in `Start()` before watcher goroutines, so no concurrency issues with `ensureProcess`'s starting mutex. Post-convergence the watcher sees updated status but won't re-trigger.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (verified by Foreman gate Job)
- [x] `make lint` passes locally (verified by Foreman gate Job)
- [x] Documentation updated (n/a, internal behavior)

## Provenance

- Pipeline: Foreman v0.1 Workload `memorial-day-v5` on shadowstack (2026-05-25)
- Coder: `qwen36-35b-carnice-mtp-coder` Agent on `m5max-coder` FleetNode (4m 55s wall, GO verdict)
- Gate: `gate` Agent on shadowstack (Job-based, GATE-PASS)
- Reviewer: `qwen36-35b-a3b-reviewer` Agent (M5-lite, 84s wall, APPROVE verdict)

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -367,6 +367,15 @@ func (a *MetalAgent) Start(ctx context.Context) error {
 		a.logger.Infow("cleaned up orphaned endpoints from prior sessions", "count", cleaned)
 	}
 
+	// Startup reconcile: sweep all InferenceServices in the watched namespace
+	// and converge their status against observed reality. The watcher only
+	// fires on CHANGES, so pre-existing zero-replica state from a previous
+	// agent run would otherwise keep stale readyReplicas / phase indefinitely.
+	// See https://github.com/defilantech/LLMKube/issues/506.
+	if err := a.reconcileOnStartup(ctx); err != nil {
+		a.logger.Warnw("startup reconcile failed", "error", err)
+	}
+
 	// Start health server. An unexpected exit here (port binding lost,
 	// listener crashed) is fatal — the management plane is how operators
 	// observe and recover the agent, so running blind is worse than
@@ -967,6 +976,28 @@ func (a *MetalAgent) markStopped(ctx context.Context, isvc *inferencev1alpha1.In
 // scheduleRestart increments the restart counter and re-runs ensureProcess
 // for the named InferenceService. It is called by HealthMonitor when a process
 // becomes unhealthy.
+
+// reconcileOnStartup lists all InferenceServices in the watched namespace and
+// calls ensureProcess for each one. This converges their status against
+// observed reality on agent boot, because the watcher only fires on CHANGES
+// and pre-existing zero-replica state from a previous agent run would
+// otherwise keep stale readyReplicas / phase indefinitely.
+// See https://github.com/defilantech/LLMKube/issues/506.
+func (a *MetalAgent) reconcileOnStartup(ctx context.Context) error {
+	var list inferencev1alpha1.InferenceServiceList
+	if err := a.config.K8sClient.List(ctx, &list, client.InNamespace(a.config.Namespace)); err != nil {
+		return fmt.Errorf("list InferenceServices for startup reconcile: %w", err)
+	}
+	for i := range list.Items {
+		svc := &list.Items[i]
+		if err := a.ensureProcess(ctx, svc); err != nil {
+			a.logger.Warnw("startup reconcile failed for InferenceService",
+				"namespace", svc.Namespace, "name", svc.Name, "error", err)
+		}
+	}
+	return nil
+}
+
 // runWatcherLoop drives a.watcher.Watch in a loop, retrying transient errors
 // with exponential backoff (handles the "CRDs not installed yet" startup
 // race) but bubbling ErrWatchStalled up via fatalErrChan immediately.
--- a/pkg/agent/agent_test.go
+++ b/pkg/agent/agent_test.go
@@ -1061,3 +1061,98 @@ func TestEnsureProcess_InFlightGuard(t *testing.T) {
 		t.Errorf("starting map not cleared after spawn, has %d entries", leftover)
 	}
 }
+
+// TestReconcileOnStartup_CorrectsStaleStatus verifies that reconcileOnStartup
+// patches the status of InferenceServices whose spec.replicas=0 but whose
+// status still shows ReadyReplicas > 0 from a prior agent run. See #506.
+func TestReconcileOnStartup_CorrectsStaleStatus(t *testing.T) {
+	scheme := newTestScheme()
+	_ = corev1.AddToScheme(scheme)
+
+	zero := int32(0)
+	stale := &inferencev1alpha1.InferenceService{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:       "stale-service",
+			Namespace:  "default",
+			Generation: 5,
+		},
+		Spec: inferencev1alpha1.InferenceServiceSpec{
+			ModelRef: "any-model",
+			Replicas: &zero,
+		},
+		Status: inferencev1alpha1.InferenceServiceStatus{
+			Phase:           "Ready",
+			ReadyReplicas:   1,
+			DesiredReplicas: 1,
+		},
+	}
+
+	k8sClient := fake.NewClientBuilder().
+		WithScheme(scheme).
+		WithStatusSubresource(&inferencev1alpha1.InferenceService{}).
+		WithRuntimeObjects(stale).
+		Build()
+
+	agent := NewMetalAgent(MetalAgentConfig{K8sClient: k8sClient, Namespace: "default"})
+	agent.executor = NewMetalExecutor("/fake/llama-server", "/tmp/models", newNopLogger())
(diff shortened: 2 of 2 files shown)


Defilan:
Closing this PR. It was authored by the Foreman v0.1 pipeline during the Memorial Day v5 batch and passed the v0.1 gate (all `make` targets clean).

A subsequent human review identified a cross-file correctness bug the v0.1 reviewer step did not catch: the new `reconcileOnStartup` sweep iterates every `InferenceService` in the namespace and calls `ensureProcess` on each, while the existing watcher (`pkg/agent/watcher.go::shouldWatch`) filters to metal-backed services before reconciling. In a mixed-fleet namespace (metal-agent-managed plus operator-Pod-managed services sharing one namespace) the new sweep would call `markStopped` on non-metal services with `replicas=0` and silently overwrite status the operator owns.

For Chris's all-metal clusters this is harmless; in any shared-fleet namespace it's a real bug. The v0.1 reviewer did not detect the missing filter because cross-file invariant checking requires reasoning about how `pkg/agent/watcher.go` interacts with the new code in `pkg/agent/agent.go`, which the 24-26B-class local reviewer demonstrably cannot do reliably.

We're treating this as empirical evidence that the v0.1 reviewer architecture is insufficient for structural-bug detection and rebuilding it for v0.2; see #546 for the plan. Issue #506 remains open and will be re-attempted under the v0.2 architecture.

PR #557, opened 2026-05-26
feat(gpu): add Intel GPU support (oneAPI/SYCL path) across controller, CLI, docs, and e2e

adebrie:
## What

- Adds Intel GPU support in scheduling and deployment generation.
- Adds Intel accelerator support to Model API and CRD schema.
- Adds Intel-aware runtime image selection in CLI deploy and benchmark flows.
- Adds Intel quickstart documentation and updates GPU setup guide/README.
- Expands tests for Intel GPU resource resolution and deployment behavior.
- Improves e2e GPU script to auto-detect NVIDIA vs Intel clusters and validate the matching resource key.

## Why

Issue #4 requests Intel GPU support (Arc/Data Center) with oneAPI/SYCL while keeping compatibility with the existing GPU CRD model.

### Controller:

- Detects Intel from either hardware.accelerator=intel or hardware.gpu.vendor=intel.
- Resolves Intel resource key to gpu.intel.com/i915 by default.
- Supports overriding Intel resource key with LLMKUBE_INTEL_GPU_RESOURCE (for xe or custom setups).
- Uses resolved GPU resource key for limits, tolerations, and insufficient-resource detection.

### API/CRD:

- Extends accelerator enum to include intel.
- Regenerates CRD schema so Intel accelerator is accepted by validation.

### CLI:

- Uses Intel llama.cpp image defaults for Intel accelerator paths.
- Keeps benchmark catalog and deploy behavior consistent with Intel runtime selection.

### Docs:

- Adds Intel GPU quickstart covering:

  - Intel device plugin expectations

  - i915 vs xe key handling

  - controller env override

  - SYCL/offload verification steps

  - throughput measurement

- Tests:

  - Adds/updates unit tests for Intel resource name resolution and scheduling event parsing.
  - Adds/updates deployment tests for Intel GPU limits/tolerations.
  - Updates e2e GPU script to validate Intel and NVIDIA paths.

## Checklist

- [x] Tests added/updated


config/crd/bases/inference.llmkube.dev_models.yaml
```diff
@@ -82,6 +82,7 @@ spec:
                     - metal
                     - cuda
                     - rocm
+                    - intel
```
Defilan, line 85: The matching helm chart CRD copy at `charts/llmkube/templates/crds/models.yaml` still has the old enum (`cpu, metal, cuda, rocm`). The repo's **CRD Sync Check** job fails because of this drift. Please run `make chart-crds` and commit the regenerated file. Otherwise helm-installed clusters will reject `accelerator: intel` even though the kustomize-installed CRD accepts it.

internal/controller/gpu_resources.go
```diff
@@ -0,0 +1,66 @@
...
+	"os"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+
+	inferencev1alpha1 "github.com/defilantech/llmkube/api/v1alpha1"
+)
+
+const (
+	acceleratorIntel       = "intel"
```
Defilan: `defaultInsufficientGPUHint` is never reassigned, so it should live in the `const (...)` block alongside `acceleratorIntel` and `intelGPUResourceEnvVar` rather than in `var (...)`.

test/e2e/gpu_test.sh
```diff
@@ -36,34 +42,68 @@ cleanup() {
...
 trap cleanup EXIT
 
 # Test 1: Deploy GPU Model
 test_deploy_gpu_model() {
     log_info "Test 1: Deploying GPU model..."
 
-    cat <<EOF | kubectl apply -f -
+        cat <<EOF | kubectl apply -f -
 apiVersion: inference.llmkube.dev/v1alpha1
 kind: Model
```
Defilan, line 88: The 2-space to 4-space reindent of these YAML heredocs (here and in the InferenceService block below) is unrelated to the Intel work and inflates the diff. YAML accepts both, so functionally it's fine, but reverting the indent change would keep the diff focused on the GPU detection logic.

docs/intel-gpu-quickstart.md
```diff
@@ -0,0 +1,182 @@
...
+
+### WaitingForGPU or Insufficient GPU events
+
+- Verify node allocatable resource key matches controller configuration.
+- If nodes expose `gpu.intel.com/xe`, set `LLMKUBE_INTEL_GPU_RESOURCE=gpu.intel.com/xe`.
+- Check pod events:
+
+```bash
+kubectl describe pod -l app=intel-phi4
+```
```
Defilan: Nitpick: file is missing a trailing newline (`\ No newline at end of file` in the diff). The other docs in this directory end with one, so adding it keeps the diff-noise consistent.

Defilan (changes requested):
Hey Alex, thanks for tackling this! Intel GPU support has been on the list since #4 was filed, and this is a clean, well-scoped take on it. The `gpu_resources.go` resolver in its own file is exactly the right shape, the tests cover the interesting branches (no hardware, non-Intel vendor, accelerator=intel, env override), and the e2e auto-detect script is a nice touch.

Two small must-fixes before this can land:

1. **CRD Sync Check is red.** `charts/llmkube/templates/crds/models.yaml` still has the old accelerator enum without `intel`. The kustomize CRD at `config/crd/bases/` and the chart copy need to ship in lockstep; `make chart-crds` will regenerate it.
2. `defaultInsufficientGPUHint` in `internal/controller/gpu_resources.go` should be `const`, not `var`, since it is never reassigned.

A couple of non-blocking polish items I left as inline comments:

- The 2-space to 4-space reindent of the YAML heredocs in `gpu_test.sh` looks like an editor reformat; reverting it would keep the diff focused.
- `docs/intel-gpu-quickstart.md` is missing a trailing newline.

Once CI is green I will squash-merge. Appreciate the contribution!!

Defilan (approved):
Thanks for the quick turnaround, Alex. All four items from the prior review are addressed cleanly: CRD Sync Check is green (chart and kustomize enums match), `defaultInsufficientGPUHint` is now a `const`, the e2e heredoc indent is restored, and the docs file has its trailing newline. Nice catch on the `\u2013` literal that `make chart-crds` had introduced; that's exactly the kind of detail I'd want a contributor to notice and fix. Squash-merging now.

Welcome to LLMKube!

PR #613, opened 2026-06-02
feat(controller): add first-class whisper (speaches) audio transcription runtime

### LINKED ISSUE #612: [FEATURE] First-class whisper (speaches) audio-transcription runtime
## Feature Description

Add a first-class `whisper` runtime to the operator that serves OpenAI-compatible audio transcription inside Kubernetes, backed by [speaches](https://speaches.ai) (the faster-whisper successor, CTranslate2). An `InferenceService` with `runtime: whisper` would deploy a transcription service reachable at a ClusterIP on `/v1/audio/transcriptions`, with the same lifecycle, scaling, GPU scheduling, and air-gapped story as every other LLMKube workload.

## Problem Statement

LLMKube today only serves text generation (llama.cpp / vLLM / TGI / generic / personaplex). Speech-to-text is a common on-prem need (meeting transcription, call analysis, accessibility, pipeline preprocessing) and partners integrating against LLMKube's OpenAI-compatible endpoints want audio too.

> As a platform operator, I want to deploy a Whisper transcription service via an `InferenceService` so that in-cluster workloads can POST audio to an OpenAI-compatible `/v1/audio/transcriptions` endpoint without standing up bespoke infrastructure.

## Proposed Solution

A new `WhisperBackend` runtime backed by `ghcr.io/speaches-ai/speaches`. speaches speaks the OpenAI audio API natively on port 8000, is configured by environment variables, fetches CTranslate2 models from HuggingFace at runtime, and ships a `/health` endpoint. The operator's architecture already accommodates this: the GGUF parser is architecture-tolerant, HF-repo model sources are handled without a controller download, and the endpoint path is parameterizable.

**Example YAML:**
```yaml
kind: Model
metadata:
  name: whisper-large-v3
spec:
  source: Systran/faster-whisper-large-v3
  format: custom
  hardware:
    accelerator: cuda
    gpu:
      enabled: true
      count: 1
---
kind: InferenceService
metadata:
  name: whisper
spec:
  modelRef: whisper-large-v3
  runtime: whisper
  whisperConfig:
    computeType: float16
    task: transcribe
  # endpoint defaults to /v1/audio/transcriptions for the whisper runtime
```

**Implementation outline:**
- New `WhisperBackend` (`internal/controller/runtime_whisper.go`): port 8000, `/health` probes, `NeedsModelInit=false` (speaches fetches from HF at runtime), env-driven config via `BuildEnv`.
- Widen the optional `EnvBuilder` interface to `BuildEnv(isvc, model)` so env values can derive from the Model CR (source, accelerator, quantization).
- Add an optional `EndpointPathProvider` interface and drop the `EndpointSpec.Path` CRD default so the whisper runtime resolves `/v1/audio/transcriptions` automatically while existing runtimes stay on `/v1/chat/completions` (backward compatible).
- New typed `WhisperConfig` (compute type, device, language, task, HF/API token secret refs), modeled on `TGIConfig`.
- Add `whisper` to the `Runtime` enum; regenerate CRDs + chart CRDs.
- Unit + reconcile tests, `examples/whisper-quickstart/`, and a `docs/contributors/adding-a-runtime.md` table update.

## Alternatives Considered

- **whisper.cpp (GGUF):** strongest fit with the GGUF/Metal story and air-gapped narrative, but newer OpenAI-endpoint support and no metrics. Good candidate for a second engine later.
- **vLLM-Whisper:** reuses the existing `vllm` runtime with the least new code, but safetensors/GPU-heavy and least edge friendly.
- **Generic runtime + docs only:** works today but is not first-class and is a footgun (manual endpoint path, no typed config).

speaches was chosen for the first cut because it speaks the OpenAI audio API natively, which aligns with LLMKube's existing OpenAI-compatible integration surface.

## Additional Context

- Out of scope for v1 (follow-ups): persistent model cache (speaches re-downloads on pod restart since there is no volume hook on the runtime interface yet, tracked separately as a `VolumeProvider` interface), and Prometheus metrics (speaches exposes none; the universal PodMonitor will benignly 404-scrape whisper pods).
- Engine: faster-whisper / speaches.

## Priority

- [ ] Medium - Nice to have
- [ ] Low - Minor improvement

## Willingness to Contribute

- [x] Yes, I can submit a PR
- [ ] Yes, I can help test
- [ ] No, just requesting

Defilan:
## What

Add a first-class `whisper` runtime to the operator that serves OpenAI-compatible audio transcription, backed by [speaches](https://speaches.ai) (faster-whisper / CTranslate2). An `InferenceService` with `runtime: whisper` deploys a transcription service reachable at a ClusterIP on `/v1/audio/transcriptions`, with the same lifecycle, scaling, GPU scheduling, and probe story as every other runtime.

## Why

LLMKube only served text generation. Speech-to-text is a common on-prem need (meeting transcription, call analysis, pipeline preprocessing) and partners integrating against the OpenAI-compatible surface want audio too. speaches was chosen because it speaks the OpenAI audio API natively.

Fixes #612

## How

- **`WhisperBackend`** (`internal/controller/runtime_whisper.go`): container `speaches`, port 8000, HTTP `/health` probes, `NeedsModelInit=false` (speaches fetches CTranslate2 models from HuggingFace at request time), config via env vars rather than CLI flags.
- **Widened the optional `EnvBuilder` interface** to `BuildEnv(isvc, model)` so a backend can derive env values from the Model spec (device from `hardware.accelerator`, compute type from `quantization`). Updated the vllm/tgi/personaplex implementers and the single `deployment_builder` call site; they ignore the new arg.
- **Runtime-aware endpoint path**: added an optional `EndpointPathProvider` interface (whisper returns `/v1/audio/transcriptions`) and dropped the `+kubebuilder:default` on `EndpointSpec.Path`. `constructEndpoint` and `routerProxyEndpoint` already default an empty path to `/v1/chat/completions`, so existing runtimes are unchanged and this is backward compatible. A user-set `spec.endpoint.path` still wins.
- **Typed `WhisperConfig`** (compute type, inference device, model TTL, UI toggle, HF/API token secret refs), modeled on `TGIConfig`. Added `whisper` to the `Runtime` enum and regenerated CRDs + chart CRDs.
- Image pinned to `ghcr.io/speaches-ai/speaches:0.8.3-cuda` (CPU users override `spec.image` with `:0.8.3-cpu`).

**Scope (v1):** targets connected clusters; speaches downloads the model on first request. Persistent model cache + air-gapped support need a volume hook on the runtime interface and are deferred to a follow-up. speaches exposes no Prometheus metrics, so the universal PodMonitor benignly 404-scrapes these pods (`DefaultHPAMetric` returns `""`).

Verified speaches v0.8.3 upstream: `/health` healthcheck, port 8000, env names (`WHISPER__INFERENCE_DEVICE`, `WHISPER__COMPUTE_TYPE`, `WHISPER__TTL`, `ENABLE_UI`, `API_KEY`), and that there is no `WHISPER__MODEL` (models are per-request).

## Checklist

- [x] Tests added/updated
- [x] `make lint` passes locally (and `GOOS=linux golangci-lint run ./...`)


Defilan:
Live-tested on a GPU cluster (Shadowstack, RTX). Two follow-up commits push the runtime to genuinely first-class:

1. **Service/endpoint port fix.** The live deploy exposed that the Service/endpoint defaulted to 8080 while speaches listens on 8000, so in-cluster routing was broken. Added `resolveServicePort` (containerPort → endpoint.port → backend.DefaultPort), used in constructService/constructEndpoint/deployment builder. Also fixes the latent mismatch for vllm (8000) and tgi (80); llamacpp (8080) unchanged.

2. **Model preload.** speaches v0.8.3 does **not** download models on first request: it 400s with "Model is not installed locally, use POST /v1/models". Added an optional `LifecycleProvider` interface; WhisperBackend injects a postStart hook that installs `model.Spec.Source` once `/health` is up. The hook gates Ready, so the Service only takes traffic once transcription will succeed.

**End-to-end result:** deploying just a Model + InferenceService(runtime: whisper) → operator preloads large-v3, pod goes Ready, and `POST /v1/audio/transcriptions` through the ClusterIP returns the correct transcript with no manual steps. GPU confirmed in use (~3.9 GB, float16).

Docs/example updated accordingly (no endpoint block needed; operator preloads rather than lazy-download). Persistent model cache + air-gapped remain the tracked follow-up.

Defilan:
Closing to match the roadmap triage: whisper is cut from the active roadmap (see #612). The branch is preserved if we revisit whisper as an optional sidecar runtime later.

PR #689, opened 2026-06-14
ci: pin GitHub Actions to exact commit SHAs

matiasinsaurralde:
Pin all third-party actions to full commit SHAs instead of floating major-version tags in the lint, test, and test-e2e workflows to harden the CI supply chain against upstream tag-mutation attacks. Other workflows will be updated in a follow-up.

## What
Pin all third-party GitHub Actions to immutable commit SHAs in lint, test, and test-e2e workflows

## Why
Floating major-version tags (e.g. `@v6`) are mutable — a compromised upstream repo could swap the tag to inject malicious code into our CI pipeline

## How
Replace each `uses: action@vN` reference with `uses: action@<full-sha> # vX.Y.Z`, preserving the version in a trailing comment for readability

## Checklist

- [x] Tests added/updated

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,14 +12,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Clone the code
-        uses: actions/checkout@v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
       - name: Setup Go
-        uses: actions/setup-go@v6
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version-file: go.mod
 
       - name: Run linter
-        uses: golangci/golangci-lint-action@v9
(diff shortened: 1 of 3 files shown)


Defilan (commented):
Thanks for this, pinning to SHAs is a great hardening step and the pins all check out (verified each against upstream). One miss: `test-e2e.yml:182` still has `azure/setup-helm@v5` on a floating tag. Pin it to the same SHA you used above (`dda3372f752e03dde6b3237bc9431cdc2f7a02a2` # v5.0.0) and this is good to go. Appreciate the follow-up plan for the other workflows too!

matiasinsaurralde:
> Thanks for this, pinning to SHAs is a great hardening step and the pins all check out (verified each against upstream). One miss: `test-e2e.yml:182` still has `azure/setup-helm@v5` on a floating tag. Pin it to the same SHA you used above (`dda3372f752e03dde6b3237bc9431cdc2f7a02a2` # v5.0.0) and this is good to go. Appreciate the follow-up plan for the other workflows too!

Great Chris, I think I've missed that one during a conflict resolution. Just fixed it!

Defilan (approved):
This is great. Thanks again for your help @matiasinsaurralde !

Issue #701, opened 2026-06-15
[FEATURE] ROCm 7.2 HIP runtime tier for AMD nodes (per-model opt-in, follow-on)


Defilan:
## Feature Description

A ROCm 7.2 (HIP) runtime tier for AMD nodes, as a per-model opt-in alongside the Vulkan default. This is the "ROCm proper" follow-on under the AMD epic (#696): for the specific models where ROCm beats Vulkan, allow an `InferenceService` to select the ROCm runtime image.

## Problem Statement

> As a fleet operator, I want to run the models that perform better under ROCm/HIP on AMD via the same CRDs, so that I am not locked to Vulkan when ROCm wins.

Vulkan is the right v1 default on `gfx1151` (faster generation, far more stable), but ROCm 7.x can edge it on some small-dense and MoE workloads with a fully-built HIP stack (rocWMMA + hipBLASLt). Once the Vulkan tier is solid, ROCm is worth offering for those cases.

## Proposed Solution

- A ROCm `llama-server` image built from `rocm/dev-ubuntu-24.04:7.2-complete` with `-DGPU_TARGETS=gfx1151 -DGGML_HIP_ROCWMMA_FATTN=ON -DGGML_HIP_NO_VMM=ON -DGGML_HIP_MMQ_MFMA=ON` (or COPY prebuilt `lemonade-sdk/llamacpp-rocm` gfx1151 binaries to avoid long HIP builds).
- Pod mounts `/dev/dri/render*` + `/dev/kfd` via the device plugin (#395); the ROCm node prerequisites extend the node runbook.
- Per-model/per-InferenceService runtime selection so an operator opts a specific model into ROCm.
- Re-benchmark ROCm vs Vulkan per model and document which wins where; track ROCm release notes for when `gfx1151` leaves Preview and the `hipGraph`/KV-cache bugs are fixed.

## Alternatives Considered

- **Make ROCm the default:** rejected for now given `gfx1151` instability (dense-model crashes, KV-cache-to-host). Vulkan stays the default; ROCm is opt-in.

## Additional Context

- Related issues: #696 (epic), Vulkan runtime image, node runbook, #395
- This is explicitly a follow-on: do not start until the Vulkan tier (image + node + example) is working in the lab.

## Priority

- [x] Medium - Nice to have

## Willingness to Contribute

- [x] Yes, I can submit a PR

Issue #697, opened 2026-06-15
[FEATURE] AMD llama.cpp runtime image (Vulkan/RADV)


Defilan:
## Feature Description

Add an AMD-capable `llama-server` container image to the runtime build matrix, using the **Vulkan/RADV** backend. Today the operator ships a CUDA image (in-cluster) and a native Metal binary (off-cluster agent); there is no image an AMD node can run. This is the load-bearing gap for the AMD tier (#696): without it, the scheduling work in #395 has nothing to schedule.

## Problem Statement

> As a fleet operator with an AMD node, I want an `llama-server` image that uses the GPU via Vulkan, so that an `InferenceService` scheduled onto that node actually offloads to the iGPU.

On `gfx1151` (Strix Halo), Vulkan/RADV is the pragmatic backend: faster generation than ROCm, stable across dense and MoE models, no ROCm install, and a much smaller image. It also covers AMD discrete GPUs, Intel iGPUs, and other `/dev/dri` devices through one image.

## Proposed Solution

- A new Dockerfile target (e.g. `.devops`/CI matrix entry) building llama.cpp with `-DGGML_VULKAN=ON` on a slim Debian/Ubuntu base with `mesa-vulkan-drivers` + `libvulkan1` (RADV). No ROCm layers.
- Publish to the same registry/tag scheme as the CUDA image, with an `-vulkan` (or `-amd`) suffix.
- The pod consumes the GPU by mounting `/dev/dri/render*` via the device-plugin resource from #395 (no `nvidia.com/gpu` request).
- Verify `--metrics` and the three-probe pattern work unchanged in this image.

**Example (InferenceService targeting the AMD node):**
```yaml
apiVersion: inference.llmkube.dev/v1alpha1
kind: InferenceService
metadata:
  name: qwen3-amd
spec:
  modelRef: qwen3-30b
  runtimeImage: ghcr.io/defilantech/llmkube-llama-vulkan:<tag>  # or selected by vendor=amd
  hardware:
    gpu:
      vendor: amd
      resourceName: squat.ai/dri-render   # from #395
```

## Alternatives Considered

- **ROCm/HIP image (`rocm/dev-ubuntu-24.04:7.2-complete`):** the follow-on tier, tracked separately under #696. 6-12GB image and `gfx1151`-fragile today; deferred.
- **Prebuilt lemonade `llamacpp-rocm` binaries:** an option for the ROCm tier to avoid long HIP builds; not needed for the Vulkan v1.

[...]

PR #724, opened 2026-06-16
docs: add Strix Halo node onboarding quickstart

### LINKED ISSUE #698: [FEATURE] AMD/Strix Halo Kubernetes node enablement runbook (kernel, GTT, /dev/dri device plugin)
## Feature Description

A runbook for onboarding an AMD node (initially Strix Halo / Ryzen AI Max+ 395, `gfx1151`) into an LLMKube cluster: kernel and driver prerequisites, unified-memory (GTT) tuning, and the `squat/generic-device-plugin` DaemonSet that advertises `/dev/dri/renderD128` as a schedulable resource. This is the "how do I plug in an AMD box" doc, analogous to the NVIDIA GPU Operator setup docs.

## Problem Statement

> As an operator adding an AMD node, I want a step-by-step that gets the iGPU visible to Kubernetes with enough memory to serve real models, so that LLMKube can schedule onto it.

The Strix Halo iGPU does not expose its unified memory to the GPU by default, and the official `amd.com/gpu` device plugin is Instinct-oriented and does not cleanly enumerate the APU. There is real node-level setup that has to be documented or every AMD onboarding rediscovers it.

## Proposed Solution

A docs page (and example manifests) covering:

- **Kernel + driver:** kernel >= 6.15 (prefer 6.19), in-tree `amdgpu` (avoid DKMS).
- **Unified-memory / GTT kernel args** to unlock ~90GB to the iGPU:
  `amdgpu.gttsize=126976 ttm.pages_limit=32505856 ttm.page_pool_size=25165824 amdgpu.vm_fragment_size=8 amd_iommu=off`
  (set a small fixed BIOS UMA and let GTT do the work).
- **Device plugin:** a `squat/generic-device-plugin` DaemonSet advertising `/dev/dri/renderD128` (mount `/dev/kfd` too if the ROCm tier is also used). Example DaemonSet manifest checked into `deployment/`.
- **Verification:** confirm the node advertises the resource, schedule a probe pod, confirm the Vulkan image sees the GPU and the memory ceiling.

## Alternatives Considered

- **Official AMD GPU Operator + `amd.com/gpu`:** works on the iGPU only via custom NFD-by-kernel-module rules and a ROCm stack on the node (the Talos pattern). Heavier and ROCm-coupled; revisit when adding Instinct/MI hardware. Noted in the doc as the "uniform-resource-name" alternative.

## Additional Context

- Related issues: #696 (epic), #395 (scheduling resource name), AMD Vulkan image issue
- The GTT args and ~90GB visible pool are corroborated by the llama.cpp Strix Halo known-good-stack discussion and a working Talos+K8s writeup.

## Priority

## Willingness to Contribute

- [x] Yes, I can submit a PR

joryirving:
Closes #698

## Summary

Adds a Strix Halo (AMD Ryzen AI Max+ 395, gfx1151) onboarding quickstart covering:

- Host prerequisites: kernel ≥ 6.15, `amdgpu` module, GTT kernel args (`amdgpu.gttsize=126976`, `ttm.pages_limit=32505856`, etc.)
- Two device-plugin paths:
  - **Primary**: ROCm GPU Operator (per Talos docs and strixhalo.wiki), with the Strix-Halo-specific NFD rule (kernel-module-based, since default NFD misses the iGPU)
  - **Alternative**: upstream `rocm/k8s-device-plugin` DaemonSet for a lighter setup (NFD still recommended for node labeling)
- Verify GPU resources advertise `amd.com/gpu`
- Deploy a Model + InferenceService (`accelerator: amd`, `vendor: amd`)
- Verify scheduling, GPU offload, and token throughput
- Troubleshooting for common issues (renderD missing, NFD label missing, GTT ceiling, OOM under load)

References the Talos AMD GPU guide, strixhalo.wiki Kubernetes guide, and kyuz0's amd-strix-halo-toolboxes for upstream context.

### DIFF
--- a/docs/strix-halo-onboarding.md
+++ b/docs/strix-halo-onboarding.md
@@ -1,0 +1,318 @@
+# Strix Halo Quickstart
+
+This guide walks through onboarding an AMD Strix Halo (Ryzen AI Max+ 395, gfx1151) node into an LLMKube cluster and running an LLM on the integrated GPU.
+
+## Prerequisites
+
+- Strix Halo node with at least 4 GB UMA reservation in BIOS
+- Linux kernel ≥ 6.15 (6.18.4+ recommended for gfx1151 stability)
+- `amdgpu` kernel module loaded
+- LLMKube operator installed
+- kubectl configured for your cluster
+
+## Step 1: Verify Host Prerequisites
+
+Check the kernel version and that `amdgpu` is loaded:
+
+```bash
+uname -r
+lsmod | grep amdgpu
+ls -la /dev/dri/
+```
+
+Expected: kernel ≥ 6.15, `amdgpu` listed, and at least `renderD128` present.
+
+If kernel args aren't yet applied, add them to your bootloader. Strix Halo requires:
+
+```
+amd_iommu=off
+amdgpu.gttsize=126976
+ttm.pages_limit=32505856
+ttm.page_pool_size=25165824
+amdgpu.vm_fragment_size=8
+```
+
+Reboot after applying. Confirm they're live:
+
+```bash
+cat /proc/cmdline | tr ' ' '\n' | grep -E 'gttsize|ttm|amd_iommu'
+```
+
+For Talos nodes, set these in the machine config `customization.extraKernelArgs` and include the `siderolabs/amdgpu` and `siderolabs/amd-ucode` system extensions. See the [Talos AMD GPU guide](https://docs.siderolabs.com/talos/v1.13/configure-your-talos-cluster/hardware-and-drivers/amd-gpu) for the full Talos recipe.
+
+## Step 2: Install the ROCm GPU Operator
+
+The [ROCm GPU Operator](https://instinct.docs.amd.com/projects/gpu-operator/en/latest/index.html) is the recommended way to expose AMD GPUs to Kubernetes. It bundles the device plugin, Node Feature Discovery rules, and (optionally) a metrics exporter.
+
+Add the Helm repo and install:
+
+```bash
+helm repo add rocm https://rocm.github.io/gpu-operator
+helm repo update
+helm install rocm-gpu-operator rocm/gpu-operator \
+  --namespace kube-system --create-namespace
+```
+
+Strix Halo is integrated, not discrete, so disable the operator's in-cluster driver loading (the kernel module is already loaded by the host):
+
+```yaml
+# values.yaml
+kmm:
+  enabled: true
+installdefaultNFDRule: true
+deviceConfig:
+  spec:
+    driver:
+      enable: false
+      blacklist: false
+metricsExporter:
+  enable: true
+  prometheus:
+    serviceMonitor:
+      enable: true
+      interval: 30s
+```
+
+For Strix Halo, the default NFD rules detect the iGPU by device ID and miss it. Add a kernel-module-based rule instead:
+
+```yaml
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureRule
+metadata:
+  name: amdgpu-rules
+spec:
+  rules:
+    - name: "amdgpu kernel module"
+      labels:
+        feature.node.kubernetes.io/amd-gpu: "true"
+      matchFeatures:
+        - feature: "kernel.loadedmodule"
+          matchExpressions:
+            amdgpu: {op: Exists}
+```
+
+This labels any node with the `amdgpu` module loaded with `feature.node.kubernetes.io/amd-gpu: "true"`, which the ROCm operator matches on.
+
+For a deeper Strix Halo walkthrough including a PyTorch verification job, see the [strixhalo.wiki Kubernetes guide](https://strixhalo.wiki/Guides/Kubernetes).
+
+### Alternative: rocm/k8s-device-plugin
+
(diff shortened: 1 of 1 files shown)


Defilan:
This is fantastic @joryirving ! I am actually setting mine up as we speak and I'm going to test this guide out. Appreciate you doing this. Will merge later tonight.

Defilan:
Strong, useful guide, thanks for this. The host-tuning sections are the reusable gold: the GTT kernel args, the NFD-by-kernel-module rule (the default device-ID rule really does miss the iGPU), the troubleshooting list, and especially the GTT-pins-unreclaimable-RAM OOM note.

One genuine question before I suggest anything: **did you get the ROCm `amd.com/gpu` path serving end-to-end on your Strix Halo box?** If so I'd love the details: which serving image (a ROCm/HIP llama.cpp build, or a Vulkan binary in a ROCm-provisioned pod?), kernel version, and rough tokens/sec. gfx1151 ROCm has been fragile in our testing, so a working recipe is genuinely valuable and would change how we frame the tiers.

Context on why I ask: LLMKube's AMD tier (#697) is currently **Vulkan/RADV-first**, and the Vulkan runtime consumes `/dev/dri` via a generic device plugin rather than `amd.com/gpu`. Your Step 5 already expects `Vulkan0` log markers, so the serving path looks like Vulkan even though the device-plugin section provisions `amd.com/gpu`.

Rather than re-architecting your PR, I'd suggest **adding a Vulkan + generic-device-plugin section alongside** the ROCm one so the doc covers both paths (ROCm tracked under #701, Vulkan under #697). I just validated the Vulkan path end-to-end on a Strix Halo node (Llama-3.2-3B, 29/29 layers on `RADV GFX1151`, ~87 tok/s decode) and am happy to share manifests for that section.

Three concrete things that bit me, useful either way:
- **Pin the image.** The floating `ghcr.io/ggml-org/llama.cpp:server-vulkan` tag is currently broken (undefined `libggml-vulkan.so` symbol, silent CPU fallback): #725. `server-vulkan-b5974` works.
- The generic-device-plugin default resource domain is `devic.es/dri-render` now (not `squat.ai`).
- The device group needs **both `/dev/dri/renderD128` and `/dev/dri/card1`**: renderD128 alone and RADV won't initialize. Worth checking whether `rocm/k8s-device-plugin` injects `card1`; if it only gives renderD128, a Vulkan container won't light up even with `amd.com/gpu`.

joryirving:
I have ROCm serving on my box. I'm running 4 concurrent models, all ROCM-6.4.4. I very specifically only use the kyuz0 toolboxes. I specifically need to run them privileged, and I don't use the operator, only the AMD plugin.

Qwen3.6-35B-A3B example: https://github.com/joryirving/home-ops/blob/5521252e7d81b99cb5917a06e4465e718c321a96/kubernetes/apps/base/llm/litellm/Qwen3.6-35B-A3B/helmrelease.yaml
Gemma-4-26B-A4B example: https://github.com/joryirving/home-ops/blob/main/kubernetes/apps/base/llm/litellm/Gemma4-26B-A4B/helmrelease.yaml

Defilan:
@joryirving this is great work, and your reply confirming ROCm running 4 concurrent MoE models on gfx1151 is genuinely valuable. Thank you. The node-enablement parts of this doc (kernel >= 6.15, the GTT args, the NFD-by-kernel-module rule, the device plugin, and especially the GTT-pins-unreclaimable-RAM OOM note) are the hard-won gold and I'd love to land them.

One thing changed since you opened this: we just shipped the **operator-integrated AMD/Vulkan tier** (#727 + #729). `vendor: amd` + `runtime: vulkan` now makes the operator pull our own `llmkube-llama-vulkan` image, request `devic.es/dri-render`, and (with per-node caches) schedule and serve on the GPU node. Validated end-to-end on a real gfx1151 box at ~310 tok/s. So Vulkan is now the path that works out of the box via the operator's `Model`/`InferenceService`.

ROCm is still very much wanted, but the *operator-integrated* ROCm runtime is tracked as #701 and isn't built yet: the operator has no ROCm image to schedule. Your working setup (kyuz0 toolboxes + the AMD plugin, privileged) is the manual path, which is great but distinct from a `kubectl apply` of an LLMKube `Model`.

Proposed reshape, keeping your contribution front and center:

1. **Keep your node-enablement sections as-is.** They apply to both tiers.
2. **Lead the deploy section with the Vulkan path** (operator default, works today). Happy to push the validated `Model` + `InferenceService` (`runtime: vulkan`, `supplementalGroups`) manifests to your branch, or pair on it.
3. **Reframe the ROCm deploy as "advanced / manual, pending #701,"** citing your kyuz0/litellm recipe as the reference. It becomes the operator path once #701 lands.
4. Small fix: `hardware.accelerator: amd` is rejected by the CRD enum (`cpu`, `metal`, `cuda`, `rocm`, `intel`); drop it (the Vulkan path keys off `gpu.vendor` + `gpu.runtime`). Tracked in #730.

Want me to push the Vulkan section to your branch so we can land this together?

joryirving:
Yep, that sounds good to me, let's do that.

joryirving:
FYI, this is me migrating onto LLMKube for strix halo/ROCm:
https://github.com/joryirving/home-ops/pull/7626

Defilan (approved):
Excellent onboarding doc. Verified the Model CRD example fields (enabled/count/vendor/layers) against the current schema and the kernel-args / NFD kernel-module rule / Vulkan log markers all match how our Strix Halo node is set up. Merging. Follow-up: we'll add an `amdgpu.lockup_timeout` note to the kernel-args section (the default 2s GPU watchdog can kill long flash-attention dispatches at deep context); niche enough not to block this.

PR #709, opened 2026-06-15
feat(crd): make GPU resource name configurable to support AMD/Vulkan/Intel scheduling

joryirving:
## What

Makes `Model.spec.hardware.gpu.resourceName` and `tolerationKey` first-class CRD fields, with a new deployment-builder helper that resolves the GPU extended resource and toleration key from `vendor` (or an explicit override) instead of hardcoding `nvidia.com/gpu`.

## Why

`deployment_builder.go` hardcoded `nvidia.com/gpu` at the resource limit and the GPU toleration, even though the CRD already exposed a `nvidia | amd | intel` vendor enum. That made AMD and Intel nodes unreachable through LLMKube — surfaced by a community user running llama.cpp's Vulkan backend on an AMD APU (#395). The vendor field was effectively dead on the pod-spec path.

A free-form `resourceName` override is also exposed for non-default device plugins (`squat/generic-device-plugin` advertising `squat.ai/dri-render`, NVIDIA MIG slices, time-slicing variants, DRA).

## How

Resolution order in the new `gpuResourceNameForSpec` / `gpuTolerationKeyForSpec` helpers:

1. `spec.hardware.gpu.resourceName` (escape hatch) — wins over everything.
2. `spec.hardware.gpu.vendor` maps to the device-plugin default (`nvidia` → `nvidia.com/gpu`, `amd` → `amd.com/gpu`, `intel` → `gpu.intel.com/i915`).
3. Unset / unknown → `nvidia.com/gpu` (backwards-compatible default).

`tolerationKey` defaults to the resolved resource name and is independently overridable.

The accelerator-aware `resolveGPUResourceName` used by `Model.checkAcceleratorAvailability` is intentionally left alone — it serves a different consumer and the two paths can evolve independently.

### Files

- `api/v1alpha1/model_types.go` — new optional fields on `GPUSpec` with kubebuilder pattern validation.
- `internal/controller/gpu_resources.go` — new helpers; existing `resolveGPUResourceName` unchanged.
- `internal/controller/deployment_builder.go` — the two hardcoded `nvidia.com/gpu` references route through the helpers.
- `internal/controller/deployment_builder_test.go` — new table-driven tests (14 subtests) covering: nil/empty GPU, each vendor mapping, case-insensitivity, unknown vendor fallback, explicit `resourceName` override (with and without vendor), `tolerationKey` fallback to `resourceName`, and explicit `tolerationKey` override.
[...]

PR #750, opened 2026-06-19
feat(controller): vendor-neutral DRA (resource.k8s.io/v1) scheduling for InferenceService

### LINKED ISSUE #733: [FEATURE] Schedule GPUs via DRA (resource.k8s.io ResourceClaims), not just device-plugin extended resources
## Problem

#395 made the GPU resource name configurable, but only for the **device-plugin** model — the operator sets `container.resources.limits[<resourceName>]` plus a matching toleration. There is no path to schedule GPUs exposed via **Dynamic Resource Allocation (DRA)** (`resource.k8s.io`): the `InferenceService` pod spec has no `resourceClaims` field, and `deployment_builder.go` never sets `PodSpec.ResourceClaims` or a container's `resources.claims`.

#395 explicitly listed DRA as out of scope — *"Kubernetes DRA is still maturing; can revisit when stable."* That note is from ~May 2026; DRA has since reached GA, and it's now the way several GPU stacks expose devices (e.g. the Intel GPU resource driver and the NVIDIA DRA driver). Clusters that run a DRA driver — instead of, or alongside, a classic device plugin — currently have **no way to schedule those GPUs through LLMKube**.

## Real-world case

An Intel integrated GPU exposed via the Intel GPU **DRA** driver — a `ResourceClaimTemplate` with `deviceClassName: gpu.intel.com`, consumed by pods through `spec.resourceClaims` + `container.resources.claims` — serving embedding/rerank models with the llama.cpp **Vulkan** backend. This runs fine bare-metal and via a hand-rolled Deployment, but cannot be expressed as an LLMKube `InferenceService` because there is no DRA passthrough.

The device-plugin `resourceName` from #395 doesn't cover it: setting `resourceName: gpu.intel.com` would emit a `limits[gpu.intel.com]` request, which a DRA driver never advertises as node capacity, so the pod is unschedulable.

## Proposed change

Add a DRA passthrough to the `InferenceService` pod spec, orthogonal to the existing device-plugin `resourceName` path:

```yaml
spec:
  resourceClaims:                                # -> PodSpec.resourceClaims
    - name: gpu
      resourceClaimTemplateName: my-igpu-claim   # or resourceClaimName for a shared claim
  # the operator references each claim from the inference container:
  #   container.resources.claims: [{ name: gpu }]
```

- `deployment_builder.go`: when `spec.resourceClaims` is set, populate `PodSpec.ResourceClaims` and add the corresponding `{name}` entries to the inference container's `resources.claims`. Skip the device-plugin `limits[<resourceName>]` path — DRA and device-plugin are mutually exclusive per workload.
- CRD: add `ResourceClaims []ResourceClaim` (mirroring the upstream `PodResourceClaim` shape: `name` + exactly one of `resourceClaimName` / `resourceClaimTemplateName`) to `InferenceServiceSpec` (or under `hardware.gpu`). Regenerate manifests + deepcopy.

## Interaction with gpuCount / nodeSelector (needs a decision)

Today `--n-gpu-layers`, `nodeSelector`, and `tolerations` are only applied when `gpuCount > 0` (i.e. a device-plugin GPU is requested). A DRA workload requests no device-plugin resource, so:

- The DRA claim itself drives placement (the scheduler puts the pod where the claim is satisfiable), so `nodeSelector` is generally unnecessary.
- But the operator should still treat "a DRA claim is set" as **GPU present**, so GPU-dependent args (`--n-gpu-layers`, `--flash-attn`, …) are emitted. Otherwise the model loads CPU-only unless the user hand-passes `--n-gpu-layers` via `extraArgs`.

So `resolveGPUCount` (or whatever gates the GPU args) should consider a DRA claim as GPU-present.

## Acceptance

- [ ] `InferenceService.spec.resourceClaims` (`name` + `resourceClaimName | resourceClaimTemplateName`) added, with validation that exactly one reference is set
- [ ] Operator populates `PodSpec.ResourceClaims` and the inference container's `resources.claims` from it
- [ ] DRA path and device-plugin `resourceName` are mutually exclusive (validated, or DRA wins when both set — documented)
- [ ] GPU-dependent args (`--n-gpu-layers`, flash-attn) are emitted when a DRA claim is present
- [ ] Sample YAML showing an Intel iGPU + Vulkan + a DRA `ResourceClaimTemplate`
- [ ] Tests cover the DRA pod-spec path and the gpuCount-via-DRA gate

## Out of scope

- Writing or installing DRA drivers (cluster prerequisite).
- The existing device-plugin path stays as-is.

## Context

- #395 — configurable resource name (device-plugin model); deferred DRA as out of scope.
- Device-plugin scheduling path: `internal/controller/deployment_builder.go`, `internal/controller/gpu_resources.go`.
- Fits the multi-vendor GPU scheduling goal in the M2 milestone.

joryirving:
## What

Add a vendor-neutral Dynamic Resource Allocation (DRA) passthrough so an InferenceService can be scheduled onto GPUs advertised via `resource.k8s.io` ResourceClaims. This is the DRA counterpart to the existing device-plugin extended-resource path (#709), and works for **any DRA driver** — Intel GPU, AMD DRA, and NVIDIA DRA — on the existing runtimes.

## Why

#709 made the GPU resource name configurable, but only for the device-plugin model (`limits[<resourceName>]`). Clusters that run a DRA driver instead of (or alongside) a classic device plugin currently have no way to schedule those GPUs through LLMKube, because the operator never sets `PodSpec.ResourceClaims` or `container.resources.claims`. DRA reached GA in Kubernetes 1.34.

### CRD field
A new `resourceClaims` field on the **Model GPUSpec** (`model.spec.hardware.gpu.resourceClaims`, type `[]corev1.PodResourceClaim`), directly beside the existing `resourceName` field. The controller reads it from the model and:
- sets `PodSpec.ResourceClaims` from it, and
- adds matching `{Name: <claimName>}` entries to the inference container `resources.claims`.

### Validation (CEL-only, webhook-free)
Two `+kubebuilder:validation:XValidation` rules on `GPUSpec`:
1. Every `resourceClaims` entry must have **exactly one** of `resourceClaimName` / `resourceClaimTemplateName`.
2. `resourceClaims` and `resourceName` are **mutually exclusive** on the same GPU spec.

### GPU-present gate
The `--n-gpu-layers` / `--flash-attn` emission is currently gated on `resolveGPUCount() > 0`, which is `0` for a DRA-only workload (no device-plugin GPU count). The new `hasGPUPresent()` helper generalizes the gate to `device-plugin resource OR resourceClaims set`, so a DRA-only pod still loads on GPU instead of silently falling back to CPU.

### Placement
For DRA pods, the claim itself drives scheduling, but an explicit `nodeSelector` and any user `tolerations` are still honored. `Recreate` strategy is used (same as device-plugin GPU pods) to avoid the scheduling deadlock where a new pod cannot get the claim while the old one holds it.

### RBAC
Added `get;list;watch` on `resourceclaims` and `resourceclaimtemplates` in `resource.k8s.io`. The chart ClusterRole is hand-maintained; `make check-helm-rbac` verifies it covers the kubebuilder-generated rules.

### Example

```yaml
kind: Model
metadata:
  name: my-embedding-model
spec:
  source: https://example.com/models/embedding.gguf
  format: gguf
  hardware:
    accelerator: vulkan
    gpu:
      enabled: true
      resourceClaims:
        - name: gpu
          resourceClaimTemplateName: example-gpu-claim
---
kind: InferenceService
metadata:
  name: embedding-service
spec:
  modelRef: my-embedding-model
  replicas: 1
```

The same `resourceClaims` field works for Intel-GPU, AMD, and NVIDIA DRA drivers.

## Kubernetes version note

This PR targets `resource.k8s.io/v1` (GA in Kubernetes **1.34**). The vendored k8s deps are >=1.34, so `corev1.PodResourceClaim` and the `resource.k8s.io/v1` types are available without redefinition. If the maintainer wants pre-1.34 support, I am happy to add `v1beta1` compatibility in a follow-up.

Closes #733


Defilan (approved):
This is a fantastic contribution! Excellent work! Really appreciate your help on this. PR approved!

PR #753, opened 2026-06-19
fix(crd): guard GPU resourceClaims/resourceName CEL rule against absent field

joryirving:
## What

Add a `has(self.resourceClaims)` guard to the `GPUSpec` mutual-exclusivity CEL rule so `resourceClaims.size()` is never evaluated when the field is absent.

## Why

The rule added in #750 (shipped in 0.8.9) is:

```
!(has(self.resourceName) && self.resourceClaims.size() > 0)
```

Setting `resourceName` forces CEL to evaluate `self.resourceClaims.size()`. For a Model on the device-plugin path (`resourceName` set, `resourceClaims` unset) there is no such key, so the apiserver rejects it on apply/dry-run:

```
spec.hardware.gpu: Invalid value: "object": no such key: resourceClaims
  evaluating rule: resourceClaims and resourceName are mutually exclusive
```

This rejects **every** `resourceName`-only Model (generic-device-plugin, NVIDIA MIG, any non-default extended resource) — a regression from #750. It slipped through because the DRA tests construct `GPUSpec` in-process via `constructDeployment` and never round-trip through the apiserver, where CEL actually runs.

## How

- Added the `has(self.resourceClaims)` guard; both fields still reject when set together, either alone is admitted.
- Regenerated CRDs (`make manifests` + `make chart-crds`) — `config/crd/bases` and the Helm chart CRD.
- Added an envtest spec that creates Models **through the apiserver** (closing the coverage gap): `resourceName`-only admitted, `resourceClaims`-only admitted, both-set rejected.

```
Ran 3 of 408 Specs — 3 Passed | 0 Failed   (envtest, k8s 1.36)
make lint — 0 issues
```

## Checklist

- [x] Tests added/updated
- [x] New envtest spec passes locally (focused run, envtest 1.36)
- [ ] Documentation updated (not user-facing)

### DIFF
--- a/api/v1alpha1/model_types.go
+++ b/api/v1alpha1/model_types.go
@@ -122,7 +122,7 @@ type HardwareSpec struct {
 }
 
 // GPUSpec defines GPU-specific requirements.
-// +kubebuilder:validation:XValidation:rule="!(has(self.resourceName) && self.resourceClaims.size() > 0)",message="resourceClaims and resourceName are mutually exclusive: use one or the other for GPU scheduling"
+// +kubebuilder:validation:XValidation:rule="!(has(self.resourceName) && has(self.resourceClaims) && self.resourceClaims.size() > 0)",message="resourceClaims and resourceName are mutually exclusive: use one or the other for GPU scheduling"
 type GPUSpec struct {
 	// Enabled indicates whether GPU acceleration is enabled
 	// +optional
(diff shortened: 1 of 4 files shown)


Defilan (approved):
Approving. The `has(self.resourceClaims)` guard is the correct fix for the CEL `no such key: resourceClaims` error that was rejecting every device-plugin (resourceName-only) Model.

Verified the rule across all cases: resourceName-only admitted, resourceClaims-only admitted, both-set rejected. The CRDs are in sync across the Go marker, `config/crd/bases`, and the Helm chart. The new envtest is the important part: it round-trips through the apiserver where CEL actually runs, which the in-process DRA tests did not, so it closes the gap that let the original rule ship broken.

Filed #754 to track two pre-existing DRA follow-ups surfaced during review (`checkAcceleratorAvailability` not being DRA-aware for resourceClaims-only Models, and the `resourceName` field doc still listing DRA as a use case). Neither blocks this fix.

PR #758, opened 2026-06-20
feat(ci): publish Helm charts as OCI artifacts to GHCR

eleboucher:
## Summary

Adds OCI publishing of the Helm charts to GHCR as part of the existing release pipeline.

## Changes

- **`.github/workflows/release-please.yml`** — in the `release-helm` job (runs only when release-please creates a release):
  - Log in to GHCR for Helm OCI using `GITHUB_TOKEN` (the job already has `packages: write`). Credentials are passed via `env:` + `--password-stdin` to avoid shell interpolation.
  - Package and push the `llmkube` and `foreman` charts at the release version to `oci://ghcr.io/defilantech/charts`.
- **`charts/llmkube/README.md`** — document installing from the OCI registry.

## Resulting artifacts per release

- `ghcr.io/defilantech/charts/llmkube:<version>`
- `ghcr.io/defilantech/charts/foreman:<version>`

## Install

```bash
helm install llmkube oci://ghcr.io/defilantech/charts/llmkube \
  --namespace llmkube-system --create-namespace
```

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -101,6 +101,22 @@ jobs:
       - name: Sync Chart CRDs Before Packaging
         run: make chart-crds
 
+      - name: Log in to GHCR for Helm OCI
+        env:
+          GHCR_USER: ${{ github.actor }}
+          GHCR_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: echo "$GHCR_TOKEN" | helm registry login ghcr.io --username "$GHCR_USER" --password-stdin
+
+      - name: Package and push charts to GHCR (OCI)
+        env:
+          VERSION: ${{ needs.release-please.outputs.version }}
+          OCI_REPO: oci://ghcr.io/defilantech/charts
+        run: |
+          for chart in llmkube foreman; do
+            helm package "charts/$chart" --destination dist-charts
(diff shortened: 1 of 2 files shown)


Defilan:
Thanks for this, and welcome to LLMKube! 🎉

This is a really clean first contribution. OCI publishing is a great addition alongside the existing GitHub Pages chart repo, and the implementation is solid: logging in via `--password-stdin` (no token interpolation), reusing the existing `packages: write` permission, and packaging after the CRD sync step are all exactly right. I checked the version handling too: `release-please.outputs.version` matches the bumped `Chart.yaml` versions, so the `helm package` / `helm push` filenames line up. Nice work.

One thing to fix before we can merge: the **DCO check is failing** because the commit isn't signed off. Since it's a single commit, you can fix it with:

```bash
git commit --amend --signoff
git push --force-with-lease
```

That adds the `Signed-off-by:` trailer the [DCO](https://developercertificate.org/) check requires. Once it's green we'll get CI to run on the workflow/README changes and take it from there.

No action needed from you on the operational side: newly published GHCR OCI packages default to private, so we'll handle making `ghcr.io/defilantech/charts/{llmkube,foreman}` public after the first release that runs this. We'll follow up on that.

Thanks again for jumping in.

Defilan (approved):

PR #766, opened 2026-06-20
fix(router): inject backend Model on dispatch so fallback chains degrade

arychj:
The router-proxy matched routing rules on the client-facing `model` string and then forwarded the request body verbatim, never applying a backend's configured `Model`. `Backend.Model` is documented as "the upstream model identifier passed to the provider" and is surfaced in `/v1/models`, but it was inert on the dispatch path. This had two consequences:

- **External providers received the client alias** (e.g. `ha`, `opus`) instead of a real upstream id, so any backend whose alias differs from its upstream model name failed.
- **A `primary-fallback` chain across cloud backends was a silent no-op:** every attempt re-sent the same frozen client model, so `[opus, sonnet, haiku]` retried `opus` three times instead of degrading across models.

It also blocks the one fallback that has no other home: **local-primary → cloud-fallback.** A downstream proxy (e.g. LiteLLM) only ever sees the cloud leg, so cross-tier degradation can only live in the router.

## What changed

`Dispatch` now rewrites the `model` field of an *outbound copy* of the request body to the chosen backend's `Model`, via a new `applyModelOverride` helper. The override is gated on a non-empty `Model`: local backends leave it empty and pass through verbatim, and non-JSON bodies are returned unchanged. The shared inbound buffer (reused across fallback attempts) is never mutated.

This is a behavior change in that `Backend.Model` was previously inert and is now applied — but it makes runtime behavior match the field's documented contract rather than changing an intended one.

## Validation

`go test ./internal/router/`, `go vet`, and `gofmt` all clean; full module builds. New coverage:

- `TestApplyModelOverride` — rewrites existing model, injects when absent, preserves sibling fields, no-op on empty Model, no-op on non-object/non-JSON bodies.
- `TestDispatchRewritesModelForExternalBackend` / `TestDispatchLeavesBodyForLocalBackend` — wire-level assertion that an external backend's `Model` reaches the upstream and a local backend's body passes through verbatim.
- `TestProxyDegradesModelAcrossFallback` — end-to-end: when the primary cloud backend 5xxs and the chain falls back, the cloud backend receives its configured `Model` and the local fallback receives the verbatim client alias.

_Assisted by Claude Opus 4.8_

### DIFF
--- a/internal/router/backend.go
+++ b/internal/router/backend.go
@@ -13,6 +13,7 @@ package router
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -212,9 +213,9 @@ func (d *Dispatcher) MarkUnhealthy(name string) {
 // upstream response. Caller is responsible for streaming the body to
 // the inbound client and closing it. On error the response is nil.
 //
-// requestBody is the already-buffered inbound body. The proxy reads the
-// body once (it needs to parse the "model" field) and reuses the bytes
-// across fallback attempts.
+// requestBody is the shared inbound body, reused across fallback attempts.
+// Dispatch rewrites the "model" field of an outbound copy (see
+// applyModelOverride) without mutating it.
 func (d *Dispatcher) Dispatch(
 	ctx context.Context,
 	backend *Backend,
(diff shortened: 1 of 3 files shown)


Defilan (approved):
Welcome to LLMKube, and thanks for a really sharp first contribution!

This is a great catch. `Backend.Model` being inert on the dispatch path is a subtle bug with real consequences, and your write-up nails exactly why it matters: cloud `primary-fallback` chains were silently re-sending the same alias, and local-primary -> cloud-fallback (the cross-tier degradation that can only live in the router) was impossible. That last case is right in line with where we're taking the heterogeneous-fleet story, so this is genuinely useful.

The implementation is clean. Parsing into `map[string]json.RawMessage` to preserve sibling fields byte-for-byte (rather than a lossy round-trip of the messages array), operating on an outbound copy so the shared inbound buffer is never mutated, and degrading safely on empty `Model` / non-JSON bodies are all exactly the right calls. The tests are thorough too: `TestProxyDegradesModelAcrossFallback` asserting the cloud backend gets its configured model while the local fallback gets the verbatim alias is precisely the proof this fix needs.

CI is green and this looks ready to merge. Thanks again for taking the time to do this well.

Issue #790, opened 2026-06-22
[QUESTION] Backend naming character regex


arychj:
Hi! First, thanks for starting this project, it looks great!

Question: The naming regex for the `backend.name` field is limited to `^[a-z0-9][a-z0-9-]{0,62}$`. I'm trying to come up with some sort to explicit indicator on the API surface to signal where a model "lives". This could be abstracted as more of a "category" than a "locality". Usage could be as simple as local/cloud, the model provider, some sort of billing indicator, etc.

Basically I'd like to present the `/v1/models` list with something like:
```
data.0.id: gemma-4-e4b # local model
data.1.id: cloud/opus-4-8 # cloud model
```

I considered tracking this information in a separate field that's exposed in the route, but a lot of downstream clients consume the model list and the present only the id to the user for model selection, so this indicator has to carried in the id field.

I also considered using a dash (à la `cloud-opus-4-8`), but wanted a clear distinction between the category and the model.

I know these names are passed as labels to k8s resources, but I figured they could be slugified at point.

Thoughts on whether this would be a helpful change?

Defilan:
Thanks for this, and for digging into the router internals! This is a real need and I want to support it.

I don't want to get there by loosening the `backend.name` regex though. Names flow through to k8s as label values, and `/` isn't legal in a label value (it's only allowed in the key prefix), so widening the regex would mean slugifying on every label mapping. That's leaky and it'll bite in surprising spots later.

I'd rather split the two concerns. Keep `name` strict and k8s-safe like it is today, and add a separate optional field that composes the id we publish on `/v1/models`. That way you get `cloud/opus-4-8` on the API surface while the internal identifier stays clean. The published id and the k8s-safe name don't have to be the same string.

Nice bonus: this composes cleanly with your #791. Matching still happens on the safe `name`, and the rich id is just presentation.

Couple things I'd want your take on before we lock the shape:
- Single freeform field (say `displayName`) vs structured (`category` + name joined with `/`)? Freeform is simpler; structured lets us validate and group later.
- When the field's unset, the published id should default to `name` so existing configs don't change. I'd lean yes there.

I've opened #792 to scope it (with the BackendNameMatch interaction captured so the two features don't fight). Happy to have you take a crack at it there if you're up for it. Either way, appreciate the thoughtful writeup.

arychj:
I like that solution! It's a lot cleaner than overloading the `name` field.

Closing this in favour of the feature issue.

PR #791, opened 2026-06-22
feat: add defaultRouteStrategy with BackendNameMatch to ModelRouter

arychj:
## What

Add an optional `spec.defaultRouteStrategy` field to the `ModelRouter` CRD — an enum of `Static` (default) and `BackendNameMatch` — controlling what happens to a request that matches no routing rule:

- **`Static`** preserves current behavior: route to `defaultRoute` (or HTTP 503 if it is unset).
- **`BackendNameMatch`**: first resolve the request's `model` field against a backend of the same `name`; if one exists, route straight to it; otherwise fall back to `defaultRoute`, then HTTP 503.

The lookup is against backend `name` — the identifier published to clients — never the upstream provider `model`, keeping the proxied API surface on its own side of the boundary.

## Why

A ModelRouter configuration tends to accumulate a large block of pure identity-passthrough rules: one rule per backend whose only job is to match `models: [X]` and route to `backends: [X]` for a backend also named `X`. That boilerplate grows linearly with backend count, is easy to get wrong, and buries the rules that carry real routing semantics (failover chains, fail-closed gates, semantic aliases) in noise.

`BackendNameMatch` collapses every one of those identity rules into a single strategy flag. Rules that carry real semantics stay exactly as they are; the per-backend passthrough rules disappear.

## How

- New `DefaultRouteStrategy` typed enum and `spec.defaultRouteStrategy` field (kubebuilder enum + default `Static`, optional). CRDs regenerated and synced to the Helm chart.
- **Proxy data plane:** the matcher resolves an unmatched request's model to a backend by name between the rule loop and the `defaultRoute` fallback (`internal/router/match.go`). The wire config carries the field with a defensive validation guard.
- **Gateway data plane:** `compileRouterRules` emits one `model == <name>` rule per backend, ordered after explicit rules and before the `defaultRoute` catch-all — preserving first-match parity with the proxy.
- Explicit rules, including `failClosed`, are evaluated first and continue to gate before any name match. The behavior, the trust-boundary implication (every backend, including cloud-tier, becomes directly addressable by name), and the fail-fast-on-unhealthy semantics are documented in the concept guide and the field doc.

Notes for reviewers:

- **Back-compat:** additive optional field defaulting to `Static`; existing `ModelRouter`s are unaffected.
- A request naming a backend that is currently unhealthy fails fast (502/503) rather than silently rerouting to `defaultRoute` — naming a backend is a request for *that* backend, and this matches how an explicit single-backend rule behaves on outage. Parity holds across both data planes.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated

_Assisted by Claude Opus 4.8_

### DIFF
--- a/api/v1alpha1/modelrouter_types.go
+++ b/api/v1alpha1/modelrouter_types.go
@@ -34,6 +34,20 @@ const (
 	ModelRouterDataPlaneGateway ModelRouterDataPlane = "Gateway"
 )
 
+// DefaultRouteStrategy selects what the router does when no rule matches a
+// request, before falling back to DefaultRoute.
+type DefaultRouteStrategy string
+
+const (
+	// DefaultRouteStrategyStatic always routes unmatched requests to the named
+	// DefaultRoute. This is the default.
+	DefaultRouteStrategyStatic DefaultRouteStrategy = "Static"
+
+	// DefaultRouteStrategyBackendNameMatch matches an unmatched request to a
+	// backend whose Name equals the request "model" field before DefaultRoute.
+	DefaultRouteStrategyBackendNameMatch DefaultRouteStrategy = "BackendNameMatch"
+)
+
 // ModelRouterSpec defines the desired state of a ModelRouter.
 //
 // A ModelRouter exposes a single OpenAI-compatible HTTP endpoint that
@@ -77,8 +91,9 @@ type ModelRouterSpec struct {
 	Backends []RouterBackend `json:"backends"`
 
 	// Rules are evaluated in declaration order. The first matching rule wins.
-	// If no rule matches, DefaultRoute is used. If neither a matching rule
-	// nor DefaultRoute is set, the request is rejected with HTTP 503.
+	// If no rule matches, the fallback is governed by DefaultRouteStrategy,
+	// ending at DefaultRoute. If nothing resolves, the request is rejected
+	// with HTTP 503.
 	// +optional
 	Rules []RouterRule `json:"rules,omitempty"`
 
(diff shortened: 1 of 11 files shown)


Defilan (approved):
This is a great PR, thanks. Genuinely one of the cleaner feature contributions we've had.

A few things worth calling out because they're exactly right:
- You implemented it on both data planes and kept them equivalent: the proxy matcher and the gateway compile (one `model==name` rule per backend, inserted ahead of the catch-all). Easy to do one and miss the other.
- The precedence is spot on and well tested: explicit rules win, then name match, then `defaultRoute`, then 503. I especially like the empty-model and rules-win-over-name cases, and that you covered both the compiler and the runtime matcher.
- Defaulting to `Static` (and the zero value being `Static`) keeps existing configs untouched.
- The security framing is the right call. `failClosed` rules evaluate first, you tested that a pii rule beats a name match, and the docs are honest that this isn't a global PII-egress invariant. I'd rather see that stated plainly than hand-waved.

One thing to flag, and it ties into your #790: BackendNameMatch keys on backend `name`. If we go the way I proposed over there (keep `name` strict/k8s-safe, add a separate field that composes the published `/v1/models` id), clients will select and send back the *published* id, not `name`. So `{"model": "cloud/opus-4-8"}` would miss the `cloud-opus` lookup and fall through to `defaultRoute`. Once that field lands, BackendNameMatch should match on the published id (the new field when set, else `name`). I captured that in #792 so it doesn't get lost.

Not a blocker here, this is correct for the current model. Approving. I'm good to merge as-is and handle the published-id match as the follow-up in #792, or you can fold it in when we spec that, your call.

Nice work.

arychj:
Responded in #792 , but I think that additional work belongs in a separate PR. This one should be good to go!

PR #794, opened 2026-06-22
feat: add topologySpreadConstraints and affinity to InferenceService (#793)

### LINKED ISSUE #793: InferenceService: support topologySpreadConstraints / affinity for pod placement
### Context

`InferenceService.spec` exposes `nodeSelector`, `tolerations`, `podLabels`, and `podAnnotations` for pod placement, but **not** `topologySpreadConstraints` or `affinity`.

### Problem

When several InferenceServices target the same small pool of GPU nodes (e.g. 3 nodes, each with an iGPU), the scheduler bin-packs them onto whichever node first satisfies the device claim. With DRA selecting a node but nothing expressing *spread*, multiple model servers stack onto one GPU while peer nodes sit idle — concentrating VRAM/UMA pressure and hurting throughput.

The only placement lever today is `nodeSelector`, which forces **static, hard pinning** per model — brittle, and it loses rescheduling flexibility.

### What I'd like

Pass `topologySpreadConstraints` (and ideally `affinity`) from `InferenceService.spec` through to the generated pod template, alongside the existing `nodeSelector`/`tolerations`. That enables the standard soft-spread pattern:

```yaml
spec:
  topologySpreadConstraints:
    - maxSkew: 1
      topologyKey: kubernetes.io/hostname
      whenUnsatisfiable: ScheduleAnyway
      labelSelector:
        matchLabels:
          gpu-workload: "true"
```

So model servers (and other GPU pods sharing the label) prefer separate nodes but never block scheduling.

### Note / current workaround

`podLabels` already lets an InferenceService pod be **counted** by *other* workloads' topologySpread (e.g. media transcoders that carry the constraint will schedule away from a model-heavy node). But the InferenceService pods can't carry a constraint of their own, so they don't self-spread. Since the pod-template plumbing for `nodeSelector`/`tolerations`/`podLabels` already exists, adding `topologySpreadConstraints` (and `affinity`) looks like a natural, low-risk extension. Happy to send a PR if you're open to it.

joryirving:
Closes #793.

## Problem

`InferenceService.spec` exposes `nodeSelector`, `tolerations`, and `podLabels` for placement, but not `topologySpreadConstraints` or `affinity`. When several model servers target the same small GPU-node pool, the scheduler picks whatever node satisfies the device/DRA claim and bin-packs them onto one node while peers sit idle — and the only lever to fix it is static per-model `nodeSelector` pinning (brittle, no rescheduling flexibility).

## Change

Add `topologySpreadConstraints` and `affinity` as passthrough fields on `InferenceServiceSpec`, applied to the pod template **regardless of the GPU / DRA / CPU path** (they're general pod scheduling, not GPU-specific). Combined with the existing `podLabels`, this enables the standard soft-spread:

```yaml
spec:
  podLabels:
    gpu-workload: "true"
  topologySpreadConstraints:
    - maxSkew: 1
      topologyKey: kubernetes.io/hostname
      whenUnsatisfiable: ScheduleAnyway
      labelSelector:
        matchLabels:
          gpu-workload: "true"
```

so model servers (and other GPU pods sharing the label) prefer separate nodes but never block scheduling.

## Notes

- Pure passthrough into `PodSpec`; no behavior change when the fields are unset.
- Applied after the GPU-device-plugin and DRA scheduling blocks so it covers all three placement paths.
- Regenerated deepcopy + CRD (`make generate manifests`).
- Added a deployment-builder test asserting both fields reach `Spec.Template.Spec`. `make test` passes (full envtest suite, `internal/controller` included).

Builds on the existing `nodeSelector`/`tolerations`/`podLabels` plumbing — same shape, just two more standard `corev1` scheduling fields.

### DIFF
--- a/api/v1alpha1/inferenceservice_types.go
+++ b/api/v1alpha1/inferenceservice_types.go
@@ -120,6 +120,20 @@ type InferenceServiceSpec struct {
 	// +optional
 	PodLabels map[string]string `json:"podLabels,omitempty"`
 
+	// TopologySpreadConstraints control how inference Pods are spread across
+	// topology domains (e.g. one model server per GPU node). Passthrough to the
+	// Pod spec; combine with PodLabels so the constraint's labelSelector can
+	// match sibling GPU workloads for a soft, cross-app spread.
+	// +optional
+	TopologySpreadConstraints []corev1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+
+	// Affinity constrains inference Pod placement (node/pod affinity and
+	// anti-affinity). Passthrough to the Pod spec, for finer control than
+	// NodeSelector — e.g. preferring or avoiding nodes already running other
+	// GPU workloads.
+	// +optional
+	Affinity *corev1.Affinity `json:"affinity,omitempty"`
+
 	// RuntimeClassName selects a Kubernetes RuntimeClass for the inference Pod.
 	// Most commonly set to "nvidia" on clusters where the NVIDIA Container
 	// Runtime is not configured as the cluster default. Without it, GPU pods
--- a/api/v1alpha1/zz_generated.deepcopy.go
+++ b/api/v1alpha1/zz_generated.deepcopy.go
@@ -485,6 +485,18 @@ func (in *InferenceServiceSpec) DeepCopyInto(out *InferenceServiceSpec) {
 			(*out)[key] = val
 		}
 	}
+	if in.TopologySpreadConstraints != nil {
+		in, out := &in.TopologySpreadConstraints, &out.TopologySpreadConstraints
+		*out = make([]v1.TopologySpreadConstraint, len(*in))
+		for i := range *in {
+			(*in)[i].DeepCopyInto(&(*out)[i])
+		}
+	}
+	if in.Affinity != nil {
+		in, out := &in.Affinity, &out.Affinity
+		*out = new(v1.Affinity)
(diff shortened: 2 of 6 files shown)


Defilan (approved):
Thanks for this, really clean PR. Exposing `topologySpreadConstraints` and `affinity` as passthrough is the right call: these are general pod-scheduling knobs, not GPU-specific, so they belong on the spec as-is.

What I like:
- Minimal and idiomatic: straight passthrough of the standard `corev1` types, deepcopy regenerated, both CRD copies (config + chart) in sync.
- Applying them after the GPU/DRA branch, unconditionally, is correct. I checked `applyDRAPodScheduling` and it only sets Strategy/NodeSelector/Tolerations, so there's no clobber of your affinity/topology. Good instinct on placement.
- The test bites: without the wiring those fields come through nil, so the assertions would fail. Real values too (`ScheduleAnyway`, the `gpu-workload` selector, an Exists nodeAffinity term).

Two small things, neither blocking:
- The test exercises the GPU path. The description says "regardless of GPU/DRA/CPU path," which the code does, but a quick CPU-model case would lock that claim so a future refactor can't silently move the passthrough into a GPU-only branch.
- The affinity assertion stops at `NodeAffinity != nil`. Fine for a whole-object assignment today, but asserting the `MatchExpressions` survived would bite harder against a partial-copy regression later.

Optional: a short note in `docs/site` on how affinity composes with the GPU `nodeSelector` (they're ANDed) would help folks who set both.

Nice, focused contribution.

PR #798, opened 2026-06-22
Add revisionHistoryLimit to ModelRouter and InferenceService

arychj:
The router-proxy and inference Deployments offer no way to cap how many old ReplicaSets Kubernetes retains, so they fall back to the default of 10. The router-proxy re-rolls its pods on every compiled-config change (the ConfigMap hash lands on the pod template), so superseded ReplicaSets pile up quickly on a frequently-edited router.

## What changed

- Add an optional `revisionHistoryLimit` field to both CRDs:
  - `ModelRouter` → `spec.proxy.revisionHistoryLimit`
  - `InferenceService` → `spec.revisionHistoryLimit`
- Plumb the value through to each Deployment. The router-proxy reconciler copies mutable fields individually, so it gets an explicit copy in the update path; the InferenceService reconciler replaces `Spec` wholesale, so it propagates automatically.
- Pure passthrough: unset defers to the Kubernetes default (10); `0` keeps no history. No operator-imposed default.
- Regenerated CRD manifests + deepcopy, synced the Helm chart CRDs, and documented the knob in the model-router concept doc.

## Validation

`go build`, `go vet`, the full `internal/controller` envtest suite, and the `api/...` tests pass. Added unit tests covering the nil-when-unset and explicit-value (including `0`) paths for both builders.

_Assisted by Claude Opus 4.8_

### DIFF
--- a/api/v1alpha1/inferenceservice_types.go
+++ b/api/v1alpha1/inferenceservice_types.go
@@ -73,6 +73,13 @@ type InferenceServiceSpec struct {
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
 
+	// RevisionHistoryLimit caps how many old ReplicaSets the inference
+	// Deployment keeps for rollback. Unset uses the Kubernetes default (10);
+	// 0 keeps none.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	RevisionHistoryLimit *int32 `json:"revisionHistoryLimit,omitempty"`
+
 	// Autoscaling configures horizontal pod autoscaling for the inference service.
 	// When set, the controller creates and manages an HPA resource targeting the
 	// inference Deployment. Requires Prometheus Adapter for custom metrics.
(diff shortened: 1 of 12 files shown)


Defilan (approved):
Another clean one, thanks. `revisionHistoryLimit` is a sensible knob, and the motivation is real: the proxy re-rolls on every config change, so superseded ReplicaSets pile up fast.

What's right:
- Pure passthrough, no operator-imposed default: unset defers to the k8s default (10), `0` keeps none, `Minimum=0` blocks negatives. Correct semantics.
- You handled the easy-to-miss part: the router reconciler copies mutable fields individually, so you added the explicit `existing.Spec.RevisionHistoryLimit = desired.Spec.RevisionHistoryLimit` in the update path next to the Replicas copy. That's what keeps an edit propagating to an existing router instead of only landing on create. The InferenceService side replaces `Spec` wholesale so it rides along. Good read of both paths.
- CRD manifests + deepcopy regenerated, both chart + config synced, doc updated. Tests bite.

One small non-blocking suggestion: the router test covers `newRouterDeployment`. A test that reconciles an *existing* router Deployment with a changed `revisionHistoryLimit` and asserts it propagated would lock down that update-path copy specifically, since the individual-field-copy surface is the one easy to regress later.

Nice work, and thanks for the run of solid router PRs today.

PR #801, opened 2026-06-22
fix(chart): disable model cache when modelCache.enabled is false

eleboucher:
## Problem

Setting `modelCache.enabled: false` does not actually disable the model cache. The operator's `--model-cache-path` flag defaults to `/models` (`cmd/main.go`), and caching is gated on that path being non-empty:

- `internal/controller/inferenceservice_controller.go`: `model.Status.CacheKey != "" && r.ModelCachePath != ""`
- `internal/controller/deployment_builder.go`: `useCache := model.Status.CacheKey != "" && r.ModelCachePath != ""`

The chart only emits the `--model-cache-*` flags when `enabled: true`, so when disabled the operator falls back to the non-empty `/models` default and keeps provisioning + mounting the cache (a shared 100Gi PVC) anyway.

## Fix

Emit an explicit empty `--model-cache-path=` in the disabled branch so the controller's `path != ""` gate evaluates false and caching is actually off.

## Test

Added a helm-unittest regression case to `tests/model-cache_test.yaml` asserting that with `modelCache.enabled: false` the deployment passes `--model-cache-path=` (empty), no `--model-cache-path=/models`, and no `/models` volume mount. Full suite passes (10/10).

### DIFF
--- a/charts/llmkube/templates/deployment.yaml
+++ b/charts/llmkube/templates/deployment.yaml
@@ -53,6 +53,9 @@ spec:
         - --model-cache-storage-class={{ .Values.modelCache.storageClass }}
         {{- end }}
         - --model-cache-access-mode={{ .Values.modelCache.accessMode }}
+        {{- else }}
+        # Empty path disables caching (the flag defaults to /models otherwise).
+        - --model-cache-path=
         {{- end }}
         - --init-container-image={{ include "llmkube.initContainerImage" . }}
         # The default value is set in values.yaml. We do NOT use `| default 102`
--- a/charts/llmkube/tests/model-cache_test.yaml
+++ b/charts/llmkube/tests/model-cache_test.yaml
@@ -67,6 +67,23 @@ tests:
           path: spec.template.spec.containers[0].args
           content: --model-cache-mode=perService
 
+  - it: passes an empty model-cache-path to disable caching when disabled
+    template: deployment.yaml
+    set:
+      modelCache.enabled: false
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: "--model-cache-path="
+      - notContains:
+          path: spec.template.spec.containers[0].args
+          content: --model-cache-path=/models
+      - notContains:
+          path: spec.template.spec.containers[0].volumeMounts
+          content:
+            name: model-cache
+            mountPath: /models
+
   # In the default shared mode the chart creates the single cluster-wide PVC so
   # all InferenceServices share one cache and `cache list` can inspect it.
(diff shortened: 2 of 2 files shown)


eleboucher:
Btw you shouldn't put the github actions to run unless you approve it, to prevent from cache poisoning and attacker getting your secrets

Defilan:
Thanks for this contribution! Good callout with actions. I set up the initial policy when the project was smaller that I had to approve actions for new contributors but allow it for returning ones. Now that the project has grown, it's worth revisiting that. Appreciate your callout on this.

Defilan (approved):
Nice fix, and thanks, the root-cause writeup is spot on. I verified both halves: `--model-cache-path` does default to `/models` (`cmd/main.go:120`), and caching is gated on `r.ModelCachePath != ""` in both the controller and the deployment builder. So with `enabled: false` emitting no flag at all, the operator fell back to `/models` and kept caching. Emitting an explicit `--model-cache-path=` flips that gate off, and Go's flag parser reads the `=` form as an empty string, so it does exactly what you want.

The regression test bites, too: without the fix the disabled branch emits no `--model-cache-path` flag, so the `contains: --model-cache-path=` assertion fails.

One tiny non-blocking note: the third assertion (no `model-cache` volumeMount at `/models`) checks the operator's own deployment, which never mounts the cache, the real mount happens in the inference pods the operator creates at runtime, which this chart template doesn't render. So that one passes regardless of the fix. The two arg assertions are doing the real work.

Approving. Clean catch, thank you.

eleboucher:
> Nice fix, and thanks, the root-cause writeup is spot on. I verified both halves: `--model-cache-path` does default to `/models` (`cmd/main.go:120`), and caching is gated on `r.ModelCachePath != ""` in both the controller and the deployment builder. So with `enabled: false` emitting no flag at all, the operator fell back to `/models` and kept caching. Emitting an explicit `--model-cache-path=` flips that gate off, and Go's flag parser reads the `=` form as an empty string, so it does exactly what you want.
> 
> The regression test bites, too: without the fix the disabled branch emits no `--model-cache-path` flag, so the `contains: --model-cache-path=` assertion fails.
> 
> One tiny non-blocking note: the third assertion (no `model-cache` volumeMount at `/models`) checks the operator's own deployment, which never mounts the cache, the real mount happens in the inference pods the operator creates at runtime, which this chart template doesn't render. So that one passes regardless of the fix. The two arg assertions are doing the real work.
> 
> Approving. Clean catch, thank you.

thanks for the quick review if you are interested we are already a bunch of people using your product in our homelab in this community https://discord.gg/home-operations if you want to join us that would be amazing

PR #853, opened 2026-06-26
feat(model): stage multi-file artifacts

### LINKED ISSUE #847: [FEATURE] Native multi-file model staging (sharded GGUF + mmproj) on the Model CRD
## Feature Description

Make the `Model` CRD able to stage **multiple files** for a single model, and treat a multimodal projector (`mmproj`) as a first-class field. Today the model-cache fetches exactly one file per `Model` (a single `curl` keyed to one source URL), so anything multi-file (sharded GGUF, or a GGUF + its `mmproj`) has to be pre-staged out-of-band and the projector wired through `extraArgs`.

## Problem Statement

> As an operator deploying a multimodal (or sharded) GGUF model, I want LLMKube to stage all of the model's files and wire the projector for me, so that I don't have to run a separate download Job and hand-thread `--mmproj` through `extraArgs`.

Two concrete gaps:

1. **Multi-file weights.** Sharded GGUF (`model-00001-of-0000N.gguf`) can't be fetched by the cache: it pulls one file per `Model`. llama.cpp can load a split if the shards sit together, but LLMKube won't download the siblings.
2. **mmproj / multimodal projector.** There's no field for a projector and no `--mmproj` wiring. Vision models need the `mmproj-*.gguf` staged separately and passed as a raw `extraArg`, so the `Model` CR has no knowledge of it (no integrity check, no cache, no drift detection for that file).

This pushes adopters to bypass the cache with a one-off staging Job. That works, but the projector ends up as a loose runtime flag the control plane can't reason about.

## Proposed Solution

Let a `Model` declare a file set and a projector. Sketch (exact shape TBD):

```yaml
kind: Model
metadata:
  name: qwen3.6-35b-a3b
spec:
  # Option A: a repo + file selection (mirrors `hf download <repo> <files>`)
  source: hf://Qwen/Qwen3.6-35B-A3B
  files:
    - "Qwen3.6-35B-A3B-UD-Q8_K_XL.gguf"   # primary (sharded set globbable)
  mmproj: "mmproj-F16.gguf"               # first-class projector
  format: gguf
```

- The download Job fetches the declared file set (globs for shard sets), the same way a hand-rolled `hf download` Job does today.
- When `mmproj` is set, the serving layer adds `--mmproj <staged-path>` automatically (no manual `extraArgs`).
- Cache/integrity/drift tracking extends to the whole set, not just one blob.

## Alternatives Considered

- **Current workaround (works today):** an out-of-band Job runs `hf download repo files` into an RWX PVC, the `Model` points at `pvc://.../primary.gguf`, and the projector is passed via `extraArgs: [--mmproj, /model-source/.../mmproj-F16.gguf]`. Clean and supported (`pvc://` is a first-class source scheme), but the projector is invisible to the `Model` CR and there's no cache/integrity for it.
- Keep `mmproj` as `extraArgs` only (status quo). Rejected: the control plane can't validate or cache a file it doesn't know about.

## Additional Context

- **Motivating real-world config** (community user running LLMKube on AMD Strix Halo / ROCm, serving a multimodal Qwen3.6-35B):
  - Staging Job: https://github.com/joryirving/home-ops/blob/699840c9da55924cbca4d022dd13a574569eae27/kubernetes/apps/base/llm/llmkube/stage/stage.yaml
  - Model + InferenceService (note the `--mmproj` in `extraArgs`): https://github.com/joryirving/home-ops/blob/699840c9da55924cbca4d022dd13a574569eae27/kubernetes/apps/base/llm/llmkube/models/llama-strix.yaml
- Related issues: #
- Workaround in use: pre-stage on an RWX PVC via `hf download`, `source: pvc://...`, projector via `extraArgs --mmproj`.

## Priority

- [ ] Medium - Nice to have
- [ ] Low - Minor improvement

## Willingness to Contribute

- [ ] Yes, I can submit a PR
- [x] Yes, I can help test
- [ ] No, just requesting

joryirving:
## What

Adds native multi-file model staging on `Model`:
- `spec.files[]` declares the ordered model artifact set; `files[0]` is the primary runtime model path.
- `spec.mmproj` declares a first-class multimodal projector that is staged and wired to llama.cpp as `--mmproj`.
- `hf://` sources are accepted and normalized for runtime/download paths.

## Why

Multimodal and sharded GGUF models need sibling files staged together, and projector files should be visible to the control plane instead of hidden in `extraArgs`.

Fixes #847

## How

The controller resolves declared file sets into an ordered staging plan, validates repo-relative paths, preserves subdirectories in cache/emptyDir storage, and records staged files in Model status. The init container downloads each artifact with per-file refresh handling, and llama.cpp derives managed `--model` / `--mmproj` paths from the staged layout while preserving `extraArgs` precedence.

## Checklist

- [x] Tests added/updated

Verification note: `make test` was run with a clean temporary `HOME`/`XDG_CONFIG_HOME` so repo tests that invoke `git commit` are not affected by my global 1Password git signing config.


Defilan (changes requested):
Really nice work on this, and it's fitting it came from you given you've actually got the multimodal use case driving it. The shape is right: the API matches what I had in mind for #847, `files[0]` to `--model` is wired correctly, extraArgs precedence is handled (and tested), the single-source path is untouched, and the codegen/CRDs are all in sync. The happy path (HF source, `files[]` + `mmproj`, llama.cpp) works and the unit coverage on it is solid. Appreciate the thoroughness, including the clean-HOME test note.

A few things I'd like tightened before merge, all at the edges rather than the core:

**1. The storage and runtime layers disagree about mmproj, and it can produce a pod that won't start.** `modelStagingPlan` returns `nil` when `ResolveFileSet` errors (mmproj set without `files`, or a rejected path), so the init container falls back to the legacy single-file command and never stages the projector. But `LlamaCppBackend.BuildArgs` emits `--mmproj <staged path>` for any model with `Mmproj != ""`, independent of whether staging succeeded. So in those cases the pod gets `--mmproj` pointing at a file that was never downloaded, and llama-server fails to start. Two asks:
- Only emit the mmproj arg inside the `ResolveFileSet` success guard. Right now `appendMmprojArgs(...)` runs after the `if plan, err := ...; err == nil` block, so it fires even on error with the raw `Spec.Mmproj`. That also closes a path-validation gap: the validation rejects `..`/absolute mmproj values, but the runtime arg emits them anyway.
- Make `modelStagingPlan` fail closed instead of returning `nil` on error, so the storage and runtime sides agree on when multi-file staging is active.

**2. `file://` and `pvc://` sources are unguarded.** Validation and staging only run in `reconcileRuntimeResolvedSource` (HF/HTTP/metal). A local or PVC Model with `files[]`/`mmproj` reaches Ready, silently ignores the extra files at the storage layer, and for `file://` can still emit `--mmproj` for an unstaged path. I'd either route those source types through the same staging path or reject `files[]`/`mmproj` for them at validation, so it fails loud instead of deploying something broken.

Smaller things, not blockers:
- `hf://` hardcodes `resolve/main`, so `hf://org/repo@rev` silently 404s. Either reject a `@rev` in the repo id or note that we only resolve `main` for now.
- The glob/existence checks in `fileset.go` are effectively dead in production: every call passes `repoFileList` as `nil`, so the "matched exactly one" and "not found in repo" guards never fire, and a typo'd filename ends up as a runtime 404 instead of a clean `PhaseFailed`. Also worth knowing `path.Match` doesn't cross `/`, so a glob meant to catch shards in a subdir matches nothing. If wiring a real file listing is more than you want here, requiring explicit filenames and dropping the glob path would be cleaner than shipping validation that doesn't run.
- Worth adding one reconcile-level test asserting an invalid `files[]` entry surfaces as `PhaseFailed`/`InvalidFileSet`. The rejection is proven at the `ResolveFileSet` unit level, but not that the controller actually fails the Model.

None of this is far off, and the core is genuinely the right design. Once the mmproj staging/runtime mismatch is consistent and local/PVC either work or fail loud, I'm happy to take it. Thanks again for picking this up.

joryirving:
Pushed a commit to address those. Let me know if you want some further tweaking!

Defilan (approved):
This is exactly the rework I was hoping for, thanks for seeing it through. Both blockers are closed at the right layers, and you picked up the smaller stuff too.

What I checked against the code:

- **mmproj runtime/storage now agree.** `appendMmprojArgs` only fires inside the `ResolveFileSet` success guard, so `--mmproj` can't point at an unstaged path, and `modelStagingPlan` fails closed with an init container that exits 1 with a clear message instead of silently dropping to legacy single-file. The two negative tests (`mmproj without files...`, `mmproj not emitted when ResolveFileSet errors...`) pin it so a refactor can't quietly regress it.
- **Non-HF sources fail loud.** `validateMultiFileStagingSource` routes any `file://`/`pvc://`/local model carrying `files[]`/`mmproj` to `PhaseFailed` + `InvalidFileSet` at reconcile, with envtest coverage.
- **Bonus:** `hf://...@rev` is rejected instead of 404ing on `resolve/main`, and globs hard-error in production rather than dead-validating.

CI is green across the kind/MicroShift/Longhorn matrix. Happy to merge. Nice work unblocking the multimodal staging path for #847.

PR #857, opened 2026-06-26
fix(controller): make model cache writable on fsGroupPolicy=None CSIs (#855)

### LINKED ISSUE #855: [BUG] model-downloader init fails (permission denied) on shared cache when CSI fsGroupPolicy=None (cephfs/NFS)
## Bug Description

When `modelCache` is enabled with a **shared RWX** PVC backed by a CSI whose `fsGroupPolicy` is `None` (e.g. Ceph CephFS via rook-ceph, and many NFS drivers), the `model-downloader` init container fails with `permission denied` and the InferenceService never starts.

The init image is `curlimages/curl`, which runs as **uid 100 (non-root)**. The operator sets the pod `fsGroup` (via `--default-fsgroup`), but a CSI with `fsGroupPolicy: None` **never applies it** to the volume, so the freshly-provisioned cache PVC root stays `root:root` mode `0755`. The non-root downloader then can't create its cache subdirectory:

```
mkdir: can't create directory '/models/<cacheKey>': Permission denied
```

This affects the multi-file / `hf://` staging path (#853) and any cache-backed model on such a cluster. It only works if the whole pod is forced to run as root.

## Steps to Reproduce

1. A cluster whose RWX StorageClass uses a CSI with `fsGroupPolicy: None` (CephFS is the common case — `kubectl get csidriver <driver> -o jsonpath='{.spec.fsGroupPolicy}'` → `None`).
2. Enable the shared cache:
   ```yaml
   modelCache:
     enabled: true
     mode: shared
     accessMode: ReadWriteMany
     storageClass: <cephfs-rwx-class>
   ```
3. Apply a remote-source Model so the init downloads:
   ```yaml
   kind: Model
   metadata: { name: demo }
   spec:
     source: hf://Qwen/Qwen3-Embedding-0.6B-GGUF
     files: [Qwen3-Embedding-0.6B-Q8_0.gguf]
   ```
4. `model-downloader` crashloops with the error above.

## Expected Behavior

The downloader populates the shared cache out-of-the-box, without requiring the pod to run as root, on any RWX CSI — including `fsGroupPolicy: None` ones.

## Actual Behavior

`model-downloader` (uid 100) gets `permission denied` on `mkdir /models/<cacheKey>` because the cache PVC root is `root:root 0755` and the pod's `fsGroup` was never applied by the CSI. InferenceService stays `Init:Error`.

## Environment

- **LLMKube:** 0.8.18
- **Kubernetes:** v1.36.2 (Talos)
- **Cluster Type:** Other: Talos / on-prem
- **Storage:** rook-ceph CephFS, RWX shared cache
- **GPU:** N/A — the failure is in the init container, independent of accelerator

CSI driver `fsGroupPolicy`:
```
cephfs.csi.ceph.com / rook-ceph.cephfs.csi.ceph.com : None
rbd.csi.ceph.com    / rook-ceph.rbd.csi.ceph.com    : File   # but block = RWO, can't back a multi-node shared cache
```

## Logs

`model-downloader` init:
```
mkdir: can't create directory '/models/60b3650478843845': Permission denied
```

Operator-generated security contexts:
- **Pod:** `fsGroup: <default-fsgroup>`, `seccompProfile: RuntimeDefault`, **no `runAsUser`**
- **Init container:** `allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]`, `readOnlyRootFilesystem: false`, **no `runAsUser`** → runs as the curl image default uid 100

## Additional Context

**Root cause:** `fsGroup` is the only thing wiring the cache volume to a non-root writer, and `fsGroupPolicy: None` CSIs ignore it. The chart docs recommend an RWX class for multi-node shared caches, but the two most common RWX backends (CephFS, NFS) frequently ship `fsGroupPolicy: None`, so the default non-root downloader can't write.

**Workaround:** set `podSecurityContext.runAsUser: 0` on the InferenceService so the init runs as root (root owns the volume → owner-write succeeds even with all caps dropped). Confirmed working, but it forces the whole pod to root.

**Proposed fixes (rough order of preference):**
1. Inject a tiny **root-run prep init** (`chown`/`chmod` the cache mount to the pod `fsGroup`/downloader uid) **before** the non-root `model-downloader` — the standard pattern for fsGroup-ignoring volumes, keeps downloader + server unprivileged.
2. Make the `model-downloader` `runAsUser` configurable (chart value and/or Model/InferenceService field) so the init can be root without the *server* running as root.
3. At minimum, document the `fsGroupPolicy: None` caveat in the modelCache docs with the `runAsUser: 0` workaround.

Happy to send a PR for (1) if that's the preferred direction.

Defilan:
## What

Adds a root-run `model-cache-permissions` init container before `model-downloader` in the cache-backed storage path. It chowns/chmods the cache mount root so the non-root downloader (uid 100) can write on CSIs that ignore `fsGroup` (CephFS/NFS).

## Why

On a `fsGroupPolicy: None` CSI the kubelet never applies the pod `fsGroup` to the volume, so a freshly-provisioned cache PVC root stays `root:root 0755` and the non-root `model-downloader` gets `permission denied` on `mkdir`, leaving the InferenceService stuck at `Init:Error`. Affects shared RWX caches on the two most common RWX backends (CephFS, NFS), which frequently ship `fsGroupPolicy: None`. Fixes #855.

## How

The prep container chowns the mount root **non-recursively** (so a shared cache does not get every other model's files rewritten on each pod start) to the **resolved** pod fsGroup:

- `isvc.Spec.PodSecurityContext.FSGroup` when the InferenceService sets it, else the operator default `r.DefaultFSGroup` (the `--default-fsgroup` flag, default 102). The resolution mirrors `inferPodSecurityContext` so the chown target matches the fsGroup the pod actually runs with (this is load-bearing: chowning to the wrong group leaves the downloader, which carries fsGroup as a supplemental group, unable to write).
- When fsGroup is disabled (`<= 0`), there is no supplemental group, so it chowns to the downloader's own uid (100) instead.

It reuses the configurable `initContainerImage` (air-gap friendly, no hardcoded busybox) and runs least-privilege: `runAsUser 0` (required to chown a root-owned mount), drop `ALL` and add back only `CHOWN`/`FOWNER`, `allowPrivilegeEscalation: false`, `readOnlyRootFilesystem: true`, seccomp `RuntimeDefault`. The downloader and llama-server containers stay unprivileged.

## Checklist

- [x] `go build`, `go vet`, `gofmt`, `golangci-lint` clean
- [x] Unit tests: ordering before downloader; default fsGroup 102; explicit isvc override (3000 over default 102); fsGroup-disabled -> uid 100; image reuse; hardened SecurityContext; not emitted for emptyDir
- [ ] Live validation on a real `fsGroupPolicy: None` CSI (CephFS): manual, follow-up. The init-container path is not reproducible in CI.

Note: the prep is root-by-necessity (chowning a root-owned mount), so it does not satisfy PSA `restricted` namespaces; that constraint is inherent to the fix.

### DIFF
--- a/internal/controller/deployment_builder.go
+++ b/internal/controller/deployment_builder.go
@@ -247,7 +247,7 @@ func (r *InferenceServiceReconciler) constructDeployment(
 	var modelPath string
 	if backend.NeedsModelInit() && !skipInit {
 		useCache := model.Status.CacheKey != "" && r.ModelCachePath != ""
-		storageConfig = buildModelStorageConfig(model, isvc, isvc.Namespace, useCache, r.ModelCacheMode, r.CACertConfigMap, r.InitContainerImage)
+		storageConfig = buildModelStorageConfig(model, isvc, isvc.Namespace, useCache, r.ModelCacheMode, r.CACertConfigMap, r.InitContainerImage, r.DefaultFSGroup)
 		modelPath = storageConfig.modelPath
 	}
 
--- a/internal/controller/inferenceservice_controller.go
+++ b/internal/controller/inferenceservice_controller.go
@@ -73,6 +73,11 @@ func sanitizeDNSName(name string) string {
 
 func boolPtr(b bool) *bool { return &b }
 
+// int64Ptr returns a pointer to the given int64. Used to construct
+// *int64 fields on SecurityContext / PodSecurityContext from a literal value
+// without an intermediate variable.
+func int64Ptr(v int64) *int64 { return &v }
+
 // initContainerSecurityContext returns the SecurityContext applied to the
 // model downloader init container. It inherits runAsUser/runAsGroup from the
 // pod-level Spec.PodSecurityContext when the user supplied one. The default
--- a/internal/controller/inferenceservice_storage_test.go
+++ b/internal/controller/inferenceservice_storage_test.go
@@ -31,7 +31,7 @@ import (
 )
 
 var _ = Describe("buildCachedStorageConfig", func() {
-	It("should configure PVC volume and init container for remote model", func() {
+	It("should configure PVC volume and init containers for remote model", func() {
 		model := &inferencev1alpha1.Model{
 			Spec: inferencev1alpha1.ModelSpec{
 				Source: "https://example.com/model.gguf",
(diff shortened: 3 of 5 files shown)


Defilan:
Superseded by #858. This branch was cut from a stale fork main (before #853 merged) and conflicts with the current buildCachedStorageConfig; #858 is rebased on current main and integrates the prep into both the single-file and multi-file download paths.

PR #866, opened 2026-06-27
feat: add spec.mode to auto-serve embedding and rerank models

eleboucher:
## Summary

Adds a `spec.mode` field (`chat` | `embedding` | `rerank`) to `InferenceService` so embedding and reranker models are first-class instead of requiring hand-wired runtime flags. Today, serving an embedding or reranker model means knowing and passing the right llama.cpp flags (`--embedding --pooling last`, `--reranking --pooling rank`, plus the reranker also needs `--embedding`) via `spec.extraArgs`, and nothing surfaces the model's task type to consumers.

## What it does

- **`spec.mode`** (enum `chat|embedding|rerank`, intentionally **undefaulted**): for the `llamacpp` runtime, setting it auto-appends the required flags — `embedding` → `--embedding --pooling last`; `rerank` → `--reranking --embedding --pooling rank`. Any flag already present in `spec.extraArgs` wins and is not duplicated.
- **`status.mode`**: the resolved mode — `spec.mode` when set, otherwise inferred from `spec.extraArgs` / `spec.endpoint.path` — so downstream routers and external operators can key off model type without parsing runtime flags.
- **Backward compatible**: existing manifests that wire `--embedding`/`--reranking` by hand keep working unchanged, and their mode is still inferred and published to `status.mode`. The field is left undefaulted on purpose so a `chat` default cannot clobber that inference.

## Resolution precedence

`spec.mode` → runtime flags (`--reranking` wins over `--embedding`, since rerankers pass both) → `spec.endpoint.path` (`/v1/rerank`, `/v1/embeddings`) → `chat`.

## Tests

`resolveServingMode` and `appendModeArgs` table tests in `internal/controller/runtime_mode_test.go` (default, explicit-wins, explicit-over-conflicting-flags, flag inference, rerank-over-embedding, generic `Args`, endpoint path, no-duplicate, chat-noop). `make manifests generate` regenerated the CRD; build, controller tests, and golangci-lint pass.

## Motivation

This unblocks routing embedding/reranker `InferenceService`s through an OpenAI-compatible proxy (e.g. LiteLLM): `status.mode` maps directly to litellm's `model_info.mode`, so the proxy serves `/v1/embeddings` and `/v1/rerank` correctly instead of treating every model as chat.

### DIFF
--- a/api/v1alpha1/inferenceservice_types.go
+++ b/api/v1alpha1/inferenceservice_types.go
@@ -21,6 +21,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// Serving modes for InferenceService spec.mode and status.mode.
+const (
+	ServingModeChat      = "chat"
+	ServingModeEmbedding = "embedding"
+	ServingModeRerank    = "rerank"
+)
+
 // InferenceServiceSpec defines the desired state of InferenceService
 // RopeScalingType selects the RoPE context-extension method. Mirrors
 // llama.cpp's --rope-scaling values.
@@ -88,6 +95,17 @@ type InferenceServiceSpec struct {
 	// +kubebuilder:default=llamacpp
 	// +optional
 	Runtime string `json:"runtime,omitempty"`
+
+	// Mode selects how the model is served: "chat" (default) for
+	// chat/completion, "embedding" for /v1/embeddings, "rerank" for /v1/rerank.
+	// For the llamacpp runtime it auto-appends the required flags (embedding ->
+	// --embedding --pooling last; rerank -> --reranking --embedding --pooling
+	// rank); any flag already set in spec.extraArgs wins. When unset, the mode is
+	// inferred from spec.extraArgs / spec.endpoint.path. The resolved value is
+	// always reported in status.mode.
+	// +kubebuilder:validation:Enum=chat;embedding;rerank
+	// +optional
+	Mode string `json:"mode,omitempty"`
 
 	// Replicas is the desired number of inference pods
 	// +kubebuilder:validation:Minimum=0
@@ -815,6 +833,11 @@ type InferenceServiceStatus struct {
 	// +kubebuilder:validation:Enum=Pending;Creating;Progressing;Ready;WaitingForGPU;Stopped;Failed
 	// +optional
 	Phase string `json:"phase,omitempty"`
+
+	// Mode is the resolved serving mode (chat, embedding, or rerank): spec.mode
+	// when set, otherwise inferred from the runtime flags and endpoint path.
+	// +optional
+	Mode string `json:"mode,omitempty"`
 
 	// Replicas tracks the number of ready vs desired pods
 	// +optional
--- a/charts/llmkube/templates/crds/inferenceservices.yaml
+++ b/charts/llmkube/templates/crds/inferenceservices.yaml
@@ -1476,6 +1476,20 @@ spec:
                 items:
                   type: string
                 type: array
+              mode:
+                description: |-
+                  Mode selects how the model is served: "chat" (default) for
+                  chat/completion, "embedding" for /v1/embeddings, "rerank" for /v1/rerank.
+                  For the llamacpp runtime it auto-appends the required flags (embedding ->
+                  --embedding --pooling last; rerank -> --reranking --embedding --pooling
+                  rank); any flag already set in spec.extraArgs wins. When unset, the mode is
+                  inferred from spec.extraArgs / spec.endpoint.path. The resolved value is
+                  always reported in status.mode.
+                enum:
+                - chat
+                - embedding
+                - rerank
+                type: string
               modelRef:
                 description: ModelRef references the Model CR that contains the model
                   to serve
@@ -3246,6 +3260,11 @@ spec:
               lastUpdated:
                 description: LastUpdated is the timestamp of the last status update
                 format: date-time
+                type: string
+              mode:
+                description: |-
+                  Mode is the resolved serving mode (chat, embedding, or rerank): spec.mode
+                  when set, otherwise inferred from the runtime flags and endpoint path.
                 type: string
               modelReady:
                 description: ModelReady indicates if the referenced Model is in Ready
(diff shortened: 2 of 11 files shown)


Defilan:
### Code review

Nice feature: making embedding/rerank first-class via `spec.mode` is a clean improvement over hand-wiring flags. Found 2 issues:

1. The Helm chart CRD was not regenerated, so it is missing the new `spec.mode` / `status.mode` fields (`AGENTS.md` requires `make chart-crds` after editing CRD types in `api/v1alpha1/`). The CI `CRD Sync Check` is failing on this. Fix: run `make chart-crds` and commit the updated `charts/llmkube/templates/crds/inferenceservices.yaml`.

https://github.com/defilantech/LLMKube/blob/6d931124b3ceb3b5d81888055e4000c534329ebc/api/v1alpha1/inferenceservice_types.go#L91-L101

2. `spec.mode` is applied only on the in-cluster controller path (`LlamaCppBackend.BuildArgs`), not on the off-cluster metal-agent. The metal-agent builds llama-server args through its own `buildExecutorConfig` -> `buildLlamaServerArgs` path, which never reads `Spec.Mode`, and `computeSpecHash` does not include it either. On Apple Silicon nodes, `spec.mode: embedding` would serve a plain chat endpoint while `status.mode` reports `embedding` (a silent mismatch; a mode-only change also would not respawn the process). The codebase already documents this mirroring contract for the cache-type fields. Suggest mirroring the mode flags into the metal-agent path (add `Mode` to `ExecutorConfig`, map it in `buildExecutorConfig`, apply it in `buildLlamaServerArgs`, and add it to `computeSpecHash`).

Controller-only wiring:
https://github.com/defilantech/LLMKube/blob/6d931124b3ceb3b5d81888055e4000c534329ebc/internal/controller/runtime_llamacpp.go#L117-L121

Metal-agent path that omits it (note the mirror convention in the comment above):
https://github.com/defilantech/LLMKube/blob/6d931124b3ceb3b5d81888055e4000c534329ebc/pkg/agent/agent.go#L566-L618

Defilan:
Thanks for this @eleboucher, making embedding/rerank first-class is a genuinely useful addition.

Two quick CI fixes to get it green:
- `make chart-crds` and commit the updated `charts/llmkube/templates/crds/inferenceservices.yaml` (the chart CRD has to be regenerated alongside `config/crd/bases`).
- `git commit --amend -s` to add the DCO sign-off.

On the metal-agent point from the review: that's a known systemic seam, not something you should have to reverse-engineer. We just filed #867 to add a parity guard so this stops being a footgun. If you're up for finishing it in this PR, it's four spots:
1. `ExecutorConfig` (`pkg/agent/executor.go`) - add a `Mode` field.
2. `buildExecutorConfig` (`pkg/agent/agent.go`) - map `isvc.Spec.Mode` onto it.
3. `buildLlamaServerArgs` (`pkg/agent/executor.go`) - apply the flags, mirroring `appendModeArgs`.
4. `computeSpecHash` (`pkg/agent/agent.go`) - include `Mode` so a mode-only change respawns the process.

Happy to pair on it if that's easier. And if you'd rather keep this PR controller-only, just say the word and we'll handle the metal side as a fast-follow.

eleboucher:
Thanks for the thorough review — all addressed, and I took the metal-agent parity since it's the part that actually makes the feature correct.

**CI fixes**
- `make chart-crds` run; `charts/llmkube/templates/crds/inferenceservices.yaml` now carries `mode` (spec + status), regenerated alongside `config/crd/bases`.
- Amended with `-s` for the DCO sign-off.

**Metal-agent parity (#867), all four spots**
- `ExecutorConfig.Mode` (executor.go)
- `buildExecutorConfig` maps `isvc.Spec.Mode` (agent.go)
- `buildLlamaServerArgs` applies the flags via a local `appendModeArgs`, mirroring the controller (embedding → `--embedding --pooling last`; rerank → `--reranking --embedding --pooling rank`), skipping anything already in `extraArgs`
- `computeSpecHash` now includes `Mode`, so a mode-only edit changes the hash and respawns the process

One small thing to flag: to avoid the mode strings drifting between the controller and agent paths (the footgun you mentioned), I hoisted the three values into exported constants in the api package (`ServingModeChat/Embedding/Rerank`) and pointed both sides at them. The arg-building itself is still mirrored per-path, consistent with how the agent already duplicates the other flag builders.

Tests: `appendModeArgs`/`resolveServingMode` table tests on the controller side, and `buildLlamaServerArgs` mode tests on the agent side (embedding, rerank, extraArgs-not-overridden, chat-noop). Ran the controller envtest suite and the agent tests locally; green.

Happy to split the api-constants bit out or adjust naming if you'd prefer.

Defilan:
### Re-review

Both findings from the earlier review are resolved and verified, and CI is green. Nice work.

- **Chart CRD synced** (`charts/llmkube/templates/crds/inferenceservices.yaml` now carries `spec.mode` + `status.mode`); CRD Sync Check passes.
- **Metal-agent now honors `spec.mode`**: `ExecutorConfig.Mode` + `buildExecutorConfig` mapping, `appendModeArgs` wired into `buildLlamaServerArgs`, and `Mode` is in `computeSpecHash` so a mode change actually respawns the metal process. The metal-agent `appendModeArgs`/`hasMatchingExtraArg` are behaviorally identical to the controller's. No divergence.

This is good to merge from my side. Two optional, non-blocking follow-ups (happy to take them as separate issues so they don't hold up this PR):

1. **`status.mode` on non-llamacpp runtimes.** `resolveServingMode` sets `status.mode` for every runtime, but flag injection is llamacpp-only. A `runtime: vllm` (or tgi/generic) service with `spec.mode: embedding` would report `embedding` in status while serving chat. Consider gating the `status.mode` write to the llamacpp backend, or a webhook/doc note that other runtimes ignore the field.
2. **Respawn test.** The load-bearing guarantee (a mode change respawns the metal process via `computeSpecHash`) has no test. A `TestComputeSpecHash_ChangesWithMode` mirroring `TestComputeSpecHash_ChangesWithContextSize` would lock it in cheaply.

Heads up on the duplication: the metal-agent now carries a verbatim copy of `appendModeArgs`/`hasMatchingExtraArg`. That is structurally unavoidable today (`pkg/agent` can't import `internal/controller`), and we track the shared-helper refactor in #867; no action needed here.

Defilan (approved):
Alt Text

PR #888, opened 2026-06-28
fix(controller): run model-cache-prep init as root again (0.8.20 regression)

### LINKED ISSUE #887: [BUG] model-cache-prep init errors out (chown EPERM) on 0.8.20 with fsGroupPolicy=None CSIs
## Bug Description

On 0.8.20 the `model-cache-prep` init container fails on CSIs with `fsGroupPolicy: None` (CephFS, NFS): the `chown` errors with `Operation not permitted` and the init container crashes, so cached-model InferenceServices never start. Reported by a user upgrading to 0.8.20. Regression from #883 (commit a8049f1c).

## Root Cause

#883 changed the prep init's `RunAsUser` from `0` to `100` (non-root) while keeping `capabilities.add: [CHOWN, FOWNER]`. The container runs `sh -c "chown 0:<fsGroup> /models && chmod ..."`.

Capabilities added to a **non-root** container are not effective for processes the shell `exec`s: in containerd a non-root process clears its capabilities across `execve` (Kubernetes has no ambient-capability field, and containerd does not set them), so the exec'd `chown` runs with an empty effective set and fails `EPERM`. As **root (uid 0)**, capabilities are retained across exec, so chown succeeds. The init must run as root.

## Steps to Reproduce

1. Cluster with an `fsGroupPolicy: None` CSI (CephFS / NFS), `modelCache` enabled (`shared` or `perService`).
2. Deploy an InferenceService that uses the cache on 0.8.20.
3. Observe the `model-cache-prep` init container: `chown: /models: Operation not permitted`, CrashLoopBackOff.

## Expected Behavior

The prep chowns/chmods the mount and the downloader starts (0.8.19 behavior).

## Fix

Revert `RunAsUser` to `0` (drop `RunAsNonRoot`), keeping `Drop: [ALL]` + `Add: [CHOWN, FOWNER]` + `allowPrivilegeEscalation: false` + read-only rootfs + seccomp. Non-privileged root with two minimal caps. Patch in 0.8.21.

## Workaround (pre-patch)

Pin 0.8.19, or use an `fsGroupPolicy: File` CSI / `emptyDir` model store. The init's securityContext is not overridable via values.

Note: `privileged` is NOT required and never was.

Defilan:
## What

Revert the `model-cache-prep` init container to `RunAsUser: 0`, fixing a crash introduced in 0.8.20.

## Why

Fixes #887 (regression from #883 / 0.8.20).

#883 ran the prep non-root (`RunAsUser: 100`) with `capabilities.add: [CHOWN, FOWNER]`. That does not work in Kubernetes: the container runs `sh -c "chown ... && chmod ..."`, and in containerd a **non-root** process clears its capabilities across `execve` (there is no ambient-capability field in the k8s securityContext, and containerd does not set them). So the `chown` the shell execs runs with an empty effective capability set and fails `EPERM`, crashing the init container on `fsGroupPolicy: None` CSIs (CephFS, NFS) -- the exact case the prep exists for. As **root**, capabilities are retained across exec, so chown succeeds.

This was the precise risk flagged in the #883 review ("envtest validates the spec, not a live chown") that did not get the belt-and-suspenders cluster check.

## How

- `internal/controller/model_storage.go`: `RunAsUser: 100 -> 0`, remove `RunAsNonRoot: true`. Keep `Drop: [ALL]` + `Add: [CHOWN, FOWNER]` + `allowPrivilegeEscalation: false` + read-only rootfs + seccomp (non-privileged root, two minimal caps). Doc comment explains why root is required (cap-on-exec).
- `inferenceservice_storage_test.go`: assert `RunAsUser == 0` with a regression-guard comment ("do not flip non-root without ambient caps").
- `docs/MODEL-CACHE.md`: correct the security note -- non-root + caps cannot work without ambient capabilities; the init runs as non-privileged root.

`privileged` is not required and never was.

## Testing

- Build + vet clean.
- `go test ./internal/controller/` cache-prep suite passes with the `RunAsUser == 0` assertion.
- `make lint`: 0 issues.

## Release

Needs a **0.8.21** patch. Workaround for users on 0.8.20: pin 0.8.19, or use an `fsGroupPolicy: File` CSI / `emptyDir` store.

## Checklist

- [x] Tests pass
- [x] Lint passes
- [x] Commit signed off (DCO)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -129,3 +129,16 @@ deployment config).
   and `charts/`). Change the source and regenerate.
 - Do not commit secrets, kubeconfigs, or `.env` files.
 - Do not skip the pre-commit checks or the CRD sync step.
+- Do not run an init/main container that performs a privileged syscall (`chown`,
+  `chmod` of an unowned file, `mount`, etc.) non-root with `capabilities.add`
+  when the syscall runs in an exec'd child (`command: ["sh", "-c", "chown ..."]`).
+  A non-root process loses its capabilities across `execve` unless the runtime
+  sets ambient capabilities, which is runtime-dependent (it works on kind but
+  not on every cluster, so envtest and CI will pass while real users hit
+  `EPERM`). Either run the container as root (uid 0, still `Drop: [ALL]` + the
+  minimal caps + no privilege escalation), or do the syscall directly in the
+  container's entrypoint (a small helper binary) so the caps are not cleared.
+  This caused the 0.8.20 `model-cache-prep` regression (#887). Any change to an
+  init/main container `securityContext`, `command`, `image`, or user that affects
+  a privileged syscall must be reasoned about at the runtime level, not just by
+  the rendered-spec envtest.
--- a/docs/MODEL-CACHE.md
+++ b/docs/MODEL-CACHE.md
@@ -315,15 +315,19 @@ This will revert to the legacy behavior where each pod downloads the model via i
 
 ## Security Considerations
 
-### PSA `restricted` incompatibility
+### Automatic shared cache on `fsGroupPolicy: None` backends (CephFS, NFS)
 
-The model cache init container cannot satisfy Pod Security Admission (PSA) `restricted` policy. PSA `restricted` forbids both running as root and adding any capability except `NET_BIND_SERVICE`. Chowning a root-owned mount (which the cache-prep init container does on CSIs with `fsGroupPolicy: None`) requires `CAP_CHOWN`/`CAP_FOWNER`, and the init must run as **root (uid 0)** to use them: it shells out to `chown`, and a non-root process clears its capabilities across `execve` (Kubernetes has no ambient-capability field, and containerd does not set them), so the exec'd `chown` would fail with `EPERM`. The init therefore runs as non-privileged root with only `CHOWN`+`FOWNER`, which cannot satisfy `restricted` no matter how it is tuned. (Running it non-root broke the cache prep in 0.8.20; reverted in 0.8.21.)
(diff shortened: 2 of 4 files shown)


Defilan:
@joryirving please have a look at this

joryirving:
`#888` looks right to me as the immediate fix for the 0.8.20 regression.

For the longer-term direction, I would frame this as a storage-compatibility problem, not a “users should avoid shared cache” problem.

I use CephFS-backed shared cache specifically because I want the automatic model-cache behavior. I do not want the long-term answer to be “pre-stage models yourself” or “manually manage your own PVC”.

What I’d rather see as the follow-up direction:

- Keep the automatic shared-cache workflow as a first-class supported path, including on `fsGroupPolicy: None` backends like CephFS/NFS.
- Treat the current `uid 0` + minimal capabilities (`CHOWN`, `FOWNER`, no privilege escalation, read-only rootfs, seccomp) prep as the compatibility implementation for those backends.
- Make the docs explicit that this is a backend-driven requirement, not an avoidable user misconfiguration.
- Add a real cluster/e2e test that exercises the shared cache on an `fsGroupPolicy: None` CSI backend so this does not regress again.

If there is appetite for a follow-up implementation, I think the cleaner improvement would be to replace the current `sh -c "chown ... && chmod ..."` prep with a tiny dedicated helper that performs the ownership fix directly, while preserving the same automatic shared-cache UX.

So my read is:

- `#888` should merge as the patch fix.
- The longer-term goal should be “automatic shared cache remains supported on CephFS/NFS-class backends, with stronger implementation/tests”, not “push those users onto manual PVC management”.

PR #912, opened 2026-06-30
fix: cache hf:// multi-file models instead of emptyDir

### LINKED ISSUE #909: [BUG] hf:// source + spec.files/mmproj strands the model on emptyDir instead of the shared cache (re-downloads every restart)
## Summary

A `Model` with an `hf://<org>/<repo>` source **and** explicit `spec.files` (and/or
`spec.mmproj`) is classified as runtime-resolved and assigned an empty
`Status.CacheKey`. The generated InferenceService Deployment then falls back to an
ephemeral `emptyDir` instead of the shared model-cache PVC, even though `modelCache`
is enabled and the file list clearly asks for cache staging. The full model is
re-downloaded into the emptyDir on every pod restart, and the controller emits a
continuous stream of `MissingSkipModelInit` warnings.

Switching the same model to `source: https://huggingface.co/<org>/<repo>` (identical
repo, different scheme) restores caching, which points squarely at the scheme-based
classification.

## Environment

- Operator / chart: v0.8.22
- `modelCache`: enabled, mode `shared` (RWX PVC)

## Root cause

`reconcileBySourceType` (`internal/controller/model_controller.go`) routes by scheme:

```go
case isHFRepoSource(model.Spec.Source):      // hf://org/repo
    result, err = r.reconcileRuntimeResolvedSource(ctx, model, "")                       // cacheKey = ""
case isRemoteHTTPSource(model.Spec.Source):  // https://...
    result, err = r.reconcileRuntimeResolvedSource(ctx, model, computeCacheKey(model.Spec.Source))
```

The `hf://` branch leaves `Status.CacheKey` empty. The deployment builder gates
caching on exactly that field (`internal/controller/deployment_builder.go`):

```go
useCache := model.Status.CacheKey != "" && r.ModelCachePath != ""
```

So `useCache=false` -> `buildModelStorageConfig` returns `buildEmptyDirStorageConfig`.

But `buildCachedStorageConfig` already contains a complete multi-file staging branch
(`modelStagingPlan` / `buildMultiFileInitCommand`) that handles `hf://` sources
correctly — `multiFileInitEnvVars` normalizes `hf:// -> https://huggingface.co/` and
the init builds `${SOURCE%/}/resolve/main/$rel`. That branch is simply never reached
for `hf://` sources because the empty-CacheKey gate fails first.

The controller already understands that `hf://` + files is a staging case
(`validateMultiFileStagingSource` calls `isHFRepoSource`, and `hasMultiFileStaging`
exists) — only the `CacheKey` assignment is missing on this path.

## Impact

- The shared cache PVC is provisioned but mounted by nothing; weights are re-pulled
  into emptyDir on every restart.
- Continuous `MissingSkipModelInit` warning events on every reconcile (which then also
  trip the events RBAC gap filed as a sibling issue).

## Repro

1. `modelCache` enabled (shared).
2. Apply a `Model` + referencing `InferenceService`:

```yaml
kind: Model
metadata:
  name: demo
spec:
  source: hf://<org>/<repo>-GGUF
  files:
    - model-Q8_0.gguf
  format: gguf
```

3. Observe the generated Deployment: volume `model-storage` is an `emptyDir`, a
   `model-downloader` init writes into it, and the shared `llmkube-model-cache` PVC is
   unmounted. `Model.status.cacheKey` is empty; the Available condition reason is
   `RuntimeResolved`.

**Workaround:** use `source: https://huggingface.co/<org>/<repo>` — `isRemoteHTTPSource`
short-circuits `isHFRepoSource`, a CacheKey is assigned, and the cached multi-file
staging path runs.

## Suggested fix

In `reconcileBySourceType`, when `hasMultiFileStaging(model)` is true, assign
`computeCacheKey(model.Spec.Source)` on the `hf://` branch as well (treat it like the
`https://` path), so `useCache` is true and the cached multi-file staging branch runs.
Pure runtime-resolved sources (no `files`/`mmproj`, e.g. vLLM fetching a repo at
startup) keep the empty-key behavior.

Defilan:
## What

Make hf:// Models that use `spec.files`/`spec.mmproj` (multi-file staging) use the shared model cache PVC instead of falling back to an `emptyDir` volume.

## Why

Fixes #909

A user reported that all their multi-file hf:// models swapped to `emptyDir`, re-downloading every declared file on each pod restart. Root cause: these models reach `Ready` via the runtime-resolved path, which deliberately leaves `Status.CacheKey` empty and keeps the Available condition reason `RuntimeResolved`. But both the PVC-ensure gate (`InferenceService` reconcile) and the deployment builder (`useCache`) keyed the "use the cache PVC?" decision off `Status.CacheKey` alone, so a keyless-but-multi-file model got an `emptyDir`.

A naive fix that mints a `Status.CacheKey` in the controller breaks the `RuntimeResolved` condition contract (an existing test guards it). This fixes the decision at the storage layer instead, leaving the controller's condition semantics untouched.

## How

- Add `effectiveModelCacheKey(model)` as the single source of truth for the cache decision: returns `Status.CacheKey` when the controller set one, otherwise derives a stable key from the source for **non-metal** multi-file models.
- Route all three deciders through it so they can no longer disagree:
  - `InferenceServiceReconciler` PVC-ensure gate
  - `deployment_builder` `useCache`
  - `buildCachedStorageConfig` cache dir
- Metal models keep no cache PVC (the host metal-agent loads files directly); the controller's `RuntimeResolved` condition is unchanged.
- Tests: unit coverage for `effectiveModelCacheKey` (controller key, runtime-resolved single-file, hf:// multi-file derivation, metal exclusion) and a `buildCachedStorageConfig` case asserting the keyed cache dir. Full `internal/controller` suite (469 specs) green, including the `RuntimeResolved` multi-file test.

## AI assistance

Assisted by an AI coding agent under human review (band 3 per CONTRIBUTING.md). The author reviewed, tested, and is accountable for all changes.

## Checklist

- [x] Tests added/updated
- [x] `make lint` passes locally (incl. `GOOS=linux`)
- [ ] Documentation updated (no user-facing surface change)

### DIFF
--- a/internal/controller/deployment_builder.go
+++ b/internal/controller/deployment_builder.go
@@ -246,7 +246,7 @@ func (r *InferenceServiceReconciler) constructDeployment(
 	var storageConfig modelStorageConfig
 	var modelPath string
 	if backend.NeedsModelInit() && !skipInit {
-		useCache := model.Status.CacheKey != "" && r.ModelCachePath != ""
+		useCache := effectiveModelCacheKey(model) != "" && r.ModelCachePath != ""
 		storageConfig = buildModelStorageConfig(model, isvc, isvc.Namespace, useCache, r.ModelCacheMode, r.CACertConfigMap, r.InitContainerImage, r.DefaultFSGroup)
 		modelPath = storageConfig.modelPath
 	}
--- a/internal/controller/inferenceservice_controller.go
+++ b/internal/controller/inferenceservice_controller.go
@@ -152,7 +152,7 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		desiredReplicas = *inferenceService.Spec.Replicas
 	}
 
-	if model.Status.CacheKey != "" && r.ModelCachePath != "" {
+	if effectiveModelCacheKey(model) != "" && r.ModelCachePath != "" {
 		if err := r.ensureModelCachePVC(ctx, inferenceService); err != nil {
 			log.Error(err, "Failed to ensure model cache PVC exists", "namespace", inferenceService.Namespace)
 			return r.updateStatusWithSchedulingInfo(ctx, inferenceService, PhaseFailed, modelReady, 0, desiredReplicas, "", "Failed to create model cache PVC", nil)
--- a/internal/controller/inferenceservice_storage_test.go
+++ b/internal/controller/inferenceservice_storage_test.go
@@ -357,6 +357,68 @@ var _ = Describe("hasMultiFileStaging", func() {
 	})
 })
 
+var _ = Describe("effectiveModelCacheKey", func() {
+	It("returns the controller-set Status.CacheKey when present", func() {
+		model := &inferencev1alpha1.Model{
+			Spec:   inferencev1alpha1.ModelSpec{Source: "https://example.com/model.gguf"},
+			Status: inferencev1alpha1.ModelStatus{CacheKey: "controllerkey"},
+		}
+		Expect(effectiveModelCacheKey(model)).To(Equal("controllerkey"))
+	})
+
+	It("returns empty for a runtime-resolved single-file model with no cache key", func() {
+		model := &inferencev1alpha1.Model{
+			Spec: inferencev1alpha1.ModelSpec{Source: "hf://unsloth/gemma-4-31B-it-GGUF"},
+		}
+		Expect(effectiveModelCacheKey(model)).To(BeEmpty())
+	})
+
+	It("derives a stable key from the source for an hf:// multi-file model (#909)", func() {
+		// The runtime-resolved path leaves Status.CacheKey empty for hf://
+		// sources, but multi-file staging must still land in the cache PVC.
+		source := "hf://unsloth/gemma-4-31B-it-GGUF"
+		model := &inferencev1alpha1.Model{
+			Spec: inferencev1alpha1.ModelSpec{
+				Source: source,
+				Files:  []string{"a.gguf", "MTP/b.gguf"},
+				Mmproj: "mmproj-F16.gguf",
+			},
+		}
+		Expect(effectiveModelCacheKey(model)).To(Equal(computeCacheKey(source)))
+	})
+
+	It("does not cache a metal multi-file model (no init container / cache PVC)", func() {
+		model := &inferencev1alpha1.Model{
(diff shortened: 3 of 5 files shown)


Defilan:
Hey @joryirving please give this a quick once over

joryirving (commented):
Reviewed the diff (didn't run the suite locally). The approach looks correct and is nicer than minting a `Status.CacheKey` in the controller — centralizing on `effectiveModelCacheKey` and routing the three deciders (PVC-ensure gate, `useCache`, `cacheDir`) through it fixes the storage decision while preserving the `RuntimeResolved` condition contract. It also keeps the source `hf://`, so it still passes `validateMultiFileStagingSource` — which is exactly what made a naive `hf://`->`https://` swap fail with `InvalidFileSet` (multi-file staging only accepts hf:// sources). Test coverage hits the right cases.

One non-blocking follow-up: the decision is now centralized in the controller, but the CLI still reads `Status.CacheKey` directly, and the two sites diverge for the exact models this PR now caches:

- `pkg/cli/cache.go` already falls back to `computeCacheKey(source)` (~L199-202), so `cache list` shows these entries correctly.
- `pkg/cli/delete.go --purge-cache` (~L104) reads `Status.CacheKey` with no fallback. For an `hf://` multi-file model it prints "Model has no cache key; nothing to purge" and skips it, leaving an orphaned cache dir under the derived key. It is guarded against an empty-key `/models/` wipe (advisory print only), so this is a consistency/UX gap rather than a safety issue.

Might be worth routing `delete.go` (and `cache.go`) through the same key logic so serve/list/purge never disagree — probably lifting `effectiveModelCacheKey` out of `internal/controller`, since `computeCacheKey` is already shared. Happy to file a separate issue for that so this PR stays focused.

Thanks for the quick turnaround on both.

Defilan:
Good catch on the purge path, thanks. You're right that it's a direct consequence of this change: we now cache these models but `delete --purge-cache` couldn't clean them.

Pushed a minimal fix (1442e55): `delete.go` now uses the same `Status.CacheKey` -> `computeCacheKey(source)` fallback that `cache list` already has, so serve/list/purge agree on the key for hf:// multi-file models.

Filed #921 for the fuller cleanup you suggested: lift `effectiveModelCacheKey` (and `computeCacheKey`, which is currently duplicated) into a shared package so all three sites route through one scoped function, instead of the CLI's looser unconditional fallback. Kept it out of this PR to stay focused.

Issue #911, opened 2026-06-30
Extend drain-before-roll idle checks to vLLM, TGI, and multi-replica services


itsmiso-ai:
## Follow-up to #856

#856 adds controller-level drain-before-roll support for the llama.cpp-compatible `/slots` path, which covers the single-replica local-model case that motivated the original issue.

This follow-up tracks broadening the idle detector so `spec.rolloutPolicy.waitForIdle` can make runtime-specific decisions beyond that first implementation.

## Scope

- Add runtime-specific idle detectors for vLLM and TGI using their exposed metrics or health endpoints.
- Define a generic fallback/contract for custom runtimes, for example an annotation or spec field declaring an idle endpoint and parser.
- For multi-replica Services, avoid relying on a single Service load-balanced request; inspect each serving pod or endpoint so rollout waits until every current replica is idle.
- Document which runtimes support drain-before-roll and what signal each uses.

## Why separate from #856

The original issue is primarily about avoiding dropped generations for single-replica llama.cpp workloads. vLLM/TGI/generic support is valuable, but it needs runtime-specific signal validation and likely per-pod endpoint iteration, so it is better reviewed as a focused follow-up rather than expanding the initial PR.

PR #913, opened 2026-06-30
feat(controller): defer rollouts until idle

### LINKED ISSUE #856: [FEATURE] Drain-before-roll: wait for an InferenceService to be idle before applying pod-template changes
## Feature Description

When a change to a `Model`/`InferenceService` triggers a pod-template update (e.g. tweaking llama.cpp `extraArgs`, swapping the model file, bumping the runtime image), have the operator **wait until the model is idle (no in-flight inference) before tearing down the running pod**, rather than rolling immediately. Add an opt-out (e.g. `force: true`) to preserve today's roll-now behavior.

## Problem Statement

> As an operator of **single-replica local models**, I want config changes to drain gracefully so that iterating on a model doesn't drop in-flight generations or cause an outage on every edit.

Today, any change to the pod template rolls the Deployment immediately. For a single-replica GPU-backed server (the common edge/homelab case — one model owns one device, so the new pod can't even start until the old releases the GPU), that means:

- Every benign change (a sampling-arg tweak, a context-size bump) is a hard restart.
- Any in-flight generation at that moment is dropped mid-stream.
- There's a download/load gap before the new pod serves, so it's a real outage window, not a rolling handoff.

This bites hardest exactly when you're tuning a model and applying frequent small changes.

## Proposed Solution

Gate the rollout on the server being idle. When the operator detects a pod-template change, it holds the update until the running server reports no active work, then proceeds.

Idle is observable from the runtime llama.cpp already exposes it on `/metrics` (`llamacpp:requests_processing`, slot state) and `/health` / `/slots`. The operator can poll and treat `requests_processing == 0` (optionally for a short debounce window) as idle.

**Example API:**
```yaml
kind: InferenceService
metadata:
  name: example
spec:
  modelRef: example
  rolloutPolicy:
    waitForIdle: true          # hold the roll until the server is idle
    idleTimeoutSeconds: 1800   # cap the wait; roll anyway past this
    force: false               # force: true => current behavior (roll immediately)
```

Behavior:
- `waitForIdle: true` (or making it the considerate default) → operator surfaces a condition like `RolloutDeferred: WaitingForIdle` and rolls once `requests_processing == 0`.
- `idleTimeoutSeconds` → upper bound so a perpetually-busy model still eventually rolls.
- `force: true` → skip the wait, roll now (the current pattern).

## Alternatives Considered

- **Native K8s rolling update**: doesn't help for single-replica + single-GPU — the new pod can't start until the old frees the device, so there's no overlap to roll through. The gate has to be controller-level (when to *begin* the roll), not Deployment strategy.
- **`preStop` drain hook**: helps a little (let in-flight finish on SIGTERM) but doesn't prevent the operator from starting the teardown while requests are active, and doesn't cover the load gap.
- **Manually scaling/pausing before edits**: works but is fully manual and easy to forget.

## Additional Context

- Workaround today: time changes for known-quiet windows / accept the blip.
- Applies cleanly to the `llamacpp` runtime (metrics already exist); vLLM/TGI expose analogous in-flight metrics if extended later.
- Relatedly, a `RolloutDeferred` status condition would make the wait observable to GitOps tooling (Flux health checks, etc.).

## Priority

## Willingness to Contribute

- [x] Yes, I can submit a PR
- [x] Yes, I can help test

joryirving:
## What

Adds `spec.rolloutPolicy` to `InferenceService` so pod-template changes can wait for an idle backend before updating the Deployment. The controller currently checks llama.cpp-compatible `/slots`, exposes `RolloutDeferred`, and preserves the existing roll-now behavior with `force: true`.

## Why

Single-replica local/GPU models currently drop in-flight generations whenever a config change rolls the pod immediately.

Fixes #856
Follow-up runtime breadth is tracked in #911.

## How

The reconciler stores a desired pod-template hash on Deployments to distinguish real template changes from API-server defaulting noise. When `waitForIdle` is enabled and a real template change is pending, it polls `/slots`; busy slots defer and requeue until idle or `idleTimeoutSeconds` expires. Generated deepcopy/CRD/Helm CRD output is updated.

Assisted-by: OpenCode/Codex with local fixer agents (generated implementation and tests; I reviewed the diff and ran the checks below).

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (`HOME=/var/folders/jh/4ss4bk3x1zgd8ff7nm_j4r0c0000gn/T/opencode/llmkube-home PATH="/opt/homebrew/bin:$PATH" make test`)
- [x] Documentation updated (CRD schema/user-facing fields regenerated)


Defilan:
### Code review

Nice feature and the controller wiring (template-hash change detection, `force` escape hatch, condition lifecycle) is well structured. Three issues worth resolving before merge, the first is a blocker:

1. The `/slots` idle detector parses a schema real llama.cpp does not emit, so it never detects idle. The struct keys off `processing_mode` / `slot_id`, but the upstream `GET /slots` response uses `id` and `is_processing` (bool) with no `processing_mode` field ([server README](https://github.com/ggml-org/llama.cpp/blob/master/tools/server/README.md)). Against a real server `ProcessingMode` is always `""`, so `"" != "idle"` marks every slot busy forever; `waitForIdle` degrades to "always wait the full `idleTimeoutSeconds`, then roll." The tests pass only because every mock emits the same invented shape. Fix: parse `is_processing` and treat idle as all-`false`, with fixtures matching the real schema.

https://github.com/defilantech/LLMKube/blob/3b0f3f820892012d77d0fda5fac15ac33a5027d0/internal/controller/drain_before_rollout.go#L55-L62

https://github.com/defilantech/LLMKube/blob/3b0f3f820892012d77d0fda5fac15ac33a5027d0/internal/controller/drain_before_rollout.go#L99-L106

2. Deferring a rollout short-circuits the entire reconcile, not just the Deployment template update. When `reconcileRolloutPolicy` defers, `reconcileDeployment` returns a non-nil `*ctrl.Result`, and the main `Reconcile` bails on it before `reconcileService`, `reconcileHPA`, and `updateStatusWithSchedulingInfo`. For the whole defer window (re-checked every 5s up to 5 min) the Service/HPA go un-reconciled and `Status.Phase`/`Endpoint`/`ReadyReplicas`/metrics go stale. The deferral should hold only the pod-template `Update`, then fall through to Service/HPA/status with `RequeueAfter` attached to the final result.

https://github.com/defilantech/LLMKube/blob/3b0f3f820892012d77d0fda5fac15ac33a5027d0/internal/controller/inferenceservice_controller.go#L185-L189

https://github.com/defilantech/LLMKube/blob/3b0f3f820892012d77d0fda5fac15ac33a5027d0/internal/controller/drain_before_rollout.go#L176-L186

3. The idle check fails open. Any `/slots` error (unreachable, non-200, the 5s timeout, or `--slots` disabled so it 404s) logs "proceeding with rollout" and rolls immediately, which is exactly the dropped-generation case `waitForIdle` is meant to prevent. Prefer fail-closed (defer + requeue) until the `idleTimeoutSeconds` budget is spent, and document that the backend must run with `--slots`.

https://github.com/defilantech/LLMKube/blob/3b0f3f820892012d77d0fda5fac15ac33a5027d0/internal/controller/drain_before_rollout.go#L176-L186

Two smaller scope notes (not blockers): `checkServiceIdle` hits the single Service VIP, so on `replicas > 1` it samples one random pod while the CRD comment promises "each replica" (the per-pod sweep is the tracked #911 follow-up, but the field lets you enable it on multi-replica today); and a metal-backed InferenceService silently ignores `rolloutPolicy` with no warning.

itsmiso-ai:
Thanks for the review — all three addressed in `04c991f`.

**1. `/slots` schema (blocker) — fixed.** Confirmed against the upstream README: the response uses `id` (int) and `is_processing` (bool) with no `processing_mode`/`slot_id`/`current_user`. The struct now parses `is_processing` and treats idle as all-`false`; every fixture was rewritten to the real shape. You're right that the old keys meant idle was never detected and `waitForIdle` degraded to "wait the full `idleTimeoutSeconds`, then roll" — that's gone now.

**2. Deferral short-circuiting the reconcile — fixed.** A deferred rollout now holds *only* the pod-template `Update` (the live Deployment keeps serving the old template) and returns without a result, so `reconcileService`/`reconcileHPA`/`updateStatusWithSchedulingInfo` still run and `Status.Phase`/`Endpoint`/`ReadyReplicas`/metrics stay fresh. The requeue is derived from the `RolloutDeferred=True` condition (persisted by the status update) in `Reconcile` rather than returned through the early-bail path.

**3. Failing open — fixed.** `/slots` errors (unreachable, non-200, `--no-slots` 404) now fail closed: the rollout defers under the same `idleTimeoutSeconds` budget as a genuinely busy backend instead of rolling immediately. New `IdleCheckFailed` reason distinguishes "couldn't tell" from `PodsBusy`, and the `WaitForIdle` field doc now notes the backend must expose `/slots` (llama.cpp enables it by default via `--slots`). The old envtest test that asserted proceed-on-error is flipped to assert defer + `IdleCheckFailed`, with a new unit test covering the 404 path.

Verified: `make fmt`/`vet`/`lint`/`lint-all` (darwin + linux) clean, `make test` green (controller 83.1%), CRD description + Helm chart CRD regenerated and in sync.

On the two scope notes (not blockers): the multi-replica single-VIP sampling is exactly the #911 follow-up, and the field stays single-replica-focused for this PR; metal-backed services have no Deployment to roll, so `rolloutPolicy` is a no-op there — happy to add a warning event if you'd like it surfaced.

Defilan:
Looks like the helm CI failed. Mind having a look @joryirving ?

joryirving:
> Looks like the helm CI failed. Mind having a look @joryirving ?

It's a TLS handshake timeout. Likely a transient network issue. Can you retry the job? I don't have access.

Defilan (approved):

Issue #915, opened 2026-07-01
foreman-agent coder is single-repo (static --git-remote-url); support per-Workload repo for multi-repo fleets


joryirving:
## Summary

The foreman-agent coder path uses a single **static** `--git-remote-url` for cloning and pushing, so a given agent can only ever work one repository. This makes multi-repo fleets (a `Workload` queue spanning many repos) require one agent per repo. It would be much more scalable for the coder to clone/push the **`Workload.spec.repo`** it was dispatched for.

## Current behavior

`Workload` (and the resulting `AgenticTask`) already carry the target repo (`spec.repo` / `repoFullName`), but the coder ignores it for git operations:

- `pkg/foreman/agent/tools/run_coder_job.go` — the clone/push URL is `cfg.GitRemoteURL`, sourced from the agent's static config (mirrors the watcher's `--git-remote-url`).
- `pkg/foreman/agent/tools/coder_job_template.yaml` — the Job runs `--git-remote-url={{ .GitRemoteURL }}` (static).
- Chart `agent.gitRemoteURL` is a single value per agent deployment.

So an agent started for repo A that receives a `Workload` for repo B will clone/push A while working B's issue.

## Impact

Multi-repo setups (e.g. a scheduler/bridge that turns issues from many repos into Workloads) can't use a single coder agent. The workaround is one agent per repo, each with its own `gitRemoteURL`, plus per-repo `Agent` CRs and `requiredCapability` routing so each repo's task lands on the right agent — boilerplate that grows linearly with repo count.

## Request

Let the coder derive the clone/push remote from the dispatched `Workload.spec.repo` / `repoFullName` (e.g. `https://github.com/<repoFullName>.git` with the existing token), or accept a per-Workload/per-AgenticTask remote override, falling back to `agent.gitRemoteURL` when unset. One agent could then serve many repos.

## Workaround today

One agent per repo (static `gitRemoteURL` each) + per-repo `Agent` CRs with `requiredCapability.nodeSelector` routing.

PR #920, opened 2026-07-01
feat(foreman): propagate gateProfile from Workload to decomposed tasks

### LINKED ISSUE #919: Workload cannot set gateProfile: bridge/Workload-driven tasks are locked to the Go gate
### Summary

`gateProfile` (language/image/commands for the verify gate) can only be set on `AgenticTask.spec`. Anything that drives Foreman through **`Workload`** — the normal orchestration path, and what an external work-source/bridge creates — has no way to set it, so every decomposed task falls back to the **Go preset**. On a non-Go repo the coder's in-loop fast gate and the clean-room verify Job both fail immediately (`go build ./...` / `go test ./...` on a repo with no Go module).

### Where it breaks (v0.8.22)

- `api/foreman/v1alpha1/workload_types.go` — `WorkloadSpec` has no `GateProfile` field (intent, repo, issues, coder/verifier/reviewer refs, pipeline, maxTasks, plannerModel, allowCloudReviewers only). Confirmed in the shipped `charts/foreman/templates/crds/workloads.yaml` (0 occurrences of `gateProfile`; the AgenticTask CRD has it).
- `internal/foreman/controller/workload_controller.go` — builds `AgenticTaskSpec{...}` for each decomposed task but never sets `GateProfile` and never infers language from the repo.
- `pkg/foreman/agent/executor_native.go` — `makeCoderGateVerifier(workspace, issueText, log, task.Spec.GateProfile)`; a nil `GateProfile` resolves to `golang:1.26` (see `executor_native_internal_test.go` "nil GateProfile resolves to golang image"). So the coder self-gate is Go too, not just the verify Job.

Net effect: `Workload` orchestration works only for Go repos. `AgenticTask`-direct callers can set `gateProfile`, but that bypasses the Workload pipeline (coder→gate→reviewer decomposition, escalation, reviewer sequencing).

### Request

Let a `Workload` specify a `gateProfile` (and/or have the reconciler infer language from the repo) and propagate it to every decomposed `AgenticTask`, so non-Go repos can run the full pipeline without hand-authoring AgenticTasks. Language presets already exist (`node`, `python`, `rust`, `generic`) per `docs/site/foreman/language-gates.md` — this is purely a propagation seam:

- Add `WorkloadSpec.GateProfile *GateProfile` (optional), copied onto each `AgenticTaskSpec` the `WorkloadReconciler` creates; and/or
- Per-issue/per-step override via `PipelineStep` for mixed-language workloads.

### Context

External work-sources (a dispatch→Workload bridge, in my case) create `Workload`s per issue across many repos, most of them TypeScript/Python. This is the language sibling of the multi-repo git limitation in #915: both make the `Workload` path unusable for a heterogeneous, multi-repo fleet even though the underlying agent features (language gates, coder loop) already exist.

joryirving:
## Summary

Fixes #919. `gateProfile` only existed on `AgenticTask.spec`, so anything driving Foreman through a **`Workload`** (the normal orchestration path, and what an external work-source/bridge creates) had no way to set it — every decomposed task fell back to the **Go gate**. On a non-Go repo the coder's in-loop fast gate *and* the clean-room verify Job both fail immediately (`go build ./...` / `go test ./...` on a repo with no Go module), even though the `node`/`python`/`rust`/`generic` presets already ship.

This is purely a propagation seam; the presets and resolver are unchanged.

## Changes

- **`WorkloadSpec.gateProfile`** — workload-wide default, copied onto every task the reconciler renders.
- **`PipelineStep.gateProfile`** — per-step override for mixed-language pipelines.
- **`renderAndCreate`** resolves `step → workload → go preset` via a small pure helper (`effectiveGateProfile`) and stamps it on each `AgenticTaskSpec`. Because every path (issue-batch, explicit pipeline, and escalation) funnels through `renderAndCreate`, all inherit it uniformly.
- Regenerated deepcopy + CRDs (`config/crd/bases` + foreman chart), doc section in `language-gates.md`.

Unset on both step and workload leaves `GateProfile` nil → the `go` preset, i.e. byte-for-byte current behavior.

## Tests

- `TestEffectiveGateProfile` — pure unit test of the resolution table (step wins / workload fallback / both-nil → nil).
- New envtest spec: an issue-batch `Workload` with `gateProfile.language: node` reconciles to `code-*` **and** `verify-*` tasks that both carry the node profile.
- Full controller suite green: `go test ./internal/foreman/controller/` (62 specs) + `go build ./...` + `go vet` + `gofmt -l` clean.

## Context

External work-sources create `Workload`s per issue across many repos, most of them TypeScript/Python; this makes the `Workload` path usable there without hand-authoring `AgenticTask`s. Sibling of the multi-repo git limitation in #915.


Defilan (approved):
LGTM. Clean propagation seam, and I like that the presets/resolver are untouched, this is purely wiring the profile from Workload down to the rendered tasks.

Verified the key bits:
- `WorkloadSpec.GateProfile` / `PipelineStep.GateProfile` reuse the same `*GateProfile` type as `AgenticTaskSpec`, so it copies cleanly.
- `effectiveGateProfile` resolution (step -> workload -> nil = go preset) is correct and nil leaves current behavior byte-for-byte unchanged.
- Single stamp in `renderAndCreate` with `.DeepCopy()` avoids pointer aliasing across sibling tasks and is nil-safe. And since issue-batch, explicit pipeline, and escalation all funnel through `renderAndCreate` (escalation via workload_escalation.go), the "every path inherits it" claim actually holds.
- Deepcopy + both CRD sets + docs regenerated; CI green.

Two non-blocking thoughts, feel free to take or leave:
1. The envtest exercises the issue-batch path. A follow-up test asserting an escalation task (or a per-step override) carries the profile would fully lock the "all paths" claim, right now those two lean on the shared funnel + the pure resolver test.
2. Minor: `WorkloadSpec.GateProfile` could carry an explicit `+optional` marker to match `PipelineStep.GateProfile` (functionally optional either way via pointer+omitempty).

Approving. Thanks for closing the non-Go gap on the Workload path.

PR #923, opened 2026-07-01
fix: preserve init-container SecurityContext in podTemplatesDiffer (#922)

### LINKED ISSUE #922: [BUG] 0.8.23: model-cache-prep init renders empty securityContext -> chown EPERM -> all cache-backed models fail init
## Summary

After upgrading to 0.8.23, every cache-backed model (shared model cache PVC) fails
in init:

```
model-cache-prep  chown: /models: Operation not permitted   (exit 1)
```

The generated Deployment's `model-cache-prep` init container renders an **empty
`securityContext: {}`** on 0.8.23. On 0.8.21/0.8.22 the same container rendered
`runAsUser: 0` + `capabilities.add: [CHOWN, FOWNER]`, which is what let the `chown`
succeed on `fsGroupPolicy: None` CSIs (e.g. rook-ceph CephFS). Without root +
CAP_CHOWN the `chown` gets EPERM and the pod crashloops. This regresses the fix
from #888 ("run model-cache-prep init as root again").

## Environment

- Operator / chart: 0.8.23 (worked on 0.8.22)
- Shared model cache enabled; backing CSI has `fsGroupPolicy: None`

## Evidence

ReplicaSet history for one InferenceService across successive operator versions
(`model-cache-prep` init container `securityContext`):

```
rev 38 (0.8.22): {runAsUser: 0, capabilities: {add: [CHOWN, FOWNER], drop: [ALL]},
                  allowPrivilegeEscalation: false, readOnlyRootFilesystem: true,
                  seccompProfile: {type: RuntimeDefault}}
rev 41 (0.8.23): {}
```

Init log: `chown: /models: Operation not permitted`.

## Curious bit

`cachePrepInitContainer()` in the v0.8.23 source still sets `RunAsUser: int64Ptr(0)`
+ `Capabilities{Add: [CHOWN, FOWNER], Drop: [ALL]}`, and `buildCachedStorageConfig`
still calls it unchanged (diff against v0.8.22 is empty for that assembly). So the
builder function is correct — something in the deployment assembly/apply path drops
the init container's securityContext on 0.8.23. The most likely delta is the new
"defer rollouts until idle" logic (#913), if it reconstructs or patches the pod
template in a way that loses per-container securityContext. Flagging as speculative,
but it's the notable 0.8.22 -> 0.8.23 change touching deployment rollout.

## Impact

All cache-backed models fail to start — including `hf://` multi-file models now that
#912 routes them through the cache. Effectively takes down every served model on a
cache-backed deployment after the upgrade.

## Workaround

Set `spec.podSecurityContext.runAsUser: 0` on the InferenceService so the prep
container inherits root (regains CAP_CHOWN). Reintroduces root pods, but unblocks.

## Suggested fix

Ensure the deferred/idle rollout path (or whichever assembly step regressed)
preserves init-container securityContexts, so `model-cache-prep` keeps
`runAsUser: 0` + CHOWN/FOWNER as `cachePrepInitContainer()` intends. A regression
test asserting the rendered `model-cache-prep` securityContext survives a
deferred/idle rollout would lock it down.

joryirving:
## What

Deep-copy both pod templates at the top of `podTemplatesDiffer` so the
comparison-time normalization can no longer mutate the caller's `desired`
template.

## Why

Fixes #922.

After 0.8.23, every cache-backed model crashlooped in init with
`model-cache-prep chown: /models: Operation not permitted`. The rendered
Deployment's `model-cache-prep` init container had an empty
`securityContext: {}` (it was `runAsUser: 0` + `CHOWN`/`FOWNER` on 0.8.22),
so the `chown` lost `CAP_CHOWN` and failed with EPERM on `fsGroupPolicy: None`
CSIs.

Root cause is an aliasing bug in `podTemplatesDiffer` (added with the
drain-before-rollout work in #913):

```go
a, b := existing.Spec, desired.Spec   // shallow struct copy
...
normalizeContainers(b.InitContainers) // nils RunAsUser/Capabilities/... in place
```

The struct copy is shallow, so `b.InitContainers` and each
`Container.SecurityContext` pointer are still shared with the caller's
`desired` template. `normalizeContainers` nils `RunAsUser`, `Capabilities`,
etc. for comparison purposes, but does so *through the shared pointers* —
gutting the real desired `model-cache-prep` securityContext to `{}`. The
reconciler then applies that mutated `desired` template to the Deployment
(`reconcileDeployment`), so the init container ships without root/CHOWN. The
same shallow-copy issue also nils the pod-level `SecurityContext.SeccompProfile`.

## How

- `existing = *existing.DeepCopy()` / `desired = *desired.DeepCopy()` at the
  top of `podTemplatesDiffer`, so all in-place normalization operates on copies
  and never leaks into the applied object.

## Testing

- Added a regression spec in `drain_before_roll_test.go` asserting the desired
  init-container `SecurityContext` (`RunAsUser`, `Capabilities`) survives a
  `podTemplatesDiffer` call, that a rollout-worthy change (image) is still
  detected, and that identical templates report no diff.
- Verified the fix + test logic locally by running the `podTemplatesDiffer`
  assertions as a standalone `go test` (bypassing the envtest `BeforeSuite`):
  passes with the fix, and without it fails with
  `desired RunAsUser mutated: ...RunAsUser:nil...`, reproducing the bug.
- `go build ./...`, `go vet ./internal/controller/`, `gofmt` all clean.
- I could not run the full envtest-backed `internal/controller` suite locally
  (the darwin-arm64 1.36.2 envtest control-plane binaries wouldn't exec in my
  environment, so `BeforeSuite` failed and all specs skipped). CI should run
  the Ginkgo suite.

## AI assistance

Assisted by an AI coding agent under human review (band 3 per CONTRIBUTING.md).
The author reviewed, tested, and is accountable for all changes.

## Checklist

- [x] Tests added/updated
- [ ] `make test` passes locally (blocked by local envtest binaries; see Testing)
- [x] `go build` / `go vet` / `gofmt` clean
- [ ] Documentation updated (no user-facing surface change)

### DIFF
--- a/internal/controller/drain_before_roll_test.go
+++ b/internal/controller/drain_before_roll_test.go
@@ -798,6 +798,49 @@ var _ = Describe("RolloutPolicy drain-before-rollout", func() {
 	})
 })
 
+var _ = Describe("podTemplatesDiffer input immutability (#922)", func() {
+	newPrepTemplate := func() corev1.PodTemplateSpec {
+		root := int64(0)
+		return corev1.PodTemplateSpec{
+			Spec: corev1.PodSpec{
+				InitContainers: []corev1.Container{{
+					Name:  "model-cache-prep",
+					Image: "curl:8.18.0",
+					SecurityContext: &corev1.SecurityContext{
+						RunAsUser: &root,
+						Capabilities: &corev1.Capabilities{
+							Add:  []corev1.Capability{"CHOWN", "FOWNER"},
+							Drop: []corev1.Capability{"ALL"},
+						},
+					},
+				}},
+			},
+		}
+	}
+
+	It("does not mutate the desired template's init container SecurityContext", func() {
+		desired := newPrepTemplate()
+		existing := newPrepTemplate()
+		// A rollout-worthy difference so the function runs its full
+		// normalization path before returning.
+		existing.Spec.InitContainers[0].Image = "curl:8.17.0"
+
+		changed := podTemplatesDiffer(existing, desired)
+
+		Expect(changed).To(BeTrue(), "an image change should still be detected")
+		sc := desired.Spec.InitContainers[0].SecurityContext
+		Expect(sc).NotTo(BeNil())
+		Expect(sc.RunAsUser).NotTo(BeNil(), "RunAsUser must survive the diff (see #922)")
+		Expect(*sc.RunAsUser).To(Equal(int64(0)))
+		Expect(sc.Capabilities).NotTo(BeNil())
+		Expect(sc.Capabilities.Add).To(ContainElement(corev1.Capability("CHOWN")))
+	})
+
+	It("reports no difference for identical templates", func() {
+		Expect(podTemplatesDiffer(newPrepTemplate(), newPrepTemplate())).To(BeFalse())
+	})
+})
+
 var _ = Describe("checkServerIdle", func() {
 	var reconciler *InferenceServiceReconciler
 
--- a/internal/controller/drain_before_rollout.go
+++ b/internal/controller/drain_before_rollout.go
@@ -267,6 +267,15 @@ func computePodTemplateForComparison(t corev1.PodTemplateSpec) corev1.PodTemplat
 // scheduling fields) while ignoring API-server-applied defaults like
 // TerminationGracePeriodSeconds that cause false positives on full DeepEqual.
 func podTemplatesDiffer(existing, desired corev1.PodTemplateSpec) bool {
+	// Deep-copy both templates up front. The normalization below mutates
+	// container SecurityContexts and other fields in place, and a shallow
+	// struct copy still shares the underlying slices and pointers with the
+	// caller's templates. Without this, normalizing `desired` for comparison
+	// nils out the real desired template's SecurityContext (e.g. the
+	// model-cache-prep init container's RunAsUser/Capabilities), which then
+	// gets applied to the Deployment. See #922.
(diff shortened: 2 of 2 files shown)


Defilan (approved):
LGTM, and thanks for running this one down, the shared-pointer aliasing is subtle.

Root cause is exactly right: `a, b := existing.Spec, desired.Spec` is a shallow struct copy, so `b.InitContainers` and each `SecurityContext` pointer are still shared with the caller's `desired` template. `normalizeContainers` then nils `RunAsUser`/`Capabilities` in place through those pointers, gutting the real desired `model-cache-prep` securityContext, which `reconcileDeployment` applies. Deep-copying both templates at the top is the correct, minimal fix, and the comparison result is unchanged.

The regression spec is the right one (securityContext survives, image change still diffs, identical templates no-diff), and reproducing the pre-fix failure with the standalone assertion is convincing. The envtest-local limitation is the same darwin-arm64 1.36.2 control-plane issue others have hit; CI runs the Ginkgo suite.

One thing worth calling out for release urgency: the blast radius is bigger than drain-before-roll. `podTemplatesDiffer` runs on every non-metal reconcile to compute `templateChanged`, so this stripped the init securityContext on every cache-backed in-cluster model (we saw qwopus-strix, gemma4, and llama32-1b-cpu all crashloop on `chown ... Operation not permitted`), not just services using `waitForIdle`. This should go into 0.8.24 as a priority patch.

Approving.

Issue #813, opened 2026-06-23
[FEATURE] Foreman coder: branch from upstream/main, not the fork's default branch (stale-base fix)


Defilan:
## Feature Description

Make the Foreman coder executor branch each task off the **current upstream base** (`upstream/main`) instead of the fork's default branch. Today the executor clones the fork (`--git-remote-url`) and branches from whatever the fork's `main` happens to be; if the fork's `main` lags the upstream project, every coder run starts from a stale base.

## Problem Statement

> As a fleet operator, I want each coder run to start from the latest upstream `main` regardless of the fork's sync state, so the change is built and tested against current code and PRs don't drift from `main`.

Workspace prep in `pkg/foreman/agent/executor_native.go` (~lines 261-275) does:

```go
repo.Clone(ctx, repo.CloneOptions{ RemoteURL: e.GitRemoteURL, Dest: workspace, Auth: auth }) // clones the FORK at its default branch
branch := branchNameForTask(task)
repo.CreateAndCheckoutBranch(ctx, workspace, branch)                                          // branches from the fork's main HEAD
```

`Clone` checks out the remote's default branch (the fork's `main`), and the coder branch is cut from there. Nothing refreshes that base against the upstream project, so a fork whose `main` lags upstream produces a stale-base branch.

**Observed:** PR #812 (the #804 reviewer-prompt sync) was authored from a fork `main` that was ~35h behind upstream, predating #802's two new `reviewer.md` rubric sections (I. Real values, J. Wired-up). The coder synced the reviewer manifests against the stale `reviewer.md`; the PR's merge commit pulled in the current `reviewer.md`, and the new `check-reviewer-prompts` drift gate correctly failed. It was caught after the fact and fixed by hand (merge main + re-sync). This will recur for any file that moves upstream between fork syncs.

## Proposed Solution

After cloning the fork and before creating the task branch, fetch the upstream base and branch from it. The upstream slug is already available as `task.Spec.Payload.Repo` (e.g. `defilantech/LLMKube`), so no new flag is required:

```go
// payload.repo is the upstream "owner/name"; derive the upstream URL.
upstreamURL := "https://github.com/" + task.Spec.Payload.Repo + ".git"
base := task.Spec.Payload.BaseBranch // new optional field; default "main"
git remote add upstream <upstreamURL>
git fetch upstream <base>
git checkout -B <branch> upstream/<base>   // base = CURRENT upstream main
```

[...]

PR #924, opened 2026-07-01
feat(foreman): coder clones/pushes each task's own repo when no static remote

### LINKED ISSUE #915: foreman-agent coder is single-repo (static --git-remote-url); support per-Workload repo for multi-repo fleets

joryirving:
## Summary

Fixes #915. The coder cloned and pushed the agent's single static `--git-remote-url`, so one agent could only serve one repository — a Workload-driven source feeding issues across many repos had to run one agent per repo.

When `--git-remote-url` is **empty**, the coder now derives its clone + push target from the task's `payload.repo` (`github.com/<owner>/<name>`, via the existing `upstreamURLForRepo` helper). So a **single agent serves every repo** the Workloads name — pair it with `Workload.spec.repo` + `Workload.spec.gateProfile` (#920) and the whole fleet runs off one agent.

## Behavior

- **Static `--git-remote-url` set** → unchanged: every task pins to that one fork/repo (fork-PR model).
- **Empty** → each coder task clones + pushes its own `payload.repo`.
- **Neither set nor a valid `payload.repo`** → fails cleanly with `GitRemoteNotConfigured` (was: same failure, now with an accurate message).

The gate and post-push envtest paths already construct their clone URL from `payload.repo` when the executor URL is empty (`buildDeterministicArgs`, `evaluatePostPushEnvtest`), so they find the branch the coder pushed without any change. Push follows the clone's `origin`, so it already targets the right repo. The branch is still cut from the current upstream base (#813); when the clone was itself derived from `payload.repo`, origin and upstream resolve to the same URL and the base branch is still current.

## Changes

- `executor_native.go`: derive `cloneURL` from `payload.repo` when `GitRemoteURL` is empty (single resolution point; `resolveUpstream` hoisted above the clone).
- `cmd/foreman-agent/main.go`: the startup info-log for an unset `--git-remote-url` now describes the per-task-repo behavior.
- `charts/foreman/values.yaml` + `runbook-m3.md`: `gitRemoteURL` documented as optional.

## Tests

- `TestNativeExecutor_MultiRepoClonesTaskRepoWhenNoStaticRemote` — empty static remote + `payload.repo` → GO with the branch pushed to the task's repo.
- `TestNativeExecutor_NoStaticRemoteAndNoRepoIsHardError` — neither set → `GitRemoteNotConfigured`.
- Full `pkg/foreman/agent` suite green (incl. the unchanged fork-model happy path); `go build`/`go vet` clean.

## Context

Sibling of #919/#920 (Workload `gateProfile` propagation). Together they let one shared agent run the full coder→gate→reviewer pipeline across a heterogeneous, multi-repo fleet via `Workload.spec.{repo,gateProfile}` — no per-repo agent fan-out.

### DIFF
--- a/charts/foreman/values.yaml
+++ b/charts/foreman/values.yaml
@@ -216,22 +216,20 @@ agent:
   # the deterministic gate Agent's workspace prep step.
   workspaceDir: /workspace
 
-  # Git remote the executor (both native LLM-driven coder and the
-  # deterministic gate, which clones the upstream coder's branch from
-  # the fork) reads from and pushes branches to. Required for ALL
-  # native-mode nodes including verifier-only nodes:
+  # Git remote the executor reads from and pushes branches to. OPTIONAL
+  # as of #915:
   #
-  #   - coder-role nodes (M5 Max running the Carnice Agent) need it to
-  #     push the resulting branch.
-  #   - verifier-role nodes (ShadowStack running the M4 gate Agent)
-  #     need it because run_gate_job clones the branch the coder pushed
-  #     to the fork, not the upstream repo where the branch does not
-  #     yet exist.
+  #   - Set: pins every coder task to this one fork/repo (the single-repo
+  #     fork-PR model). The coder pushes the branch here and the gate
+  #     clones it from here.
+  #   - Empty: each coder task clones and pushes its OWN payload.repo
+  #     (github.com/<owner>/<name>), so a single agent serves many repos.
+  #     The gate then constructs the same clone URL from payload.repo, so
+  #     it still finds the branch the coder pushed.
   #
-  # Leaving this empty puts the agent into "deterministic-only-without-
-  # network-git" mode: the M4 V3 two-step pipeline fails at the gate's
-  # clone step with "foreman-agent was not started with --git-remote-url;
-  # cannot clone". Stub-mode agents do not need it.
+  # Set this only to force every task onto one remote (a shared fork).
+  # A coder task with neither this NOR a payload.repo fails at the clone
+  # step with GitRemoteNotConfigured. Stub-mode agents ignore it.
   gitRemoteURL: ""
 
   # Git author + committer identity stamped onto commits the coder-
--- a/cmd/foreman-agent/main.go
+++ b/cmd/foreman-agent/main.go
@@ -273,8 +273,9 @@ func main() {
 		// leaves other tasks unaffected. The flags stay required at
 		// the per-task level, not at startup.
 		if gitRemoteURL == "" {
-			setupLog.Info("--git-remote-url is unset; coder-role tasks will fail with GitRemoteURLNotConfigured. " +
-				"Deterministic Agents (e.g. M4 gate) work fine without it.")
+			setupLog.Info("--git-remote-url is unset; coder-role tasks clone and push " +
+				"each task's own payload.repo (#915). Tasks with no payload.repo fail " +
+				"with GitRemoteURLNotConfigured; deterministic Agents work fine without it.")
 		}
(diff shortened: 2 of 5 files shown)


Defilan:
Hey @joryirving we had a linting error. Once you get that fixed, I'll get this going through the CI again and pushed out for an emergency fix.

Defilan (approved):
LGTM. Nice unlock, and it composes cleanly with the Workload gateProfile work (#920) so one agent can run the whole multi-repo fleet.

Verified the parts that matter:
- Backward-compatible: `cloneURL := e.GitRemoteURL` first, payload.repo only when it's empty, so agents with a static remote are unchanged (fork-PR model intact).
- Safe against redirection: `upstreamURLForRepo` validates the slug against `^[A-Za-z0-9._-]+/[A-Za-z0-9._-]+$` and hardcodes the `https://github.com/` host, so a task can't point the clone/push at an arbitrary host or a non-owner/name path. An invalid slug returns "" and fails cleanly as `GitRemoteNotConfigured`.
- Single resolution point: `resolveUpstream` is hoisted so the clone target and the #813 upstream base both use it; origin == upstream when derived from payload.repo, so branching off the base stays consistent, and push follows origin into the task's repo.
- Tests cover the matrix (multi-repo GO, hard-error when neither is set, and the unchanged static-remote path).

One non-blocking note on the trust boundary this shifts: in empty-remote mode the agent will push a branch to any `github.com/<owner>/<name>` a task names, using its git token. That's a deliberate opt-in and it's well guarded (host-pinned, slug-validated), but it does mean the effective security posture becomes (a) who can create tasks/Workloads and (b) the agent token's scope. Since you're already touching the `gitRemoteURL` docs, a sentence noting "leaving this empty lets tasks direct pushes to any repo the agent's token can write, so scope the token to the fleet's repos and control task creation" would make the implication explicit. Not a blocker.

Approving.

Issue #478, opened 2026-05-17
[FEATURE] Tutorial: metrics-driven autoscaling for InferenceService


Defilan:
## Feature Description

A documentation tutorial showing how to autoscale an `InferenceService` based on metrics: a HorizontalPodAutoscaler driving the replica count, plus cluster/node autoscaling so replicas that do not fit land on additional GPU nodes.

## Problem Statement

> As an LLMKube user, I want a step-by-step guide for metrics-driven autoscaling so that I can scale inference capacity automatically under load without designing the HPA and node-autoscaler wiring myself.

The `InferenceService` scale subresource (#474) makes HPA targeting possible, but there is no end-to-end doc that ties HPA, the metrics source, and GPU node autoscaling together. Requested by a viewer on the project's YouTube walkthrough.

## Proposed Solution

A tutorial in the docs covering:

- Targeting the `InferenceService` scale subresource with a HorizontalPodAutoscaler
- Which signals to scale on, and how to wire them: LLMKube's Prometheus metrics and GPU utilization (DCGM) via prometheus-adapter or KEDA
- Node autoscaling so scaled-out replicas get GPU capacity: Cluster Autoscaler, Karpenter, or GKE node auto-provisioning
- A worked example with YAML, from zero load to scaled-out

## Alternatives Considered

- Leave users to assemble it from generic Kubernetes HPA docs. This is the current state and has high friction for the GPU-specific parts.
- A built-in autoscaling CRD field. Heavier, and unnecessary now that standard HPA works through the scale subresource.

## Additional Context

- Related issues: #474 (scale subresource)
- Requested via a YouTube comment on the project walkthrough video

## Priority

- [ ] Critical - Blocking my use case
- [ ] High - Would significantly improve my workflow
- [x] Medium - Nice to have
- [ ] Low - Minor improvement

## Willingness to Contribute

- [x] Yes, I can submit a PR
- [ ] Yes, I can help test
- [ ] No, but I can provide feedback
- [ ] No, just requesting

Defilan:
Tracked as part of the documentation sprint umbrella #628.

PR #931, opened 2026-07-02
feat(foreman): two-tier gate-check suite for the coder gate

### LINKED ISSUE #930: [FEATURE] Two-tier gate-check suite for the Foreman coder gate
## Feature Description

Extend the Foreman coder gate with a tiered gate-check registry and a set of deterministic checks that catch defects which pass the existing gate (gofmt/vet/build/lint/tests) but are not actually mergeable.

## Problem Statement

A recent 5-issue Foreman batch returned GATE-PASS on all five branches, yet rigorous review found every one not mergeable as-is: a hallucinated metric name and invalid YAML in a docs change, missing RBAC on a new controller path, a subtle verification-logic bug with tests tuned to green, a shared-function change that regressed a second caller, and a refactor that introduced an unwanted `pkg/ -> internal/` dependency. The gate verifies mechanics; it does not verify substance.

> As a maintainer, I want the harness to catch these deterministic defect classes at gate time so review shrinks to genuine judgment and (long term) local review support suffices.

## Proposed Solution

A `gateCheck{name, tier, lang, fn}` registry with two tiers:
- **Block tier** (fails the gate, fed back to the coder loop): `rbac-use` (client verb without a matching `+kubebuilder:rbac` marker), `import-graph` (new `pkg/ -> internal/` edge), `embedded-artifact` (invalid YAML/manifest in changed docs).
- **Advisory tier** (surfaced to the reviewer, non-blocking): `grounding-breadth` (hallucinated metric/resource names in docs), `caller-impact` (blast radius of a changed shared function), `issue-example` (the issue's own repro surfaced for the reviewer to verify).

Also tightens the existing test-presence check to per-symbol, and wires advisories through the AgenticTask payload to the reviewer prompt. Each check has a `FOREMAN_<NAME>_GATE=0` kill switch.

## Alternatives Considered

Making the review model smarter instead of the harness; rejected in favor of shifting deterministic findings out of review into the gate (the harness-not-model approach). Judgment-tier findings (e.g. "reinvented an existing feature") remain the reviewer's job.

## Additional Context

- Motivating issues (the batch that surfaced the gap): #478, #809, #904, #907, #921
- Design + plan committed internally; regression fixtures reproduce all five.

## Priority

## Willingness to Contribute

- [x] Yes, I can submit a PR

Defilan:
## What

Adds a two-tier gate-check registry to the Foreman coder gate and six deterministic checks that catch defects passing the existing gate (gofmt/vet/build/lint/tests) but not actually mergeable. Block-tier failures feed the coder loop; advisory-tier findings flow to the reviewer.

- Registry: `gateCheck{name, tier, lang, fn}` in `pkg/foreman/agent/` with per-check `FOREMAN_<NAME>_GATE=0` kill switches.
- **Block tier:** `rbac-use` (controller client verb without a matching `+kubebuilder:rbac` marker), `import-graph` (new `pkg/ -> internal/` import edge), `embedded-artifact` (invalid YAML/manifest in changed docs).
- **Advisory tier:** `grounding-breadth` (hallucinated metric/resource names in docs), `caller-impact` (changed shared function blast radius), `issue-example` (the issue reproducer surfaced for the reviewer).
- Tightened test-presence from per-package to per-symbol.
- Advisories are persisted to the code task result and wired through a new `AgenticTaskPayload.GateAdvisories` field into the reviewer prompt.

## Why

Fixes #930

A 5-issue Foreman batch returned GATE-PASS on every branch, yet review found none mergeable (hallucinated metric + broken YAML, missing RBAC, a verification-logic bug with tests tuned to green, a shared-function change that regressed a second caller, and an unwanted `pkg/ -> internal/` dependency). This shifts those deterministic defect classes out of review and into the gate: the harness-not-model approach applied to the review layer.

## How

Each check is a `func(ctx, workspace, run) (failed bool, output string)` plugged into `RunCoderGate`; `runGateChecks` splits results by tier in one pass. Block failures append to the existing feedback path (3-retry loop); advisories go to `Result.Extra["gateAdvisories"]`, then `patchReviewAdvisories` (WorkloadReconciler) copies them into the review task payload and `buildUserPrompt` renders them for the reviewer. All checks are fail-open (any git/parse error returns no-fire) so a block-tier check can never spuriously stall a good branch. The five motivating issues (#478/#809/#904/#907/#921) are captured as integration regression fixtures that materialize real temp git repos and assert each check fires on its defect and stays silent on a known-good counterpart.

## Checklist

- [x] Tests added/updated (unit + integration fixtures + envtest specs)
- [x] `make test` passes locally (agent + grounding packages; controller envtest 66 specs)
- [x] `make lint` passes locally (`GOOS=linux golangci-lint`, 0 issues)
- [x] Documentation updated (`docs/site/foreman/gate-check-suite.md`)

## Known minor follow-ups (non-blocking)

- `advisory` (internal) vs `GateAdvisory` (CRD) json-tag `omitempty` asymmetry: safe on the current decode path; worth aligning.
- Advisories accumulate across gate retries by design (reviewer instruction handles it); a clarifying comment would help.
- The gate-check doc does not yet mention the pre-existing `FOREMAN_MUTATION_GATE` toggle.

---

Assisted-by: Claude Code (Anthropic) generated the implementation via a task-by-task subagent workflow; each task went through independent spec-compliance and code-quality review passes that caught several real defects (a working-tree-vs-index read bug, RBAC false positives, a block-tier grounding contamination). I reviewed the result, ran the tests and cross-arch lint, and am accountable for the change.


Defilan:
@joryirving can you please review?

joryirving (approved):
## Code Review

### Medium — Off-by-one in fenceStart (issue_example_gate.go:45)

```go
fenceStart := i - len(buf) - 1 // index of opening fence
for lookback := 1; lookback <= 5 && fenceStart-lookback >= 0; lookback++ {
```

When buf is non-empty, `fenceStart` points at the opening fence line itself, not the line before it. The lookback then skips the fence line and examines `fenceStart-1` first. This works when there are blank lines between the marker and fence, but **fails when content immediately follows the opening fence with no blank line separating them**.

Repro: `"## Example\n\`\`\`\ncode\n\`\`\`"` — the "## Example" line is not found because lookback=1 checks the fence line itself, not the heading.

The existing test has blank lines between "## Example" and the fence, so it passes. The simpler case without blank lines would silently fail to find the issue-example block.

### Low — Unused variable (issue_example_gate.go:31)

```go
closing := strings.TrimLeft(line, "`")
// The info string follows the backticks; closing fence is "```" alone.
_ = closing
```

`closing` is assigned but never used. Either distinguish info-string fences from bare closing fences with it, or remove the assignment.

### Design concern — ExporterMetricPrefixes not seeded in LoadGroundTruth

`checkGroundingBreadth` explicitly sets `ExporterMetricPrefixes` after `LoadGroundTruth` returns, so the block tier gets nil and the check is inert there (correct behavior). However, this is a silent contract: any future caller of `LoadGroundTruth` who uses `gt.ExporterMetricPrefixes` directly without the advisory-only setter will get nil/empty and the check will silently do nothing. The block-tier regression test covers the outcome, but not the contract.

### Potential bug — splitListValue skips groups="" to []

`splitListValue` skips any token that becomes `""` after `Trim(tok, "\"")`. But `groups=""` (kubebuilder convention for the core API group) should become `[""]` not `[]`. The test fixture uses `groups=batch`, not `groups=""`, so this path is untested. If `groups=""` appears in real markers, the RBAC check would silently skip the empty group and produce false negatives.

---

### What's working well

- Two-tier architecture (block → loop retry, advisory → reviewer) is cleanly implemented.
- All new checks are fail-open on git/read/exec errors — appropriate for an in-loop gate.
- `patchReviewAdvisories` correctly bails for explicit pipeline workloads and is idempotent via `GateAdvisories` and phase guards.
- All call sites updated for the new `RunCoderGate(bool, string, []advisory)` signature.
- Integration fixture tests using real temp git repos provide good coverage for the git status/diff/code paths.

Defilan:
Thanks Jory, good pass. Handled:

- **Unused `closing` var** — removed.
- **`ExporterMetricPrefixes` contract** — added a `CONTRACT:` note on the field itself spelling out that `LoadGroundTruth` intentionally leaves it nil and only the advisory-tier check sets it. Worth adding: the block-tier `checkReferenceGrounding` also filters to non-minor severity, so even if a future caller seeded prefixes on that path, minor exporter-token findings still could not block. Belt and suspenders now.

On the two potential-bug ones, I believe these are already covered:

- **`fenceStart` / marker directly above the fence** — the lookback inspects the line *above* the opening fence, not the fence line itself, so the repro `## Example` + fence + `code` + fence does return the block. `fenceStart = i - len(buf) - 1` lands on the opening fence, and `lines[fenceStart-1]` is the heading. I added your exact case as an explicit regression test (`TestHarvestIssueExample_MarkerImmediatelyAboveFence`) so it stays that way.
- **`groups=""`** — `splitListValue` strips surrounding quotes and keeps `[""]` (skipping only tokens that were empty *before* stripping), and there are two tests for it: `TestCheckRBACUse_QuotedCoreGroupMarkerPasses` and `TestCheckRBACUse_CoreGroupSpellingMarkerPasses` (the latter also covers `groups=core`). Let me know if you are seeing a case those miss.

Pushed as e973e43.

joryirving:
## Second review — no blocking issues

Three non-blocking items:

1. **`externalCallers` substring false positive** (`pkg/foreman/agent/caller_impact_gate.go:147`) — `strings.Contains(text, funcName+"(")` matches suffixes. `"Renew("` contains `"New("`, so a call to `Renew()` would be reported as a caller of `New`. Same for short names like `Get`, `Set`, `Run`. Advisory-only impact (noisy output), but fix is simple: anchor with a non-identifier boundary on the left, e.g. check that the char before the match is not an identifier character.

2. **`capBytes` truncates by bytes, not runes** (`pkg/foreman/agent/issue_example_gate.go:127`) — `s[:maxExampleBytes]` can split mid UTF-8 character. Practically harmless (examples are ASCII), but technically incorrect for non-ASCII content.

3. **`checkGroundingBreadth` mutates shared ground truth** (`pkg/foreman/agent/grounding_breadth_gate.go:51-53`) — sets `ExporterMetricPrefixes` on the object returned by `LoadGroundTruth`. Currently safe (fresh alloc per call), but fragile if that function ever caches.

PR #939, opened 2026-07-02
fix(foreman): chart exposes --foreman-namespace so gate Jobs land with the cache PVC

### LINKED ISSUE #933: charts/foreman: expose the agent's --foreman-namespace flag (gate Jobs land in a namespace the chart never creates)
## Problem

The foreman-agent submits gate Jobs into `--foreman-namespace`, which defaults to `foreman-system` (`cmd/foreman-agent/main.go:209`, `pkg/foreman/agent/tools/run_gate_job.go:454`). The chart never renders this flag (`templates/agent-deployment.yaml` has no way to set it) and never creates `foreman-system`.

So for any install outside a namespace literally named `foreman-system`, every verify task fails:

```
gate did not run: create job: namespaces "foreman-system" not found
```

→ verdict `GATE-ERROR` → downstream review tasks cascade-fail with `UpstreamFailed`. Hit in production on a release installed in `llm`.

Note the gate Job also mounts the `foreman-gate-cache` PVC, which the chart creates in the **release namespace** (`templates/gate-cache-pvc.yaml`) — so even manually creating `foreman-system` doesn't help; the Job's pod would hang on the missing PVC. The release namespace is the only namespace where the default config can actually work.

## Proposed fix

Expose the flag in the chart, defaulting to the release namespace (where the PVC lives):

```yaml
# values.yaml
agent:
  # Namespace the agent submits gate Jobs into. Defaults to the release
  # namespace, which is where the chart creates the foreman-gate-cache PVC.
  foremanNamespace: ""
```

```yaml
# templates/agent-deployment.yaml
- --foreman-namespace={{ .Values.agent.foremanNamespace | default (include "foreman.namespace" .) }}
```

Chart-only change — the binary flag already exists. Arguably the binary default should also prefer the pod's own namespace (downward API) over a hardcoded `foreman-system`, but the chart fix alone unbreaks every non-`foreman-system` install.

## Workaround

Flux/kustomize postRenderer patching the arg onto the agent Deployment:

```yaml
postRenderers:
  - kustomize:
      patches:
        - target: {kind: Deployment, name: foreman-agent}
          patch: |
            - op: add
              path: /spec/template/spec/containers/0/args/-
              value: --foreman-namespace=<release-namespace>
```

Defilan:
## What

Expose the agent's `--foreman-namespace` flag in the foreman Helm chart, defaulting to the release namespace.

## Why

Fixes #933

The agent binary defaults `--foreman-namespace` to `foreman-system`, a namespace the chart never creates. Any install not named `foreman-system` lands gate Jobs where the gate-cache PVC does not exist, so every verify task GATE-ERRORs.

## How

Adds `agent.foremanNamespace` to values.yaml (empty default resolves via the `foreman.namespace` helper to `.Release.Namespace`) and wires it unconditionally into the agent Deployment args, so gate Jobs always share the namespace of the cache PVC (which uses the same helper). Verified default and override paths with `helm lint` and `helm template`.

## Checklist

- [x] Tested (helm lint + helm template on default and override paths)
- [x] Documentation updated (values.yaml documents the new key)
---

Assisted-by: Foreman (LLMKube coder harness, local Ornith-35B on the Strix) generated the change; it passed the post-push clean-room envtest gate and an independent external-contributor review pass before I opened this. I reviewed the result, ran the tests and cross-arch lint, and am accountable for it.

---

**Upgrade note (for release notes):** operators who applied the `--foreman-namespace=<release-ns>` postRenderer workaround from #933 will briefly have the flag set twice after upgrading to this chart (chart-rendered + patch, same value, last-wins — harmless). They should drop the postRenderer patch once on this version. (Thanks @joryirving for flagging.)

### DIFF
--- a/charts/foreman/templates/agent-deployment.yaml
+++ b/charts/foreman/templates/agent-deployment.yaml
@@ -53,6 +53,7 @@ spec:
             {{- if .Values.agent.taskNamespace }}
             - --task-namespace={{ .Values.agent.taskNamespace }}
             {{- end }}
+            - --foreman-namespace={{ .Values.agent.foremanNamespace | default (include "foreman.namespace" .) }}
             {{- if .Values.agent.installedModels }}
             - --installed-models={{ join "," .Values.agent.installedModels }}
             {{- end }}
--- a/charts/foreman/templates/agent-deployment.yaml
+++ b/charts/foreman/templates/agent-deployment.yaml
@@ -53,6 +53,7 @@ spec:
             {{- if .Values.agent.taskNamespace }}
             - --task-namespace={{ .Values.agent.taskNamespace }}
             {{- end }}
+            {{- /* Always set: the binary default is "foreman-system", a namespace the chart never creates, so gate Jobs would land where the cache PVC does not exist. */}}
             - --foreman-namespace={{ .Values.agent.foremanNamespace | default (include "foreman.namespace" .) }}
             {{- if .Values.agent.installedModels }}
             - --installed-models={{ join "," .Values.agent.installedModels }}
--- a/charts/foreman/values.yaml
+++ b/charts/foreman/values.yaml
@@ -197,6 +197,12 @@ agent:
   # tenants that should not see each other's tasks.
   taskNamespace: ""
 
+  # --foreman-namespace flag. Namespace the agent submits gate Jobs
+  # into. Defaults to the release namespace, which is where the chart
+  # creates the foreman-gate-cache PVC. Override only when the PVC
+  # and the agent live in different namespaces.
+  foremanNamespace: ""
+
   # Models the foreman-agent advertises in its FleetNode.spec
   # heartbeat. Use kebab-case slugs (e.g. "qwen36-35b-carnice-mtp").
   installedModels: []


joryirving (approved):
Exactly the shape proposed in #933 — verified the `foreman.namespace` helper resolves to `.Release.Namespace` (with the top-level `namespace` override still honored), and rendering the flag unconditionally is the right call since the binary default can never be correct for a chart install that doesn't create `foreman-system`.

Battle-tested shape: we've been running precisely `--foreman-namespace=<release-ns>` via a Flux postRenderer in prod since filing #933, and it took verify tasks from 100% GATE-ERROR to GATE-PASS.

One note for the release notes: operators using the postRenderer workaround from #933 will briefly have the flag twice after upgrading (chart + patch, same value, last-wins — harmless) and should drop the patch.

Defilan:
Thanks Jory — no code change needed here (the shape is exactly #933), but I've added your double-flag transition note to the PR body as an upgrade note so it carries into the release notes: postRenderer users get the flag twice for one upgrade (same value, last-wins, harmless) and should drop the patch on this version. Good catch on the prod postRenderer validation — glad it took verify from 100% GATE-ERROR to GATE-PASS.

PR #940, opened 2026-07-02
fix(foreman): substitute placeholder for empty assistant replies to avoid poisoning history

### LINKED ISSUE #935: agent loop: empty assistant message poisons conversation history — every later turn 400s ('Assistant message must contain either content or tool_calls')
## Problem

Mid-loop, a local model (llama.cpp via an OpenAI-compatible endpoint) occasionally returns an assistant message with neither `content` nor `tool_calls` (e.g. reasoning-only or empty completion). The native agent loop appends it to the conversation history verbatim. Every subsequent turn then replays the poisoned history and the backend rejects the whole request:

```
turn 40: chat: oai: status 400: {"error":{"code":400,"message":"Assistant message must contain either 'content' or 'tool_calls'!","type":"invalid_request_error"}}
```

The task fails (INCOMPLETE) and cascade-fails its verify/review dependents — after burning 40 turns of real work. Hit in production on an issue-fix task; the failure is unrecoverable within the loop because the bad message stays in history forever.

## Proposed fix

When the model returns an assistant message with no content and no tool calls, don't append it as-is. Options (in preference order):

1. Substitute a placeholder (`content: "(empty response)"`) before appending, and nudge the model with a user-role correction — keeps the turn count/history consistent and the loop alive.
2. Drop the empty message and retry the turn (bounded), treating it like a transport error.

Either way the loop should never construct a history the backend is guaranteed to reject.

Hit in production on foreman 0.8.25, native agent, llama.cpp backend via litellm.

Defilan:
## What

Stop an empty assistant reply from poisoning the agent conversation history.

## Why

Fixes #935

When the model returns an assistant message with no content and no tool_calls, appending it to the transcript makes every later turn 400 with "Assistant message must contain either content or tool_calls", stalling the run.

## How

Before the assistant turn is appended to the transcript, if its content is empty (after trim) AND it has no tool_calls AND no reasoning_content, substitute a short placeholder. The guard is conjunctive, so valid tool-call turns are left untouched and reasoning-only turns continue to be handled by the existing `stripReasoningForWire` path. Adds an end-to-end test (empty reply on turn 1, successful submit on turn 2) plus a regression test asserting an empty-content tool-call turn is not clobbered.

## Checklist

- [x] Tests added/updated (e2e + regression)
- [x] `make lint` passes locally (GOOS=linux, 0 issues)
- [ ] Documentation updated (n/a: internal agent-loop fix)
---

Assisted-by: Foreman (LLMKube coder harness, local Ornith-35B on the Strix) generated the change; it passed the post-push clean-room envtest gate and an independent external-contributor review pass before I opened this. I reviewed the result, ran the tests and cross-arch lint, and am accountable for it.

### DIFF
--- a/pkg/foreman/agent/loop.go
+++ b/pkg/foreman/agent/loop.go
@@ -792,11 +792,21 @@ func (l *Loop) runOneTurn(
 	// every subsequent turn (#935). Substitute a placeholder so the
 	// transcript stays valid and the loop can recover with a corrective
 	// nudge rather than burning turns until MaxTurns.
+	//
+	// The placeholder is a NEUTRAL factual marker, not an instruction: it
+	// is replayed as assistant content on every later turn, and a
+	// self-addressed "please call a tool" reads to smaller models as the
+	// model instructing itself. The corrective instruction lives in the
+	// user-role NoToolCallNudgeMessage() the no-tool-call path appends
+	// below, which is the right voice for it.
 	if msg.Role == oai.RoleAssistant &&
 		strings.TrimSpace(msg.Content) == "" &&
 		len(msg.ToolCalls) == 0 &&
 		msg.ReasoningContent == "" {
-		msg.Content = "(empty response from model; please call a tool or submit_result to finish)"
+		msg.Content = "(empty response from model)"
+		// Surface chronic empty replies (a backend or chat-template
+		// problem) on the turn span instead of silently absorbing them.
+		span.SetAttributes(attribute.Bool("empty_assistant_reply", true))
 	}
 	res.Transcript = append(res.Transcript, msg)
 
--- a/pkg/foreman/agent/loop.go
+++ b/pkg/foreman/agent/loop.go
@@ -784,6 +784,20 @@ func (l *Loop) runOneTurn(
 	if msg.Role == "" {
 		msg.Role = oai.RoleAssistant
 	}
+	// Guard against a completely empty assistant reply (no content, no
+	// tool_calls, no reasoning_content). Stricter backends (llama.cpp via
+	// litellm, Devstral, OpenAI) reject a history that contains such a
+	// message on the next turn with "Assistant message must contain
+	// either 'content' or 'tool_calls'", poisoning the conversation for
+	// every subsequent turn (#935). Substitute a placeholder so the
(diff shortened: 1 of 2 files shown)


joryirving (approved):
Reviewed the guard and ran the new tests locally (`go test ./pkg/foreman/agent/ -run 'EmptyAssistant|ToolCallTurnNotClobbered|StrippedReasoning'` → ok). Approving.

What convinced me:
- The conjunctive condition is right: content-empty AND no tool_calls AND no reasoning_content. Tool-call turns with empty content (the normal OAI pattern) are explicitly protected by `TestRunOneTurn_ToolCallTurnNotClobbered`, and reasoning-only turns stay on the existing `stripReasoningForWire` path.
- Substituting *before* the transcript append fixes it at the source — every later turn replays a valid history, which is exactly where #935 bit us (40 turns of work lost to one empty reply).
- The e2e test proves recovery, not just substitution: empty reply → placeholder → existing corrective-nudge path → GO on the next turn.

Two non-blocking thoughts, fine as follow-ups:
1. The placeholder reads as the model instructing itself ("please call a tool…") in later history. A neutral marker like "(model returned an empty response)" with the corrective instruction kept in the user-role nudge might confuse smaller models less.
2. A log line / counter when the guard fires would make chronic empty replies (a backend or template problem) visible instead of silently absorbed.

Thanks for the fast turnaround on both — this was the last of the overnight production blockers.

Defilan:
Thanks Jory — folded both of your follow-ups into `c7b7a91`:

1. The placeholder is now the neutral marker `(empty response from model)`; the corrective "call a tool / submit_result" instruction stays in the user-role `NoToolCallNudgeMessage()` the no-tool-call path already appends, so nothing self-instructs in replayed assistant history. This matters most for the smaller models we run (Ornith-35B), which was your read exactly.
2. Added an `empty_assistant_reply` attribute on the turn span when the guard fires — no Prom counter in the loop today, so the OTel span is the idiomatic signal; chronic empty replies now show up in Tempo instead of being silently absorbed.

Existing tests still assert on the `empty response from model` substring, so they pass unchanged; `GOOS=linux` lint clean. Heads-up that the push may have dismissed your approval — a re-glance when you get a sec.

Issue #573, opened 2026-05-27
[FEATURE] AgenticTask branch names should be unique across reruns


Defilan:
## Feature Description

AgenticTask branch names collide across reruns. Two tasks on the same GitHub issue (different Workloads, different days, different model runs) both produce the same branch name `foreman/issue-N`, and the second one's push gets rejected non-fast-forward by the fork. The executor loses real commits and writes verdict=NO-GO with outcome=PUSH-FAILED.

Empirical: the rerun-5 batch (2026-05-27 09:30 UTC, post-#571 issue-body fetch) had Carnice make a real commit on issue #449 (`da76f23f0c3148b86b46d559e169e9b92e630e81`) that the executor could not land because `refs/heads/foreman/issue-449` already exists on the fork from the V5 batch on 2026-05-25. Same shape on #510.

Push error verbatim:

```
push rejected by remote: ! [rejected]        foreman/issue-449 -> foreman/issue-449 (non-fast-forward)
error: failed to push some refs to 'https://github.com/Defilan/LLMKube.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart.
```

The executor is right to refuse a force-push (it would silently overwrite published work). The fix is to make branch names unique by construction so two foreman tasks on the same issue never produce the same branch.

## Problem Statement

> As a Foreman operator running multiple batches on the same set of issues, I want each AgenticTask's commit to land on its own branch so the empirical artifact (the foreman-authored branch) survives even when an earlier run already produced a branch on the same issue.

## Proposed Solution

Two-part change:

**1. WorkloadReconciler synthesizes unique branch names.**

The M6 stub planner currently sets:

```go
Payload: AgenticTaskPayload{
    Issue:  n,
    Branch: fmt.Sprintf("foreman/issue-%d", n),
},
```

Replace with:

```go
Branch: fmt.Sprintf("foreman/%s/issue-%d", workload.Name, n),
```

Result: `foreman/v03-validation-batch-rerun-5/issue-510`. Reads naturally, groups by Workload, never collides with prior runs.

**2. branchNameForTask fallback also gets the workload prefix when present.**

For tasks NOT created by the M6 reconciler (explicit Pipeline mode, manual one-off), branchNameForTask should still produce something unique. Use the OwnerReferences chain: if the task has a Workload owner, prefix with the workload name; otherwise fall back to the current `foreman/issue-N` for true one-off tasks.

## Acceptance Criteria

[...]

PR #948, opened 2026-07-02
fix(foreman): replace stale agent-owned branches on push rejection (force-with-lease)

### LINKED ISSUE #934: coder: branch push is not idempotent across Workload re-runs (non-fast-forward NO-GO); error message says 'fork' but pushes to origin
## Problem 1: retries collide with their predecessor's branch

The coder derives its branch name deterministically (`foreman/<workload>/issue-<n>`) and pushes with a plain `git push --set-upstream origin <branch>`. If an earlier run of the same Workload already pushed that branch (e.g. coder GO'd, then a downstream task failed and the Workload was re-created — the exact flow the bridge retry loop produces), the re-run's push is rejected:

```
push rejected by remote: git push --set-upstream origin foreman/wl-…/issue-123: exit status 1:
 ! [rejected]  foreman/wl-…/issue-123 -> foreman/wl-…/issue-123 (non-fast-forward)
```

→ verdict NO-GO, outcome PUSH-FAILED. Every retry of a previously-pushed task fails forever until a human deletes the remote branch.

**Proposed fix:** the coder owns the `foreman/*` branch namespace — push with `--force-with-lease` (or delete-then-push) so a re-run replaces its own predecessor's branch. A human never commits to these branches, so the lease only protects against concurrent foreman runs, which is the right semantics.

## Problem 2: message says "fork" but the push targets origin

`pushFailedResult` reports "push to fork failed" (`pkg/foreman/agent/executor_native.go:1167`), but since #924 the coder pushes directly to the task's own repo (`origin`). The wording sent us down a "why is it forking?" rabbit hole in prod. Suggest "push to remote failed" (and the same for any other fork-era wording in executor messages).

Hit in production: retried Workloads after an unrelated infra failure; 3 of 3 previously-GO'd tasks NO-GO'd on re-run.

joryirving:
## What

`repo.Push` gains `ReplaceOnReject`: on a non-fast-forward rejection, re-push with `--force-with-lease` pinned to the remote SHA read via `ls-remote` — a compare-and-swap that replaces exactly the ref we observed and still fails if a concurrent writer moves it. Gated by a new opt-in flag, `Workload.spec.allowOverwrite`, stamped onto the synthesized issue-fix payload.

## Why

Fixes #934 (the behavioral half; the "push to fork" wording is left for a cosmetic pass)

Workload re-runs derive the same deterministic branch name as their predecessor, so the retry's push dies non-fast-forward on the previous run's ref: every delete-and-recreate retry ends NO-GO / PUSH-FAILED until a human deletes the branch. Observed in production: 3 of 3 previously-GO'd tasks failed exactly this way on re-run.

**On #573's design rejection** (thanks for the catch — I hadn't traced that history): this now adopts exactly the opt-in shape #573 prescribed ("an opt-in `Workload.spec.allowOverwrite: true` flag could come later if real automated retries need it") rather than superseding it. The default stays false — a plain Workload never force-replaces a published ref, preserving the audit-artifact property #573 protected. Retry systems that re-run Workloads under the same name (the "real automated retries" #573 anticipated — ours exists now) set the flag explicitly on the re-run, where the stale ref is their own previous attempt.

## How

- `isPushRejection` now inspects only the failed invocation's **stderr**, carried on a typed `gitError` from `runGit` (review point 1: the full message embeds the git args, and branch names carry caller-controlled Workload names — a workload literally named `non-fast-forward` must not classify auth/network failures as rejections)
- plain push first; on rejection with `payload.allowOverwrite`: `ls-remote` → `--force-with-lease=refs/heads/<branch>:<sha>`; a vanished ref leases against empty; lease failure re-wraps as `ErrPushRejected`
- `Workload.spec.allowOverwrite` → `synthesizeIssueBatch` stamps `payload.allowOverwrite` on the issue-fix step → executor sets `ReplaceOnReject` from the payload (explicit pipelines set it per step); CRDs + chart CRDs regenerated

Tests: real-git `push_test.go` reproduces the production collision (rejected without the flag / replaced with it, remote ref verified equal to the retry's HEAD / clean push never forces); envtest pins the operator stamping and the false default.

AI assistance: authored by Claude (Opus 4.8) operating under my direction against our production cluster; I reviewed the change and test output before submitting.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (repo + controller suites green)
- [x] `make lint` passes locally (`golangci-lint run` 0 issues)
- [x] Documentation updated (if user-facing change) — new `allowOverwrite` field documented in the CRD/API comments

### DIFF
--- a/api/foreman/v1alpha1/agentictask_types.go
+++ b/api/foreman/v1alpha1/agentictask_types.go
@@ -254,6 +254,19 @@ type AgenticTaskPayload struct {
 	// +optional
 	BaseBranch string `json:"baseBranch,omitempty"`
 
+	// AllowOverwrite permits the coder to replace a stale remote ref for
+	// its own branch (force-with-lease pinned to the observed SHA) when a
+	// plain push is rejected non-fast-forward. Opt-in, per the design
+	// decision in #573: silently overwriting published branches is the
+	// wrong default for a system that leaves an audit trail, but
+	// automated retry systems that delete-and-recreate Workloads derive
+	// the same branch name every run and are otherwise permanently
+	// wedged by their predecessor's ref (#934). Stamped from
+	// Workload.spec.allowOverwrite in issue-batch mode; explicit
+	// pipelines set it per step.
+	// +optional
+	AllowOverwrite bool `json:"allowOverwrite,omitempty"`
+
 	// BranchPrefix overrides the branch name prefix on issue-fix tasks
 	// (default derived from the issue's labels via conventional commit
 	// prefixes: fix/, feat/, chore/, etc.).
--- a/api/foreman/v1alpha1/workload_types.go
+++ b/api/foreman/v1alpha1/workload_types.go
@@ -164,6 +164,16 @@ type WorkloadSpec struct {
 	// +optional
 	EscalationReviewerAgentRefs []corev1.LocalObjectReference `json:"escalationReviewerAgentRefs,omitempty"`
 
+	// AllowOverwrite lets this Workload's coder replace a stale remote
+	// ref for its own foreman/* branch (force-with-lease compare-and-swap)
+	// instead of failing non-fast-forward. Opt-in — #573 deliberately
+	// rejected force-on-collision as a default; retry systems that
+	// re-run Workloads under the same name set this on the re-run, where
+	// the stale ref is their own previous attempt. Copied onto every
+	// synthesized issue-batch task payload.
+	// +optional
+	AllowOverwrite bool `json:"allowOverwrite,omitempty"`
+
 	// AllowCloudReviewers gates whether reviewer Agents whose
 	// spec.provider is "cloud-proxy" (or any non-"local" value) may be
 	// dispatched for this Workload. Three-valued via the *bool:
--- a/charts/foreman/templates/crds/agentictasks.yaml
+++ b/charts/foreman/templates/crds/agentictasks.yaml
@@ -160,6 +160,19 @@ spec:
                       Agent is the named agent to invoke. Required for freeform; defaults
                       to "issue-fixer" for issue-fix and "verify" for verify.
                     type: string
+                  allowOverwrite:
+                    description: |-
+                      AllowOverwrite permits the coder to replace a stale remote ref for
+                      its own branch (force-with-lease pinned to the observed SHA) when a
+                      plain push is rejected non-fast-forward. Opt-in, per the design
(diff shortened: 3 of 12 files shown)


Defilan:
### Code review

Found 3 issues:

1. `isPushRejection` matches on the whole error string, which includes the joined git ARGS (`runGit` builds `"git %s: %w: %s"` args-first). Branch names embed Workload names, and `non-fast-forward`/`rejected` are legal Kubernetes names, so a workload named e.g. `non-fast-forward` makes every push failure on its branch (auth, DNS, network) classify as a rejection and trigger the force-with-lease retry. Sniffing only the captured stderr fixes it.

https://github.com/defilantech/LLMKube/blob/96eebef93808aba0ac554d53da0fc98ee4155c3e/pkg/foreman/agent/repo/push.go#L149-L158

2. This reverses a documented design rejection without acknowledging it: issue #573's Out of Scope says verbatim "Force-with-lease on collision. Considered. Rejected: silently overwriting published branches is the wrong default for a system designed to leave an audit trail. An opt-in `Workload.spec.allowOverwrite: true` flag could come later if real automated retries need it." This PR is exactly that automated-retry case, but ships the behavior hardcoded-on for the whole `foreman/*` namespace rather than the opt-in flag #573 prescribed, and the replaced ref can contain previously-GO'd commits (the artifact #573 protected). Worth either adopting the opt-in shape or explicitly superseding #573's decision in the PR/comments with the rationale.

https://github.com/defilantech/LLMKube/blob/96eebef93808aba0ac554d53da0fc98ee4155c3e/pkg/foreman/agent/executor_native.go#L645-L654

3. The PR body is missing the template's `## Checklist` section (AGENTS.md: What / Why / How / Checklist), including the AI-assistance disclosure attestation.

https://github.com/defilantech/LLMKube/blob/96eebef93808aba0ac554d53da0fc98ee4155c3e/.github/PULL_REQUEST_TEMPLATE.md#L15-L23

The lease mechanics themselves verified clean: empty-expect lease on a vanished branch is correct git semantics, no shell/injection surface (argv exec), and lease failure re-wraps as ErrPushRejected.

joryirving:
Review feedback addressed — see updated PR body (template + AI disclosure) and 210d7c0 (stderr-only rejection sniffing via typed gitError; force-with-lease is now the opt-in Workload.spec.allowOverwrite flag #573 prescribed, default off, stamped onto the issue-fix payload).

Defilan (approved):

Issue #942, opened 2026-07-02
[FEATURE] Foreman edit-fidelity: make str_replace resilient to local-model old_string drift (fleet-wide)


Defilan:
## Feature Description

Harden the Foreman coder edit path so a local model that cannot reproduce an exact `old_string` still lands its intended edit, instead of looping to `EditFreeStreak`/INCOMPLETE. This is a fleet-wide harness improvement, not a per-model tweak.

## Problem Statement

`str_replace` matches `old_string` by exact substring. Local/open-weight coder models frequently retype `old_string` from memory and drift (wrong tokens, not just whitespace), so the match returns 0 and the turn is wasted. PR #917 added two recovery layers (a whitespace-normalized fallback that auto-applies on a unique normalized match, and an `anchorContext` hint that returns the real bytes for the model to copy), but two gaps remain:

1. The whitespace fallback only rescues **whitespace** drift, not **token** drift.
2. The `anchorContext` hint hands the model the correct text, but the model has to copy it verbatim on the next turn, and low-active-param MoEs often re-drift instead.

Evidence (two different local models, same failure):
- **Ornith-35B, today** (run-20260702-001622): #478 and #907 both went INCOMPLETE via `EditFreeStreak`. #907 read the target file 9 times and still emitted `old_string found 0 times` repeatedly. #478 also hit an absolute-path rejection and chose `str_replace` on a file it never read when `write_file` was appropriate.
- **Qwopus-A3B, earlier (#867)**: identical `old_string` hallucination drove RepeatedToolCall over ~78 turns.

Most-affected models are the low-active-param MoEs the fleet favors for cheap prefill (Ornith-35B, Qwen3.6-35B-A3B, Qwopus-A3B); dense coders (Qwopus-27B, Devstral-24B) fumble less but still benefit. Because the fix lives in the edit tool, hardening it once lifts every coder on the fleet.

> As a fleet operator, I want the edit tool to tolerate a models near-miss `old_string` so the coder lands its edit, rather than needing each model to be perfect at verbatim copying.

## Proposed Solution

Three levers, roughly in impact order:

[...]

PR #957, opened 2026-07-03
fix(foreman): accept workspace-internal absolute paths in resolveInside

### LINKED ISSUE #945: [FEATURE] Foreman tools: accept workspace-internal absolute paths instead of rejecting them
## Feature Description

Path-bounded coder tools (`str_replace`, `read_file`, `write_file`, `grep`) reject absolute paths outright, even when the path is inside the workspace: `resolveInside` returns `path escapes workspace: "<path>" is absolute`. Accept workspace-prefixed absolute paths by stripping the workspace root and continuing with the existing containment checks.

## Problem Statement

Local coder models constantly see absolute paths in their own bash output (`find`, `ls`, `git status` run against `$WORKSPACE_ROOT`) and then paste those paths into edit tools. The rejection costs a full turn each time, which is fatal under aggressive stuck-loop forcing (3 restricted turns).

Observed live twice in one day, both on the same fleet:

- run-20260702-110319-code-478, turn 5: `str_replace` with the absolute workspace path of the target file, rejected, one of only 3 restricted turns burned.
- revise-478-r1-code-478, turn 5: identical pattern (`str_replace: path escapes workspace: "/Users/defilan/foreman-workspaces/default/revise-478-r1-code-478/docs/site/guides/metrics-driven-autoscaling.md" is absolute`); the task died two turns later.

> As a coder-harness operator, I want workspace-internal absolute paths accepted so that a model quoting its own bash output does not lose a turn to a rejection that carries no safety value.

## Proposed Solution

In `resolveInside` (pkg/foreman/agent/tools/workspace.go): when the incoming path is absolute AND has the workspace root as a prefix (after cleaning/symlink-normalizing both sides), strip the prefix and proceed exactly as for a relative path. Absolute paths OUTSIDE the workspace stay rejected with the current error. The containment guarantee is unchanged; this only widens the accepted spelling of paths that already resolve inside.

Also worth updating the rejection message for genuinely-external absolute paths to state the rule ("paths must be workspace-relative"), since the current wording tells the model what is wrong but not what to do.

## Alternatives Considered

- Prompt-only mitigation (tell models to use relative paths): already implicit in the tool schema; empirically insufficient for 35B-class models quoting bash output.
- Rejecting but echoing the workspace-relative form in the error ("did you mean docs/site/guides/x.md?"): helpful, but still burns the turn; accepting is strictly better and no less safe.

## Additional Context

Follow-up from the #942 edit-fidelity work (PR #943); both reproductions came from its live fleet validation. Related: #944 (no-anchor feedback).

Defilan:
## What

Widens `resolveInside` (the path-containment boundary for every file tool) to accept absolute paths that resolve *inside* the workspace, instead of rejecting all absolute paths outright. External absolute paths are still rejected, now with a clearer message.

## Why

Fixes #945

Local coder models constantly quote absolute paths from their own bash output (`find`, `ls`, `git status` against `$WORKSPACE_ROOT`) and paste them into edit tools. The blanket rejection cost a full turn on each — fatal under the 3-turn restricted-edit budget of an aggressive stuck-loop profile. Observed live twice on the fleet (the #942/#943 validation runs).

## How

When the input is absolute, compute `filepath.Rel(wsResolved, filepath.Clean(p))`; reject if the result escapes (`..`), otherwise strip the prefix and fall through to the **unchanged** relative-path containment path (Join → Rel check → terminal `Lstat`+`EvalSymlinks` symlink-escape check). Containment is computed with `filepath.Rel` (a real path relationship), never a naive string prefix, so a sibling like `<ws>-evil` is correctly rejected.

Adversarially verified: prefix-collision, absolute traversal, non-existent-path lexical escape, and absolute-path-through-an-escaping-symlink (with real on-disk symlinks) all reject; only the accepted *spelling* of already-inside paths widens. The two highest-value cases (prefix-collision, absolute+symlink escape) are pinned as regression tests so a future refactor can't reintroduce a `strings.HasPrefix` bug.

Authored by **Foreman**, LLMKube's agentic coder harness (local model on the lab fleet); reviewed adversarially and hardened with the two extra containment tests by the maintainer, who takes responsibility for the change per the AI-assisted contribution policy.

## Checklist

- [x] Tests added/updated - `TestResolveInside_Cases` gains absolute-inside/outside, prefix-collision, and absolute-symlink-escape subtests
- [x] `make lint` passes locally (+ `GOOS=linux golangci-lint`, 0 issues)
- [x] Documentation updated (if user-facing change) - not user-facing (internal tool boundary)

### DIFF
--- a/pkg/foreman/agent/tools/tools_test.go
+++ b/pkg/foreman/agent/tools/tools_test.go
@@ -177,8 +177,19 @@ func TestResolveInside_Cases(t *testing.T) {
 		}
 	})
 
-	t.Run("absolute rejected", func(t *testing.T) {
-		_, err := resolveInside(ws, filepath.Join(ws, "ok.txt"))
+	t.Run("absolute inside workspace accepted", func(t *testing.T) {
+		abs := filepath.Join(ws, "ok.txt")
+		got, err := resolveInside(ws, abs)
+		if err != nil {
+			t.Fatalf("resolveInside absolute inside: %v", err)
+		}
+		if got != abs {
+			t.Errorf("absolute inside: got %q want %q", got, abs)
+		}
+	})
+
+	t.Run("absolute outside workspace rejected", func(t *testing.T) {
+		_, err := resolveInside(ws, "/etc/passwd")
 		if !errors.Is(err, ErrPathEscapesWorkspace) {
 			t.Errorf("expected ErrPathEscapesWorkspace, got %v", err)
 		}
--- a/pkg/foreman/agent/tools/tools_test.go
+++ b/pkg/foreman/agent/tools/tools_test.go
@@ -218,6 +218,46 @@ func TestResolveInside_Cases(t *testing.T) {
 		}
 	})
 
+	// Prefix-collision: a sibling directory whose name has the workspace path
+	// as a string prefix (ws + "-evil") must NOT be treated as inside. This
+	// pins the guarantee that containment uses filepath.Rel (a real path
+	// relationship), not a naive strings.HasPrefix on the cleaned path.
+	t.Run("absolute prefix-collision sibling rejected", func(t *testing.T) {
+		evil := ws + "-evil"
+		if err := os.MkdirAll(evil, 0o755); err != nil {
+			t.Fatalf("mkdir sibling: %v", err)
+		}
+		t.Cleanup(func() { _ = os.RemoveAll(evil) })
+		secret := filepath.Join(evil, "secret.txt")
+		if err := os.WriteFile(secret, []byte("x"), 0o644); err != nil {
+			t.Fatalf("seed sibling secret: %v", err)
+		}
+		if _, err := resolveInside(ws, secret); !errors.Is(err, ErrPathEscapesWorkspace) {
+			t.Errorf("prefix-collision sibling must be rejected, got %v", err)
+		}
+	})
+
+	// Absolute + symlink escape: the new absolute-path branch must still be
+	// caught by the terminal symlink-resolution check. An in-workspace symlink
+	// to an outside dir, referenced by its ABSOLUTE path, must be rejected.
+	t.Run("absolute path through escaping symlink rejected", func(t *testing.T) {
+		if runtime.GOOS == "windows" {
+			t.Skip("symlink semantics differ on windows")
+		}
+		outside := t.TempDir()
+		if err := os.WriteFile(filepath.Join(outside, "secret.txt"), []byte("x"), 0o644); err != nil {
+			t.Fatalf("seed outside secret: %v", err)
+		}
+		link := filepath.Join(ws, "abs-escape-link")
+		if err := os.Symlink(outside, link); err != nil {
+			t.Fatalf("symlink: %v", err)
+		}
+		abs := filepath.Join(ws, "abs-escape-link", "secret.txt")
+		if _, err := resolveInside(ws, abs); !errors.Is(err, ErrPathEscapesWorkspace) {
+			t.Errorf("absolute path through escaping symlink must be rejected, got %v", err)
+		}
+	})
+
 	t.Run("nonexistent path ok for write_file", func(t *testing.T) {
 		_, err := resolveInside(ws, "subdir/new.txt")
 		if err != nil {
--- a/pkg/foreman/agent/tools/workspace.go
+++ b/pkg/foreman/agent/tools/workspace.go
@@ -43,10 +43,6 @@ var ErrPathEscapesWorkspace = errors.New("path escapes workspace")
 // is skipped. The workspace itself is EvalSymlinks-resolved so a
 // symlinked workspace is treated as its real path.
 func resolveInside(workspace, p string) (string, error) {
-	if filepath.IsAbs(p) {
-		return "", fmt.Errorf("%w: %q is absolute", ErrPathEscapesWorkspace, p)
-	}
-
 	wsAbs, err := filepath.Abs(workspace)
 	if err != nil {
 		return "", fmt.Errorf("workspace abs(%s): %w", workspace, err)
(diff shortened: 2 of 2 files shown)


joryirving (changes requested):
Two issues to address before merge:

**1. Misleading error message on relative-path escape (`workspace.go:65`)**

The new error `"paths must be workspace-relative"` fires when a relative path like `../etc/passwd` escapes the workspace. The path IS already relative; the problem is that it escapes containment. The model will read this and be confused — exactly the failure mode this PR exists to fix. The pre-PR `"%w: %q"` was clearer because it just showed the offending path.

Suggested: `"%w: %q resolves outside workspace"` (or similar that names the actual problem).

The sibling error in the absolute branch — `"absolute paths must be inside the workspace"` — is well-formed; the issue is only the relative-branch message.

**2. Stale doc comment (`workspace.go:36`)**

```
//  1. p is not absolute.
```

Rule 1 no longer applies: absolute paths resolving inside the workspace are now accepted. The comment should reflect the new contract — e.g. split into "absolute paths inside the workspace are accepted; absolute paths outside are rejected" alongside the existing rules 2 and 3.

Both are small; happy to re-review once addressed.

Defilan:
Thanks Jory — both fixed in `1db7e3f`:

1. The relative-branch escape error now reads `path escapes workspace: "../etc/passwd" resolves outside workspace` — names the actual problem and shows the offending path, matching the pre-PR clarity. Left the absolute-branch message as-is since you noted it was well-formed.
2. Refreshed doc rule 1 to state the new contract: absolute paths must resolve inside the workspace (outside is rejected) and are then treated as the equivalent relative path; relative paths as-is.

Tests + `GOOS=linux` lint green. Ready for re-review.

joryirving (commented):
Re-reviewing after the latest push. First, my CHANGES_REQUESTED findings — both addressed:

- **Misleading error message** — now `"%w: %q resolves outside workspace"`. Fixed.
- **Stale doc comment** (rule 1) — now reflects that absolute paths inside the workspace are accepted. Fixed.

Mea culpa: I posted this re-review without re-reading the diff after the author's update; the framing "Three additional findings" wrongly implied the original two were still open.

Three new findings against the current diff:

**1. Symlinked-workspace case still broken (medium, config-dependent)**

The new absolute branch compares `filepath.Rel(wsResolved, cleaned)` where `wsResolved = EvalSymlinks(wsAbs)`. When `Workspace` itself is reached via a symlink (e.g. `ws = /tmp/.../link` → `/tmp/.../real`), an absolute input using the **link spelling** is rejected even though it resolves inside:

```
wsResolved=/tmp/.../real
abs=/tmp/.../link/file.txt    # what $WORKSPACE_ROOT literally echoes
rel → "../link/file.txt"      # REJECTED
```

This is exactly the path the PR describes ("model quotes `find`/`ls`/`git status` output against `$WORKSPACE_ROOT` and pastes it") and the one the fix doesn't cover. `Workspace` is constructed as `filepath.Join(workspaceRoot, ns, name)` (`internal/controller/...`, `pkg/foreman/agent/executor_native.go:249`) and is *not* `EvalSymlinks`'d before being passed to `BashTool`; bash then echoes the un-resolved spelling back into `$WORKSPACE_ROOT` (`pkg/foreman/agent/tools/bash.go:275`).

Severity depends on whether your deployment has a symlinked workspace root. Typical `$HOME/foreman-workspaces/...` Linux layout won't trigger, but if the workspace is ever created through a symlink (or any parent component is) the fix regresses to the pre-PR behavior for that path.

Cheap fix: when the first `Rel` escapes, retry against `wsAbs` (un-resolved spelling); if it doesn't escape there, fall through to the existing `EvalSymlinks(p)` check for the symlink-escape case. A regression test mirroring the prefix-collision / absolute-symlink-escape shape, but with a symlinked workspace, would pin this.

**2. Minor doc inaccuracy (`workspace.go`)**

> Strip the workspace prefix and continue with the same containment checks as for a relative path.

The code computes `filepath.Rel`, not a literal prefix strip. Since the new prefix-collision test is *explicitly* about why `filepath.Rel` is not `strings.HasPrefix`, calling the operation "strip" reads as the simpler mechanism the test is pinning against. Suggest: "Compute the workspace-relative path…".

**3. Test coverage gap**

The "absolute inside workspace accepted" subtest uses `ws` from `makeWorkspace`, which is already `EvalSymlinks`'d (`tools_test.go:37`) — so `wsResolved == wsAbs` and the new branch never exercises the case it was added for. The hard spelling (`Workspace ≠ EvalSymlinks(Workspace)`) is the one the absolute-path branch is supposed to make work, and it's untested. A test in the same shape as the prefix-collision / absolute-symlink-escape ones — workspace through a symlink, input using the link spelling — would pin this.

**Non-issues confirmed**

- All four call sites (`read_file.go:116`, `str_replace.go:84`, `write_file.go:70`, `grep.go:100`) just wrap with `errors.Is(err, ErrPathEscapesWorkspace)` — error-message text change doesn't break consumers.
- Removed `"absolute rejected"` subtest cleanly subsumed by `absolute inside accepted` + `absolute outside rejected`.
- TOCTOU between `EvalSymlinks(wsAbs)` and the terminal `EvalSymlinks(joined)` is pre-existing, not introduced here.
- Bash `cd-guard` (`bash.go:84-86`) is a separate mechanism, unaffected.

**Verdict**

Ship-able. Address (1) only if a symlinked workspace root is realistic in your deployment matrix; otherwise add a one-liner near the new branch documenting the limitation so the next person debugging the same `$WORKSPACE_ROOT` paste failure has a pointer.

joryirving (approved):
Approve. Read this as a sandbox-boundary change, not just a UX fix, and traced the containment:

- Lexical escape via `..` is caught pre-symlink (`Clean` collapses `/ws/../etc/passwd` → `/etc/passwd` → `Rel` → `..` → reject).
- Prefix-collision uses `filepath.Rel`, not `HasPrefix`, so `<ws>-evil` → `../ws-evil/...` → reject (your test pins it).
- Crucially the abs branch strips to a relative path and **falls through to the unchanged terminal `EvalSymlinks` check** rather than short-circuiting it, so an abs path through an escaping symlink still dies (your test pins that too). It only widens the accepted *spelling* of already-inside paths — the reject set is unchanged. Solid.

One non-blocking observation, since it can quietly make the fix a no-op in some deploys rather than a security issue (it fails safe — rejects, never wrongly accepts): the abs branch computes `Rel(wsResolved, cleaned)` where `wsResolved` is `EvalSymlinks`-resolved, but the model's absolute paths come from `$WORKSPACE_ROOT`, which the executor sets to the *unresolved* workspace string (`bash.go` `WORKSPACE_ROOT = t.Workspace`, same string passed here). If `WorkspaceDir` has any symlinked component (a not-uncommon container mount shape), the model's own `$WORKSPACE_ROOT`-prefixed paths resolve to the symlink spelling and get rejected against the resolved root — i.e. the exact paths this PR means to accept. Do you canonicalize `WorkspaceDir` at startup? If so, moot. If not, resolving the input's symlinks before the `Rel` (or comparing against both `wsAbs` and `wsResolved`) would close it. Happy to confirm against our fleet's workspace path if useful.

PR #958, opened 2026-07-03
fix(foreman): defer generic self-gate when runtime is missing from coder image

### LINKED ISSUE #929: Coder in-pod self-gate fails non-Go GateProfiles when the coder image lacks the language runtime
### Summary

A non-Go `gateProfile` runs fine in the clean-room verify Job (it uses `gateProfile.image`), but the **coder's in-loop fast self-gate** runs the same commands **in the coder pod** via `sh -c` (`RunGenericGate` ← `execCommandRunner`). The published coder image is `FROM golang:1.26` (+ make/helm/golangci-lint) with no `node`/`python`/`npm`/`pip`, so a real `node`/`python` gate command fails with "command not found". `RunGenericGate` treats any non-zero exit as a failure, and `makeCoderGateVerifier` retries `coderGateMaxRetries` (3) times then **downgrades the GO to INCOMPLETE** — the coder never pushes, so the clean-room Job (which *would* run the command in the right image) never runs.

Net: a correct, working `gateProfile` for a non-Go repo makes the coder produce no PR, unless the operator builds a bespoke polyglot coder image.

### Repro

- Agent runs native/in-process on the stock coder image (`ghcr.io/defilantech/llmkube-foreman-agent-coder`).
- `Workload.spec.gateProfile: {language: node, commands: {test: "npm ci && npm test"}}` (or python).
- Coder reaches GO → self-gate runs `npm ci …` in the Go-only pod → `sh: npm: not found` → fail ×3 → INCOMPLETE, no branch.

### Proposed fix

Mirror the existing golangci-lint-bootstrap precedent (executor_native.go: "could not bootstrap golangci-lint; terminal stands, gate Job is the backstop"): when a generic self-gate command **cannot run because the tool/runtime is absent** (e.g. `command not found` / exec ENOENT), treat it as **could-not-verify** — let the terminal stand and defer to the clean-room verify Job, which runs in `gateProfile.image` and has the runtime. Distinguish that from a command that ran and genuinely failed (real gate failure → keep the current retry/feedback behavior).

Options:
- In `RunGenericGate`, classify exec-not-found / 127 as could-not-verify rather than failure (return a distinct signal), and have `makeCoderGateVerifier` pass the terminal through in that case.
- Or a lighter check: if none of the resolved gate commands' entrypoints are on `PATH` in the coder pod, skip the fast self-gate for this task and rely on the verify Job.

Either way the clean-room Job stays the authoritative gate; this only stops the fast pre-check from hard-failing on a missing runtime.

### Impact

Unblocks real per-language in-cluster gates for multi-repo fleets running one stock coder image, without a bespoke polyglot image. Sibling to #915/#924 (per-task repo) and #919/#920 (Workload gateProfile). Happy to take this one (Foreman area).

joryirving:
## What

When a generic (non-Go) self-gate command cannot run because its runtime/binary is missing from the coder image (exit 127 / exec ENOENT / shell "command not found"), skip the check with a "self-gate deferred: runtime missing" advisory instead of failing the gate. A command that ran and genuinely failed still fails as before; the Go gate path is untouched.

## Why

Fixes #929

The published coder image is Go-only, so a correct node/python `gateProfile` command fails "command not found" in the in-pod fast self-gate, gets retried 3 times, and downgrades the GO to INCOMPLETE — the clean-room verify Job (which runs in `gateProfile.image` and would pass) never gets a branch to verify.

## How

- `RunGenericGate` classifies each command error via a new `isMissingRuntime` (exit 127, `exec.ErrNotFound`, or the shell's "command not found" / ": not found" message — matched narrowly so a test failure that merely mentions "not found" is not misread) and returns deferrals as advisories alongside pass/feedback.
- `makeCoderGateVerifier` appends those advisories to the existing `gateAdvisories` accumulator, so the deferral surfaces to the reviewer via `Result.Extra["gateAdvisories"]`, and logs each deferral.
- Tests: table-driven `isMissingRuntime` cases (incl. non-deferring false positives), fake-runner gate tests (127 → deferred, real failure → still fails, mixed), and a real-exec test through `execCommandRunner` (missing binary defers, `exit 1` fails, `true` passes).

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (`go test ./pkg/foreman/... -count=1` green)
- [x] `make lint` passes locally (`golangci-lint run ./pkg/foreman/...` — 0 issues)

### DIFF
--- a/pkg/foreman/agent/executor_native.go
+++ b/pkg/foreman/agent/executor_native.go
@@ -701,7 +701,16 @@ func makeCoderGateVerifier(
 		// resolved commands. The Go path below is left byte-identical: a nil,
 		// empty-language, or explicit-"go" profile takes it unchanged.
 		if usesGenericGate(profile) {
-			pass, feedback := RunGenericGate(ctx, workspace, profile.Resolve(), execCommandRunner)
+			pass, feedback, advisories := RunGenericGate(ctx, workspace, profile.Resolve(), execCommandRunner)
+			if acc != nil {
+				*acc = append(*acc, advisories...)
+			}
+			for _, a := range advisories {
+				// A missing runtime in the coder image (#929) defers the check
+				// to the clean-room verify Job instead of failing the GO.
+				log.Info("coder gate (generic): check deferred; verify Job is the backstop",
+					"language", string(profile.Language), "check", a.Check, "detail", a.Detail)
+			}
 			if !pass {
 				log.Info("coder gate (generic): fast checks failed; returning feedback to the loop for a fix",
 					"language", string(profile.Language))
--- a/pkg/foreman/agent/generic_gate.go
+++ b/pkg/foreman/agent/generic_gate.go
@@ -18,6 +18,9 @@ package agent
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"os/exec"
 	"strings"
 
 	foremanv1alpha1 "github.com/defilantech/llmkube/api/foreman/v1alpha1"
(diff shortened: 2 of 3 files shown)


Defilan:
### Review

**Approve.** The false-defer risk is architecturally excluded for the dangerous case: the whole deferral path sits behind `usesGenericGate(profile)` (`pkg/foreman/agent/generic_gate.go:31`), which returns false for a `go`/empty profile, so a Go change on the Go image always takes the untouched `RunCoderGate` path and can never be wrongly deferred. For non-Go targets the primary signal is exit 127, which is precise (`sh -c` returns exactly 127 on a missing interpreter, preserved by `execCommandRunner`), and a deferral is safe because the clean-room gate Job re-runs the full generic suite in `gateProfile.image` (which has the runtime) and every deferral is surfaced as a `self-gate-deferred` advisory in `Result.Extra["gateAdvisories"]`. Test coverage is strong in both directions, including the false-positive traps (`pytest fixture not found`, `npm 404 Not Found`).

One non-blocking nit: the bare `": not found"` output-substring match (`generic_gate.go:77`) is the one branch that can misclassify a **real** non-Go failure as a deferral — a command that runs and exits non-zero whose output contains e.g. `Error: not found` or `handler.go:42: resource: not found`. The clean-room Job still catches it, but the coder loses the fast in-pod feedback cycle for that fix, which defeats the point of the tier. Since exit 127 already covers genuine missing-runtime for both bash and dash, consider dropping the bare `": not found"` (keep the full `"command not found"` phrase) and adding a true-negative test row for it. Not a merge blocker.

Defilan (approved):

Issue #546, opened 2026-05-25
[FEATURE] Foreman v0.2 / v0.8.0: close the reviewer-step empirical gap


Defilan:
## What

Foreman v0.2 closes the reviewer-step empirical gap that v0.1 surfaced under load. Tracks the full work for the **v0.8.0 release**, which lands Foreman's reviewer architecture in a state that actually delivers the "drop a backlog, walk away, come back to merge-ready PRs" claim.

This epic is a planning + tracking issue. Each sub-task lands as its own PR.

## Why

The Memorial Day v5 batch (run 2026-05-25) plus a controlled reviewer-comparison experiment (same day) produced a clear empirical finding:

- The v0.1 pipeline's same-model reviewer (Qwen 3.6 35B-A3B reviewing Qwen 3.6 35B-A3B output) approves structurally-broken PRs.
- Different-family local reviewers at the 24-26B scale (Devstral, Gemma 4) also approve the same PRs.
- Tools, more context (TurboQuant 65K), structured deliberation prompts, and 3-of-3 unanimous ensemble all produce a 0/3 catch rate on three hand-curated structural bugs.
- A frontier cloud model (Claude, tool-using) catches 3/3.
- Falsification framing + PR-metadata redaction (on Devstral) and thinking-mode (on Qwen3) DO shift verdicts to REQUEST-CHANGES on different PRs each, for false-positive-ish reasons. Combined any-NO-GO escalation flags 3/3 PRs for human review without catching the actual bugs.

Full experiment writeup: `llmkube-internal/diagnostics/reviewer-comparison-2026-05-25/SUMMARY.md` (private repo; will be summarized publicly in a research-note blog post separate from this release).

The v0.2 architecture this implies:

1. **Multi-strategy local reviewer ensemble** (3 prompt strategies, any-NO-GO escalates) closes the gap for air-gap-only orgs in the "flag for human review" sense.
2. **Optional hybrid cloud reviewer** (sovereignty-toggled) closes the gap structurally for orgs that can use it.
3. **Pipeline reliability fixes** unblock the production reliability gaps that surfaced under load.

## How

Seven sub-tasks, each its own PR. Day-by-day execution plan with rough scoping:

### Reliability fixes (the bugs filed during the v5 dogfood)

[...]

PR #956, opened 2026-07-03
feat(foreman): open the pull request on review GO

joryirving:
## What

A reviewer GO now opens the Workload's pull request — idempotently — instead of leaving a reviewed branch unopened on the remote. New `githubpr` package (mirrors `githubissue`); `Workload.spec.openPullRequest` gates it (nil = **true** in issue-batch mode, `false` opts out; explicit pipelines set `payload.openPullRequest` per step).

## Why

Fixes #937

The pipeline ended one hop short of its artifact: coder GO → gate PASS → review GO → …a branch sitting on the remote until a human noticed. We have hand-opened 12 PRs for completed Workloads in production since filing #937.

## How

- `githubpr.EnsurePR`: list-by-head first (any state → reuse), then create; a 422 create race (two reviewer tasks GOing concurrently) resolves to the winner's PR. Title = branch head's commit subject (fallback `Fix #<n>`), base = `payload.baseBranch` (default main), body = `Fixes #<n>` + workload provenance.
- Executor: after the reviewer enforcement passes settle a GO, `maybeOpenPullRequest` runs best-effort with the same token the coder pushed with — the approval stands if GitHub is down; `pullRequestURL`/`pullRequestError` surface in result extra.
- Operator: `synthesizeIssueBatch` stamps `payload.openPullRequest` on review steps from the spec field (three-valued `*bool`, default true — an issue-batch Workload exists to produce a PR). CRDs + chart CRDs regenerated.
- `PREnsurer` is injected (like `IssueFetcher`); nil disables the feature entirely.

Follow-up candidates kept out of scope: draft PR on NO-GO carrying review findings (noted in #937), GHES base-URL config.

AI assistance: authored by Claude (Opus 4.8) operating under my direction; I reviewed the change and test output before submitting.

## Checklist

- [x] Tests added/updated (httptest client suite: create / reuse / 422 race / auth failure / title fallback; envtest: default-true, opt-out, review-step-only stamping)
- [x] `make test` passes locally (agent + controller suites green)
- [x] `make lint` passes locally (`golangci-lint run` 0 issues)
- [x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)
- [x] All commits are signed off (`git commit -s`) per [DCO](https://developercertificate.org/)
- [x] AI assistance (if any) is disclosed above, per [CONTRIBUTING.md](../CONTRIBUTING.md#ai-assisted-and-agent-contributions)
- [x] Documentation updated (if user-facing change) — new fields documented in CRD/API comments

PR #959, opened 2026-07-03
feat(foreman): bounded fix iteration on reviewer NO-GO instead of terminal failure

### LINKED ISSUE #946: Workload: review NO-GO should iterate — re-run the coder with the review feedback instead of terminally failing
## Problem

A reviewer NO-GO terminally fails the Workload. The review result carries exactly what a fix needs — structured `findings` (`missing_tests`, `scope_creep` + `*_details`), a summary, and the branch — but nothing consumes it. External retry systems can only delete + recreate the Workload, which re-runs the coder *blind*: same issue, same prompt, no knowledge of why it was rejected. Observed in production: 4 issues burned 3 attempts each producing near-identical rejected patches (e.g. reviewer: `scope_creep: commit reduces ACCESS_TOKEN_EXPIRE_MINUTES from 10080 to 30, unrelated to #141` — the retry did it again).

## Proposed fix

On review NO-GO, the WorkloadReconciler appends an iteration instead of failing (bounded):

- new step `code-<n>-r<k>` (same coder AgentRef), `payload.prompt` = the review's findings/details/summary, branch unchanged (the coder amends its own branch)
- fresh `verify-<n>-r<k>` + `review-<n>-<i>-r<k>` chained after it
- bounded by e.g. `spec.maxReviewIterations` (default 1–2); exhausting it fails the Workload as today
- `status` records the iteration count so operators can see convergence vs thrash

This mirrors what a human does with request-changes and is where the pipeline's economics live: a NO-GO with feedback is a nearly-done task, and throwing the context away wastes the most expensive part of the run.

## Workaround

Our dispatch bridge (v0.5.0) now reads the failed Workload's task results and recreates it as an explicit `spec.pipeline` with the review feedback injected via the code step's `payload.prompt` — works, but reimplements `synthesizeIssueBatch` outside the operator and loses the same-Workload continuity.

Related: #934 (branch push idempotency — an in-Workload iteration amending its own branch avoids that class entirely), #937 (a NO-GO's final state could also be a draft PR carrying the feedback).

### LINKED ISSUE #951: [FEATURE] Foreman: first-class review-feedback revision cycles (revise from prior branch, review-as-spec, revision-tuned profile)
## Feature Description

First-class review-feedback revision cycles: when a reviewer returns REQUEST-CHANGES (or a maintainer supplies review findings), Foreman should spawn a follow-up coder task that starts FROM the prior attempt's branch with the review as its spec, instead of requiring a hand-authored Pipeline Workload per round.

## Problem Statement

Today the loop from review back to code is manual. It works (validated live: a maintainer REQUEST-CHANGES review with 6 findings was hand-packed into a Pipeline Workload's `payload.prompt` and converged to GO + GATE-PASS on the third round), but the manual orchestration surfaced three problems the feature should own:

1. **Branch restore does not belong in the prompt.** Round 1 asked the model to `git fetch` + `git checkout` its prior work via bash; the multi-step restore collided with the stuck-loop forcing window (bash was disabled mid-restore) and the task died at turn 7. The executor owns git everywhere else; it should own this too.
2. **Forcing profiles are task-shape dependent.** Round 2 fixed the restore (single mandated first command, worked) but died anyway: the issue-fix ModelProfile's `editFreeTurnsLimit: 4` cannot hold a restore plus the source-verification reads a review demands, and the model collapsed under the nudge (degenerate `str_replace("placeholder","placeholder")` x3). A revision-tuned profile (`editFreeTurnsLimit: 12`, verification-oriented addendum) converged in round 3. A revision task dispatched with an issue-fix forcing profile is a config bug the controller can prevent.
3. **Per-round Workload authoring is boilerplate.** Rounds differ only in prompt text and branch names.

> As an operator running local-model coder fleets, I want review feedback to flow back into a bounded revision loop natively, so that REQUEST-CHANGES converges to a mergeable branch without hand-authored Workloads per round.

## Proposed Solution

Sketch, open to design review:

- **Payload:** add `reviseFromBranch` to `AgenticTaskPayload`. On an issue-fix task, the executor creates the working branch FROM that ref (fetched from the fork remote) instead of from `baseBranch`, so the prior attempt's files are simply present. No prompt-driven git.
- **Review-as-spec:** keep the existing semantics where a non-empty `payload.prompt` suppresses the lazy issue-body fetch; the revision prompt is the review findings (each item ideally citing the source file that proves the correct answer).
- **Profile pairing:** allow the revision path to select a revision-tuned Agent or ModelProfile (e.g. `Workload.spec.revisionAgentRef`, or a `revisionModelProfileRef` on the Agent). The controller rejects or warns when a revision task would run under an issue-fix forcing profile.
- **Controller loop:** when a reviewer task on an issue returns REQUEST-CHANGES/NO-GO with structured findings, the WorkloadReconciler emits one revision code+verify pair (findings -> prompt, prior branch -> reviseFromBranch), bounded by a `maxRevisionRounds` cap (default 1-2). The verdict-triggered fan-out pattern already exists in the escalation-reviewer tier.

## Alternatives Considered

- Keep hand-authored Pipeline Workloads per round (status quo): works, validated, but repeats the three failure modes above for every operator who discovers them the hard way.
- Prompt-only restore (no `reviseFromBranch`): empirically fragile under forcing windows (round 1), and makes the executor's "owns git" invariant leaky.

## Additional Context

Live validation trail (2026-07-02): revision rounds r1/r2/r3 on issue #478; round 3 delivered GO + GATE-PASS with 5.5 of 6 review findings verified landed. Related: #944 (no-anchor feedback), #945 (workspace-absolute paths), both filed from the same runs. The escalation-reviewer tier (#546) provides the verdict-triggered fan-out precedent.

joryirving:
> **Depends on #967** (`feat/revise-from-branch`, the executor restore extracted from #951). This branch is rebased on top of it, so the diff includes #967's commit until it merges; after that it shrinks to just the iteration loop.

## What

On a reviewer NO-GO the WorkloadReconciler now appends a bounded fix iteration instead of terminally failing the Workload: a new coder step `code-<n>-r<k>` on the same branch whose `payload.prompt` carries the reviewer's summary + structured findings, chained to fresh `verify-<n>-r<k>` + `review-<n>-<i>-r<k>` steps. Bounded by new `Workload.spec.maxReviewIterations` (`*int32`; unset defaults to 1, explicit 0 restores fail-on-first-NO-GO). `status.reviewIterations` records the emitted count and a `ReviewIterationTriggered` condition names the re-dispatched issues.

Rework after the hold-for-#951 review — the retry now amends the prior attempt instead of rebuilding from main:

- Iteration coder payloads stamp `reviseFromBranch` = the task's own branch (alongside `allowOverwrite`), so the executor (#967) restores the r0 attempt from the push remote before the model runs. The feedback prompt now states the real semantics: the workspace starts from the prior attempt, apply the findings as a delta, don't rebuild.
- New `Workload.spec.revisionCoderAgentRef` implements the profile-pairing piece of #951: iteration coder steps use it when set, pairing revisions with a revision-tuned profile instead of the issue-fix forcing profile. When unset they fall back to `coderAgentRef` and the reconciler emits a Warning event (`RevisionUnderIssueFixProfile`) on the Workload.
- `maxReviewIterations` default stays 1: with the restore in place, an on-by-default iteration can no longer destroy a passing base attempt.

## Why

The review result carries exactly what a fix needs (findings + summary + branch), but a NO-GO discarded it; external retry systems could only delete + recreate the Workload and re-run the coder blind, reproducing the same rejected patch.

Fixes #946. With #967, closes #951.

## How

- `internal/foreman/controller/workload_iteration.go`: emission modeled on `emitEscalations` — pure step-synthesis function (`reviewIterationSteps`), idempotency by step-name existence (partial creates repaired next reconcile), MaxTasks accounting (`MaxTasksIterationCap`), sovereignty gates on iteration reviewers, steady-state condition short-circuit, labeled in-flight placeholders so rollup/escalation see the new round in the same pass.
- Feedback prompt distills the review task's `status.result.extra.modelExtra` via `pkg/foreman/agent/reviewer.ParseFindings`, with a raw-JSON fallback for non-conforming (legacy map-shaped) findings.
- Rollup and the escalation trigger now judge each issue by its latest iteration only (`activeChildren`): a converged retry completes the Workload, a superseded round's NO-GO no longer pins it at Failed, and the escalation tier defers until iterations settle. Escalation semantics are unchanged for workloads that opt out (existing #546 tests now set `maxReviewIterations: 0` since the default changed to 1).
- `patchReviewAdvisories` wires each iteration's reviews to their own round's coder task (escalate reads the latest) instead of always `code-<n>`.
- `WorkloadReconciler` gains the operator-standard `events.EventRecorder` (nil-safe for tests), wired in `cmd/foreman-operator/main.go`; operator chart RBAC gains `events.k8s.io` create/patch.
- CRD manifests + chart CRDs regenerated.

Notable default change: unset `maxReviewIterations` means one fix iteration, so a NO-GO now fails the Workload one round later than before.

Verification: `KUBEBUILDER_ASSETS=... go test ./internal/foreman/controller/ -count=1` (envtest specs for NO-GO-with-budget → r1 steps with feedback prompt + `reviseFromBranch` + `allowOverwrite` and the fallback Warning event, `revisionCoderAgentRef` selected when set with no warning, exhausted budget → Failed/ChildrenIncomplete, GO → no iteration, iteration count in status, idempotency; plus table tests for step synthesis, coder-ref selection, prompt rendering, and superseded-round filtering); `go test ./pkg/foreman/...` ok; `golangci-lint run ./internal/... ./pkg/... ./api/...` 0 issues.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/api/foreman/v1alpha1/workload_types.go
+++ b/api/foreman/v1alpha1/workload_types.go
@@ -185,6 +185,27 @@ type WorkloadSpec struct {
 	// +optional
 	AllowOverwrite bool `json:"allowOverwrite,omitempty"`
 
+	// MaxReviewIterations bounds how many fix iterations the
+	// WorkloadReconciler may append per issue after a reviewer NO-GO
+	// (#946). On a NO-GO, instead of failing the Workload, the
+	// reconciler emits a new coder task ("code-<N>-r<k>", same
+	// CoderAgentRef, same branch, payload.allowOverwrite=true) whose
+	// payload.prompt carries the reviewer's structured findings and
+	// summary, then chains a fresh verify + reviewer fan-out behind
+	// it ("verify-<N>-r<k>", "review-<N>-<i>-r<k>"). Superseded
+	// iterations stop counting toward the rollup; the Workload
+	// completes when the final iteration's reviewers say GO and
+	// fails (today's behavior) when the last allowed iteration is
+	// still NO-GO.
+	//
+	// nil (unset) defaults to 1 iteration. An explicit 0 disables
+	// iteration and restores the fail-on-first-NO-GO behavior.
+	// Issue-batch mode only; ignored in explicit Pipeline mode.
+	// Iteration tasks count against MaxTasks.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	MaxReviewIterations *int32 `json:"maxReviewIterations,omitempty"`
+
 	// AllowCloudReviewers gates whether reviewer Agents whose
 	// spec.provider is "cloud-proxy" (or any non-"local" value) may be
 	// dispatched for this Workload. Three-valued via the *bool:
@@ -303,6 +324,15 @@ type WorkloadStatus struct {
 	// branch didn't pass the checks."
 	// +optional
 	IncompleteTasks int32 `json:"incompleteTasks,omitempty"`
+
+	// ReviewIterations counts the fix iterations the reconciler has
+	// emitted after reviewer NO-GO verdicts (#946), summed across
+	// issues. Zero (or absent) means no reviewer ever bounced a
+	// branch back to the coder; operators watch this to distinguish
+	// convergence (small counts) from thrash (counts near
+	// issues * maxReviewIterations).
+	// +optional
+	ReviewIterations int32 `json:"reviewIterations,omitempty"`
 
 	// PlannerModel records which frontier model the planner actually used
 	// for this workload. Set after the planner runs.
--- a/api/foreman/v1alpha1/workload_types.go
+++ b/api/foreman/v1alpha1/workload_types.go
@@ -188,11 +188,14 @@ type WorkloadSpec struct {
 	// MaxReviewIterations bounds how many fix iterations the
 	// WorkloadReconciler may append per issue after a reviewer NO-GO
 	// (#946). On a NO-GO, instead of failing the Workload, the
-	// reconciler emits a new coder task ("code-<N>-r<k>", same
-	// CoderAgentRef, same branch, payload.allowOverwrite=true) whose
-	// payload.prompt carries the reviewer's structured findings and
-	// summary, then chains a fresh verify + reviewer fan-out behind
-	// it ("verify-<N>-r<k>", "review-<N>-<i>-r<k>"). Superseded
+	// reconciler emits a new coder task ("code-<N>-r<k>", using
+	// RevisionCoderAgentRef when set and CoderAgentRef otherwise, same
+	// branch, payload.reviseFromBranch naming that branch so the
+	// executor restores the prior attempt (#951), and
+	// payload.allowOverwrite=true) whose payload.prompt carries the
+	// reviewer's structured findings and summary, then chains a fresh
+	// verify + reviewer fan-out behind it ("verify-<N>-r<k>",
+	// "review-<N>-<i>-r<k>"). Superseded
 	// iterations stop counting toward the rollup; the Workload
 	// completes when the final iteration's reviewers say GO and
 	// fails (today's behavior) when the last allowed iteration is
@@ -205,6 +208,20 @@ type WorkloadSpec struct {
 	// +kubebuilder:validation:Minimum=0
 	// +optional
 	MaxReviewIterations *int32 `json:"maxReviewIterations,omitempty"`
+
+	// RevisionCoderAgentRef optionally names the same-namespace Agent
+	// the fix-iteration coder steps ("code-<N>-r<k>") reference instead
+	// of CoderAgentRef. A revision task amends its restored prior
+	// attempt (payload.reviseFromBranch, #951) rather than building a
+	// fix from scratch, and the issue-fix Agent's forcing profile is
+	// tuned for the latter — #951's live validation showed a revision
+	// task collapsing under the issue-fix profile's forcing windows.
+	// When unset, iteration coder steps fall back to CoderAgentRef and
+	// the reconciler emits a Warning event on the Workload (reason
+	// RevisionUnderIssueFixProfile). Issue-batch mode only; ignored in
+	// explicit Pipeline mode.
+	// +optional
+	RevisionCoderAgentRef *corev1.LocalObjectReference `json:"revisionCoderAgentRef,omitempty"`
 
 	// AllowCloudReviewers gates whether reviewer Agents whose
 	// spec.provider is "cloud-proxy" (or any non-"local" value) may be
--- a/api/foreman/v1alpha1/zz_generated.deepcopy.go
+++ b/api/foreman/v1alpha1/zz_generated.deepcopy.go
@@ -1002,6 +1002,11 @@ func (in *WorkloadSpec) DeepCopyInto(out *WorkloadSpec) {
 		*out = new(int32)
 		**out = **in
 	}
+	if in.RevisionCoderAgentRef != nil {
+		in, out := &in.RevisionCoderAgentRef, &out.RevisionCoderAgentRef
+		*out = new(v1.LocalObjectReference)
+		**out = **in
+	}
 	if in.AllowCloudReviewers != nil {
 		in, out := &in.AllowCloudReviewers, &out.AllowCloudReviewers
 		*out = new(bool)
--- a/api/foreman/v1alpha1/zz_generated.deepcopy.go
+++ b/api/foreman/v1alpha1/zz_generated.deepcopy.go
@@ -997,6 +997,11 @@ func (in *WorkloadSpec) DeepCopyInto(out *WorkloadSpec) {
 		*out = make([]v1.LocalObjectReference, len(*in))
 		copy(*out, *in)
 	}
+	if in.MaxReviewIterations != nil {
+		in, out := &in.MaxReviewIterations, &out.MaxReviewIterations
+		*out = new(int32)
+		**out = **in
+	}
 	if in.AllowCloudReviewers != nil {
 		in, out := &in.AllowCloudReviewers, &out.AllowCloudReviewers
 		*out = new(bool)
--- a/charts/foreman/templates/crds/workloads.yaml
+++ b/charts/foreman/templates/crds/workloads.yaml
@@ -229,6 +229,28 @@ spec:
                   format: int32
                   type: integer
                 type: array
+              maxReviewIterations:
+                description: |-
+                  MaxReviewIterations bounds how many fix iterations the
+                  WorkloadReconciler may append per issue after a reviewer NO-GO
+                  (#946). On a NO-GO, instead of failing the Workload, the
+                  reconciler emits a new coder task ("code-<N>-r<k>", same
+                  CoderAgentRef, same branch, payload.allowOverwrite=true) whose
+                  payload.prompt carries the reviewer's structured findings and
+                  summary, then chains a fresh verify + reviewer fan-out behind
+                  it ("verify-<N>-r<k>", "review-<N>-<i>-r<k>"). Superseded
+                  iterations stop counting toward the rollup; the Workload
+                  completes when the final iteration's reviewers say GO and
+                  fails (today's behavior) when the last allowed iteration is
+                  still NO-GO.
+
+                  nil (unset) defaults to 1 iteration. An explicit 0 disables
+                  iteration and restores the fail-on-first-NO-GO behavior.
+                  Issue-batch mode only; ignored in explicit Pipeline mode.
+                  Iteration tasks count against MaxTasks.
+                format: int32
+                minimum: 0
+                type: integer
               maxTasks:
                 description: |-
                   MaxTasks caps how many AgenticTasks may be emitted. Zero means no
@@ -653,6 +675,16 @@ spec:
                   PlannerModel records which frontier model the planner actually used
                   for this workload. Set after the planner runs.
                 type: string
+              reviewIterations:
+                description: |-
+                  ReviewIterations counts the fix iterations the reconciler has
+                  emitted after reviewer NO-GO verdicts (#946), summed across
+                  issues. Zero (or absent) means no reviewer ever bounced a
+                  branch back to the coder; operators watch this to distinguish
+                  convergence (small counts) from thrash (counts near
+                  issues * maxReviewIterations).
+                format: int32
+                type: integer
               succeededTasks:
                 description: |-
                   SucceededTasks counts child tasks that reached a positive terminal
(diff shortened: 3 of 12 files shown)


joryirving:
Interaction note for merge ordering: this PR and #956 (open-PR-on-review-GO) touch adjacent payload/synthesis code, and this branch intentionally does **not** stamp `openPullRequest` onto iteration review steps because #956 isn't in its base. Whichever merges second should get a one-line follow-up (stamp `openPullRequest` on the `review-<n>-<i>-r<k>` payloads, mirroring the base-round stamp) so an iterated-then-approved Workload still opens its PR. Happy to push that follow-up once both land.

Defilan:
### Review

**Request changes.** The controller state-machine work is excellent — but the feature isn't functional yet because the execution half (restoring the prior attempt) is missing, and as written the retry can destroy the base attempt's work.

**Revision coder rebuilds from `main` and force-overwrites the prior attempt (blocking):** `workload_iteration.go:217,323` sets the iteration coder's `Payload.Branch` to the same base name with `AllowOverwrite: true`, and `reviewFeedbackPrompt` (:281) tells the model "your previous attempt is already pushed on this task's branch." But the executor never restores it: each task runs in a fresh workspace, `resolveUpstream(payload.Repo)` is non-empty, so `repo.CreateBranchFromUpstream` runs → `git checkout -B <branch> FETCH_HEAD` (`pkg/foreman/agent/repo/branch.go:164`) which **resets the branch to a fresh fetch of `main`**. The prior commits live only on `origin/foreman/<w>/issue-<n>` and are never fetched. `AllowOverwrite` only feeds `push.ReplaceOnReject` (force-with-lease). So on the common "you changed too much / revert X" NO-GO (exactly the envtest fixture), the retry starts from main, sees only the feedback, produces a feedback-only diff, and force-with-lease overwrites the base attempt — the branch ends with **less** than it started. This is the precise failure mode #951 was filed to prevent (its `reviseFromBranch`: executor fetches + checks out the prior ref so the files are simply present). The envtest can't see it — it asserts task-spec fields and marks the retry GO by hand, never running the executor.

**Related:** the retry also loses the original ask — `fetchIssueBodyIfNeeded` (`executor_native.go:1791`) early-returns when `Payload.Prompt != ""`, so the iteration coder gets the review feedback but never the issue body/acceptance criteria. And it reuses the issue-fix `CoderAgentRef` (`workload_iteration.go:320`), i.e. the issue-fix forcing profile — #951's live validation (#478 r1/r2/r3) showed that profile's `editFreeTurnsLimit: 4` collapses a revision task; only a revision-tuned profile converged.

**Recommendation:** land this as the loop skeleton, but it does **not** supersede #951 — keep #951 open, rescoped to the two load-bearing pieces (executor-owned `reviseFromBranch` restore + revision profile pairing), and make #946's closure note explicit that today the retry rebuilds rather than amends. Until the restore lands I'd gate the CRITICAL or drop the "previous attempt is on this branch" claim from the prompt.

Verified clean and genuinely well done: the bound truly bounds (round `k` derived from observed child step-names, no mutable counter, superseded rounds filtered not deleted so re-emission is permanently blocked — no unbounded/off-by-one path), feedback plumbing, state-machine ordering vs escalation/rollup, placeholder anti-double-dispatch, CRD/chart/deepcopy sync.

Defilan (changes requested):
### Review: Request changes (hold for #951)

The controller state-machine is genuinely excellent and I want to land it: the bound truly bounds (round derived from observed child step-names, no mutable counter), superseded rounds are filtered not deleted so re-emission is permanently blocked, and the feedback/placeholder plumbing and ordering-vs-escalation are all clean. But I am holding it until the execution half exists, because as shipped it is on by default and the retry can destroy a passing base attempt.

**Rebuild-from-main + force-overwrite (blocking).** `maxReviewIterations` defaults to 1 (nil resolves to 1), so iteration is on for every Workload. The iteration coder targets the same base branch with `AllowOverwrite: true`, and the prompt tells the model "your previous attempt is already pushed on this task's branch." But the executor never restores it: each task runs in a fresh workspace and `repo.CreateBranchFromUpstream` does `git checkout -B <branch> FETCH_HEAD`, resetting the branch to a fresh fetch of `main`. The r0 commits live only on `origin/foreman/<w>/issue-<n>` and are never fetched; `AllowOverwrite` only feeds the force-with-lease push. So on the common "you changed too much / revert X" NO-GO (exactly the envtest fixture), the retry starts from main, produces a feedback-only diff, and force-overwrites r0 with less than it started with. The envtest hand-sets task status and never runs the executor, so it cannot observe this.

Related: the retry also loses the original ask (`fetchIssueBodyIfNeeded` early-returns on a non-empty `Prompt`, so the iteration coder gets the review feedback but not the issue body), and it reuses the issue-fix `CoderAgentRef`, i.e. the issue-fix forcing profile whose `editFreeTurnsLimit: 4` collapsed the revision task in the #478 r1/r2 validation.

This is precisely the failure mode #951 (`reviseFromBranch`: the executor fetches and checks out the prior ref so the files are simply present) is filed to prevent. Plan: keep #951 open and land it first, then this skeleton becomes functional and safe, and it can merge. To unblock:
- Land the executor branch-restore for iteration tasks (fetch + checkout the prior branch instead of resetting to main).
- Pair iteration with a revision-tuned profile rather than the issue-fix forcing profile.

If you would rather land the skeleton sooner, the minimum safe change is to flip the default to opt-in (0) so default Workloads carry no regression, but I would prefer to pair it with the restore so the feature actually works when enabled. Great work on the loop itself; this is purely about not enabling it before the restore exists.

joryirving:
Reworked per your "land #951 first" plan, split in two:

- **#967** (extracted from #951, this branch is now rebased on top of it): the executor-owned restore — `payload.reviseFromBranch` makes `setupTaskBranch` fetch + checkout the prior ref from the push remote instead of resetting to main, with fallback to the base-branch path when the ref is gone. It also fixes the lost original ask: `fetchIssueBodyIfNeeded` appends the issue body behind a revision task's feedback prompt instead of early-returning on non-empty `Prompt`.
- **This PR**: iteration coder payloads now stamp `reviseFromBranch` = the task's own branch, so the retry starts from the r0 files and amends them rather than producing a feedback-only diff from main. The prompt was reworded to match the now-true semantics (workspace contains the prior attempt; apply findings as a delta, don't rebuild).

Profile pairing: new `spec.revisionCoderAgentRef` selects a revision-tuned Agent for iteration coder steps. When unset it falls back to `coderAgentRef` and emits a Warning event (`RevisionUnderIssueFixProfile`) on the Workload, since a revision under the issue-fix forcing profile is the #478 r1/r2 failure mode. I didn't make it required — most deployments won't have a second profile yet — but the warning makes the mismatch visible.

Default `maxReviewIterations` stays 1: with the restore, an on-by-default iteration amends the prior attempt instead of destroying it, which I read as your preferred option over flipping to opt-in.

Envtests now assert `reviseFromBranch` on the retry payload, the ref selection, and the warning; the executor restore itself is covered in #967's tests.

Defilan:
Update: #967 (executor-owned revise-from-branch restore) is merged, which resolves the blocking concern here. The force-overwrite problem existed because the executor reset the workspace to a fresh `main` and the retry rebuilt a feedback-only diff over the prior attempt. That lever now exists.

To unblock this PR, rebase on main and stamp the new field on the iteration coder task: `payload.ReviseFromBranch = <the task's push-remote branch>` (the same branch the prior attempt pushed to). With that set, the executor fetches and checks out the prior tip, the coder edits on top of real prior work, and the subsequent force-with-lease is both safe (lease matches the remote tip) and correct (prior + new edits).

Two smaller items from my original review still stand, neither blocking the mechanism:
- Pair the revision with a revision-tuned profile rather than the issue-fix forcing profile (the `editFreeTurnsLimit: 4` profile collapsed revision tasks in the #478 validation). #951 flags this as the config-pairing half.
- With the data-loss fixed, on-by-default (`maxReviewIterations` nil -> 1) is far less risky than before. I'd still like to see one live revision cycle converge end-to-end before defaulting it on, but I no longer consider the default itself a blocker.

Happy to re-review once ReviseFromBranch is wired in.

joryirving:
Rebased on main — #967's commit dropped out, so the diff is now just the loop + pairing.

- `ReviseFromBranch` was already stamped on the iteration coder payload from the rework (`workload_iteration.go`, alongside `AllowOverwrite`); it now compiles against the merged field and the envtests assert it end-to-end.
- Since #956 landed first, this PR also picks up the interaction fix I flagged earlier: iteration review payloads (`review-<n>-<i>-r<k>`) now carry the same `openPullRequest` stamp as the base round, with an envtest assertion — an iterated-then-approved issue still opens its PR.
- Profile pairing: `spec.revisionCoderAgentRef` selects the revision-tuned Agent; unset falls back to `coderAgentRef` with a `RevisionUnderIssueFixProfile` warning event.

On the live-cycle validation: our cluster produces real NO-GO traffic daily, and we'll have a revision-tuned Agent CR deployed alongside the next release. Happy to run one end-to-end revision cycle there and report results on this PR before you merge, if you'd like it as a gate — or immediately after release if not.

Lint 0 issues; `go test ./pkg/foreman/...` and the controller envtest suite pass on the rebase.

Defilan (approved):
### Review

Approve. The blocking concern from my earlier review is resolved, and the two secondary points are addressed.

Verified against the updated branch:
- **Restore, not rebuild-from-main:** the fix-iteration coder step now stamps `payload.ReviseFromBranch = <the task's push-remote branch>` (`workload_iteration.go`), so #967's executor fetches and checks out the prior attempt's tip before the model runs. The coder edits on top of real prior work and the subsequent force-with-lease is both safe (lease matches the remote tip) and correct (prior + new edits, not a feedback-only diff). This is exactly the mechanism #951 called for; the force-overwrite failure mode is gone.
- **Revision profile pairing:** new `spec.revisionCoderAgentRef` is used for iterations when set (revision-tuned profile) and falls back to `coderAgentRef` otherwise, with a warning event when unset that steers the operator to set it (avoids the issue-fix `editFreeTurnsLimit` collapse #951 flagged).
- **PR on success:** iteration review steps now stamp `openPullRequest` (coherent with #956).
- CRDs in sync (config + chart), deepcopy correctly regenerated for both `MaxReviewIterations` and `RevisionCoderAgentRef` pointer fields, tests cover the `ReviseFromBranch == Branch` stamping and the coder-ref selection. Merges cleanly onto current main (the `ReviseFromBranch` field git-dedupes against #967, single field, compiles).

One follow-up, not blocking: `maxReviewIterations` defaults to 1, so iteration is on by default. With the restore in place that carries no data-loss risk, but I'd still like to see one live revision cycle converge end-to-end before we lean on it. Worth a note in the release.

Nice work turning this around.

PR #960, opened 2026-07-03
feat(inference): add spec.modelCache.claimName for user-owned cache PVCs

### LINKED ISSUE #928: [FEATURE] Per-InferenceService model cache PVC override (spec.modelCache.claimName)
## Summary

Add an optional per-`InferenceService` field that points the built-in model
cache (prep + download init containers) at a **user-managed PVC**, instead of
the operator's shared/perService cache PVC:

```yaml
kind: InferenceService
spec:
  modelCache:
    claimName: my-model-cache   # pre-existing, user-owned PVC
  ...
```

When set, the operator uses `claimName` as the writable cache volume for that
workload and downloads into it exactly as it does today for the shared cache.
When unset, behavior is unchanged (the global `shared`/`perService` mode
applies). This lets a cluster **mix**: most InferenceServices ride the shared
cache, while specific ones use their own PVC.

> Naming is the maintainer's call — `spec.modelCache.claimName` keeps the
> existing "model cache" vocabulary and leaves room to grow; flat
> `spec.modelCacheClaimName` or a more generic `spec.persistence.claimName`
> were also considered.

## Motivation

The model cache backend is currently an **operator-global** choice
(`--model-cache-mode` / chart `modelCache`): either one shared PVC (`shared`)
or an operator-provisioned per-workload PVC (`perService`), both using a single
chart-level `storageClass`. There is no way to say "this *one* model should
cache on a *different* volume."

Concrete case: on a multi-node cluster the shared cache is an RWX class
(e.g. a networked filesystem). For most models that's fine, but for a large
model on a specific node it's desirable to cache on **node-local** storage —
the networked filesystem's cold-load (first prefill reading the weights) is
markedly slower, and the model is pinned to that node anyway. Today the only
ways to get node-local weights are:

- Switch the **whole operator** to `perService` — but that moves *every* model
  off the shared cache and still uses one storage class, so you can't target
  just the one workload or give it a different (local) class.
- Pre-stage onto a PVC and use `source: pvc://…` — but that is **read-only with
  no download**, so you have to build and maintain your own staging Job, which
  is exactly the toil the built-in downloader was meant to remove.

A per-InferenceService `claimName` override closes that gap: you bring a PVC
backed by whatever storage class you want, and the operator's existing
prep+download machinery fills and serves it — no staging Job, no giving up the
managed download path.

## Proposed behavior

When `spec.modelCache.claimName` is set on an InferenceService:

- The operator uses that PVC as the `model-cache` volume in
  `buildCachedStorageConfig` — same code path as the shared/perService cache:
  `model-cache-prep` (chown) + `model-downloader` run against it, and the
  server is started with `--model` pointing into it. No new download logic.
- Weights land under the existing **`<cacheKey>/` subdirectory** (not the PVC
  root), matching current layout so `RefreshPolicy` and cache-key semantics are
  unchanged, and pointing two models at one PVC can't collide.
- The main container mounts the PVC **read-only** (as today); the init
  containers mount it read-write.

When unset: unchanged — the global `shared`/`perService` mode applies.

## Lifecycle

The operator **never creates or deletes** a `claimName` PVC — the user owns it
end-to-end (contrast with `perService`, where the operator provisions and may
GC `<isvc>-model-cache`). The operator only mounts and populates it.

## Edge cases / guardrails

- **Missing PVC:** surface a clear `Degraded` condition / event rather than
  silently falling back to the shared cache.
- **`source: pvc://…`** (already-staged, read-only, no download) +
  `claimName`: mutually exclusive — `claimName` targets the *download* path, so
  it's meaningless for a pre-staged source. Reject at validation or document as
  ignored.
- **Node alignment is the user's responsibility:** for a node-local RWO PVC,
  the InferenceService's `nodeSelector` must land the pod where the PVC binds
  (a `WaitForFirstConsumer` local class binds on first consumer = the pod's
  node; a pre-bound RWO PVC pins the pod). The operator does not enforce this.
- **fsGroup/chown:** no new handling needed — the same root `model-cache-prep`
  runs, so `fsGroupPolicy: None` local storage classes are covered exactly like
  the shared path.

## Out of scope (possible follow-ups)

- `llmkube cache list` / `--purge-cache` would not inspect BYO cache PVCs (they
  read the shared cache). A natural later enhancement; relates to the CLI
  cache-key handling noted on #912.
- Operator-provisioned per-model cache (a per-InferenceService
  `storageClassName` + `size` instead of a pre-made PVC) — a reasonable sibling
  feature, but intentionally left out here to keep this focused on the
  bring-your-own-PVC case.

joryirving:
## What

Adds optional `InferenceService.spec.modelCache.claimName`: when set, the named pre-existing, user-owned PVC is mounted as the writable model cache for that workload instead of the operator's shared/perService cache PVC.

## Why

The cache backend is currently operator-global, so there is no way to give one workload (e.g. a large model pinned to a node) node-local cache storage while everything else rides the shared cache.

Fixes #928

## How

- `modelCachePVCName` returns the user claim when set, so all cache volume references (single-file, multi-file, invalid-fileset branches) and the same prep + downloader init containers use it; weights still land under `<cacheKey>/` and the serving container mounts read-only.
- `ensureModelCachePVC` never creates/adopts/GCs the user claim — it only verifies existence. A missing claim yields a `Degraded` condition plus a `ModelCachePVCNotFound` warning event (no silent fallback to the shared cache).
- `claimName` + pre-staged `pvc://` source: claimName is ignored (no download path) and a `ModelCacheClaimIgnored` warning event is emitted.
- Unset claimName: behavior unchanged (global shared/perService mode).
- Tests: builder units cover claim override in both modes, layout/init-container parity, and unchanged default; envtest covers no operator PVC created, no adoption/mutation of the user PVC, and clear error when missing. Verified with `go test ./internal/controller/... ./api/... -count=1` and `golangci-lint run` (0 issues).

## Checklist

- [x] Tests added/updated


Defilan:
### Review

**One fix, then good to go.** The riskiest concern is handled correctly: a user-owned `claimName` PVC gets **only a `Get` and `return nil`** in `ensureModelCachePVC` — never `Create`, never a controller owner-ref, never operator labels (`internal/controller/model_storage.go:669-683`), so it can't be garbage-collected with the InferenceService. Verified and tested. Precedence (claimName wins over `mode` before the shared/perService branch, `model_storage.go:91-100`), RBAC (existing `persistentvolumeclaims` get is sufficient), and CRD/deepcopy/chart sync all check out.

**Silent drop when caching is disabled or the model has no cache key (should fix):** `useCache := effectiveModelCacheKey(model) != "" && r.ModelCachePath != ""` (`internal/controller/deployment_builder.go:249`) gates the whole cache path. When it's false — chart `modelCache.enabled: false`, a `file://`/local source, or a remote single-file model before `Status.CacheKey` is populated — `buildModelStorageConfig` routes to `buildEmptyDirStorageConfig` (`model_storage.go:369`) which never consults `claimName`. So a user who sets `spec.modelCache.claimName` gets an emptyDir and re-downloads every restart, with **no event, no condition, no signal**. The PR deliberately added a `ModelCacheClaimIgnored` warning for the `pvc://`-source case (`inferenceservice_controller.go:178`) to avoid exactly this silent drop, but missed the symmetric one. Recommend emitting the same warning (or failing closed) when `claimName != ""` but caching is off / no cache key, so the behavior matches the PR's own "no silent fallback" promise — plus a test (neither new event is currently asserted).

Nice-to-haves: the missing-claim degraded condition uses the generic `Reason: PhaseFailed` rather than `ModelCachePVCNotFound` (the specific reason is only in the message/event); `claimName` is mutable on a running service with no immutability marker (no data loss since the old claim isn't GC'd, but worth a docs note). `docs/MODEL-CACHE.md` is otherwise accurate.

Defilan (approved):
### Review

Approve. The riskiest part is handled correctly and tested: a user-owned `claimName` PVC gets only a `Get` and `return nil` in `ensureModelCachePVC` (never `Create`, no owner-ref, no operator labels), so it survives InferenceService deletion instead of being garbage-collected. Precedence (claimName wins before the shared/perService branch), RBAC (existing `persistentvolumeclaims` get suffices), and CRD/deepcopy/chart sync all check out.

Merging now. One non-blocking follow-up, filed as #965: the symmetric silent-drop when caching is off (`claimName` set but `useCache` false routes to `buildEmptyDirStorageConfig`, which ignores `claimName`) wants the same `ModelCacheClaimIgnored` warning you already added for the `pvc://` case. Purely observability, no data-loss risk.

Thanks for this, the safe-by-default handling of user PVCs is exactly right.

PR #962, opened 2026-07-03
fix(foreman): scope-overlap check catches Go files in new directories

### LINKED ISSUE #907: [BUG] Coder-gate scope-overlap check is blind pre-commit (diffs committed history, but the gate runs before commit)
## Bug Description

The coder gate's **scope-overlap** check (#782) reads the coder's changed files via `repo.DiffNameOnly`, which diffs **committed history** (`git diff --name-only <base>...HEAD`, `pkg/foreman/agent/repo/diff.go:51`). But the gate runs as the `VerifyTerminal` at the coder's `submit_result` (`makeCoderGateVerifier` -> `RunCoderGate`, `executor_native.go`), i.e. **before any commit or stage**. At gate time the coder's edits are uncommitted working-tree changes and any new file is untracked, so `<base>...HEAD` returns **none of them** and the scope-overlap check sees an empty changed-file set and silently no-ops.

This was found while adding a new gate check (the reference-grounding check in #906), which hit the same trap and was fixed there by diffing the staged working tree (`git add -A` + `git diff --cached ... HEAD`) instead of `...HEAD`. `DiffNameOnly` and its scope-overlap caller were not changed and remain affected.

The static checks (gofmt/vet/build/lint) and the `git status`-based checks (unit-test tier, mutation gate) are NOT affected: they read the working tree directly. Only the `...HEAD`-based checks are blind pre-commit.

## Steps to Reproduce

1. Run a coder AgenticTask whose changes drift to an unrelated subsystem (the case scope-overlap is meant to catch).
2. The coder reaches a GO `submit_result`; the gate runs before the executor commits.
3. `DiffNameOnly(workspace, "main")` returns an empty list (changes are uncommitted/untracked), so `checkScopeOverlap` finds no changed Go files and reports no drift.

## Expected Behavior

The scope-overlap check inspects the coder's actual (uncommitted) working-tree changes and flags drift to an unrelated subsystem.

## Actual Behavior

The check operates on committed history (empty pre-commit), so it no-ops and never flags drift. It has likely been inert in production since #782; it stays quiet by design, so the no-op was not noticed.

## Environment

Not deployment-specific: this is coder-gate logic in `pkg/foreman/agent`. Affects any Foreman coder run (metal or AMD). LLMKube main as of #906.

## Additional Context

Fix pattern (mirror #906): stage then diff the working tree, e.g. `git add -A` followed by `git diff --name-only --cached <base>` (or `git status`-based enumeration), with `base = HEAD` (the branch point at gate time, since the coder has not committed). Note the `git diff --cached` mnemonic `c/`/`i/` path-prefix gotcha if a unified-diff parser is involved (force `--src-prefix=a/ --dst-prefix=b/`).

Related: #782 (scope-overlap), #906 (reference-grounding, where the same trap was fixed for the new check).

Defilan:
## What

Fixes the coder gate's scope-overlap check so it sees a Go file created in a **brand-new untracked directory**.

## Why

Fixes #907

The scope-overlap check enumerated the coder's changed Go files via `git status --porcelain`. Porcelain lists untracked files in already-tracked directories fine, but it **collapses a brand-new untracked directory to a single `newdir/` entry** — which fails the `.go` suffix filter, so an out-of-scope Go file created in a new directory slipped past the check.

(Note: the original issue described this as the check "diffing committed history" / being "blind pre-commit." On inspection the pre-change code was `git status --porcelain`, not a `base...HEAD` diff, so it did see most uncommitted changes — the real, narrower gap is the new-directory collapse above. The fix is the same either way.)

## How

Replace the porcelain scan with `git add -A` followed by `git diff --name-only --cached HEAD`, which lists each new file individually (including files in new directories). This mirrors the established working-tree-diff approach used by the reference-grounding check (#906).

Safety of `git add -A` inside the gate: it runs pre-commit, but the executor's commit path (`repo.Commit`) runs its own `git add -A` immediately before `git commit -s`, so the committed file set is identical regardless of what the gate staged — no junk inclusion, no interference. This is the same documented-safe pattern as #906.

The regression test models the actual fixed case (a Go file in a new directory) and the diff drops two flags (`--src-prefix`/`--dst-prefix`) that are inert with `--name-only`.

Authored by **Foreman**, LLMKube's agentic coder harness (local model on the lab fleet); the maintainer adversarially reviewed the git-side-effect safety, corrected the diagnosis/test, and takes responsibility per the AI-assisted contribution policy.

## Checklist

- [x] Tests added/updated
- [x] `make lint` passes locally (+ `GOOS=linux golangci-lint`, 0 issues)
- [x] Documentation updated (if user-facing change) - internal gate check, no user-facing surface

### DIFF
--- a/pkg/foreman/agent/caller_impact_gate_test.go
+++ b/pkg/foreman/agent/caller_impact_gate_test.go
@@ -25,7 +25,7 @@ import (
 // callerImpactRunner builds a fake commandRunner keyed on argv patterns for
 // checkCallerImpact tests. It answers three call types:
 //  1. git status -z  -> returns the changedFiles NUL-separated list
-//  2. git diff -U0 -- <file>  -> returns diffOutput for that file
+//  2. git diff -U0 HEAD -- <file>  -> returns diffOutput for that file
 //  3. grep -rn --include=*.go <name> .  -> returns grepOutput for that name
 func callerImpactRunner(changedFiles, diffOutput string, grepOutputs map[string]string) commandRunner {
 	return func(_ context.Context, _ string, _ []string, name string, args ...string) (string, error) {
@@ -34,8 +34,8 @@ func callerImpactRunner(changedFiles, diffOutput string, grepOutputs map[string]
 			if len(args) >= 2 && args[0] == "status" && args[1] == "-z" {
 				return changedFiles, nil
 			}
-			if len(args) >= 3 && args[0] == "diff" && args[1] == "-U0" {
-				// args[2] is "--", args[3] is the file
+			if len(args) >= 4 && args[0] == "diff" && args[1] == "-U0" && args[2] == "HEAD" {
+				// args[3] is "--", args[4] is the file
 				return diffOutput, nil
 			}
 		case "grep":
@@ -178,19 +178,21 @@ func TestCheckCallerImpact_AddedFuncWithExternalCaller(t *testing.T) {
 			"./pkg/y/consumer.go:42:\tNewThing()\n",
 	}
 
-	// addedFuncNames uses "git diff -- <file>" (without -U0); the fake runner
-	// above answers args[0]=="diff" && args[1]=="-U0" for modifiedFuncNames.
-	// We need a separate branch for the plain "git diff -- <file>" call used by
-	// addedFuncNames.  Extend the runner to handle both.
+	// addedFuncNames uses "git diff HEAD -- <file>" (without -U0); the fake runner
+	// above answers args[0]=="diff" && args[1]=="-U0" && args[2]=="HEAD" for
+	// modifiedFuncNames. Both diff forms now carry HEAD (#907/#962) so the coder's
+	// changes survive an earlier gate's `git add -A` staging; this branch matches
+	// either shape and returns the same diff (addedFuncNames only reads "+func").
 	run := func(_ context.Context, _ string, _ []string, name string, args ...string) (string, error) {
 		switch name {
 		case "git":
 			if len(args) >= 2 && args[0] == "status" && args[1] == "-z" {
 				return changedFiles, nil
 			}
-			if len(args) >= 2 && args[0] == "diff" {
-				// Both -U0 (modifiedFuncNames) and plain (addedFuncNames) paths
-				// return the same diff; addedFuncNames only cares about "+func" lines.
+			if len(args) >= 3 && args[0] == "diff" &&
+				(args[1] == "HEAD" || (args[1] == "-U0" && args[2] == "HEAD")) {
+				// Both -U0 HEAD (modifiedFuncNames) and plain HEAD (addedFuncNames)
+				// paths return the same diff; addedFuncNames only cares about "+func".
 				return diffOutput, nil
 			}
 		case "grep":
--- a/pkg/foreman/agent/coder_gate.go
+++ b/pkg/foreman/agent/coder_gate.go
@@ -676,16 +676,13 @@ func checkReferenceGrounding(ctx context.Context, workspace string, run commandR
 // uncommitted edits rather than committed history (the gate runs before any
 // commit, so `...HEAD` would see nothing). A git error yields nil (the caller
 // treats that as "no changed files" and skips the check).
-//
-// `--src-prefix=a/ --dst-prefix=b/` is forced so the output is a clean path
-// list, not unified-diff path-prefix output (the `c/`/`i/` gotcha).
 func changedWorkingTreeGoFiles(ctx context.Context, workspace string, run commandRunner) ([]string, error) {
 	// Stage everything so untracked new files appear in the diff.
 	if _, err := run(ctx, workspace, nil, "git", "add", "-A"); err != nil {
 		return nil, err
 	}
 	out, err := run(ctx, workspace, nil, "git", "diff", "--name-only",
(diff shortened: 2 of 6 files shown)


joryirving (changes requested):
Request changes — the fix is right for #907, but `git add -A` at this point in the gate silently disables the two mutation-gate checks that run after it.

**Ordering regression (blocking).** `RunCoderGate` runs, in sequence: step 8 `checkScopeOverlap` (now `git add -A` + `git diff --cached`), then step 9 `checkTestPresence`, step 10 `checkMutationSurvival`, step 11 `checkReferenceGrounding`. The mutation gates read the coder's changes via **unstaged** diff — `addedFuncNames` does `git diff -- <file>` (`mutation_gate.go:118`) and `changedNewLines` does `git diff -U0 -- <file>` (`:347`), neither with `--cached`. After step 8 stages the whole tree, working tree == index, so those unstaged diffs return empty:

- `checkTestPresence` sees zero net-new funcs → can't flag missing tests (the #856 class it exists to catch).
- `checkMutationSurvival` sees zero changed funcs → nothing to neuter → auto-passes.

Both silently green on the GO path. This is new: pre-change `checkScopeOverlap` used read-only `git status --porcelain` and never touched the index, so steps 9–10 saw a genuine unstaged tree. `grounding.AddedLines` (#906, the cited pattern) also does `git add -A`, but it runs at step 11 — *after* the mutation gates — so it never disturbed them; this PR moves the same side effect ahead of them.

The per-check unit tests can't catch it: each mocks `run` in isolation, so none exercises step 8's staging leaking into steps 9–10. A full-`RunCoderGate` test with a real git workspace (add a new func in a new dir, no test) would fail today with the change and is the regression guard worth adding.

**Fix options:**
- Make the mutation-gate diffs order-independent: `git diff HEAD -- <file>` (covers staged + unstaged vs HEAD) in `addedFuncNames`/`changedNewLines`. Cleanest — the checks want "changes vs HEAD" regardless of index state.
- Or stage once at the top of `RunCoderGate` and switch every changed-file check to `--cached` consistently.
- Or narrowly: have `changedWorkingTreeGoFiles` restore the index after (`git reset` / diff without persistent staging) so the gate leaves the tree as it found it — but the executor commits with its own `git add -A` regardless, so making the diffs read `HEAD` is simpler and removes the ordering coupling entirely.

Everything else checks out: the new-directory gap is real and `git add -A` + `git diff --cached HEAD` closes it, the fail-open on git error matches the surrounding checks, the `--src/--dst-prefix` drop is correct under `--name-only`, and the diagnosis correction in the PR body is accurate. Just needs the downstream checks to stop depending on an unstaged tree.

Defilan:
Great catch, thank you. You were exactly right: `checkScopeOverlap`'s `git add -A` at step 8 staged the tree, and the downstream mutation gates read the coder's changes via unstaged `git diff`, so once staged they saw an empty diff and both silently passed.

Fixed with your recommended option 1: the three helpers in `mutation_gate.go` now diff vs `HEAD` (index-independent), so an earlier gate staging the tree can't blind them:
- `addedFuncNames` (:118): `git diff -- <f>` -> `git diff HEAD -- <f>`
- `modifiedFuncNames` (:200) and `changedNewLines` (:347): `-U0 --` -> `-U0 HEAD --`

Since HEAD is the pre-work base (the coder's changes are uncommitted), `git diff HEAD` is exactly "the coder's changes" whether or not they're staged. I also confirmed `checkCallerImpact` reuses `addedFuncNames`/`modifiedFuncNames`, so it was similarly blinded and is fixed by the same change; `changedWorkingTreeGoFiles` already uses `--cached HEAD` (that is the #962 staging), and the `git status`-based helpers report staged+unstaged so were never affected. No other blinded reads.

Added the regression guard you asked for: a real-git test (`materializeFixture` harness, real `git` not a mock) that commits a base, adds an untested exported func in a NEW dir, runs `checkScopeOverlap` (which stages), asserts the unstaged diff is now empty (the blinding precondition), then asserts `checkTestPresence` STILL fires and names the func. Fail-then-pass verified: reverting the HEAD change makes it fail with the #907 blinding, restoring it passes.

Green: build, full `pkg/foreman/agent` tests, lint both arches. Ready for another look.

joryirving (approved):
Approve — verified against the code, not the description.

All three mutation-gate helpers now diff against HEAD (`addedFuncNames`, `modifiedFuncNames`, `changedNewLines` → `git diff HEAD` / `git diff -U0 HEAD`), so `checkScopeOverlap`'s step-8 `git add -A` can no longer blind them: HEAD is the pre-work base, and `git diff HEAD` reflects the coder's changes whether or not they're staged. Good catch extending it to `checkCallerImpact` (reuses the two helpers) — that was blinded by the same mechanism.

The regression guard is exactly right: `TestGateFixtures_StagingDoesNotBlindDownstream` uses a real git repo (not a mock), stages a new exported func in a NEW directory via the actual `checkScopeOverlap`, asserts the blinding precondition (plain `git diff` empties after `git add -A`), then asserts `checkTestPresence` still fires and names the func. That reproduces the precise #907 sequence and, with your revert-to-confirm, proves fail-then-pass. It also confirms the subtle point that a staged new file remains visible to `git diff HEAD` — which is what makes the fix hold for the new-directory case.

No new blindness: post-staging, `git diff HEAD` is strictly more visible than the old unstaged `git diff`, and the `git status`-based helpers (staged+unstaged) were never affected. Resolves my earlier changes-requested.

Issue #937, opened 2026-07-02
Workload completion should open a pull request — a GO'd, reviewed branch currently just sits on the remote


joryirving:
## Problem

The pipeline ends one hop short of its artifact. On a fully successful Workload (coder GO → gate GATE-PASS → reviewer GO → phase `Completed`), the result is a pushed `foreman/<workload>/issue-<n>` branch — and nothing else. No PR is opened, nothing links the branch to the issue, and no human is notified. Grepping the agent, operator, and docs: there is no PR-creation code anywhere.

Observed in production: first fully-green Workload completed (`code GO → verify GATE-PASS → review GO`), operator watched the dashboard expecting a PR, and the reviewed branch sat unopened on the remote until a human found it.

## Proposed fix

On Workload completion (all tasks terminal, review verdict GO), open a PR:

- **Where:** the agent already holds the GitHub token it pushed with — one `POST /repos/{repo}/pulls` after the review task GOes (or an operator-side finalizer step, which centralizes it and survives agent restarts).
- **Content:** head = the workload branch, base = repo default branch, title from the coder's commit subject, body `Fixes #<issue>` + a summary of gate/review verdicts (the review transcript ref is already in the task result).
- **Config:** `openPullRequest: true|false` on the Workload spec (default true when `issues` is set), so callers that want branch-only behavior keep it.
- **Idempotency:** skip if a PR with the same head already exists (re-runs after LLMKube#934 force-push should update, not duplicate).

A review verdict of request-changes should arguably still open a draft PR carrying the review feedback — that's where a human (or a follow-up fix task) picks it up.

Related: #934 (branch push idempotency across re-runs).

PR #964, opened 2026-07-03
feat(foreman): coder-tier escalation, re-dispatch a failed coder to a larger model

### LINKED ISSUE #963: [FEATURE] Foreman: coder-tier escalation — re-dispatch a failed coder task to a larger model
## Feature Description

A coder-side escalation tier: when a coder task fails at its model's ceiling, the WorkloadReconciler re-dispatches that issue once to a larger/denser coder model, carrying the failed model's diagnosis as a hint. This is the symmetric twin of the escalation-reviewer tier (#546) — the same verdict-triggered second-pass fan-out, applied to coding instead of judging.

## Problem Statement

A single coder model is a fixed capability ceiling. A live four-issue batch (Ornith-35B-A3B) made the shape concrete: it GO'd the tractable bug but three failed at that model's ceiling in distinct ways. The clearest was a task where the model produced a *correct diagnosis it could not execute* and honestly NO-GO'd — exactly the case a larger model should take over.

Density is not a blanket upgrade, though: it helps reasoning-limited failures but not scope/harness failures (a slower dense model times out on a big refactor). So the goal is a **cascade**: default to the cheap, fast model for the tractable majority, and pay for the larger model only on the specific failure modes where its extra reasoning is worth the cost.

> As a coder-fleet operator, I want a failed coder task to automatically retry on a larger model, so the tractable work stays cheap and only the hard tail pays the dense-model cost.

## Proposed Solution

- `WorkloadSpec.EscalationCoderAgentRef` (singular; coders are sequential) opts a Workload in. Unset = off, fully backward-compatible. Issue-batch mode only.
- **Trigger = capability failures only:** escalate iff the base `code-<N>` task is terminal with verdict `NO-GO` (model-decided) OR outcome `CODER-GATE-FAILED`. Do NOT escalate on a model-decided `INCOMPLETE` (gave up / ran out of turns), `STUCK-LOOP-DETECTED`, or `ERROR` — those are scope/harness failures escalation cannot fix.
- **Context = fresh + prior diagnosis:** the escalated coder starts on a fresh branch (`foreman/<w>/issue-<N>-esc`), not the failed attempt's branch, and receives the prior model's summary as a prompt hint (via a new `payload.promptPrefix` that leaves the issue body fetch intact). This keeps the insight, avoids anchoring the strong model on the weak model's broken code, and ships independently of the `reviseFromBranch` machinery #951/#959 still lack.
- **Single tier, idempotent**, MaxTasks- and sovereignty-gated, emitted in a second-pass hook mirroring `emitEscalations`, run before reviewer escalation (a not-GO'd coder has no branch to review).

## Alternatives Considered

- Executor-level in-process retry with a bigger model: rejected — the executor is per-task and does not own model selection; fan-out belongs to the controller.
- A unified escalation-ladder abstraction (coder + reviewer escalation + #959 NO-GO iteration): ~3x the scope; #959 is in flight. Noted as the eventual convergence, out of scope for v1.

## Additional Context

Design + implementation plan drafted. Live validation: dispatch the NO-GO'd task with `escalationCoderAgentRef` pointing at a Qwopus coder Agent and confirm the auto-escalation. Recommended deployment is dual-box (both models hot) so escalation is a routing hop, not a cold model load.

Defilan:
## What

Adds an opt-in coder escalation tier to Foreman's Workload controller. When a base coder task fails at its model's ceiling, the WorkloadReconciler re-dispatches that issue once to a larger coder model, carrying the failed model's own diagnosis as a prompt hint. Off by default; enabled per-Workload via a new `escalationCoderAgentRef`. This is the coder-side twin of the existing escalation-reviewer tier (#546).

## Why

Fixes #963

A single coder model is a fixed capability ceiling. In a four-issue batch on one MoE coder, the model cleanly resolved the tractable bug but failed three others at its ceiling in distinct ways. The clearest case: the model produced a correct diagnosis it could not itself execute and honestly returned NO-GO. That is exactly the case a larger, denser model should take over.

Density is not a blanket upgrade, though. It helps the reasoning-limited failures but hurts convergence-heavy refactors, where a slower dense model exhausts its turn budget sooner. So this is a cascade, not a swap: keep the cheap, fast model for the tractable majority and pay for the larger model only on the specific failure modes where its extra reasoning is worth the cost.

## How

- **New `WorkloadSpec.EscalationCoderAgentRef`** (singular, unlike the plural `escalationReviewerAgentRefs`): coders are sequential, so N parallel escalation coders would produce N competing branches. One tier. Unset means the feature is off, fully backward compatible.
- **Trigger is capability failures only.** Escalate iff the terminal base coder is verdict `NO-GO` with top-level outcome `MODEL-DECIDED`, or nested `modelExtra.outcome == CODER-GATE-FAILED`. Do not escalate on a model-decided `INCOMPLETE` (the model gave up / ran out of turns), `STUCK-LOOP-DETECTED`, a trivial no-changes NO-GO, or `ERROR`: those are scope/harness failures a larger, slower model will not fix and may worsen by timing out.
  - The discriminator is the nested `modelExtra.outcome`, not the top-level `extra.outcome`, which is `MODEL-DECIDED` for the NO-GO, gate-fail, and gave-up cases alike. A naive top-level read would wrongly escalate the gave-up case. The unit truth table pins this against three real batch outcomes: a NO-GO / `MODEL-DECIDED` (escalate), an `INCOMPLETE` with nested `CODER-GATE-FAILED` (escalate), and an `INCOMPLETE` with no gate outcome (do not escalate).
- **Context is fresh branch plus diagnosis.** On a triggering failure the reconciler emits a `code-<N>-esc` + `verify-<N>-esc` pair at the escalation Agent on a fresh branch `foreman/<w>/issue-<N>-esc`, carrying the prior model's summary. A fresh branch (not a restore of the failed attempt) keeps the high-signal insight without anchoring the strong model on the weak model's broken code, and makes the feature independent of the branch-restore machinery that is still in flight elsewhere.
- **The hint reaches the coder via a new `AgenticTaskPayload.PromptPrefix`**, rendered before the fetched issue body, so the escalated coder sees both the diagnosis and the issue's acceptance criteria (leaving `Prompt` empty so the issue-body fetch still runs).
- **Second-pass emission hook `emitCoderEscalations`**, a structural twin of `emitEscalations`, wired into Reconcile before reviewer escalation (a coder that did not GO has no branch to review). One tier deep (`-esc` tasks are never re-scanned as base tasks), idempotent (permanent step names, skip when the `-esc` child already exists), MaxTasks- and sovereignty-gated.
- **The controller does not manage serving.** The escalation Agent must point at an already-reachable model, exactly as escalation reviewers do. The recommended deployment is dual-box with both models hot so escalation is a routing hop, not a cold model load. Documented in the Foreman runbook.

**AI-assisted contribution (band 3).** Implemented with AI assistance under human review; human-accountable. Commits are kept clean per the project's no-attribution-in-commits convention, with the disclosure here in the PR body per CONTRIBUTING.md.

## Checklist

- [x] Tests added/updated (unit trigger truth table + step synthesis; envtest for emission, non-escalation of gave-up/stuck, backward-compat unset, idempotency)
- [x] `make test` passes locally (new envtest green; controller coverage 80.2%)
- [x] `make lint` passes locally (0 issues; also `GOOS=linux golangci-lint run ./...` 0 issues)
- [x] AI assistance disclosed above, per CONTRIBUTING.md
- [x] Documentation updated (`docs/site/foreman/README.md`: coder escalation tier + dual-box serving)

### DIFF
--- a/api/foreman/v1alpha1/agentictask_types.go
+++ b/api/foreman/v1alpha1/agentictask_types.go
@@ -300,6 +300,15 @@ type AgenticTaskPayload struct {
 	// +optional
 	Prompt string `json:"prompt,omitempty"`
 
+	// PromptPrefix is prepended to the assembled user prompt for an
+	// issue-fix task, BEFORE the fetched issue body. Unlike Prompt
+	// (which, when set, suppresses the issue-body fetch), PromptPrefix
+	// leaves the issue-body fetch intact, so the agent sees both this
+	// hint and the issue's acceptance criteria. Used by the coder
+	// escalation tier to carry the prior model's diagnosis.
+	// +optional
+	PromptPrefix string `json:"promptPrefix,omitempty"`
+
 	// Agent is the named agent to invoke. Required for freeform; defaults
 	// to "issue-fixer" for issue-fix and "verify" for verify.
 	// +optional
--- a/api/foreman/v1alpha1/workload_types.go
+++ b/api/foreman/v1alpha1/workload_types.go
@@ -175,6 +175,28 @@ type WorkloadSpec struct {
 	// +optional
 	OpenPullRequest *bool `json:"openPullRequest,omitempty"`
 
+	// EscalationCoderAgentRef is the optional coder escalation tier: a
+	// single, typically larger/denser coder Agent that re-attempts an
+	// issue when the base CoderAgentRef fails at its model's ceiling.
+	// Issue-batch mode only. For each issue N, if the base code-<N> task
+	// is terminal with a capability failure (verdict NO-GO from the
+	// model, or outcome CODER-GATE-FAILED), the WorkloadReconciler emits
+	// an escalation code+verify(+review) set (code-<N>-esc /
+	// verify-<N>-esc / review-<N>-esc-<i>) against this Agent, on a fresh
+	// branch foreman/<w>/issue-<N>-esc, with the failed model's summary
+	// passed as a prompt hint.
+	//
+	// Singular (unlike EscalationReviewerAgentRefs, which fans out N
+	// reviewers in parallel): coders are sequential, so N parallel
+	// escalation coders would produce N competing branches. Exactly one
+	// escalation tier; an escalation task that itself fails is terminal.
+	//
+	// Does NOT trigger on STUCK-LOOP-DETECTED, a model-decided
+	// INCOMPLETE, or ERROR: those are scope/harness failures a larger,
+	// slower model will not fix. Leave unset to disable escalation.
+	// +optional
+	EscalationCoderAgentRef *corev1.LocalObjectReference `json:"escalationCoderAgentRef,omitempty"`
+
 	// AllowOverwrite lets this Workload's coder replace a stale remote
 	// ref for its own foreman/* branch (force-with-lease compare-and-swap)
 	// instead of failing non-fast-forward. Opt-in — #573 deliberately
--- a/api/foreman/v1alpha1/zz_generated.deepcopy.go
+++ b/api/foreman/v1alpha1/zz_generated.deepcopy.go
@@ -997,6 +997,16 @@ func (in *WorkloadSpec) DeepCopyInto(out *WorkloadSpec) {
 		*out = make([]v1.LocalObjectReference, len(*in))
 		copy(*out, *in)
 	}
+	if in.OpenPullRequest != nil {
+		in, out := &in.OpenPullRequest, &out.OpenPullRequest
+		*out = new(bool)
+		**out = **in
+	}
+	if in.EscalationCoderAgentRef != nil {
+		in, out := &in.EscalationCoderAgentRef, &out.EscalationCoderAgentRef
+		*out = new(v1.LocalObjectReference)
+		**out = **in
+	}
 	if in.MaxReviewIterations != nil {
 		in, out := &in.MaxReviewIterations, &out.MaxReviewIterations
 		*out = new(int32)
--- a/charts/foreman/templates/crds/agentictasks.yaml
+++ b/charts/foreman/templates/crds/agentictasks.yaml
@@ -235,6 +235,15 @@ spec:
                   prompt:
                     description: Prompt is the agent input. Required for freeform.
                     type: string
+                  promptPrefix:
+                    description: |-
+                      PromptPrefix is prepended to the assembled user prompt for an
+                      issue-fix task, BEFORE the fetched issue body. Unlike Prompt
+                      (which, when set, suppresses the issue-body fetch), PromptPrefix
+                      leaves the issue-body fetch intact, so the agent sees both this
+                      hint and the issue's acceptance criteria. Used by the coder
+                      escalation tier to carry the prior model's diagnosis.
+                    type: string
                   repo:
                     description: Repo is the "owner/name" GitHub repo. Required for
                       issue-fix and verify.
--- a/charts/foreman/templates/crds/workloads.yaml
+++ b/charts/foreman/templates/crds/workloads.yaml
@@ -112,6 +112,39 @@ spec:
                     type: string
                 type: object
                 x-kubernetes-map-type: atomic
+              escalationCoderAgentRef:
+                description: |-
+                  EscalationCoderAgentRef is the optional coder escalation tier: a
+                  single, typically larger/denser coder Agent that re-attempts an
+                  issue when the base CoderAgentRef fails at its model's ceiling.
+                  Issue-batch mode only. For each issue N, if the base code-<N> task
+                  is terminal with a capability failure (verdict NO-GO from the
+                  model, or outcome CODER-GATE-FAILED), the WorkloadReconciler emits
+                  an escalation code+verify(+review) set (code-<N>-esc /
+                  verify-<N>-esc / review-<N>-esc-<i>) against this Agent, on a fresh
+                  branch foreman/<w>/issue-<N>-esc, with the failed model's summary
+                  passed as a prompt hint.
+
+                  Singular (unlike EscalationReviewerAgentRefs, which fans out N
+                  reviewers in parallel): coders are sequential, so N parallel
+                  escalation coders would produce N competing branches. Exactly one
+                  escalation tier; an escalation task that itself fails is terminal.
+
+                  Does NOT trigger on STUCK-LOOP-DETECTED, a model-decided
+                  INCOMPLETE, or ERROR: those are scope/harness failures a larger,
+                  slower model will not fix. Leave unset to disable escalation.
+                properties:
+                  name:
+                    default: ""
+                    description: |-
+                      Name of the referent.
+                      This field is effectively required, but due to backwards compatibility is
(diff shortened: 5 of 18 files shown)


joryirving (changes requested):
Reviewing as the operator running the outer-layer equivalent of this in production (our dispatch-bridge re-lanes attempt-exhausted issues to a frontier coder), so I checked the trigger design against our live fleet's task envelopes rather than just the diff.

**Field validation of the truth table — it holds on real CRs:**
- `wl-misospace-dispatch-525-code-525`: verdict `INCOMPLETE`, top-level `extra.outcome: MODEL-DECIDED`, nested `modelExtra.outcome: STUCK-LOOP-DETECTED`. This is a live instance of exactly the trap the PR calls out — a naive top-level read escalates it; your discriminator correctly refuses.
- `wl-misospace-dispatch-528-code-528`: `NO-GO` with top `NO-CHANGES` — correctly excluded.
- `wl-misospace-kubetix-152-code-152` and `...-action-365-code-365`: `NO-GO` / `MODEL-DECIDED` — correctly escalate. The `coderResultEnvelope` shape matches production `status.result` verbatim.

Also confirmed no interaction hazard with #959: its iteration fires on a review NO-GO (coder GO'd), yours on a coder-terminal NO-GO (reviews never ran) — mutually exclusive by construction, and the `code-<n>-r<k>` step names can't collide with your exact-match `code-<n>` lookup.

**Blocking question — who reviews the `-esc` branch?** The tier emits `code-<N>-esc` + `verify-<N>-esc` only. On a Workload with reviewers (our fleet's default shape), the escalated coder can GO and gate-pass, but no `review-<N>-esc` exists — so no reviewer verdict, and post-#956 no `openPullRequest` carrier either. The escalated fix ends as a pushed, unreviewed branch with no PR, on the path where the operator has explicitly declared they want reviews. Related: what does rollup make of that issue — the base `review-<N>` tasks presumably cascade-failed when `code-<N>` failed, so does the Workload still roll up Failed even after a successful escalation? If skipping reviews on the esc tier is intentional scoping, a doc note + follow-up issue works for me and I'll approve; if not, mirroring the base review steps (with the #937 stamp) onto the esc branch looks mechanical.

**Non-blocking, from the same field data:** 2 of our 3 real `NO-GO/MODEL-DECIDED` cases were "this issue is already resolved by <commit>" honest bails. Under this design both burn an escalation run for the larger model to re-confirm staleness. Not this PR's problem — but it suggests a future dedicated outcome (e.g. `ALREADY-RESOLVED`) so the honest-bail class can split into "couldn't do it" (escalate) vs "nothing to do" (close/flag). Happy to file it upstream with the CR evidence if useful.

We'll also adopt this taxonomy in our bridge's outer tier — attempt-counting escalated our PUSH-FAILED/harness failures to the big model, which your trigger correctly never would. The two layers compose nicely: in-Workload capability escalation first, lane-level re-dispatch as the outer loop.

Defilan:
Thanks for the field-validation against live CRs, that is exactly the confidence I wanted on the trigger, and the #959 mutual-exclusivity check is reassuring.

Addressed your blocking point (full mirror, not scoped out) plus one interaction I found along the way. Rebased onto current main (your #959/#967/#956 all landed).

**Reviewers on the escalated branch.** `coderEscalationSteps` now emits `review-<N>-esc-<i>` per `ReviewerAgentRef`, `DependsOn: verify-<N>-esc`, on the esc branch, with the `openPullRequest` stamp computed the same way the base does. No reviewers configured means code+verify only, unchanged.

**Rollup.** You were right to ask. I verified the semantics rather than assume: a base `code-<N>` NO-GO ends `Succeeded/NoGo` (counts incomplete), and the cascade-failed `verify-<N>`/`review-<N>-<i>` count failed, so without intervention a successful escalation still rolls up Failed. Extended `activeChildren` so that once `code-<N>-esc` exists, the base attempt for N is superseded and dropped from the rollup slice. The envtest proves it: with the fix the escalated issue reaches `Completed`, and reverting the supersession forces `Failed` (leftover base incomplete+failed).

**Cross-tier guard (found while wiring the above).** `escalationSteps` scans `review-<N>-` by prefix, which also matches my new `review-<N>-esc-<i>`. With both `EscalationReviewerAgentRefs` and `EscalationCoderAgentRef` set (your fleet's shape), a NO-GO on an escalated review would have fanned `escalate-<N>-<j>` against the base branch. Added an exclusion so the esc reviews are skipped by the base scan; unit-tested.

All green: unit + envtest, `make test`, lint both arches, no CRD drift.

On the non-blocking `ALREADY-RESOLVED` idea: agreed, the "nothing to do" honest-bail is a distinct class from "couldn't do it," and splitting it would stop escalation burning a run to re-confirm staleness. Please do file it upstream with the CR evidence, that is a clean follow-up and your live data makes the case. And yes, the two-layer composition (in-Workload capability escalation, then your lane-level re-dispatch) is exactly the shape I was hoping for.

Ready for another look.

joryirving (approved):
Approve. Traced the three changes against the parsers rather than the descriptions; all correct.

**Reviewer fan-out.** `review-<N>-esc-<i>` per ref, `DependsOn: verify-<N>-esc`, `openPullRequest` computed identically to the base — resolves the unreviewed-branch blocker. Code+verify-only when no reviewers, unchanged.

**Supersession.** Verified the matcher does exactly what's needed and nothing more: `issueStepIteration` matches every base shape (`code-N`/`verify-N` via the `step==base` arm, `review-N-<i>` via the digit-checked review parser) so `escalated[n]` drops the base coder's NO-GO **and** its cascade-failed verify/review — but rejects all three `-esc` shapes (`verify-N-esc` misses the `-r` prefix; `review-N-esc-<i>` fails `reviewIterationOf`'s digit loop on the `e` of `esc`), so the escalation attempt's own verify/review stay in the rollup. The same-pass labeled placeholders keep the Workload in-flight while esc runs, so no premature Completed. Confirmed the placeholder `labelStep` (`TrimPrefix(name, w.Name+"-")`) equals the real task's `step.Name` under `absoluteTaskName`.

**Cross-tier guard.** Switch order puts `review-N-esc-` before `review-N-`, so esc reviews are skipped by the base reviewer-escalation scan while `review-N-<i>` still counts. Correct.

**One non-blocking edge (inherited, not introduced here).** `coderEscalationSteps` gates the whole issue on `existingEsc[code-<N>-esc]`, so if a reconcile creates `code-<N>-esc` but a transient API error aborts before `review-<N>-esc-<i>` (renderAndCreate creates in order, returns on the first non-AlreadyExists error), the next pass sees `code-<N>-esc` exists → `continue` → the missing review-esc is never re-proposed, and the branch lands unreviewed again — the exact class this commit closes. Low probability, and it's the same issue-level dedup `escalationSteps` (#546) uses. Since `renderAndCreate` already skips AlreadyExists, keying the skip per-step (or emitting all esc steps every pass and letting the create dedup) would make it self-healing. Fine as a follow-up.

Filing the `ALREADY-RESOLVED` outcome issue upstream with the CR evidence as agreed.

PR #966, opened 2026-07-03
fix(inference): warn when modelCache.claimName is silently ignored

### LINKED ISSUE #965: [BUG] InferenceService: spec.modelCache.claimName silently ignored when caching is off
## Bug Description

`spec.modelCache.claimName` (added in #960) lets an InferenceService point at a user-owned cache PVC. But when model caching is effectively OFF, a set `claimName` is silently ignored: the storage builder routes to `buildEmptyDirStorageConfig`, which never consults `claimName`, so the pod gets an `emptyDir` and re-downloads the model on every restart with no event, condition, or log.

Caching is off when `useCache := effectiveModelCacheKey(model) != "" && r.ModelCachePath != ""` is false, e.g.:
- the chart sets `modelCache.enabled: false`,
- the model uses a `file://` / local source, or
- a remote single-file model is reconciled before `Status.CacheKey` is populated.

#960 deliberately added a `ModelCacheClaimIgnored` warning for the `pvc://`-source case (`inferenceservice_controller.go`) to avoid exactly this silent drop, but the symmetric caching-off case was missed.

## Steps to Reproduce

1. Install the chart with `modelCache.enabled: false` (or use a `file://` model source).
2. Create an InferenceService with `spec.modelCache.claimName: my-cache`.
3. Observe the pod mounts an `emptyDir`, not the named PVC, and re-downloads on restart. No warning event is emitted.

## Expected Behavior

When `claimName` is set but caching is off, emit the same `ModelCacheClaimIgnored` warning event (or fail closed), so the ignored config is visible. This matches the PR's own "no silent fallback" contract.

## Actual Behavior

`claimName` is silently dropped; the pod gets an `emptyDir` and re-downloads every restart, with no event/condition/log.

## Environment

**LLMKube Version:** main (post-#960)

## Notes

Small, localized fix: add a `claimName != "" && !useCache` check next to the existing `pvc://` warning in `Reconcile` (~15-20 lines + a unit test asserting the event). Follow-up from the #960 review; not a GC/data-loss risk (the user PVC is never adopted), purely an observability gap.

joryirving:
## What

Emit a `ModelCacheClaimIgnored` warning event when `spec.modelCache.claimName` is set but the cache path is inactive — caching disabled on the operator (chart `modelCache.enabled: false` / `--model-cache-path` unset) or the model has no effective cache key (local `file://` source, or a remote model not yet fingerprinted).

## Why

Follow-up from the #960 review: in these cases the pod silently fell back to an emptyDir and re-downloaded the model on every restart, with no event or condition. #960 already warned for the `pvc://`-source conflict; this covers the symmetric case.

Fixes #965

## How

- The check mirrors the `useCache` gate in `constructDeployment`; it lives in a `warnIgnoredModelCacheClaim` helper next to the other claimName logic (extracted rather than inlined because the extra branches pushed `Reconcile` over the gocyclo limit). Event messages say why the claim was ignored and that the pod gets an ephemeral emptyDir.
- Envtest reconcile tests with a `FakeRecorder` assert every `ModelCacheClaimIgnored` variant — including the pre-existing `pvc://` case, which was previously unasserted — plus a no-warning case when caching is active.
- `docs/MODEL-CACHE.md`: updated the ignored-claim bullet and added a note that `claimName` is mutable and the old claim is not garbage-collected.
- Skipped the `ModelCachePVCNotFound` condition-reason nice-to-have: the Degraded reason is hardcoded in the shared `updateStatusWithSchedulingInfo` switch, so a specific reason means plumbing a new parameter through it — happy to do as a separate PR if wanted.

AI assistance: authored by Claude (Fable 5) operating under the maintainer's direction; reviewed before submitting.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (`go test ./internal/controller/...` with envtest assets, 506/506)
- [x] `make lint` passes locally (`golangci-lint run ./internal/... ./api/...`, 0 issues)

### DIFF
--- a/docs/MODEL-CACHE.md
+++ b/docs/MODEL-CACHE.md
@@ -175,9 +175,15 @@ Behavior:
   garbage-collects `<isvc>-model-cache`). If the claim does not exist, the
   InferenceService is marked `Degraded` with a `ModelCachePVCNotFound` event
   instead of silently falling back to the shared cache.
-- `claimName` targets the download path, so it is ignored for pre-staged
-  `pvc://` model sources (mounted read-only, no download); a warning event is
-  emitted if both are set.
+- `claimName` targets the download path, so it is ignored — with a
+  `ModelCacheClaimIgnored` warning event — for pre-staged `pvc://` model
+  sources (mounted read-only, no download), and whenever caching is inactive
+  (chart `modelCache.enabled: false`, a local `file://` source, or a remote
+  model not yet fingerprinted); in the inactive-caching cases the pod falls
+  back to an ephemeral emptyDir and re-downloads on every restart.
+- `claimName` is mutable: changing it rolls the Deployment onto the new claim,
+  but the old claim (and the weights on it) is **not** garbage-collected —
+  clean it up yourself if it is no longer needed.
 - Node alignment is your responsibility: for an RWO or node-local claim, use
   `nodeSelector` so the pod lands where the PVC binds (a
   `WaitForFirstConsumer` local class binds on the first consumer; a pre-bound
--- a/internal/controller/inferenceservice_controller.go
+++ b/internal/controller/inferenceservice_controller.go
@@ -173,15 +173,7 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		}
 	}
 
-	// spec.modelCache.claimName targets the download path, so it is meaningless
-	// for a pre-staged pvc:// source (mounted read-only, no download). The
-	// claimName is ignored in that case; surface a Warning so the conflict is
-	// visible instead of silently dropped.
-	if r.Recorder != nil && userModelCacheClaimName(inferenceService) != "" && isPVCSource(model.Spec.Source) {
-		r.Recorder.Eventf(inferenceService, nil, corev1.EventTypeWarning, "ModelCacheClaimIgnored", "Reconcile",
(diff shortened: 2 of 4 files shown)


Defilan (approved):
### Review

Approve. This closes the silent-drop gap I flagged on #960 (filed as #965): when `spec.modelCache.claimName` is set but caching is off, the user now gets a `ModelCacheClaimIgnored` warning instead of a silent emptyDir.

Verified:
- `warnIgnoredModelCacheClaim` fires the existing `ModelCacheClaimIgnored` reason across all three ignore cases (`pvc://` source, `ModelCachePath == ""`, empty cache key). The switch conditions are the exact negation of the real `useCache` gate (`effectiveModelCacheKey(model) != "" && r.ModelCachePath != ""`), so the warning mirrors the actual storage path chosen.
- Wired unconditionally in `Reconcile` before `reconcileDeployment` (not dead code), and it back-fills the previously-inline `pvc://` warning.
- No false positive on the happy path (claimName set + caching on), asserted by a dedicated `NotTo(ContainElement(...))` test; the envtest covers all four cases with a FakeRecorder.
- Tightly scoped (helper extraction + call-site swap + tests + a docs bullet), DCO signed.

One cosmetic note: the test/doc reference `#928` rather than the silent-drop follow-up `#965` this resolves; I'll close #965 pointing here.

Thanks for picking this up.

Issue #944, opened 2026-07-02
[FEATURE] Foreman edit-fidelity: actionable feedback when str_replace recovery exhausts with no unique anchor


Defilan:
## Feature Description

When `str_replace` fails with zero matches AND `anchorContext` cannot find a unique anchor line, the model currently gets the bare error `str_replace: old_string found 0 times in <path>, want 1` with no actionable information. Add a best-effort feedback fallback for this corner so the model always receives some real file content to re-anchor on.

## Problem Statement

The recovery ladder (exact -> whitespace-normalized -> fuzzy unique-window -> anchorContext truth hint) has one dead end: on large files where the drifted `old_string` has no line that appears verbatim-trimmed exactly once, `anchorContext` returns nothing and the error carries zero guidance.

Observed live: a coder run drifted `old_string` against a ~2000-line file three times in a row, received the bare count error each time (no anchor existed for its paraphrased lines), and was terminated by EditFreeStreak. The fuzzy layer correctly refused to guess (the drift was beyond threshold), but the model had no path back to truth.

> As a coder-harness operator, I want str_replace failures to always include some real file context so that a local model can re-copy actual bytes instead of re-hallucinating old_string until the stuck-loop detector kills the run.

## Proposed Solution

When the recovery ladder is exhausted and `anchorContext` finds no unique anchor, fall back to a best-effort locator instead of the bare count error, for example:

- Score every file line against the most distinctive `old_string` line with the existing `boundedLevenshtein` helper and return the surrounding real lines of the single lowest-distance hit (even if not unique-verbatim), clearly labeled as approximate.
- Alternatively (or additionally), tell the model the file's line count and suggest a `read_file` of a specific line range near the closest match.

Safety posture unchanged: this is feedback only, never an auto-apply. The unique-window rule for automatic edits stays as is.

## Alternatives Considered

- Read-before-edit enforcement (tool-level: reject str_replace unless the file was read this task). Deferred from the original edit-fidelity work; heavier UX change for all models.
- Raising fuzzy thresholds so more drift auto-applies. Rejected: threshold 6 is already the documented safety ceiling; loosening it trades wrong-target edits for convergence.

## Additional Context

Follow-up from the edit-fidelity work (#942 / PR #943), which added the fuzzy recovery and write_file steering; this covers the remaining no-anchor dead end observed during its live fleet validation.

PR #969, opened 2026-07-03
fix(foreman): reject no-op str_replace where old_string equals new_string

### LINKED ISSUE #968: [BUG] Foreman str_replace reports success on a no-op edit (old_string == new_string), driving a RepeatedToolCall loop
## Bug Description

The Foreman coder `str_replace` tool reports **success** (`{"replacements":1}`) for a no-op edit where `old_string == new_string`. Because the file is not actually changed, a model that issues such a call sees "success" but observes no change, so it re-issues the identical call, driving a degenerate loop until the RepeatedToolCall stuck-loop detector terminates the run. The tool should reject an identical-strings replacement instead of rewarding it with a phantom success.

Observed live on two different models (Ornith-35B-A3B MoE and Qwopus3.6-27B dense) on the same issue, so it is model-independent: it is a tool-behavior bug, not a model capability limit.

## Steps to Reproduce

1. Run a coder task where the model constructs a `str_replace` call with `old_string` equal to `new_string` (e.g. it intends an edit but emits identical strings).
2. Observe the tool returns `{"path":"...","replacements":1}` (success), while the file is unchanged.
3. The model, seeing no change, re-issues the identical call. Repeat until the RepeatedToolCall progress monitor fires (>= 5 identical calls) and the run ends INCOMPLETE / STUCK-LOOP-DETECTED.

## Expected Behavior

`str_replace` returns an error when `old_string == new_string`, e.g. "old_string and new_string are identical; no change made. If the file already has the desired content, proceed; otherwise correct new_string." This breaks the loop on the first no-op instead of letting the model believe it edited.

## Actual Behavior

A no-op (`old == new`) returns `{"replacements":1}` (success). The model loops on the identical call. Live transcript (Qwopus, issue #944 A/B): six byte-identical `str_replace` calls (old and new both 1856 chars, identical), each returning `replacements:1`, model reasoning "I keep writing the same thing," RepeatedToolCall alert at turn 34, run ends INCOMPLETE.

## Environment

**LLMKube Version:** foreman-agent 0.8.27 (also reproduced on the PR #964 build)

## Notes

- Distinct from #942/#943/#944, which handle `str_replace` *failing to match*; this is `str_replace` *succeeding on a no-op*.
- The RepeatedToolCall detector does catch it, but only after 5 identical calls (turn 34), by which point the turn budget is largely spent. A no-op guard catches it on the first occurrence.
- Fix is localized to the `str_replace` tool: add an `old_string == new_string` guard that returns an error before attempting the replacement, with a message steering the model to either proceed (if content is already correct) or correct `new_string`.

Defilan:
## What

Reject a no-op `str_replace` where `old_string == new_string` with an actionable error, instead of silently reporting `replacements:1` (a phantom success).

## Why

Fixes #968

A `str_replace` whose `old_string` equals its `new_string` changes nothing, but the tool's match logic still "found and replaced" the string with itself and returned `{"replacements":1}`. A model that sees success on a file that did not actually change concludes the edit did not take, re-issues the identical call, and loops until the `RepeatedToolCall` stuck-loop detector terminates the run INCOMPLETE.

This was found in the wild, not theorized: in live Foreman coder runs on issue #944, two different local models (a 35B-A3B MoE and a coder-tuned sibling) each constructed a byte-identical `old_string`/`new_string` `str_replace`, received the phantom success six times in a row, and were killed by the stuck-loop detector at turn 34. The models even narrated it ("I keep writing the same thing") but could not break out, because the tool kept telling them they had succeeded. The failure is model-independent: it is the tool rewarding a no-op.

## How

A guard at the top of `Execute`, right after the required-args check and before the file is read (a no-op is a pure argument mistake, so there is no reason to touch the file):

```go
if a.OldString == a.NewString {
    return nil, fmt.Errorf("str_replace: old_string and new_string are identical, ...")
}
```

The error steers the model to either move on (if the file already contains the desired text) or correct `new_string`. Behavior for every real edit is unchanged; only the exact-equality no-op is now rejected. Distinct from #942/#943's recovery ladder (which handles `str_replace` *failing to match*); this is `str_replace` *succeeding on a no-op*.

## Checklist

- [x] Tests added/updated (`TestStrReplace_IdenticalOldNewRejected`: asserts the error, the `identical` message, and that the file is untouched; written test-first, watched it fail on the phantom success, then pass)
- [x] `make test` scope: `go test ./pkg/foreman/agent/tools/` passes; `go build ./...` clean
- [x] `make lint` passes locally (0 issues; also `GOOS=linux golangci-lint` 0 issues)
- [x] AI assistance disclosed: implemented with AI assistance (Claude Code) under human review; human-accountable (band 3), commits kept clean per CONTRIBUTING.md
- [x] Documentation updated (n/a: internal tool behavior, no user-facing doc)

### DIFF
--- a/pkg/foreman/agent/tools/str_replace.go
+++ b/pkg/foreman/agent/tools/str_replace.go
@@ -81,6 +81,18 @@ func (t *StrReplaceTool) Execute(_ context.Context, args json.RawMessage) (*agen
 	if a.Path == "" || a.OldString == "" {
 		return nil, fmt.Errorf("str_replace: path and old_string are required")
 	}
+	// A no-op edit (old_string == new_string) changes nothing, but the match
+	// logic below would still "replace" the string with itself and report
+	// replacements:1 -- a phantom success. A model that sees success on an
+	// unchanged file re-issues the identical call and gets killed by the
+	// RepeatedToolCall stuck-loop detector (#968). Reject it explicitly so the
+	// model corrects new_string or moves on.
+	if a.OldString == a.NewString {
+		return nil, fmt.Errorf("str_replace: old_string and new_string are identical, " +
+			"so this edit would change nothing. If the file already contains the text " +
+			"you want, move on to the next step; otherwise set new_string to the " +
+			"replacement text you intend")
+	}
 	full, err := resolveInside(t.Workspace, a.Path)
 	if err != nil {
 		return nil, fmt.Errorf("str_replace: %w", err)
--- a/pkg/foreman/agent/tools/str_replace_fuzzy_test.go
+++ b/pkg/foreman/agent/tools/str_replace_fuzzy_test.go
@@ -324,3 +324,26 @@ func TestStrReplace_NonexistentFileSteersToWriteFile(t *testing.T) {
 		t.Errorf("expected existence + write_file steering in the error, got: %v", err)
 	}
 }
+
+// Regression for #968: str_replace with old_string == new_string is a no-op.
+// Before the guard the match logic would "find and replace" the string with
+// itself and report replacements:1 -- a phantom success. Models that see
+// success on an unchanged file re-issue the identical call and get killed by
+// the RepeatedToolCall stuck-loop detector. It must be rejected instead.
+func TestStrReplace_IdenticalOldNewRejected(t *testing.T) {
+	ws := makeWorkspace(t)
+	src := "func compute() int {\n\treturn userCount\n}\n"
+	seedFile(t, ws, "f.go", src)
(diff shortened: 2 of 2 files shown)


joryirving (approved):
Review:
- Placement is correct — guard sits after required-args validation, before any file I/O. Cheapest possible rejection.
- Error message is good — tells the model why it failed and what to do (move on or fix new_string). Matches the steering style of the rest of this file.
- Test is solid — verifies error returned, error contains "identical", file untouched. Follows existing helpers (execStrReplace, seedFile, readBack).
- Comment explains the why — phantom success → stuck loop chain. No fluff.

Issue #14, opened 2025-11-18
[FEATURE] Create Minikube quickstart guide


Defilan:
## Feature Description
Create a quickstart guide specifically for running LLMKube on Minikube for local development and testing.

## Motivation
- Many developers use Minikube for local Kubernetes development
- Current quickstart assumes GKE, which requires cloud resources and costs money
- A Minikube guide would lower the barrier to entry for contributors and users who want to test locally

## Proposed Solution
Create a new guide `docs/minikube-quickstart.md` that covers:

1. **Prerequisites**
   - Minikube installation
   - kubectl installation
   - Minimum system requirements (CPU, RAM, disk)

2. **Minikube Setup**
   - Starting Minikube with appropriate resources
   - Recommended Minikube driver (docker, hyperkit, etc.)
   - CPU and memory allocation for running small models

3. **LLMKube Installation**
   - Installing CRDs
   - Running the controller locally (similar to current quickstart)
   - Or deploying controller to Minikube

4. **Deploy a Model**
   - Use TinyLlama or another small model suitable for local testing
   - Resource requests appropriate for Minikube (e.g., 500m CPU, 1Gi RAM)

5. **Testing the Service**
   - Port-forwarding to access the API
   - Example curl commands
   - Performance expectations on local machines

6. **Troubleshooting**
   - Common issues (insufficient resources, scheduling failures)
   - How to check Minikube resources
   - Cleaning up resources

## Acceptance Criteria
- [ ] Documentation file created at `docs/minikube-quickstart.md`
- [ ] Tested on at least 2 different Minikube drivers
- [ ] Includes commands that can be copy-pasted
- [ ] Model deploys successfully with reasonable resource requests
- [ ] API endpoint accessible and returns valid responses
- [ ] Linked from main README.md

## Additional Context
The current quickstart (examples/quickstart/) works on any Kubernetes cluster, but a dedicated Minikube guide would:
- Provide specific resource recommendations for local machines
- Include Minikube-specific tips and troubleshooting
- Make it easier for new contributors to get started without cloud resources

## Related
- See existing quickstart: `examples/quickstart/README.md`
- Current CPU request fix: 500m CPU works on small nodes (e2-standard-2)

Defilan:
## Minikube Quickstart Guide Complete

I've created a comprehensive Minikube quickstart guide that covers all the requirements from this issue.

### What's Included

✅ **Prerequisites**
- Minikube installation instructions for macOS, Linux, and Windows
[...]

PR #971, opened 2026-07-04
ci(release): publish the coder image on release (closes the manual step behind #932)

### LINKED ISSUE #932: [FEATURE] Bump + publish the llmkube-runtimes coder image as part of the release process
## Feature Description

Make bumping and publishing the Foreman **coder image** (`ghcr.io/defilantech/llmkube-foreman-agent-coder`) a codified part of the LLMKube release process, instead of a manual, easy-to-forget step in a separate repo.

## Problem Statement

The in-cluster foreman-agent runs the coder image, which is built in the separate `defilantech/llmkube-runtimes` repo. That image pins the release it was built from via `ARG LLMKUBE_REF` in `coder/Dockerfile`, and is published by pushing a `coder-v<version>` tag. Today, after every LLMKube release, someone must manually (a) bump `ARG LLMKUBE_REF` to the new tag and (b) push a matching `coder-v<version>` tag in `llmkube-runtimes`.

Because it lives in another repo and is not part of the release checklist, it is easy to forget or to lag. When it lags, deploying the new release fails: the in-cluster foreman-agent hits `ImagePullBackOff` on the coder image tag that was never published (this happened on the 0.8.23 rollout).

> As a maintainer cutting a release, I want the coder image for that version to be built and published automatically (or via a single documented step), so a fresh deploy of the release never lands on a missing coder image.

## Proposed Solution

Preferred: on LLMKube release publish, trigger the `llmkube-runtimes` coder build for the new tag automatically. Options:

1. **Cross-repo dispatch (recommended):** the LLMKube release workflow sends a `repository_dispatch` (or `gh workflow run`) to `llmkube-runtimes` with the new `v<version>`; a workflow there bumps `ARG LLMKUBE_REF`, commits, and pushes `coder-v<version>` (which its existing `build-coder` workflow already turns into the published image). Requires a cross-repo token.
2. **release-please post-release hook:** wire the bump into the release automation that already cuts the tag.
3. **Documented release-checklist step (minimum):** add a `RELEASING.md` entry (2-step: bump `ARG LLMKUBE_REF`, push `coder-v<version>`), and/or a release-workflow check that fails/warns if `coder-v<version>` does not exist in `llmkube-runtimes` after a release.

At minimum ship option 3 so the step is never silently skipped; option 1 is the real fix.

## Alternatives Considered

Continuing to do it by hand each release (current state, error-prone). Folding the coder image build into the LLMKube repo itself (rejected: keeping the toolchain/runtime image in `llmkube-runtimes` keeps the operator repo air-gap-clean and its build fast).

## Additional Context

- Coder image repo: `defilantech/llmkube-runtimes`, `coder/Dockerfile` (`ARG LLMKUBE_REF`), tag pattern `coder-v<version>`, workflow `build-coder`.
- Related failure mode: in-cluster foreman-agent `ImagePullBackOff` on a coder tag that was never published (0.8.23).
- The `llmkubelab` Ansible deploy pins the in-cluster agent image tag to the release version, so the coder image must exist at `:<version>` before a deploy.

## Priority

## Willingness to Contribute

- [x] Yes, I can submit a PR

joryirving:
## What
Adds a `publish-coder-image` job to `release-please.yml`: on a created release it tags `llmkube-runtimes` `coder-v<version>` (which the runtimes `build-coder` workflow turns into `llmkube-foreman-agent-coder:<version>`), then polls GHCR until that image exists — failing the release if it does not within ~30m.

## Why
The coder image is the one release artifact built outside this repo, published by a manual bump+tag in llmkube-runtimes. When it lags, a fresh deploy of the release ImagePullBackOffs on the missing coder tag (the 0.8.23 incident). This makes it automatic and, failing that, loud.

## How
- **Automation:** creates the `coder-v<version>` tag via `gh api` (idempotent — skips if it exists). Paired with defilantech/llmkube-runtimes#14 (LLMKUBE_REF-from-tag), the tag push alone builds the correct ref — no Dockerfile bump.
- **Verification:** polls the GHCR manifest for `:<version>`; a missing image fails the release loudly instead of silently.
- **Token-gated:** requires a `RUNTIMES_TOKEN` secret — a PAT or GitHub App token with `contents:write` on llmkube-runtimes (needed both for the cross-repo write and to *trigger* the tag-driven build; the default `GITHUB_TOKEN` can do neither). **Without the secret the job is a warning-only no-op, so releases still succeed** — it activates the moment you add it.

## Merge / activation order
1. Merge llmkube-runtimes#14 (ref-from-tag).
2. Add the `RUNTIMES_TOKEN` secret to this repo.
3. Merge this. From then on every release publishes its coder image with no manual step; a broken coder build fails the release.

Fixes #932 (the automation half; #14 is the runtimes half).

## Verification
YAML validates. Inert until the secret exists; the tag-create + poll logic is standard `gh api` / GHCR manifest HEAD.

🤖 Authored by Claude (Fable 5).

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -166,3 +166,69 @@ jobs:
           mark_as_latest: false
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+
+  publish-coder-image:
+    needs: release-please
+    if: ${{ needs.release-please.outputs.release_created == 'true' }}
+    runs-on: ubuntu-latest
+    steps:
+      # The coder image (llmkube-foreman-agent-coder) is built in the separate
+      # llmkube-runtimes repo, triggered by a coder-v<version> tag. Create that
+      # tag on every release so its coder image publishes automatically, then
+      # verify it lands (#932). Requires RUNTIMES_TOKEN: a PAT or GitHub App
+      # token with contents:write on llmkube-runtimes (needed both to push the
+      # tag and to trigger the downstream workflow — the default GITHUB_TOKEN
+      # cannot do cross-repo writes and would not fire the tag-triggered build).
+      # Without the secret this job is a warning-only no-op, so releases still
+      # succeed until it is provisioned.
+      - name: Gate on cross-repo token
+        id: gate
+        env:
+          RUNTIMES_TOKEN: ${{ secrets.RUNTIMES_TOKEN }}
+        run: |
+          if [ -z "$RUNTIMES_TOKEN" ]; then
+            echo "::warning::RUNTIMES_TOKEN not set; skipping automatic coder-image publish for v${{ needs.release-please.outputs.version }}. Add the secret (contents:write on llmkube-runtimes) to automate, or push coder-v${{ needs.release-please.outputs.version }} manually (#932)."
+            echo "enabled=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "enabled=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Tag llmkube-runtimes to build the coder image
+        if: ${{ steps.gate.outputs.enabled == 'true' }}
+        env:
+          GH_TOKEN: ${{ secrets.RUNTIMES_TOKEN }}
+          VERSION: ${{ needs.release-please.outputs.version }}
+        run: |
+          set -euo pipefail
+          repo=defilantech/llmkube-runtimes
+          tag="coder-v${VERSION}"
+          if gh api "repos/${repo}/git/ref/tags/${tag}" >/dev/null 2>&1; then
+            echo "tag ${tag} already exists in ${repo}; not recreating"
+          else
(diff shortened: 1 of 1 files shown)


Defilan (approved):
Approving. Checked the two things that looked risky and both hold up:

- `version` is bare (used as image tags elsewhere in this workflow; existing coder tags are bare), so `coder-v${VERSION}` and the `:${VERSION}` manifest check are well-formed.
- The `llmkube-foreman-agent-coder` GHCR package is public, so the anonymous-pull verify actually returns 200 (it would loop forever on a private package).

Idempotent tag creation, the token-gated warning-only no-op, and the loud-fail verification are all done right.

Non-blocking follow-up: in the verify loop, `set -euo pipefail` + `token=$(curl -sSf ...)` means a transient token/network failure on a single poll iteration aborts the step and fails the release instead of retrying to the next attempt — worth letting transient fetch errors fall through. Low probability given the public image.

Merge order per the body: llmkube-runtimes#14 first, add `RUNTIMES_TOKEN`, then this.

PR #973, opened 2026-07-04
fix(runtime): bind inference servers to :: for IPv6-only clusters

### LINKED ISSUE #972: [BUG] llama-server binds to 0.0.0.0 in IPv6 only cluster instead of :: which binds to IPv6 and IPv4
## Bug Description

Pods running llama-server do not get ready in IPv6 only clusters.

## Steps to Reproduce

1. Deploy model with '...'
2. Run command 'kubectl --namespace llmkube-system get pods'
3. Observe error 'Pod never gets Ready'

a. scale llmkube-controller-manager to 0 so it leaves the deployment alone
b. manually edit the deployment, replace --host 0.0.0.0 with -host :: (don't forget the quotes)
c. Observe how the pod is reachable from the control plane and gets ready as expected

## Expected Behavior

Pod gets Ready

## Actual Behavior

Pod never gets Ready

## Environment

**LLMKube Version:**
```bash
llmkube version
# Output:
llmkube version 0.8.27
  git commit: 7eb640587f5b9ec2d682edafe9ce2d3a0c2b7385
  build date: 2026-07-03T00:09:26Z
```

**Kubernetes Version:**
```bash
kubectl version
# Output:
Client Version: v1.35.3
Kustomize Version: v5.7.1
Server Version: v1.36.1
```

**Cluster Type:**
- [ ] GKE
- [ ] EKS
- [ ] AKS
- [ ] minikube
- [ ] kind
- [ ] K3s
- [X] Other: Talos

**GPU (if applicable):**
- [ ] NVIDIA T4
- [ ] NVIDIA L4
- [ ] NVIDIA V100
- [ ] NVIDIA A100
- [ ] Metal (Apple Silicon)
- [ ] None (CPU only)
- [X] Other: AMD APU

## Logs

**Controller Logs:**
```bash
# Paste logs here
```

**Inference Pod Logs:**
```bash
kubectl logs <pod-name> -c llama-server
# Paste logs here
```

## YAML Manifests

**Model:**
```yaml
# Paste your Model YAML here
```

**InferenceService:**
```yaml
# Paste your InferenceService YAML here
```

## Additional Context

Add any other context about the problem here (screenshots, error messages, etc.).

Defilan:
## What

Change the inference-server bind address from `0.0.0.0` (IPv4 wildcard) to `::` (dual-stack wildcard) in the llama.cpp, vLLM, and TGI runtimes.

## Why

Fixes #972. On an IPv6-only cluster (reported on Talos + AMD APU), the pod IP is IPv6. The runtimes hardcoded `--host` / `--hostname 0.0.0.0`, so llama-server only listened on IPv4. The kubelet health probe targets the pod IP (IPv6), nothing answers there, and the pod never becomes Ready.

## How

- `internal/controller/runtime_llamacpp.go`, `runtime_vllm.go`, `runtime_tgi.go`: bind `::` instead of `0.0.0.0`.
- `::` with the default `net.ipv6.bindv6only=0` also accepts IPv4 via v4-mapped addresses, so **IPv4-only and dual-stack clusters are unaffected** (no regression for existing users).
- **Escape hatch for the rare IPv6-disabled node:** llama.cpp and vLLM append `extraArgs` last, and all three parsers keep the last occurrence of a repeated flag (verified: llama.cpp logs "only last value will be used"; vLLM argparse; TGI clap). So `extraArgs: ["--host", "0.0.0.0"]` overrides it. (TGI does not currently consume `extraArgs`; a small parity follow-up could add that if needed.)
- Added a llama.cpp regression test asserting `--host ::` and no `0.0.0.0`; updated the vLLM arg assertions.

## Checklist

- [x] `gofmt`, `go build ./...`, `go vet ./...`, `golangci-lint` (0 issues)
- [x] Unit tests pass (`TestLlamaCppBuildArgs`, `TestVLLMBuildArgs`); the one envtest-suite assertion is updated and runs in CI
- [x] No API/CRD change (no `make manifests`/`generate` needed)
- [x] No RBAC marker change

Thanks to @dekarl for the clear, reproducible report and for verifying the `::` fix.

---
Developed with AI assistance (Claude) per LLMKube's AI-assisted contribution policy (band 3).

### DIFF
--- a/internal/controller/inferenceservice_deployment_test.go
+++ b/internal/controller/inferenceservice_deployment_test.go
@@ -3660,12 +3660,12 @@ var _ = Describe("RuntimeBackend interface", func() {
 			// deprecated and intentionally omitted.
 			Expect(args[0]).To(Equal("/models/llama3"))
 			Expect(args).NotTo(ContainElement("--model"))
-			Expect(args).To(ContainElements("--host", "0.0.0.0"))
+			Expect(args).To(ContainElements("--host", "::"))
 			Expect(args).To(ContainElements("--port", "8000"))
 			// --enable-metrics is always set so vLLM pods expose /metrics for
 			// PodMonitor scraping (#409).
 			Expect(args).To(ContainElement("--enable-metrics"))
-			// Six elements: <model> --host 0.0.0.0 --port 8000 --enable-metrics.
+			// Six elements: <model> --host :: --port 8000 --enable-metrics.
 			Expect(args).To(HaveLen(6))
 		})
 
--- a/internal/controller/inferenceservice_deployment_test.go
+++ b/internal/controller/inferenceservice_deployment_test.go
@@ -4206,7 +4206,7 @@ var _ = Describe("constructDeployment Regression Tests", func() {
 			Expect(container.Ports[0].Name).To(Equal("http"))
 
 			By("verifying base args")
-			Expect(container.Args).To(ContainElements("--model", "--host", "0.0.0.0", "--port", "8080"))
+			Expect(container.Args).To(ContainElements("--model", "--host", "::", "--port", "8080"))
 			Expect(container.Args).To(ContainElement("--metrics"))
 
 			By("verifying no GPU args")
--- a/internal/controller/runtime_llamacpp.go
+++ b/internal/controller/runtime_llamacpp.go
@@ -43,7 +43,12 @@ func (b *LlamaCppBackend) DefaultHPAMetric() string { return "llamacpp:requests_
 func (b *LlamaCppBackend) BuildArgs(isvc *inferencev1alpha1.InferenceService, model *inferencev1alpha1.Model, modelPath string, port int32) []string {
 	args := []string{
 		"--model", modelPath,
-		"--host", "0.0.0.0",
+		// Bind the dual-stack wildcard so pods are reachable on IPv6-only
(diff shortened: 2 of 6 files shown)


joryirving (approved):
Approve. Verified the change does what it says and the no-regression claim holds:

- All three runtimes flip the bind wildcard: llama.cpp/vLLM `--host ::`, TGI `--hostname ::`. `::` with the Linux default `bindv6only=0` accepts IPv4 via v4-mapped addresses, so IPv4-only and dual-stack clusters keep working — correct.
- The `extraArgs` escape hatch is genuine for llama.cpp and vLLM: vLLM appends `ExtraArgs` last ("user flags win"), llama.cpp appends them after the `--host` default, and last-flag-wins in both parsers, so `extraArgs: ["--host","0.0.0.0"]` overrides. Tests updated (llama.cpp asserts `--host ::` + no `0.0.0.0`; vLLM assertions).

One non-blocking note: this flips the default, so a node with IPv6 fully *disabled* (`ipv6.disable=1` / `bindv6only=1`) now fails to bind by default rather than working — a narrow new failure class. llama.cpp/vLLM users can recover via `extraArgs`, but **TGI doesn't consume `extraArgs`, so a TGI user on an IPv6-disabled node has no override** — that's the parity follow-up you already flagged, and worth doing since it's the only combination with no escape hatch. Fine to land this as-is.

PR #975, opened 2026-07-04
feat(foreman): use the reviewer's summary as the opened-PR body

joryirving:
## What
When #956 opens a PR on review GO, use the reviewer's own `submit_result` summary as the PR body (followed by `Fixes #N` + the provenance line), instead of the hardcoded `Opened by foreman on review GO` stub. Empty summary falls back to just the issue link.

## Why
The reviewer has just read the entire diff against the issue to decide GO — a model-written description of the change is already sitting in `r.Summary` at the exact point we open the PR, and the code was throwing it away for a stub. Foreman-opened PRs currently have useless bodies; this gives them a real description from the model that reviewed the change, at zero extra cost (no new model call).

## How
- `maybeOpenPullRequest` passes `r.Summary` into `openPullRequest`.
- `openPullRequest` builds the body: reviewer summary (if non-empty) → `Fixes #N` → italic provenance.
- Title is unchanged (branch-head commit subject).

Pairs well with tuning the reviewer prompt to shape its summary as a PR description; that's a deployment-side prompt change, not needed for this.

## Verification
- New `TestMaybeOpenPullRequest_BodyCarriesReviewSummary` (fake Ensurer now captures the body) asserts the summary leads the body and the issue link remains.
- `go test ./pkg/foreman/agent/...` pass; lint clean.

🤖 Authored by Claude (Fable 5).

### DIFF
--- a/pkg/foreman/agent/executor_native.go
+++ b/pkg/foreman/agent/executor_native.go
@@ -1204,7 +1204,7 @@ func (e *NativeAgentLoopExecutor) maybeOpenPullRequest(
 		e.PREnsurer == nil {
 		return
 	}
-	if prURL, prErr := e.openPullRequest(ctx, task, auth); prErr != nil {
+	if prURL, prErr := e.openPullRequest(ctx, task, auth, r.Summary); prErr != nil {
 		log.Error(prErr, "review GO: opening pull request failed",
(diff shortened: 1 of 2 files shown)


Defilan (approved):
Great idea! Thanks for doing this.

PR #978, opened 2026-07-05
fix(foreman): populate FleetNode.Status.CurrentTask so the scheduler spreads work

### LINKED ISSUE #977: Foreman scheduler funnels all tasks to one FleetNode: Status.CurrentTask is never written
## Summary

The Foreman scheduler routes **every** AgenticTask to a single FleetNode. Additional agent replicas never receive work — effective concurrency is 1 regardless of `foreman-agent` replica count.

## Root cause

`firstFitNode` (`internal/foreman/controller/agentictask_controller.go:285`) enforces the "v0.1: one task per node" model by skipping any node whose `Status.CurrentTask` is non-empty:

```go
if n.Status.CurrentTask != "" {
    continue // v0.1: one task per node
}
```

But **`FleetNode.Status.CurrentTask` is never written anywhere in the operator.** Repo-wide it is only:
- declared — `api/foreman/v1alpha1/fleetnode_types.go:151`
- read by the scheduler guard — `agentictask_controller.go:302`
- logged / printcolumn — `fleetnode_controller.go:58`, `fleetnode_types.go:198`

`git log -S "CurrentTask ="` returns nothing — a write to the field has never existed. The FleetAgent watcher (`pkg/foreman/agent/watcher.go`) patches the *AgenticTask* status on claim/complete but never patches its own FleetNode.

Because `CurrentTask` is permanently `""`, the busy-check never trips, and `firstFitNode` — which sorts nodes by name and returns the first schedulable match — returns the **alphabetically-first** Ready node for every task. The rest of the fleet is never selected.

## Observed (v0.8.28, 3 `foreman-agent` replicas)

All three replicas Ready with identical roles `[worker, coder, verifier]` and fresh heartbeats. Task distribution:

```
50 unassigned  Pending
25 <first-node> Scheduled
 1 <first-node> Running
 0 <other two nodes>
```

The two idle pods log nothing past startup. Scaling replicas has no effect.

## Impact

- Throughput is capped at one concurrent task cluster-wide; horizontal scaling is a no-op.
- A backlog piles onto one node while Ready capacity sits idle.

## Suggested fix

Populate `CurrentTask` so the existing guard works (the "a node freeing up `CurrentTask` re-enqueues every Pending task" wiring at `agentictask_controller.go:599` already anticipates this lifecycle):

- **Set** on the target FleetNode when a task is scheduled (`scheduleToNode`).
- **Clear** when the task reaches a terminal phase and on the claim-expiry release paths (`releaseExpiredClaim` / `terminalFailExpired`, which already null `AssignedNode`).

Note a race: `firstFitNode` reads FleetNodes from the cached client, so back-to-back scheduling reconciles can both observe `CurrentTask == ""` and pick the same node before the first write propagates. A conflict-aware reservation (guarded status update on the node; on `Conflict`/already-set, fall through to the next candidate) closes it. Happy to send a PR along these lines.

---

joryirving:
## What

Populate `FleetNode.Status.CurrentTask` so the Foreman scheduler's one-task-per-node model actually spreads work across the fleet, instead of funneling every task onto a single node.

## Why

Fixes #977

`firstFitNode` skips nodes whose `Status.CurrentTask` is non-empty ("v0.1: one task per node"), but that field was **never written** anywhere in the operator — only declared, read by the guard, and logged. So the guard never tripped and every task was scheduled onto the alphabetically-first Ready node. Extra `foreman-agent` replicas never received work; effective concurrency was 1 regardless of fleet size (observed on v0.8.28 with 3 replicas: one node running, 25 queued on it, 50 Pending, two idle).

## How

- **`reserveFirstFitNode`** replaces `firstFitNode`: it scans candidates (sorted, schedulable, capability-matched) and reserves the first free one by stamping `CurrentTask` via an **optimistic-lock status patch**. The node scan reads FleetNodes from a cache that lags writes, so two back-to-back scheduling reconciles can both see a node as free; the optimistic lock means at most one wins and the loser (`Conflict`) falls through to the next candidate.
- **Stale-reservation recovery:** a `CurrentTask` pointing at a missing or terminal task is treated as free and reclaimed, so a reservation leaked by a task deleted mid-flight can't wedge a node busy forever.
- **`clearNodeCurrentTask`** releases the reservation on terminal tasks and on both claim-expiry paths (`releaseExpiredClaim`, `terminalFailExpired`), guarded on the task key so a node already re-reserved for a different task is untouched.
- `scheduleToNode` now returns only `error` (its `Result` was always the zero value; flagged by `unparam`).

No API/CRD change — `CurrentTask` already exists in the schema.

Design note: with `MaxConcurrentReconciles` at the default (1) the optimistic lock is belt-and-suspenders, but it keeps the fix correct if concurrency is ever raised. Happy to adjust the set/clear split (e.g. move clearing into a level-triggered `FleetNodeReconciler` pass) if you'd prefer that shape.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (envtest; controller suite, 83 specs)
- [x] `make lint` passes locally (`golangci-lint`: 0 issues)
- [x] AI assistance is disclosed (commit trailer + here), per CONTRIBUTING.md
- [ ] Documentation updated (no user-facing doc change)

### DIFF
--- a/internal/foreman/controller/agentictask_controller.go
+++ b/internal/foreman/controller/agentictask_controller.go
@@ -356,7 +356,7 @@ func (r *AgenticTaskReconciler) reserveFirstFitNode(ctx context.Context, task *f
 // clear cannot wedge a node busy forever.
 func (r *AgenticTaskReconciler) reserveNode(ctx context.Context, node *foremanv1alpha1.FleetNode, taskKey string) (bool, error) {
 	if cur := node.Status.CurrentTask; cur != "" && cur != taskKey {
-		live, err := r.taskIsLive(ctx, cur)
+		live, err := r.taskIsLive(ctx, cur, node.Name)
 		if err != nil {
 			return false, err
 		}
@@ -367,18 +367,25 @@ func (r *AgenticTaskReconciler) reserveNode(ctx context.Context, node *foremanv1
 	patch := client.MergeFromWithOptions(node.DeepCopy(), client.MergeFromWithOptimisticLock{})
 	node.Status.CurrentTask = taskKey
 	if err := r.Status().Patch(ctx, node, patch); err != nil {
-		if apierrors.IsConflict(err) {
-			return false, nil // another reconcile reserved it first
+		// Lost the race: another reconcile reserved it first (Conflict), or the
+		// node was deleted between the List and here (NotFound — e.g. the
+		// Draining-node reaper, #980/#979). Either way, fall through to the next
+		// candidate rather than erroring the whole reconcile.
+		if apierrors.IsConflict(err) || apierrors.IsNotFound(err) {
+			return false, nil
 		}
 		return false, err
 	}
 	return true, nil
 }
 
-// taskIsLive reports whether the namespaced-name key refers to an AgenticTask
-// that still exists and has not reached a terminal phase. A missing or terminal
-// task is not live, so a node whose CurrentTask points at it may be reclaimed.
-func (r *AgenticTaskReconciler) taskIsLive(ctx context.Context, key string) (bool, error) {
+// taskIsLive reports whether the namespaced-name key still occupies nodeName:
+// the task exists, has not reached a terminal phase, and has not been assigned
+// to a different node. A missing, terminal, or reassigned task is not live, so
+// a node whose CurrentTask points at it may be reclaimed. An empty AssignedNode
+// counts as live — that is the legitimate window between reserveNode and
+// scheduleToNode, which must not be stolen.
+func (r *AgenticTaskReconciler) taskIsLive(ctx context.Context, key, nodeName string) (bool, error) {
 	ns, name, ok := splitNamespacedName(key)
 	if !ok {
 		return false, nil // unparseable key: treat as not live so the node frees
@@ -390,8 +397,17 @@ func (r *AgenticTaskReconciler) taskIsLive(ctx context.Context, key string) (boo
 		}
 		return false, err
 	}
-	return t.Status.Phase != foremanv1alpha1.AgenticTaskPhaseSucceeded &&
-		t.Status.Phase != foremanv1alpha1.AgenticTaskPhaseFailed, nil
+	if t.Status.Phase == foremanv1alpha1.AgenticTaskPhaseSucceeded ||
+		t.Status.Phase == foremanv1alpha1.AgenticTaskPhaseFailed {
+		return false, nil
+	}
+	// A live task that has been scheduled onto a different node no longer holds
+	// this one; the reservation is stale (e.g. a clear that failed log-and-
+	// continue while the task rescheduled elsewhere). Reclaim it.
+	if an := t.Status.AssignedNode; an != "" && an != nodeName {
+		return false, nil
+	}
+	return true, nil
 }
 
 // clearNodeCurrentTask releases the named FleetNode's reservation, but only if
@@ -408,7 +424,12 @@ func (r *AgenticTaskReconciler) clearNodeCurrentTask(ctx context.Context, nodeNa
 	if node.Status.CurrentTask != taskKey {
 		return nil
 	}
-	patch := client.MergeFrom(node.DeepCopy())
+	// Optimistic lock for symmetry with reserveNode: between the Get above and
+	// this patch a concurrent reserve for a different task could land, and an
+	// unlocked merge would stomp that fresh reservation (the double-book class
+	// this fix exists to prevent). On Conflict the caller logs-and-continues and
+	// the stale-reclaim path in reserveNode self-heals.
+	patch := client.MergeFromWithOptions(node.DeepCopy(), client.MergeFromWithOptimisticLock{})
 	node.Status.CurrentTask = ""
 	return r.Status().Patch(ctx, &node, patch)
 }
--- a/internal/foreman/controller/agentictask_controller.go
+++ b/internal/foreman/controller/agentictask_controller.go
@@ -22,6 +22,7 @@ import (
 	"slices"
 	"sort"
 	"strconv"
+	"strings"
 	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -115,6 +116,13 @@ func (r *AgenticTaskReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		if err := audit.RecordTerminal(ctx, r.Client, &task, r.AuditNamespace, log); err != nil {
 			log.Error(err, "audit: failed to record terminal run (continuing)")
 		}
+		// Release the node reservation so the scheduler can dispatch the next
+		// task there. Guarded on taskKey, so a node already reserved for a
+		// different task is untouched.
+		if err := r.clearNodeCurrentTask(ctx, task.Status.AssignedNode, taskKey(&task)); err != nil {
+			log.Error(err, "failed to release node reservation on terminal task",
+				"node", task.Status.AssignedNode, "task", task.Name)
+		}
 		return ctrl.Result{}, nil
 	}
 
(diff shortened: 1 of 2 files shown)


internal/foreman/controller/agentictask_controller_test.go
```diff
@@ -345,6 +345,158 @@ var _ = Describe("AgenticTaskReconciler scheduler", func() {
...
+		Expect(k8sClient.Create(ctx, t1)).To(Succeed())
+		DeferCleanup(func() { _ = k8sClient.Delete(ctx, t1) })
+		setPhase(t1, foremanv1alpha1.AgenticTaskPhasePending)
+		t2 := newTask("solo-t2")
+		Expect(k8sClient.Create(ctx, t2)).To(Succeed())
+		DeferCleanup(func() { _ = k8sClient.Delete(ctx, t2) })
+		setPhase(t2, foremanv1alpha1.AgenticTaskPhasePending)
+
+		_, err := reconciler.Reconcile(ctx, reqFor(t1))
+		Expect(err).NotTo(HaveOccurred())
```
Defilan, line 416: **Bug:** This spec can pass vacuously. If capability matching (or reservation itself) regressed so that nothing scheduled at all, t2 would still requeue with Pending and an empty AssignedNode, and every assertion here would hold. Anchor the precondition that t1 actually holds the node:

```suggestion
		Expect(err).NotTo(HaveOccurred())

		// Anchor the precondition: t1 must actually hold the node, otherwise
		// this spec passes vacuously when nothing schedules at all.
		var busy foremanv1alpha1.FleetNode
		Expect(k8sClient.Get(ctx, types.NamespacedName{Name: node.Name}, &busy)).To(Succeed())
		Expect(busy.Status.CurrentTask).To(Equal("default/solo-t1"))
```

internal/foreman/controller/agentictask_controller.go
```diff
@@ -294,22 +323,110 @@ func (r *AgenticTaskReconciler) firstFitNode(ctx context.Context, required forem
...
+		if live {
+			return false, nil // genuinely busy
+		}
+	}
+	patch := client.MergeFromWithOptions(node.DeepCopy(), client.MergeFromWithOptimisticLock{})
+	node.Status.CurrentTask = taskKey
+	if err := r.Status().Patch(ctx, node, patch); err != nil {
+		if apierrors.IsConflict(err) {
+			return false, nil // another reconcile reserved it first
+		}
```
Defilan: **Suggestion:** With #980 in flight (which deletes orphaned Draining FleetNodes), a node can vanish between the List in reserveFirstFitNode and this patch. That surfaces as NotFound, which is not IsConflict, so the whole reconcile errors instead of falling through to the next candidate. Retried and benign, but treating it like losing the race is cleaner:

```suggestion
		if apierrors.IsConflict(err) || apierrors.IsNotFound(err) {
			return false, nil // lost the race: reserved by another reconcile, or the node was deleted
		}
```

internal/foreman/controller/agentictask_controller.go
```diff
@@ -294,22 +323,110 @@ func (r *AgenticTaskReconciler) firstFitNode(ctx context.Context, required forem
...
+		return nil
+	}
+	var node foremanv1alpha1.FleetNode
+	if err := r.Get(ctx, types.NamespacedName{Name: nodeName}, &node); err != nil {
+		return client.IgnoreNotFound(err)
+	}
+	if node.Status.CurrentTask != taskKey {
+		return nil
+	}
+	patch := client.MergeFrom(node.DeepCopy())
```
Defilan: **Suggestion:** reserveNode takes the optimistic lock but this clear does not, and the guard here is read-then-patch: between the Get (which sees CurrentTask == taskKey) and this patch, a concurrent reserve for a different task can land, and the unlocked merge patch would stomp that fresh reservation; the node then looks free while the new task runs there, which is exactly the double-book class this PR fixes. Impossible today with MaxConcurrentReconciles at 1, but your own design note says the lock exists for the day that changes, so the clear path should match:

```suggestion
	patch := client.MergeFromWithOptions(node.DeepCopy(), client.MergeFromWithOptimisticLock{})
```

Callers already log-and-continue on clear errors, and a Conflict-dropped clear self-heals via your stale-reclaim path, so no caller changes needed. The cost is an occasional spurious Conflict from a heartbeat patch bumping the resourceVersion; I think that trade is worth it, but tell me if you disagree.

internal/foreman/controller/agentictask_controller.go
```diff
@@ -294,22 +323,110 @@ func (r *AgenticTaskReconciler) firstFitNode(ctx context.Context, required forem
...
+// (try the next node) when the node is already running another live task or
+// when a concurrent writer won the node first (Conflict).
+//
+// A CurrentTask that points at a task which no longer exists or has reached a
+// terminal phase is stale: the node is treated as free and re-reserved. That
+// self-heals a reservation leaked by a task deleted mid-flight, so a missed
+// clear cannot wedge a node busy forever.
+func (r *AgenticTaskReconciler) reserveNode(ctx context.Context, node *foremanv1alpha1.FleetNode, taskKey string) (bool, error) {
+	if cur := node.Status.CurrentTask; cur != "" && cur != taskKey {
+		live, err := r.taskIsLive(ctx, cur)
```
Defilan: **Suggestion:** taskIsLive checks only the phase, not whether the task still points back at this node. A reservation referencing a live task whose AssignedNode is a *different* node stays "genuinely busy" until that task terminates. You can get there whenever one of the (log-and-continue) clear paths fails: e.g. releaseExpiredClaim releases the task but its clear patch fails, the task reschedules elsewhere, and when this node's agent revives it sits wedged busy for the full duration of a task it never ran. Bounded, but easy to close: pass node.Name in and treat the reservation as stale when the live task's AssignedNode is non-empty and differs from this node. Keep treating an empty AssignedNode as live; that protects the legitimate window between reserveNode and scheduleToNode. Non-blocking; fine as a follow-up if you prefer.

Defilan (changes requested):
Thanks for this one; it is an unusually well-built external contribution. The bug is real (CurrentTask was declared, read, and logged but never written, so everything funneled onto one node), the optimistic-lock reservation is the right tool for the cache-lag race, and I traced the guard chain in Reconcile: the scheduling block is only reachable for Pending tasks, and re-reservation of the same node is idempotent, so there is no double-reserve in the normal path. I also checked the two other FleetNode.Status writers (agent heartbeat and the FleetNodeReconciler phase flip); both use field-scoped merge patches, so your reservation survives them. And I checked the interaction with #980 (Draining-node reaping): claim expiry at 90s fires long before the 10m reap, and clearNodeCurrentTask already tolerates a deleted node, so that composes cleanly.

Requesting one small change before merge: the busy-wait spec never asserts that t1 actually reserved solo-node, so it can pass vacuously if nothing schedules at all; one assertion fixes it. I have also left three non-blocking hardening suggestions inline (NotFound tolerance in reserveNode given #980, optimistic lock on the clear path for symmetry with your own design note, and a back-pointer check in the liveness test). Happy to merge once the test anchor is in; the design shape you chose is the one I would keep, no need for the FleetNodeReconciler variant.

joryirving:
Thanks for the thorough trace. Addressed all four:

- **Must-fix (busy-spec anchor):** applied your suggestion verbatim — the spec now asserts `solo-node` holds `default/solo-t1` before checking t2 waits, so it can't pass vacuously.
- **reserveNode NotFound tolerance:** `IsConflict || IsNotFound` → fall through to the next candidate (composes with the #979 reaper deleting a node mid-List).
- **clearNodeCurrentTask optimistic lock:** switched to `MergeFromWithOptimisticLock`; callers already log-and-continue and the stale-reclaim path self-heals a Conflict-dropped clear, so no caller changes.
- **taskIsLive back-pointer:** now takes the reserving node's name and treats a live task assigned to a *different* node as stale (reclaimable), while an empty AssignedNode still counts as live to protect the reserve→schedule window.

Green: build, `golangci-lint` (0 issues), full controller suite.

Defilan (approved):
This addresses everything from my review, and thoroughly. The busy-wait spec now anchors on `busy.Status.CurrentTask == "default/solo-t1"` before asserting t2 waits, so it can no longer pass vacuously; that was the one blocker and it's cleanly resolved.

You also took all three suggestions and implemented each one correctly: `reserveNode` tolerating NotFound keeps a mid-scan node deletion from erroring the whole reconcile; the optimistic-lock clear can't stomp a concurrent reservation (and I verified the stale-reclaim path still self-heals a Conflict-failed clear, exactly as your comment describes); and the `taskIsLive` back-pointer check reclaims reassigned reservations while keeping the empty-AssignedNode carve-out that protects the reserve-to-schedule window. The doc comments you added explaining the race reasoning are genuinely valuable for whoever touches this next.

Thanks for the fast, careful turnaround. Great contribution, happy to merge once CI finishes green.

PR #980, opened 2026-07-05
fix(foreman): reap orphaned Draining FleetNodes so they stop leaking

### LINKED ISSUE #979: Orphaned Draining FleetNodes never cleaned up: one leaks per agent rollout
## Summary

Orphaned `Draining` FleetNodes are never cleaned up: every `foreman-agent` rollout, scale-down, or crash leaves one behind, and they accumulate indefinitely.

## Root cause

When an agent shuts down it marks its FleetNode `Draining`. The `FleetNodeReconciler` (`internal/foreman/controller/fleetnode_controller.go:61`) then deliberately does nothing for a Draining node:

```go
// A node the agent is intentionally draining is the agent's domain;
// the scheduler already treats Draining as ineligible. Don't fight it.
if node.Status.Phase == foremanv1alpha1.FleetNodePhaseDraining {
    return ctrl.Result{RequeueAfter: FleetNodeHeartbeatTimeout}, nil
}
```

If the pod then terminates before deleting its own FleetNode (rolling update, scale-down, SIGKILL, crash), nothing ever removes it:

- The reconciler skips all Draining nodes regardless of heartbeat age.
- FleetNodes carry no `ownerReference` to the pod, so there is no Kubernetes garbage collection (confirmed: `ownerReferences: []`, `finalizers: []`).

So the node stays `Draining` with a heartbeat that never advances. One leaks per agent restart.

## Observed (v0.8.28)

23 `Draining` FleetNodes accumulated across ~11 old ReplicaSets, heartbeats days stale, against 3 live `Ready` nodes. The scheduler ignores them (not Ready), so they are harmless to scheduling — but they clutter `kubectl get fleetnodes` and grow without bound.

## Suggested fix

Reap a Draining node whose heartbeat has been silent past a drain grace window (well above the heartbeat timeout, so a live agent draining its in-flight task — which keeps heartbeating — is never reaped). An `ownerReference` to the pod would also work via native GC, but requires agent-side plumbing (pod UID at registration) and does not cover a node created before that change. The reconciler-side reap is self-contained. PR incoming.

---

joryirving:
## What

Delete orphaned `Draining` FleetNodes so they stop accumulating — one otherwise leaks per `foreman-agent` rollout, scale-down, or crash.

## Why

Fixes #979

An agent marks its FleetNode `Draining` on shutdown. The `FleetNodeReconciler` deliberately leaves Draining nodes alone ("the agent's domain"), and FleetNodes carry no `ownerReference`, so there is no Kubernetes GC. If the pod terminates before deleting its own node, the node stays `Draining` with a frozen heartbeat forever. Observed on v0.8.28: 23 orphaned Draining nodes across ~11 old ReplicaSets against 3 live nodes.

## How

- New `FleetNodeDrainReapTimeout` (10m) and `FleetNode.DrainReapable(now)`.
- In the reconciler's Draining branch, delete a node that has been silent past the reap timeout (its agent is gone and its drain will never complete). The window sits well above the 90s heartbeat timeout, so a live agent draining its in-flight task — which keeps heartbeating — is never reaped. A live-but-draining node still short-circuits with a requeue, unchanged.

No API/CRD schema change (added a const and a method, not a field).

## Checklist

- [x] Tests added/updated (2 envtest cases: long-stale Draining deleted; fresh Draining left intact)
- [x] `make test` passes locally (controller + api suites)
- [x] `make lint` passes locally (`golangci-lint`: 0 issues)
- [x] AI assistance is disclosed (commit trailer + here), per CONTRIBUTING.md
- [ ] Documentation updated (no user-facing doc change)

### DIFF
--- a/api/foreman/v1alpha1/fleetnode_types.go
+++ b/api/foreman/v1alpha1/fleetnode_types.go
@@ -240,15 +240,20 @@ func (n *FleetNode) HeartbeatStale(now time.Time) bool {
 }
 
 // DrainReapable reports whether a Draining node has been silent long enough
-// (past FleetNodeDrainReapTimeout, or never heart-beat) that its agent is gone
-// and the drain will never complete, so the node may be deleted. The caller is
-// responsible for checking that the node is actually in the Draining phase.
+// (past FleetNodeDrainReapTimeout) that its agent is gone and the drain will
+// never complete, so the node may be deleted. The caller is responsible for
+// checking that the node is actually in the Draining phase.
+//
+// A node with no heartbeat at all falls back to its creation time, so even a
+// hand-crafted Draining object gets the same grace window rather than being
+// reaped on first sight (the agent always stamps LastHeartbeatTime in the same
+// patch that sets Draining, so this branch is unreachable via the agent).
 func (n *FleetNode) DrainReapable(now time.Time) bool {
 	if n == nil {
 		return false
 	}
 	if n.Status.LastHeartbeatTime == nil {
-		return true
+		return now.Sub(n.CreationTimestamp.Time) > FleetNodeDrainReapTimeout
 	}
 	return now.Sub(n.Status.LastHeartbeatTime.Time) > FleetNodeDrainReapTimeout
 }
--- a/api/foreman/v1alpha1/fleetnode_types.go
+++ b/api/foreman/v1alpha1/fleetnode_types.go
@@ -28,6 +28,16 @@ import (
 // declared dead.
 const FleetNodeHeartbeatTimeout = 90 * time.Second
 
+// FleetNodeDrainReapTimeout is how long a node may stay Draining without a
+// heartbeat before the controller deletes it. A live agent keeps heartbeating
+// while it finishes its in-flight task, so this only fires for a node whose
+// agent set Draining and then went away (rollout, scale-down, crash) — which
+// otherwise leaks a FleetNode per restart, since the reconciler leaves
+// Draining nodes alone and there is no ownerReference for garbage collection.
+// Set well above the heartbeat timeout so a genuinely draining agent is never
+// reaped out from under an in-flight task.
+const FleetNodeDrainReapTimeout = 10 * time.Minute
+
 // FleetNodePhase is the heartbeat-driven health state of a fleet worker.
 // +kubebuilder:validation:Enum=Ready;Draining;NotReady;Unknown
 type FleetNodePhase string
@@ -227,6 +237,20 @@ func (n *FleetNode) HeartbeatStale(now time.Time) bool {
 		return true
 	}
 	return now.Sub(n.Status.LastHeartbeatTime.Time) > FleetNodeHeartbeatTimeout
+}
+
+// DrainReapable reports whether a Draining node has been silent long enough
+// (past FleetNodeDrainReapTimeout, or never heart-beat) that its agent is gone
+// and the drain will never complete, so the node may be deleted. The caller is
+// responsible for checking that the node is actually in the Draining phase.
+func (n *FleetNode) DrainReapable(now time.Time) bool {
+	if n == nil {
+		return false
+	}
+	if n.Status.LastHeartbeatTime == nil {
+		return true
+	}
+	return now.Sub(n.Status.LastHeartbeatTime.Time) > FleetNodeDrainReapTimeout
 }
 
 // +kubebuilder:object:root=true
--- a/internal/foreman/controller/fleetnode_controller.go
+++ b/internal/foreman/controller/fleetnode_controller.go
@@ -58,9 +58,23 @@ func (r *FleetNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		"currentTask", node.Status.CurrentTask,
 	)
 
-	// A node the agent is intentionally draining is the agent's domain;
-	// the scheduler already treats Draining as ineligible. Don't fight it.
+	// A node the agent is intentionally draining is the agent's domain; the
+	// scheduler already treats Draining as ineligible, so while the agent is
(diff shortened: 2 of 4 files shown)


internal/foreman/controller/fleetnode_controller.go
```diff
@@ -58,9 +58,23 @@ func (r *FleetNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	// Draining and then went away — rollout, scale-down, crash — leaves the
+	// node Draining forever: nothing here transitions it and there is no
+	// ownerReference for garbage collection, so one FleetNode leaks per agent
+	// restart. Reap a Draining node whose heartbeat has been silent past the
+	// drain grace; its drain will never complete.
 	if node.Status.Phase == foremanv1alpha1.FleetNodePhaseDraining {
+		if node.DrainReapable(time.Now()) {
+			log.Info("reaping orphaned Draining FleetNode (agent gone)",
+				"nodeName", node.Spec.NodeName, "lastHeartbeat", node.Status.LastHeartbeatTime)
+			if err := r.Delete(ctx, &node); err != nil {
```
Defilan: **Bug:** Revival race on stable-named nodes. In-cluster agents create a fresh per-pod FleetNode, but launchd-managed agents reuse a stable NodeName and run `Upsert` only once at startup. If such an agent is down >10m while Draining and then restarts, its initial `PatchHeartbeat(Ready)` revives the node; this reconciler reads from the informer cache, which can lag that write by seconds, so it can still see the stale Draining view and delete the freshly revived node. The agent then fails every subsequent heartbeat (Get -> NotFound, "will retry") and never re-creates the node, silently dropping a live worker from the fleet until the process restarts.

Deleting with preconditions closes the window: if anything touched the node since our read, the delete 409s, the Conflict propagates as the returned error, and backoff retries with a fresh view (a truly dead node's ResourceVersion never advances, so the reap still converges).

```suggestion
			if err := r.Delete(ctx, &node, client.Preconditions{
				UID:             &node.UID,
				ResourceVersion: &node.ResourceVersion,
			}); err != nil {
```

If you'd rather not surface the Conflict as an error, special-casing `apierrors.IsConflict` to a quiet `RequeueAfter` also works; either is fine by me.

api/foreman/v1alpha1/fleetnode_types.go
```diff
@@ -229,6 +239,20 @@ func (n *FleetNode) HeartbeatStale(now time.Time) bool {
...
+// (past FleetNodeDrainReapTimeout, or never heart-beat) that its agent is gone
+// and the drain will never complete, so the node may be deleted. The caller is
+// responsible for checking that the node is actually in the Draining phase.
+func (n *FleetNode) DrainReapable(now time.Time) bool {
+	if n == nil {
+		return false
+	}
+	if n.Status.LastHeartbeatTime == nil {
+		return true
+	}
```
Defilan, line 257: **Suggestion:** (non-blocking) The nil branch is unreachable via the agent: `PatchHeartbeat` stamps `LastHeartbeatTime` in the same status patch that sets Draining, so only a hand-crafted object can be Draining with no heartbeat. Reaping that on first sight is defensible, but anchoring the nil case to `creationTimestamp` gives even a hand-set node the same 10m grace for free and keeps the "silent past the reap timeout" contract uniform:

```suggestion
	if n.Status.LastHeartbeatTime == nil {
		return now.Sub(n.CreationTimestamp.Time) > FleetNodeDrainReapTimeout
	}
```

If you take this, a third small spec covering the nil-heartbeat branch would round out the coverage.

Defilan (changes requested):
Thanks for this, and for the outstanding diagnosis in #979; the root cause, the evidence, and the choice of a reconciler-side reap over an ownerReference are all exactly right. I traced the interactions you'd worry about and they hold up: the claim-expiry ladder releases an in-flight task ~90s after silence (well before the 10m reap), a NotFound node is already handled as an expired claim, and this composes cleanly with #978 in either merge order. One must-fix before merge: the delete needs UID/ResourceVersion preconditions. In-cluster agents get per-pod node names, but launchd-managed agents reuse a stable NodeName and only Upsert at startup, so a cached Draining read can race a revival heartbeat and delete a live node whose agent then fails every heartbeat until restart. Details inline; the rest is optional. Tests assert real behavior on both sides, nicely done.

joryirving:
Thanks — both addressed:

- **Must-fix (revival race):** delete now carries `Preconditions{UID, ResourceVersion}`. On Conflict I requeue after the heartbeat timeout to re-evaluate on a fresh view (rather than surfacing an error) — a revived node reads Ready and is spared; a truly dead node's ResourceVersion never advances so the reap still converges.
- **creationTimestamp grace:** `DrainReapable` falls back to `CreationTimestamp` when `LastHeartbeatTime` is nil, so a hand-set Draining object gets the full grace window instead of instant reap.
- Added `DrainReapable` unit tests covering fresh/stale heartbeat + the nil-heartbeat recent/old-creation branches.

Green: build, `golangci-lint` (0 issues), controller + api suites.

Defilan (approved):
This is exactly right, thank you @joryirving. The precondition delete closes the revival race cleanly: a heartbeat that lands after our stale read bumps the ResourceVersion, the delete 409s, and the requeue re-evaluates on a fresh view where the revived node shows Ready and gets spared; a genuinely dead node's ResourceVersion never moves, so the reap still converges. I also appreciate that the Conflict path requeues instead of surfacing an error, and that IgnoreNotFound still covers the already-deleted case.

The creationTimestamp fallback in `DrainReapable` plus the four-case table test (and the nil-receiver check) round it out nicely; the comment explaining why that branch is unreachable via the agent is a great touch for future readers.

LGTM, happy to merge once CI goes green.

Issue #921, opened 2026-07-01
[FEATURE] Unify model cache-key derivation across controller and CLI


Defilan:
## Feature Description

There are now three sites that decide a model's cache key, and they should share one function so serve / `cache list` / `delete --purge-cache` can never disagree:

- **serve** (`internal/controller`): `effectiveModelCacheKey(model)` — the source of truth. Returns `Status.CacheKey`, else `computeCacheKey(source)` for non-metal multi-file models, else empty.
- **`cache list`** (`pkg/cli/cache.go`): `Status.CacheKey` else `computeCacheKey(source)` — unconditional fallback.
- **`delete --purge-cache`** (`pkg/cli/delete.go`): same unconditional fallback (added in #912).

The two CLI sites use an *unscoped* fallback (any empty-key model gets a derived key), while the controller scopes it to non-metal multi-file models. So the CLI can show/purge a phantom key for models that were never cached (metal, single-file runtime-resolved). Harmless today (advisory prints, guarded against a bare `/models/` wipe) but inconsistent.

## Proposed Solution

Lift `effectiveModelCacheKey` (and the `hasMultiFileStaging` / metal checks it needs) into a package importable by both `internal/controller` and `pkg/cli`, and route all three sites through it. `computeCacheKey` is currently duplicated in `internal/controller` and `pkg/cli` — fold that into the same shared home.

## Additional Context

Raised in the #912 review. #912 shipped the minimal delete.go fallback so serve/list/purge agree for the cached models; this issue tracks the proper single-source-of-truth refactor.

PR #983, opened 2026-07-06
fix(foreman): key test-presence gate on net-new functions only

Defilan:
## What

The coder gate's test-presence check (mutation gate Layer 1) now demands a named test only for net-new functions (+func lines in the diff). Body-modified functions no longer trigger the demand.

## Why

The check's own doc comment states the intent: key on net-new functions "so behavior-preserving edits to existing funcs do not require a new test." The implementation additionally unioned in every function whose body a hunk touched. On refactor-class issues this demands one named test per touched call site, including in envtest-backed packages whose tests the coder workspace cannot run. Live evidence: the #921 coder run burned all 3 gate attempts (112 turns) on exactly this, with the gate demanding tests for 8 call-site functions across internal/controller/ and pkg/cli/.

Coverage for body-modified functions is unchanged where it is enforceable: Layer 2 (mutation survival) still neuters any function intersecting changed lines when the package has a changed test, and CI runs the full suite on every PR.

## How

- checkTestPresence builds its demand set from addedFuncNames only; modifiedFuncNames is no longer unioned in.
- modifiedFuncNames itself stays (the caller-impact gate consumes it).
- Package and function doc comments updated to match.
- TestCheckTestPresence_ModifiedFuncNeedsTest rewritten as TestCheckTestPresence_ModifiedFuncDoesNotNeedTest, pinning the corrected contract.

## Checklist

- [x] Tests added/updated
- [x] No API/CRD changes

## AI assistance disclosure

Assisted-by: Claude Code (Claude Opus 4.8). Generated the implementation, tests, and PR text from a design I approved; I reviewed the diff, ran the gate locally (go test ./pkg/foreman/agent/..., make lint on darwin and GOOS=linux), and validated the change live on the lab fleet before submitting.

### DIFF
--- a/pkg/foreman/agent/mutation_gate.go
+++ b/pkg/foreman/agent/mutation_gate.go
@@ -18,11 +18,13 @@ limitations under the License.
 // changes which pass the syntactic checks (gofmt/vet/build/lint) but are not
 // actually constrained by a test (the #856 class).
 //
-//   - Layer 1 (checkTestPresence): for each changed package, every new or
-//     body-modified function in hand-written, non-test Go must be referenced
-//     by name in a changed _test.go in that package; unreferenced functions
-//     fail. Pure diff/text inspection, so it covers envtest/controller
-//     packages too.
+//   - Layer 1 (checkTestPresence): for each changed package, every NET-NEW
+//     function in hand-written, non-test Go must be referenced by name in a
+//     changed _test.go in that package; unreferenced new functions fail.
+//     Body-modified functions are exempt (behavior-preserving edits do not
+//     require a new test); they are covered by Layer 2 where package tests
+//     exist and by CI's full suite everywhere. Pure diff/text inspection, so
+//     it covers envtest/controller packages too.
 //   - Layer 2 (checkMutationSurvival): for non-envtest changed packages that DO
 //     have a changed test, blank the changed function bodies on an in-memory
 //     backup, re-run the package tests, and flag any package whose tests still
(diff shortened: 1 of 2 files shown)


Defilan:
**Live validation from last night's lab run.**

Deployed this PR (combined with #984 on an integration branch) to the lab fleet and re-dispatched the exact issue that motivated it: #921, the cache-key refactor that had failed six consecutive coder runs across two models, every one of them dying in the test-presence gate.

Result, same model (Ornith-35B), same repo map, same 3-attempt budget:

| | before (0.9.0 gate) | with this PR |
|---|---|---|
| turns | 112 | 55 |
| gate attempts | 3, all failed on test presence | passed first attempt |
| outcome | INCOMPLETE | GO + GATE-PASS |

The run produced the fix #921 actually asks for: a new shared `pkg/cachekey` package with its own tests, controller and all four CLI sites routed through it. The net-new function correctly still demanded a test (and got one); the eight body-modified call sites stopped demanding tests the workspace can't run. Gate output went from three walls of unmeetable demands to a single caller-impact advisory.

So the failure mode this PR targets is confirmed gone in production, not just in the unit tests. Coder branch from the run: `foreman/run-20260705-232312/issue-921` if you want to inspect what the gate now lets through.

joryirving (approved):
**Approve.** Clean, well-motivated change. The fix correctly narrows Layer 1's demand set to net-new functions only, and the test is rewritten to pin the corrected contract.

**Verified:**

- **`modifiedFuncNames` is not orphaned.** Still called by `caller_impact_gate.go:67` — the function is retained correctly, only `checkTestPresence` stops consuming it.
- **`addedFuncNames` parsing logic unchanged.** Still matches `^\+\s*func\b[^/]*\(` against `git diff HEAD -- file` output. The test's `diffOut` has no `+func` line → returns empty → `changedFuncs` stays empty → gate passes. Correct.
- **Layer 2 (mutation survival) is unaffected.** `checkMutationSurvival` independently calls `changedNewLines` and `modifiedFuncNames` — neither touched by this PR. The claim that modified functions remain covered by Layer 2 where package tests exist is accurate.
- **Test mock loosening is safe.** The old mock matched `args[1]=="-U0" && args[2]=="HEAD"`; the new one matches `args[1]=="-U0"`. All `git diff` calls in the file use `"HEAD"` as the second arg after `"diff"` (either `diff HEAD -- file` or `diff -U0 HEAD -- file`), so the looser pattern can't mismatch a different call path.
- **Existing `TestCheckTestPresence` (line 70) still covers the net-new path** — verifies a `+func` line with no test reference fails. The contract is maintained for the case that matters.
- **DCO ✅**, conventional commit ✅ (`fix(foreman):`), AI disclosure ✅, no CRD changes, no API changes.

**Non-blocking nit:** The doc comments and inline comments are verbose relative to the code change — the #921 incident context is useful but repeated across package doc, function doc, and inline comment. Not worth a revision cycle.

Issue #982, opened 2026-07-05
Coder self-commit / bash edits discarded: false 'no diff' NO-GO and EditFreeStreak kills


joryirving:
## Summary

When a coder self-commits its edits (runs `git commit` via the bash tool) or edits files through bash instead of the edit tools, the executor discards the work: it either reports "model emitted GO but produced no diff" (false NO-GO) or force-terminates the task as a stuck loop. In a real fleet this was ~64% of non-GO coder outcomes.

## Evidence (v0.8.28, live fleet)

Coder GO-yield ~27% (12 GO / 45 completed issue-fix tasks). Of the 33 non-GO:
- **13 INCOMPLETE — all "stuck-loop: EditFreeStreak"**
- **8 NO-GO — "model emitted GO but produced no diff"**
- (remainder: 4 push/fork-auth, 4 gate fails, 4 genuinely already-resolved)

The 21 above share one root cause: the model works through bash rather than the executor's contract (edit via `str_replace`/`write_file`, leave changes uncommitted for the executor to commit + DCO-sign + push). Example — a task that made real edits and committed them itself:

> summary: "…Local commit 75bfb9f on issue-173 branch ready to push." — `submit_result` called 12×, outcome: **no diff**.

## Root cause (two spots)

1. **`repo.HasChanges` is working-tree only.** `pkg/foreman/agent/repo/commit.go:HasChanges` runs `git status --porcelain`. `executor_native.go` (`runLLMPath`) gates the GO path on it (`if !hasChanges → noChangesResult`), then commits the working tree itself. If the model already ran `git commit`, the working tree is clean → `HasChanges` returns false → false "no diff", even though the branch has real commits ahead of base. (`repo.DiffNameOnly(base...HEAD)` *would* see them — it's used for the scope check right nearby.)

2. **The EditFreeStreak detector only counts `str_replace`/`write_file` as edits** (`progress_896_test.go`). A model editing via bash (`sed`/`echo`/`git apply`) never resets the streak → force-terminated INCOMPLETE mid-edit.

By model: base-tier (Qwen-27B) skews to false-GO-no-diff (6/13); frontier (MiniMax-M3) skews to stuck-loop (11/32).

## Suggested fix

[...]

PR #984, opened 2026-07-06
feat(foreman): append actionable steers to structural-lint gate feedback

Defilan:
## What

The coder gate's failure feedback gains deterministic one-line steers for structural lint classes: gocyclo (named function, "extract a helper"), dupl, and funlen. Unknown failure classes are untouched.

## Why

A raw linter dump is not actionable for mid-size coder models. Live evidence: on the #982 run the coder implemented a correct fix, then re-submitted the same over-complex function on all 3 gate attempts because the only feedback was "cyclomatic complexity 31 of func runLLMPath is high (> 30)". Telling the model the mechanical remedy (do not add branches; extract a named helper) is a steer this model class demonstrably follows.

## How

- New package-private lintAdvisories(output) []string: regex-captures gocyclo function names (deduped), substring-detects (dupl) and (funlen), caps at 3 steers per failing check, prefixes each with "Advice:".
- buildFeedback appends the steers under each failing check's section.
- Table-driven tests: gocyclo steer content and function naming, no advisory for non-structural failures, dupl+funlen coverage, dedupe and cap.

## Checklist

- [x] Tests added/updated
- [x] No API/CRD changes

## AI assistance disclosure

Assisted-by: Claude Code (Claude Opus 4.8). Generated the implementation, tests, and PR text from a design I approved; I reviewed the diff, ran the gate locally (go test ./pkg/foreman/agent/..., make lint on darwin and GOOS=linux), and validated the change live on the lab fleet before submitting.

### DIFF
--- a/pkg/foreman/agent/coder_gate.go
+++ b/pkg/foreman/agent/coder_gate.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -503,6 +504,10 @@ func checkGoreleaserConfig(ctx context.Context, workspace string, run commandRun
 
 // buildFeedback renders the directive and a per-check section for every
 // failing check, truncating each check's output to maxCheckOutputBytes.
+// Structural-lint failures additionally get a one-line steer (see
+// lintAdvisories): the raw linter dump alone is not actionable for
+// mid-size coders, which re-submit the same shape until the attempt
+// budget is gone (#982 run: gocyclo 31 > 30, three identical attempts).
 func buildFeedback(failures []checkFailure) string {
 	var b strings.Builder
 	b.WriteString("The verification gate failed. Fix the issues below and resubmit.\n")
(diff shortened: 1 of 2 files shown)


joryirving (approved):
**Approve.** Targeted, well-tested, motivated by real failure evidence.

**Verified:**

- **`gocycloFuncRe` correctly captures the function name** (backticked in the linter output). Verified the regex against the exact test fixture `"cyclomatic complexity 31 of func `(*NativeAgentLoopExecutor).runLLMPath` is high (> 30) (gocyclo)"` — captures `(*NativeAgentLoopExecutor).runLLMPath` cleanly, including method receivers.
- **Dedup is correct.** `seen` set keyed by func name means repeated gocyclo lines for the same func across packages don't emit duplicate steers. Test `TestLintAdvisories_DedupesAndCaps` covers both dedup (same func twice → 1 steer) and cap (4 distinct funcs → capped at 3).
- **The cap is global, not per-class.** `if len(steers) > maxLintAdvisories { steers = steers[:maxLintAdvisories] }` truncates the slice after all classes appended. If a single check has 3 gocyclo funcs *and* `(dupl)` *and* `(funlen)`, the (funlen) steer gets dropped. Probably fine — gocyclo is the dominant signal in the #982 evidence — but worth knowing. Not a blocker.
- **`buildFeedback` integration is call-site-only.** New `for _, steer := range lintAdvisories(f.output)` loop inside the existing per-failure loop. The truncation via `truncateOutput(f.output)` runs on the *raw* output before steers, so a sprawling lint report can't bloat the steer section.
- **The (dupl)/(funlen) substring matches are safe in practice.** Those tokens are golangci-lint linter tags. Other failing checks (`gofmt -l .`, `go vet ./...`, `go build ./...`, `go test ...`, `codegen drift`, `goreleaser check`, `scope overlap`, `test presence`, `mutation survival`, `reference grounding`) emit well-defined, linter-free output. Test presence output is asserted explicitly in `TestBuildFeedback_NoAdvisoryForUnknownFailure`.
- **Test coverage is table-driven and pins the contract** — gocyclo steer content (`Do not add more branches to...` + `helper function`), the absent-of-advisory path for non-structural failures, dupl+funlen coverage, dedupe, cap. Good.
- **DCO ✅**, conventional commit ✅ (`feat(foreman):`), AI disclosure ✅, no CRD/API changes.

**Non-blocking nit:** The cap-slice on line 561 is silent — if it ever fires in production, the dropped steer is just gone. A `t.Logf` or a counter in the steer message would help, but not worth a revision cycle.

PR #985, opened 2026-07-06
fix(cachekey): unify model cache-key derivation across controller and CLI

### LINKED ISSUE #921: [FEATURE] Unify model cache-key derivation across controller and CLI

Defilan:
## What

Introduces `pkg/cachekey` as the single source of truth for a model's cache key, and routes the controller and all CLI sites through it. Two functions:

- `Compute(source)`: the raw SHA-256[:16] fingerprint (was duplicated in `internal/controller` and `pkg/cli`).
- `EffectiveKey(model)`: the scoped decision. `Status.CacheKey` wins; otherwise only a non-metal, multi-file model derives a key from its source; metal and single-file models return `""`.

The controller's `effectiveModelCacheKey` now delegates to `cachekey.EffectiveKey`; `cache list` and `delete --purge-cache` use it too.

## Why

Fixes #921. There were three sites deciding a model's cache key. The controller scoped the fallback to non-metal multi-file models, but the two CLI sites applied it unconditionally. So `cache list` showed phantom entries and `delete --purge-cache` targeted directories the serving pod never used, for metal and single-file models. Sharing only the raw fingerprint would not fix that; the scoped *decision* has to be shared, which is what `EffectiveKey` does. `cache list` now skips models with no derived key instead of inventing one, and `--purge-cache` no-ops on them.

## How

- New `pkg/cachekey` package: `Compute` (byte-for-byte parity with both prior implementations) and `EffectiveKey` (mirrors the controller's prior scoping semantics exactly).
- `internal/controller`: `effectiveModelCacheKey` delegates to `cachekey.EffectiveKey`; the local `computeCacheKey` becomes a thin wrapper for the many existing call sites. `hasMultiFileStaging` / `isMetalModel` stay in the controller (used widely there for non-cache reasons).
- `pkg/cli`: `cache.go`, `cache_inspect.go`, `delete.go`, `deploy.go` route through the shared package; the duplicate CLI `computeCacheKey` is removed.
- Tests: `pkg/cachekey` covers `Compute` (length, hex, determinism, SHA-256 parity) and `EffectiveKey` (nil, Status wins, multi-file non-metal derives, metal → "", single-file → ""). Existing CLI tests migrated to the shared package.

## Checklist

- [x] Tests added and exercise real behavior (the `EffectiveKey` cases fail if the scoping regresses)
- [x] `make test` / `make lint` pass (darwin and GOOS=linux)
- [x] No API/CRD changes
- [x] Scope matches the issue; no unrelated edits

## AI assistance disclosure

Assisted-by: Foreman (LLMKube's own agentic-coding harness), coder model Ornith-35B running locally on Apple Silicon. Foreman produced the implementation and tests from the issue; I reviewed the first attempt, found it under-delivered on the scoping fix and carried a duplicated doc comment, and fed those two findings back through a Foreman revision cycle that produced this branch. I then reviewed the result, ran the gate locally, and added one doc-comment commit by hand. A human (me) is accountable for the change per CONTRIBUTING.md.

### DIFF
--- a/internal/controller/model_controller.go
+++ b/internal/controller/model_controller.go
@@ -42,6 +42,7 @@ import (
 
 	inferencev1alpha1 "github.com/defilantech/llmkube/api/v1alpha1"
 	llmkubemetrics "github.com/defilantech/llmkube/internal/metrics"
+	"github.com/defilantech/llmkube/pkg/cachekey"
 	"github.com/defilantech/llmkube/pkg/gguf"
 	"github.com/defilantech/llmkube/pkg/license"
 )
@@ -995,9 +996,13 @@ func formatBytes(bytes int64) string {
 	return fmt.Sprintf("%.1f %ciB", float64(bytes)/float64(div), "KMGTPE"[exp])
 }
 
+// computeCacheKey is retained as a thin wrapper around cachekey.Compute for
+// call sites that pre-date the shared package; it exists so the controller
+// package compiles without touching the many `computeCacheKey(source)` sites
+// that only need an unconditional fingerprint. New code should import
+// github.com/defilantech/llmkube/pkg/cachekey directly.
 func computeCacheKey(source string) string {
-	hash := sha256.Sum256([]byte(source))
-	return hex.EncodeToString(hash[:])[:16]
+	return cachekey.Compute(source)
 }
 
 func (r *ModelReconciler) parseGGUFMetadata(path string) (*inferencev1alpha1.GGUFMetadata, error) {
--- a/internal/controller/model_storage.go
+++ b/internal/controller/model_storage.go
@@ -30,6 +30,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	inferencev1alpha1 "github.com/defilantech/llmkube/api/v1alpha1"
+	"github.com/defilantech/llmkube/pkg/cachekey"
 )
 
 // Model storage wiring. The controller has three paths for making a model
@@ -232,31 +233,13 @@ func hasMultiFileStaging(model *inferencev1alpha1.Model) bool {
 	return model != nil && (len(model.Spec.Files) > 0 || model.Spec.Mmproj != "")
 }
 
-// effectiveModelCacheKey returns the key used to namespace a model's files
-// inside the cache PVC, and serves as the single source of truth for whether
-// the in-cluster serving pod should use the cache (non-empty) or an emptyDir
-// (empty). Both the PVC-ensure gate (InferenceService reconcile) and the
-// storage builder MUST consult this so they never disagree about caching.
-//
-// For sources the controller fingerprints (http/https) this is the
-// controller-set Status.CacheKey. The runtime-resolved path (hf:// repo IDs)
-// deliberately leaves Status.CacheKey empty and keeps the Available condition
-// reason "RuntimeResolved", but an hf:// model with multi-file staging still
-// needs the cache PVC: otherwise every file re-downloads into an emptyDir on
-// each pod restart (#909). Derive a stable key from the source in that case.
-// Metal models have no init container and no cache PVC (the host metal-agent
-// loads the files directly), so they never cache here.
+// effectiveModelCacheKey returns the key used to namespace a model's
+// files inside the cache PVC, and is the single source of truth for
+// whether the in-cluster serving pod should use the cache (non-empty)
+// or an emptyDir (empty). It delegates to cachekey.EffectiveKey so the
+// controller and the CLI can never disagree about caching.
 func effectiveModelCacheKey(model *inferencev1alpha1.Model) string {
-	if model == nil {
-		return ""
-	}
-	if model.Status.CacheKey != "" {
-		return model.Status.CacheKey
-	}
-	if hasMultiFileStaging(model) && !isMetalModel(model) {
-		return computeCacheKey(model.Spec.Source)
-	}
-	return ""
+	return cachekey.EffectiveKey(model)
 }
 
 // modelStagingPlan resolves the model's declared files into a staging plan.
--- a/pkg/cachekey/cachekey.go
+++ b/pkg/cachekey/cachekey.go
@@ -19,9 +19,11 @@ limitations under the License.
 // package so `serve`, `cache list`, and `delete --purge-cache` can never
 // disagree about which directory on the cache PVC owns a given model.
 //
-// The controller's effectiveModelCacheKey() additionally scopes the fallback
-// to non-metal multi-file models; that scoping lives in the controller package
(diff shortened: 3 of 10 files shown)


joryirving (approved):
**Approve.** Correct fix for #921 — the scoped decision had to be shared, not just the raw fingerprint. The diff is mechanically right and the doc update commit is a clean follow-up.

**Verified:**

- **`EffectiveKey` is semantically byte-equivalent to the old `effectiveModelCacheKey`.** Old: `hasMultiFileStaging(model) && !isMetalModel(model)`. New: `(len(Files) > 0 || Mmproj != "") && !(Hardware != nil && Accelerator == "metal")`. The nil guard is absorbed by the earlier `model == nil → ""` branch. Same behavior in every observable case.
- **`Compute` parity preserved.** `TestComputeMatchesSHA256Prefix` pins the exact hex prefix against a known URL, and the byte-for-byte logic is identical to both prior implementations. The CLI test suite (`TestComputeCacheKey*`) was migrated to call `cachekey.Compute` — same inputs → same outputs.
- **The bug fix is real.** `cache list` was unconditionally falling back to `computeCacheKey(model.Spec.Source)` when `Status.CacheKey == ""`, which produced phantom entries for metal/single-file models (CLI sites didn't replicate the controller's scoping). `delete --purge-cache` had the same defect — would target directories the serving pod never used. Both now route through `cachekey.EffectiveKey`, which correctly returns `""` and lets the call sites skip/no-op.
- **`metal` is a literal in `cachekey.EffectiveKey`, not `acceleratorMetal`.** Pragmatic to avoid a controller→cachekey import (cachekey is a low-level utility and can't import controllers). If `acceleratorMetal` were ever renamed, this would silently go stale. Worth a one-line risk acknowledgment in the package doc, but the constant is established enough that this isn't a blocking issue.
- **Two-commit split is clean.** Implementation commit, then a docs-only follow-up. Conventional commits for both. DCO on both.
- **Scope matches the issue.** No unrelated edits. `hasMultiFileStaging` / `isMetalModel` retained in the controller (the comment claims they're used widely elsewhere — confirmed: 1 hit in `inferenceservice_controller.go`, 4 hits in `model_controller.go`).
- **Test coverage pins the regression.** `TestEffectiveKey*` cases (nil, Status wins, multi-file non-metal derives, metal → "", single-file → "") will fail loud if anyone reverts the scoping.
- **No API/CRD changes** — `pkg/cachekey` is internal.

**Non-blocking nit:** The package doc could add a one-line note like "the `metal` literal mirrors `internal/controller.acceleratorMetal`; keep them in sync." Risk is real but low — not worth a revision cycle.

Defilan:
Good catch on the `metal` literal / `acceleratorMetal` coupling — that's exactly the kind of silent-drift risk worth pinning. Followed up in #993 with a doc note at the literal. Went with the comment rather than a real drift-guard test because such a test would have to live in `internal/controller` (the only package that sees both symbols) and would pull that package's envtest setup into a pure unit test; not worth the friction for a low-risk coupling. Thanks for the thorough review across all three.

itsmiso-ai (approved):
**Approve.** Correct fix for #921 — the scoped decision had to be shared, not just the raw fingerprint. The diff is mechanically right and the doc update commit is a clean follow-up.

**Verified:**

- **`EffectiveKey` is semantically byte-equivalent to the old `effectiveModelCacheKey`.** Old: `hasMultiFileStaging(model) && !isMetalModel(model)`. New: `(len(Files) > 0 || Mmproj != "") && !(Hardware != nil && Accelerator == "metal")`. `hasMultiFileStaging` is `model != nil && (len(Files) > 0 || Mmproj != "")`; the nil guard is absorbed by the earlier `model == nil → ""` branch. Same behavior in every observable case.
- **`Compute` parity preserved.** `TestComputeMatchesSHA256Prefix` pins the exact hex prefix against a known URL, and the byte-for-byte logic is identical to both prior implementations. The CLI test suite (`TestComputeCacheKey*`) was migrated to call `cachekey.Compute` — same inputs → same outputs.
- **The bug fix is real.** `cache list` was unconditionally falling back to `computeCacheKey(model.Spec.Source)` when `Status.CacheKey == ""`, which produced phantom entries for metal/single-file models (CLI sites did not replicate the controller's scoping). `delete --purge-cache` had the same defect — would target directories the serving pod never used. Both now route through `cachekey.EffectiveKey`, which correctly returns `""` and lets the call sites skip/no-op.
- **`metal` is a literal in `cachekey.EffectiveKey`, not `acceleratorMetal`.** Pragmatic to avoid a controller→cachekey import (the cachekey package is a low-level utility and cannot import controllers). If `acceleratorMetal` were ever renamed, this would silently go stale. Worth a one-line risk acknowledgment in the package doc, but the constant is established enough that this is not a blocking issue.
- **Two-commit split is clean.** Implementation commit, then a docs-only follow-up. Conventional commits for both. DCO on both.
- **Scope matches the issue.** No unrelated edits. `hasMultiFileStaging` / `isMetalModel` retained in the controller (the comment claims they are used widely elsewhere — confirmed: 1 hit in `inferenceservice_controller.go`, 4 hits in `model_controller.go`).
- **Test coverage pins the regression.** `TestEffectiveKey*` cases (nil, Status wins, multi-file non-metal derives, metal → "", single-file → "") will fail loud if anyone reverts the scoping.
- **No API/CRD changes** — `pkg/cachekey` is internal.

**Non-blocking nit:** The package doc could add a one-line note like "the `metal` literal mirrors `internal/controller.acceleratorMetal`; keep them in sync." Risk is real but low — not worth a revision cycle.

Issue #892, opened 2026-06-29
[BUG] gateway-routed (cloud-proxy) coder agent never edits: loop stalls on EditFreeStreak with no branch, even on tractable issues


Defilan:
## Bug Description

A coder-role `Agent` that reaches its model through the gateway (`provider: cloud-proxy` -> Envoy AI Gateway -> a served `InferenceService`) never lands a code edit. The agent loop reads/explores and then terminates via the stuck-loop `EditFreeStreak` detector with **no branch pushed** and verdict `INCOMPLETE`. The *same* issue dispatched to a coder Agent backed by a **local** model produces edits and a branch normally.

This makes the gateway-routed (e.g. AMD/Strix dual-box) coder path unable to produce code, even though it schedules, reasons, and calls tools.

## Steps to Reproduce

1. Have two coder Agents on the same operator:
   - `local`: a coder Agent backed by a model the agent's node serves locally (e.g. metal-agent on the same host).
   - `gateway`: a coder Agent with `provider: cloud-proxy`, `providerConfig.baseURL` pointing at an Envoy AI Gateway `/v1`, routing (via ModelRouter) to a served `InferenceService`.
2. Dispatch a **tractable** `issue-fix` AgenticTask (one a 27-35B coder can edit) to `local` -> it produces a branch with real edits.
3. Dispatch the **same** issue to `gateway` -> it runs, reasons, but never edits; ends `Succeeded`/`INCOMPLETE` with condition message `stuck-loop detector intervened: EditFreeStreak` and `status.branch` empty.

Observed repeatedly (6 runs) across two models (Ornith-35B, Qwopus-27B) and two issues (#856 tractable, #867 harder), at `editFreeTurnsLimit` 4 and 8. Every gateway-routed run failed to edit; the only local-model run edited.

## Expected Behavior

A gateway-routed coder Agent should produce edits and a branch on a tractable issue, the same as a local-model coder Agent.

## Actual Behavior

The gateway-routed coder loop never edits and always terminates via `EditFreeStreak` with no branch.

## What is NOT the cause (ruled out)

- **Tool-calling passthrough works.** A direct `/v1/chat/completions` to the gateway model with a `tools` array returns a correct `tool_call` (`finish_reason: tool_calls`, a well-formed `read_file` call). So the gateway transports function-calling fine.
[...]

PR #987, opened 2026-07-06
fix(foreman): grant coder Job pods get on secrets for cloud-proxy auth

### LINKED ISSUE #986: Job-mode coder agent RBAC missing `secrets:get` for provider auth
## Bug

The `foreman-coder` Role created by the foreman Helm chart grants `get/list/watch` on `agentictasks`, `agents`, and `inferenceservices`, but **not `secrets`**. A Job-mode coder agent fails before its first inference call because it cannot read the Secret referenced by `providerConfig.apiKeySecretRef`.

## Reproduction

1. Define a Job-mode coder Agent with a `cloud-proxy` provider that uses `apiKeySecretRef`:

```yaml
apiVersion: foreman.llmkube.dev/v1alpha1
kind: Agent
metadata:
  name: coder-go
spec:
  role: coder
  provider: cloud-proxy
  providerConfig:
    baseURL: http://litellm.llm:4000/v1
    model: nvidia
    apiKeySecretRef:
      name: foreman-agent
      key: LITELLM_API_KEY
  execution:
    mode: Job
    serviceAccountName: foreman-coder
    image: ghcr.io/defilantech/llmkube-foreman-agent-coder:0.9.0
```

2. Dispatch a task:

```bash
llmkube foreman dispatch --repo defilantech/LLMKube --agents coder-go -n llm 892
```

3. The Job terminates with `INCOMPLETE` / `InferenceServiceUnavailable`:

## Expected

The agent resolves the API key from the referenced Secret and proceeds to inference.

## Actual

The agent's executor calls `get Secret <ns>/<name>` to resolve `providerConfig.apiKeySecretRef`, and the request is denied:

```
FOREMAN-RESULT: {"verdict":"INCOMPLETE","summary":"get Secret llm/foreman-agent for provider auth: secrets \"foreman-agent\" is forbidden: User \"system:serviceaccount:llm:foreman-coder\" cannot get resource \"secrets\" in API group \"\" in the namespace \"llm\"","result":{"schemaVersion":"foreman.v1","kind":"native-agent-loop","verdict":"INCOMPLETE","failureReason":"InferenceServiceUnavailable","summary":"get Secret llm/foreman-agent for provider auth: secrets \"foreman-agent\" is forbidden: User \"system:serviceaccount:llm:foreman-coder\" cannot get resource \"secrets\" in API group \"\" in the namespace \"llm\"","extra":{"outcome":"EXECUTOR-PRECONDITION-FAILED","reason":"InferenceServiceUnavailable"},"elapsedSec":0.003746279}}
```

The `foreman-coder` Role as shipped by the chart:

```yaml
rules:
- apiGroups: ["foreman.llmkube.dev"]
  resources: ["agentictasks", "agents"]
  verbs: ["get", "list", "watch"]
- apiGroups: ["inference.llmkube.dev"]
  resources: ["inferenceservices"]
  verbs: ["get", "list", "watch"]
```

The code path: `executor_native.go` `resolveAuthToken` → `clientset.CoreV1().Secrets(namespace).Get(...)` → requires `get` on `secrets` in the core API group.

## Why InProcess works but Job doesn't

InProcess agents run inside the `foreman-agent` Deployment, which has separate, broader RBAC (including secret reads) wired up by the chart's agent section. Job-mode agents run under the least-privilege `foreman-coder` SA and never inherit the agent deployment's permissions.

## Suggested fix

Add a `secrets: get` rule to the `foreman-coder` Role in the foreman chart. The secret read is scoped to the agent's own namespace and only used to resolve `apiKeySecretRef`. An alternative is a separate Role+binding for secret reads that can be toggled via a chart value, but since every cloud-proxy coder needs this, granting it by default on the `foreman-coder` SA is the simplest correct posture.

## Context

- Introduced in #621 (Job execution strategy), which noted "coder ServiceAccount + read-only RBAC in the foreman chart" but shipped the Role without `secrets`.
- Observed on v0.9.0 with a LiteLLM-backed cloud-proxy coder.
- The `serviceAccountName` field on `Agent.spec.execution` is what wires the SA onto the Job pod (confirmed working once set); this issue is purely the Role's missing verb.

joryirving:
## What

Adds two missing rules to the `foreman-coder` Role in the foreman chart:

1. **`secrets: get`** — resolves `providerConfig.apiKeySecretRef` for cloud-proxy providers. The executor's `resolveAuthToken` calls `clientset.CoreV1().Secrets(namespace).Get(...)` to fetch the API key.
2. **`configmaps: get` + `create` + `update`** — writes the per-task transcript ConfigMap (`foreman-transcript-<task-name>`) so failing runs can be diagnosed post-hoc.

The chart mounts `GITHUB_TOKEN` as a Job env, but the API key is resolved dynamically — hence the secrets gap. The transcript writer was always failing silently (the run-task logged `continuing`), but the consequence is invisible: a failing Job-mode coder leaves no recoverable transcript, only the summary audit ConfigMap. That just bit a 52-minute `ModelReportedError` run.

## Why

Fixes #986

Repro from the issue:

```bash
llmkube foreman dispatch --repo defilantech/LLMKube --agents coder-go -n llm 892
# → INCOMPLETE / InferenceServiceUnavailable
# → get Secret llm/foreman-agent for provider auth: secrets "foreman-agent" is forbidden
```

## How

- Two new rules in `charts/foreman/templates/coder-rbac.yaml`:
  - Core API group, `secrets`, `get` only.
  - Core API group, `configmaps`, `get` + `create` + `update`.
- Header comment updated to reflect the new read+write pattern.
- Both grants scoped to the Role's namespace (the chart's namespace) and the minimum verbs needed.

No `values.yaml` toggle — every cloud-proxy Job-mode coder needs the secrets rule, every Job-mode coder run produces a transcript, so granting both by default is the simplest correct posture.

## Checklist

- [ ] Tests added/updated
- [ ] `make test` passes locally
- [ ] `make lint` passes locally
- [ ] Documentation updated (n/a — RBAC is internal)

---

Assisted-by: opencode — the assistant located the chart template, drafted the diff and this PR body. I reviewed the change, verified the code paths (`executor_native.go` `resolveAuthToken` → `Secrets().Get`; `WriteTranscript` → ConfigMap read-then-upsert), confirmed the reproductions against the live cluster, and signed off both commits myself.

### DIFF
--- a/charts/foreman/templates/coder-rbac.yaml
+++ b/charts/foreman/templates/coder-rbac.yaml
@@ -10,13 +10,18 @@ metadata:
 ---
 {{- end }}
 # Least-privilege Role for the coder Job pods. A Job-mode coder runs
-# `foreman-agent run-task`, which READS the AgenticTask, the referenced
-# Agent, the Agent's InferenceService, and (for cloud-proxy providers) the
-# Secret referenced by `providerConfig.apiKeySecretRef` from the API to
-# resolve the executor's base URL and auth, then runs the native loop
-# in-process. It never writes task status (the foreman-agent watcher
-# polls the Job's log tail and patches status), so this Role is read-only
-# and namespaced to where the chart (and the tasks) live.
+# `foreman-agent run-task`, which:
+#   - READS the AgenticTask, the referenced Agent, and the Agent's
+#     InferenceService from the API to resolve the executor's base URL.
+#   - For cloud-proxy providers, READS the Secret referenced by
+#     `providerConfig.apiKeySecretRef` to resolve the API key.
+#   - WRITES a per-task transcript ConfigMap (foreman-transcript-<task>)
+#     during the run so failing runs can be diagnosed post-hoc; without
+#     this the turn-by-turn transcript is lost and only the summary
+#     audit ConfigMap survives.
+# The foreman-agent watcher (separate process) patches AgenticTask
+# status — the in-Job run-task never writes task status itself. The
+# Role is namespaced to where the chart (and the tasks) live.
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
@@ -36,6 +41,12 @@ rules:
   - apiGroups: [""]
     resources: ["secrets"]
     verbs: ["get"]
+  # Per-task transcript write. `get` + `create`/`update` mirror the
+  # transcript writer's read-then-upsert flow; the writer targets a
+  # single ConfigMap (foreman-transcript-<task-name>) per run.
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    verbs: ["get", "create", "update"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
--- a/charts/foreman/templates/coder-rbac.yaml
+++ b/charts/foreman/templates/coder-rbac.yaml
@@ -11,11 +11,12 @@ metadata:
 {{- end }}
 # Least-privilege Role for the coder Job pods. A Job-mode coder runs
 # `foreman-agent run-task`, which READS the AgenticTask, the referenced
-# Agent, and the Agent's InferenceService from the API to resolve the
-# executor's base URL, then runs the native loop in-process. It never
-# writes task status (the foreman-agent watcher polls the Job's log tail
-# and patches status), so this Role is read-only and namespaced to where
-# the chart (and the tasks) live.
+# Agent, the Agent's InferenceService, and (for cloud-proxy providers) the
+# Secret referenced by `providerConfig.apiKeySecretRef` from the API to
+# resolve the executor's base URL and auth, then runs the native loop
+# in-process. It never writes task status (the foreman-agent watcher
+# polls the Job's log tail and patches status), so this Role is read-only
+# and namespaced to where the chart (and the tasks) live.
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
@@ -30,6 +31,11 @@ rules:
   - apiGroups: ["inference.llmkube.dev"]
     resources: ["inferenceservices"]
     verbs: ["get", "list", "watch"]
+  # Resolves providerConfig.apiKeySecretRef for cloud-proxy providers.
+  # `get` only — the agent does not list/watch secrets.
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding


joryirving:
Follow-up RBAC gap on the same coder Job ServiceAccount: the executor also gets **`configmaps: get` 403 (forbidden)** when writing the debug transcript ConfigMap during a run. It's non-fatal, but the consequence is that failing runs can't be diagnosed post-hoc — the turn-by-turn transcript is never persisted, only the summary audit ConfigMap is. This just bit a repro of #892 (the 52-min `ModelReportedError` run has no recoverable transcript).

Suggest adding the `configmaps` verbs the transcript writer needs (`get` + `create`/`update`) to the `foreman-coder` Role in `charts/foreman/templates/coder-rbac.yaml` — either folded into this PR or as an immediate follow-up.

Defilan (approved):
**Approve.** Correct, minimal, and well-motivated. I verified both grants against the actual code paths rather than the PR description:

**Verified:**
- **`secrets: get` is sufficient (not `list`/`watch`).** `resolveAuthToken` (executor_native.go) does `e.Client.Get(ctx, key, &secret)`, and the run-task path builds `e.Client` via `client.New(cfg, ...)` (cmd/foreman-agent/main.go) — a direct, non-cached controller-runtime client. So the read is a real API GET; a cache-backed client would have needed `list`+`watch` for the informer, but this isn't one. Good catch keeping it `get`-only.
- **`configmaps: get, create, update` matches the writer exactly.** `WriteTranscript` (transcript.go) does `Get` → on `IsNotFound` `Create`, else `Update`. No `list`, no `delete` — so the verb set is precise, no over-grant.
- **Least-privilege posture holds.** Namespaced `Role` (not ClusterRole), correct for namespace-scoped secrets/configmaps. Header comment accurately describes the new read+write pattern.
- Fixes a real, invisible failure (#986): a cloud-proxy Job-mode coder couldn't resolve its auth secret, and the transcript writer failed silently, leaving no recoverable transcript on a failed run.

**Non-blocking, worth a note:** `secrets: get` is namespace-wide. RBAC can't scope `get` to the specific `apiKeySecretRef` name because that name is dynamic (it comes from the Agent's `providerConfig` at run time, not known at chart render), so `resourceNames` isn't feasible here. The consequence: a coder Job — which executes model-generated bash and has its SA token auto-mounted — can now `get` any Secret in its namespace. That's necessary for this fix given the current architecture, and the threat model already trusts the coder image, so I don't think it blocks. A cleaner long-term posture would be to resolve the secret in the operator/watcher (which already reads the Agent) and inject the token as a Job env, the same way `GITHUB_TOKEN` is mounted — that would let the Job drop `secrets` entirely. Worth an issue for later, not this PR.

One tiny thing: the checklist has tests unchecked. There's no existing coder-rbac template test to extend, so I don't think it's required here, but a helm-template assertion that these two rules render would guard against an accidental removal down the line.

Ship it.

PR #988, opened 2026-07-06
feat(foreman): grounded-finding rail for reviewer NO-GO verdicts

Defilan:
## What

A deterministic rail that demotes a reviewer NO-GO to GO when none of its blocking findings (severity blocker/major) cite a line the diff actually changed. The rejected findings are archived under `extra.ungroundedFindings`. Mirror of the existing scope-overlap / issueAsk rails, in the opposite direction.

## Why

Every existing reviewer rail guards the false-negative direction (a too-lenient GO). Nothing guarded the false-positive direction: a NO-GO whose blocking findings cite fabricated or out-of-scope code. In reviewer experiments a model blocked a branch on a fabricated file (`docs/MODEL-CACHE.md`, never in the diff) and on a pre-existing out-of-scope line (a `kubectl exec` print unchanged since the prior release). Both are objections not anchored to changed code. This rail makes the reviewer trustworthy: it can no longer block on inventions.

## How

- New `enforceReviewerGroundedFindings` (pure function, injected `changedLines(file)` closure, unit-tested): line-level grounding via the diff-hunk parsing already in the package; degrade-open on git error; `FOREMAN_GROUNDED_FINDINGS=0` toggle.
- Wired into the reviewer executor block before the GO->NO-GO rails, hoisting the diff ground-truth to one `repo.DiffNameOnly` call (fed to both rails). The closure is built in a small helper `reviewerGroundedChangedLines` to keep `runLLMPath` under the gocyclo cap.
- Reviewer prompt (`config/foreman/system-prompts/reviewer.md`) now requires `file`+`line` on blocking findings.

Scope note: this makes the reviewer trustworthy, not smarter. On a subtly-flawed branch where the reviewer never found the real defect, an ungrounded NO-GO correctly demotes to GO. Catching subtle gaps is a separate (stronger-model / panel) track.

## Interaction note

Demoting NO-GO->GO puts the branch on the normal review-GO path, so if a review task sets `Payload.OpenPullRequest`, a demoted GO that also clears scope-overlap AND issueAsk will auto-open a PR. This is design-consistent (an ungrounded NO-GO is treated as APPROVE) and double-gated by the two GO->NO-GO rails. The "real defect, wrong citation" case is covered by the stronger-reviewer / panel track, not this rail.

## Checklist

- [x] Tests added (8 table-driven cases: grounded, fabricated-file, unchanged-line, no-line, minor-only, GO-untouched, toggle-off, git-degrade-open)
- [x] No API/CRD changes

## AI assistance disclosure

Assisted-by: Claude Code (Claude Opus 4.8). Generated the implementation, tests, and PR text from a design I approved; I reviewed the diff, ran the gate locally (go test, make lint on darwin and GOOS=linux), and adversarially reviewed the whole branch before submitting.

### DIFF
--- a/config/foreman/agents/claude-sonnet-cloud-reviewer.yaml
+++ b/config/foreman/agents/claude-sonnet-cloud-reviewer.yaml
@@ -414,11 +414,21 @@ spec:
                                  identification.)
           - `findings`        :: `[]` of `{severity: "blocker" | "major" |
                                  "minor", area: "scope" | "tests" | "style"
-                                 | "side-effects" | "docs", message: "..."}`
-                                 . Each finding must include enough specifics
-                                 (file path, line number, quoted issue text)
-                                 that a maintainer can act on it without
-                                 re-deriving your reasoning.
+                                 | "side-effects" | "docs", message: "...",
+                                 file: "path", line: N}`. Each finding must
+                                 include enough specifics (file path, line
+                                 number, quoted issue text) that a maintainer
+                                 can act on it without re-deriving your
+                                 reasoning.
+
+                                 Every BLOCKING finding (severity `blocker`
+                                 or `major`) MUST set `file` and `line` to a
+                                 line THIS diff changed (a line inside a
+                                 `git diff main...HEAD` hunk). A blocking
+                                 finding without a `file` + `line` on changed
+                                 code is IGNORED by the harness and cannot
+                                 sustain a REQUEST-CHANGES verdict; minor
+                                 findings may omit `file`/`line`.
           - `issueAsk`        :: a short quote (≤200 chars) capturing the
                                  operative ask of the issue body, taken from
                                  the `fetch_issue` result as closely as you
--- a/config/foreman/agents/devstral-24b-reviewer.yaml
+++ b/config/foreman/agents/devstral-24b-reviewer.yaml
@@ -360,11 +360,21 @@ spec:
                                  identification.)
           - `findings`        :: `[]` of `{severity: "blocker" | "major" |
                                  "minor", area: "scope" | "tests" | "style"
-                                 | "side-effects" | "docs", message: "..."}`
-                                 . Each finding must include enough specifics
-                                 (file path, line number, quoted issue text)
-                                 that a maintainer can act on it without
-                                 re-deriving your reasoning.
+                                 | "side-effects" | "docs", message: "...",
+                                 file: "path", line: N}`. Each finding must
+                                 include enough specifics (file path, line
+                                 number, quoted issue text) that a maintainer
+                                 can act on it without re-deriving your
+                                 reasoning.
+
+                                 Every BLOCKING finding (severity `blocker`
+                                 or `major`) MUST set `file` and `line` to a
+                                 line THIS diff changed (a line inside a
+                                 `git diff main...HEAD` hunk). A blocking
+                                 finding without a `file` + `line` on changed
+                                 code is IGNORED by the harness and cannot
+                                 sustain a REQUEST-CHANGES verdict; minor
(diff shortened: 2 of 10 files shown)


joryirving (changes requested):
**Request changes.** The rail is a no-op in production — `changedNewLines` diffs against `HEAD` (working tree), but in the reviewer workspace the branch is already committed and checked out, so the working tree equals HEAD. Every file returns zero changed lines, every blocking finding is classified ungrounded, and every NO-GO is demoted to GO unconditionally.

**Blocking: `changedNewLines` uses the wrong base for the reviewer context**

`reviewerGroundedChangedLines` (executor_native.go:748) delegates to `changedNewLines`, which runs `git diff -U0 HEAD -- file`. This compares the **working tree** against **HEAD**. In the coder path that's correct — the coder's changes are uncommitted, so HEAD is the pre-work base. But in the reviewer workspace, the model runs `git checkout origin/{branch}` (per reviewer.md Step 1), so HEAD **is** the coder's commit and the working tree is clean. `git diff HEAD` is empty. Zero changed lines for every file → zero grounded findings → every NO-GO demotes.

The rail's own doc comment (`grounded_finding_gate.go:8`) says it checks `git diff -U0 main...HEAD`, but the wired implementation doesn't do that. `DiffNameOnly` (used right next to it for scope-overlap) correctly uses `main...HEAD` (three-dot merge-base diff) and works. `changedNewLines` doesn't.

**Why the tests don't catch it:** `TestGroundedFindings_*` inject synthetic `changedLines` closures (`changed(fix)`) that return static maps. The integration point — `reviewerGroundedChangedLines` → `changedNewLines` → `git diff HEAD` — is never exercised with a real git workspace. The tests prove the pure function works; they don't prove the wiring does.

**Fix direction:** `changedNewLines` is shared with `mutation_gate.go` where the working-tree diff is correct. Either:
- Add a `changedBranchLines(ctx, workspace, base, file, run)` that runs `git diff -U0 main...HEAD -- file`, or
fish- Parameterize `changedNewLines` with a base ref (`HEAD` for coder, `main` for reviewer).

And add one integration test (or test of `reviewerGroundedChangedLines` with a real tmp git repo) that confirms the closure returns non-empty for a committed branch diff.

**Non-blocking nits:**
- The per-file cache in `enforceReviewerGroundedFindings` is good — avoids redundant git calls when multiple findings cite the same file.
- Prompt sync across all three reviewer agent YAMLs + reviewer.md is consistent and correct.

**Compliance:** DCO on all 4 commits, conventional commits, no CRD changes, lll clean, AI disclosure present.

Defilan:
Fixed in 637c194, and you were exactly right. The rail resolved changed lines via `changedNewLines` (`git diff -U0 HEAD`), which is empty in the reviewer workspace because the branch is already committed and checked out, so no finding was ever grounded and every NO-GO demoted unconditionally.

- Added `changedBranchLines` (`git diff -U0 base...HEAD`), the committed-branch mirror of `changedNewLines`, sharing a `parseAddedLines` helper. Wired `reviewerGroundedChangedLines` to it with base `main`, matching the `repo.DiffNameOnly` scope-overlap check on the line right above, so both reviewer rails agree on what the branch changed. `changedNewLines` and the coder mutation gate are untouched.
- Added the integration test you called for: it drives the real git wiring in a reviewer-shaped workspace (committed branch, clean working tree) and asserts the closure returns the committed change's lines. It fails on the old `git diff HEAD` wiring and passes now; the prior synthetic-closure tests never exercised the git call, which is why they missed it.

Full agent package + `golangci-lint` green locally. #992 is rebased on top so it inherits the fix. Nice symmetry with the base-ref bug on #995: same class, opposite direction.

PR #995, opened 2026-07-06
fix(foreman): tolerate self-committed model work and detect git apply edits

### LINKED ISSUE #982: Coder self-commit / bash edits discarded: false 'no diff' NO-GO and EditFreeStreak kills

joryirving:
## What

Two related foreman agent fixes that address false non-GO outcomes in ~64% of fleet runs:

1. **Self-commit tolerance**: Models frequently run `git commit` before calling `submit_result`, leaving the working tree clean but commits ahead of base on the branch. The executor previously only checked working-tree diff state, so it returned NO-GO ("no diff") even though real edits existed — just committed by the model instead of left uncommitted for the executor.

2. **Bash edit detection gap**: `EditFreeStreak` resets when bash commands mutate source files (sed -i, tee, mv, cp, etc.), but models editing via `git apply patch.diff` bypassed all existing token/redirect patterns — git apply modifies files directly without output redirection.

## Why

- The self-commit NO-GO false positive was the dominant cause of non-successful coder outcomes (~64% of fleet).
- Git apply edits caused force-terminate mid-edit because EditFreeStreak never reset, so consecutive edits were counted as "no edit needed" and triggered termination.

Fixes #982

## How

**Fix 1 — self-commit tolerance:**

The review caught that the first iteration of this PR targeted the fork's local `main` (a possibly-stale ref) for the commit count and soft reset. In the production fork workflow, `setupTaskBranch` cuts the task branch from `FETCH_HEAD` of the upstream base (#813), so a lagging local `main` would cause `git rev-list --count main..HEAD` to over-count and `git reset --soft main` to re-stage intervening upstream commits — squatting upstream delta into the recovered commit, invisible in a synced lab or a fresh-clone E2E.

The fix: a new `repo.BaseBranchSHA(ctx, ws, upstreamURL, baseBranch)` helper that re-fetches the upstream base and returns the resolved tip SHA (mirroring `CreateBranchFromUpstream`'s FETCH_HEAD pattern; same `gitRefSafe` guards against `-`/`..` smuggling). The executor's `recoverSelfCommitsOrNoChange` (renamed from `shouldNoChange` to signal its side effect) now resolves the upstream base, soft-resets against that SHA, and:

- Logs an error (rather than silently returning true) when upstream-base resolution or soft-reset fails, so a broken recovery is observable in the field instead of masquerading as a clean NO-CHANGES outcome.
- Returns true only for a true NO-CHANGES (zero commits ahead of the resolved upstream tip) or a logged failure.

If `upstreamURL` is empty (a self-hosted repo with no upstream fork), `BaseBranchSHA` refuses rather than falling back to the local ref, since that is exactly the failure mode it exists to prevent.

**Fix 2 — bash edit token expansion:**
- Added `"git apply"` to `fileWritingBashTokens` in `progress.go`. The redirect parser (`bashRedirectsToFile`) already catches heredocs and echo/tee redirection, but git apply uses neither pattern. Explicit token needed. A one-line comment notes that the substring match also fires on read-only `git apply --check/--stat/--numstat` — left in deliberately because the only failure mode of that over-match is a false force-terminate, and biased away from that direction.

## Tests

| Test | Verifies |
|------|----------|
| `TestCommitsAheadOfBase` | Count commits ahead of base (0 when base==HEAD, >0 otherwise) |
| `TestSoftResetToBase` | Recovery: commits moved to working tree, `ErrNothingToCommit` on no-op, workspace/base required guards |
| `TestBaseBranchSHA` | Upstream-base resolution: returns upstream tip (not lagging fork main); required-arg + `gitRefSafe` guards |
| `TestExecutorNative_SelfCommittedWorkIsRecovered` | E2E: model self-commits → executor detects commits ahead → recovers via soft reset → re-commits with DCO sign-off |
| `TestRecoverSelfCommitsOrNoChange_StaleForkBaseIsNotCountedAsCommitsAhead` | The exact stale-fork regression: bare upstream + fork clone with lagging local `main` + branch cut from upstream FETCH_HEAD + model self-commit. Asserts HEAD lands at the upstream tip and the staged diff contains `fix.go` but NOT `UPSTREAM.md` (the regression fingerprint). This test would have failed under the previous implementation. |
| `TestRecoverSelfCommitsOrNoChange_GenuinelyNothingToRecover` | True NO-CHANGES: clean branch with no self-commit → returns true, HEAD unchanged. |
| `TestBashLikelyMutatesWorkspace_GitApplyAdded` | git apply now detected; existing tokens still work; non-mutating commands unaffected |

The two `TestRecoverSelfCommitsOrNoChange_*` tests live in `executor_native_internal_test.go` (package `agent`) so they can drive the unexported `recoverSelfCommitsOrNoChange` directly. The previous test was a black-box on the underlying repo helpers, which doesn't exercise the executor's base-resolution logic.

## Checklist

- [x] Tests added/updated (6 new tests across repo + executor packages)
- [x] `make fmt` + `make vet` pass
- [x] `.gitignore` now excludes `.superpowers/` and the per-issue `docs/superpowers/{plans,specs}/` so agent scratch state cannot accidentally be committed again. Existing tracked docs under `docs/superpowers/` remain unaffected.

## Non-goals

- Does not fix the separate reviewer scope issue (#983 / #672): `baseBranch` defaults to `"main"` in the reviewer check at `executor_native.go:~625`. That's a different bug tracked separately and should be addressed in its own PR so it doesn't mix concerns with this one.
- Does not rewrite the two arg builders (`BuildArgs` for controller, `buildLlamaServerArgs` for metal-agent) into a shared function — they serve different runtimes with different defaults; the AGENTS.md parity test contract still applies unchanged.

---

AI assistance disclosure per CONTRIBUTING.md:

Assisted-by: opencode (generated both fixes, tests, and PR body; reviewed and ran make test/lint)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,10 @@ CLAUDE.md
 # Root-level screenshots and session artifacts (kept docs/images/*.png)
 /*.png
 /home-snapshot.md
+
+# Superpowers agent working state (specs/plans/ledger live outside the repo).
+# New files under these paths are agent scratch; existing tracked files
+# (e.g. shared design docs in docs/superpowers/) remain unaffected.
+.superpowers/
+/docs/superpowers/plans/
+/docs/superpowers/specs/
--- a/pkg/foreman/agent/executor_native.go
+++ b/pkg/foreman/agent/executor_native.go
@@ -659,11 +659,19 @@ func (e *NativeAgentLoopExecutor) runLLMPath(
 	// 10. GO verdict: check for changes first, then commit + push. If
 	// the model emitted GO but never edited a file, NO-CHANGES is the
 	// honest outcome regardless of whether commit_message was set.
+	baseBranch := baseBranchOrDefault(task.Spec.Payload.BaseBranch)
 	hasChanges, hcErr := repo.HasChanges(ctx, workspace)
 	if hcErr != nil {
 		return e.commitRejectedResult(start, transcriptRef, loopRes, branch, hcErr), nil
 	}
-	if !hasChanges {
+
+	// #982: tolerate a model that self-committed its work. If the working tree
+	// is clean but there are commits ahead of base on this branch, recover them
+	// by soft-resetting into the working tree so repo.Commit can re-apply with
+	// DCO sign-off and executor-owned author identity. This prevents false
+	// NO-GO ("no diff") outcomes when a model runs git commit before calling
+	// submit_result instead of leaving changes uncommitted for the executor.
+	if e.shouldNoChange(ctx, log, hasChanges, workspace, baseBranch) {
 		return e.noChangesResult(start, transcriptRef, loopRes, branch), nil
 	}
 
@@ -1332,6 +1340,29 @@ func (e *NativeAgentLoopExecutor) envtestGateFailedResult(
 		"turnCount":     lr.Turns,
 	}
 	return r
+}
+
+// shouldNoChange returns true when the executor should emit a NO-CHANGES
+// result: either the working tree is genuinely clean, or a model self-
+// committed its work and soft-reset failed to recover changes. When it
+// returns false and hasChanges was initially false, the caller should
+// proceed to commit (the model's self-committed edits are now in the
+// working tree after a successful soft reset).
+func (e *NativeAgentLoopExecutor) shouldNoChange(
+	ctx context.Context, log logr.Logger, hasChanges bool, workspace, baseBranch string,
+) bool {
+	if hasChanges {
+		return false
+	}
+	err := repo.SoftResetToBase(ctx, workspace, baseBranch)
+	if errors.Is(err, repo.ErrNothingToCommit) {
+		return true // truly no changes
+	}
+	if err != nil {
+		return true // detection or reset failed — degrade gracefully
+	}
+	log.Info("model self-committed; recovering changes via soft reset", "baseBranch", baseBranch)
+	return false
 }
 
 func (e *NativeAgentLoopExecutor) pushFailedResult(
--- a/pkg/foreman/agent/executor_native.go
+++ b/pkg/foreman/agent/executor_native.go
@@ -666,12 +666,18 @@ func (e *NativeAgentLoopExecutor) runLLMPath(
 	}
 
 	// #982: tolerate a model that self-committed its work. If the working tree
-	// is clean but there are commits ahead of base on this branch, recover them
-	// by soft-resetting into the working tree so repo.Commit can re-apply with
-	// DCO sign-off and executor-owned author identity. This prevents false
-	// NO-GO ("no diff") outcomes when a model runs git commit before calling
-	// submit_result instead of leaving changes uncommitted for the executor.
-	if e.shouldNoChange(ctx, log, hasChanges, workspace, baseBranch) {
+	// is clean but there are commits ahead of the resolved upstream base on
+	// this branch, recover them by soft-resetting into the working tree so
+	// repo.Commit can re-apply with DCO sign-off and executor-owned author
+	// identity. This prevents false NO-GO ("no diff") outcomes when a model
+	// runs git commit before calling submit_result instead of leaving changes
+	// uncommitted for the executor. The base is the upstream tip at the time
+	// the branch was cut (per #813), not a possibly-stale local ref, so we
+	// don't accidentally re-stage intervening upstream commits.
+	if e.recoverSelfCommitsOrNoChange(
+		ctx, log, hasChanges, workspace,
+		e.resolveUpstreamForRun(task), baseBranch,
+	) {
 		return e.noChangesResult(start, transcriptRef, loopRes, branch), nil
 	}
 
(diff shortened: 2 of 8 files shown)


Defilan (changes requested):
Really nice work, Jory. The dual root-cause diagnosis is right, the `git apply` token fix is clean and well-tested, and the scoping discipline (the explicit #983/#672 non-goal) is exactly right. The helper design (`CommitsAheadOfBase` / `SoftResetToBase` with the `ErrNothingToCommit` sentinel and the required-arg guards) is tidy, and the tests exercise real git behavior instead of mocks. One blocking correctness issue on the self-commit recovery, plus a couple of hardening asks, below.

## Blocking: the soft reset targets the wrong base in the fork workflow

`shouldNoChange` resets against `baseBranchOrDefault(...)`, i.e. the literal local ref `main`. But the task branch is not cut from local `main`: `setupTaskBranch` -> `CreateBranchFromUpstream` fetches the base fresh from the upstream remote into `FETCH_HEAD` and cuts the branch there. That is the #813 fix, precisely so a stale fork base cannot poison the branch point. The executor clones from and pushes to the fork (`GitRemoteURL`), so local `main` is the *fork's* main, which can lag the upstream tip the branch was actually cut from.

When it lags, `git rev-list --count main..HEAD` over-counts and `git reset --soft main` re-stages the intervening upstream commits, so `repo.Commit` squashes them into the executor's recovered commit. Reproduction (fork `main` at A, upstream advanced to B, branch cut from B, model self-commits M):

```
$ git rev-list --count main..HEAD     # what SoftResetToBase counts
2                                       # should be 1 (just M); it includes upstream commit B
$ git reset --soft main && git diff --cached f
 A
+B
+M                                      # B gets squashed into the recovered commit
```

Effect: whenever the fork base is behind upstream (the exact fragility #813 removed), a self-committed run produces a branch whose merge-base with upstream is the stale point, bloating the PR with unrelated upstream delta. It is invisible in a fully-synced lab and in the E2E test (which clones fresh, so local `main` == branch point), which is why it passes today.

Fix direction: reuse the actual branch-creation point rather than re-deriving it from `"main"`. Either capture the `FETCH_HEAD` SHA that `CreateBranchFromUpstream` resolved and thread it through as the base for both helpers, or compute `git merge-base <upstream-base> HEAD` at recovery time. The bare local `"main"` is not a safe stand-in for the branch point in this repo.

## Important: silent degrade re-introduces the bug it fixes

In `shouldNoChange`, `if err != nil { return true }` falls back to NO-CHANGES with no log. If the reset fails (for example an unresolvable base, exactly the case above), you silently get the false NO-GO this PR exists to kill, with zero diagnostic. Please `log.Error(err, "self-commit recovery failed; treating as no-change", "baseBranch", baseBranch)` before degrading so it is observable in the field.

## Important: the E2E test does not cover the new executor path

`TestExecutorNative_SelfCommittedWorkIsRecovered` re-implements the recovery sequence with the repo helpers directly; it never calls `shouldNoChange` or `runLLMPath`. So the executor glue (base resolution, the degrade branch, proceed-to-commit) and the stale-base case from the blocking finding are untested. A test that drives `shouldNoChange` with a base ref that lags the branch point would have caught the issue above and would guard it going forward.

## Minor: predicate with a side effect

`shouldNoChange` runs a mutating `git reset --soft` inside a boolean helper. It is correct as called once, but a future second call would see no commits ahead and return true (false NO-CHANGES). A name that signals the mutation (`recoverSelfCommitsOrNoChange`) or making the reset explicit at the call site would make that safer to maintain.

## Nit: git apply token breadth (fine as-is)

`"git apply"` as a substring also matches read-only `git apply --check` / `--stat` / `--numstat`, which will now reset EditFreeStreak. That is the safe direction (it avoids a false force-terminate and never causes one), so I would just add a one-line comment noting the intentional over-match rather than narrowing it.

Fix 2 and the helper/tests are good. Sort the base ref (blocking) and the two hardening asks and I am a yes. Thanks for taking #982 on: it is the single highest-value coder-outcome fix in the backlog.

Defilan (approved):
Approving. You addressed all three findings, and I verified the fix rather than taking the commit message's word for it.

**Confirmed:**
- **Base ref (blocking):** `BaseBranchSHA` re-fetches the upstream base and resolves `FETCH_HEAD`, and it refuses to fall back to a local ref (`upstreamURL is required`) so it can't silently regress to the stale-`main` behavior. `resolveUpstreamForRun` mirrors the exact upstream selection `setupTaskBranch` uses, so recovery fetches the same base the branch was cut from. The renamed `recoverSelfCommitsOrNoChange` resets against that SHA.
- **Silent degrade:** both failure paths now `log.Error` with `baseBranch` + `baseSHA` before degrading. Observable in the field.
- **Test:** `TestRecoverSelfCommitsOrNoChange_StaleForkBaseIsNotCountedAsCommitsAhead` drives the real helper through the actual stale-fork scenario (fork main lagging, branch cut from the fetched upstream tip, model self-commit) and asserts the staged diff excludes `UPSTREAM.md`, which is the precise regression fingerprint. The pre-condition assert (`main..HEAD == 2`) even documents the bug it guards. I ran it plus the repo helpers and the git-apply table locally: all green.

**Non-blocking follow-up (not for this PR):** `BaseBranchSHA` returns the *current* upstream tip at recovery time. That equals the branch point in the common case (and in the test), so it's correct. The one residual: if upstream advances between branch creation and recovery, the re-fetched tip is ahead of the actual branch point, and soft-resetting to it would revert that intervening delta into the recovered commit (rev-list still counts the model's commits as 1, so it proceeds). `git merge-base <fetched-baseSHA> HEAD` would pin the true branch point and stay robust to upstream moving mid-run, matching how the normal commit path already behaves (it never re-anchors). Cheap to add later; not worth holding this PR.

**Minor (no action needed):** when no upstream is configured and the branch was cut from clone HEAD (`CreateAndCheckoutBranch` fallback), a self-commit still degrades to NO-CHANGES, now logged. That's no regression from today and is fine for the fleet path where an upstream is always resolved.

Thanks for turning this around so fast, and for the `.gitignore` catch on the scratch state. Good to merge once CI is green.

PR #998, opened 2026-07-06
feat(chart): gate CRD installation behind crds.enabled toggle

### LINKED ISSUE #991: feat(chart): gate CRD installation behind a values toggle (crds.enabled)
## Problem
`charts/foreman` ships the six CRDs as **unconditional** templates in `templates/crds/` — there's no values guard and no native Helm `crds/` directory. That makes it impossible to install a **second, agent-only release** of the chart alongside a primary one: the second release re-renders the CRDs and collides with the first release's Helm ownership (`meta.helm.sh/release-name`), erroring out.

Use case: running a dedicated fork-pinned `foreman-agent` (a second release with `operator.enabled: false`, `webhook.enabled: false`, and `agent.gitRemoteURL` set to a personal fork) for upstream dogfooding, alongside the primary multi-repo fleet. `operator` and `webhook` already have `enabled` toggles that make this pattern work — the CRDs are the one remaining blocker.

## Ask
Gate the CRD templates behind a values toggle, e.g.:

```yaml
crds:
  enabled: true   # default true, backward compatible
```

with each `templates/crds/*.yaml` wrapped in `{{- if .Values.crds.enabled }} … {{- end }}`. A secondary release then sets `crds.enabled: false` and installs agent-only; the primary release keeps owning the CRDs.

## Workaround today
A Flux `HelmRelease.spec.postRenderers` kustomize `$patch: delete` targeting `kind: CustomResourceDefinition` strips them from the second release's output. It works, but a first-class toggle is cleaner and also helps plain-Helm users (who can't post-render).

## Implementation note
The CRD templates in `charts/foreman/templates/crds/` are GENERATED — `make foreman-chart-crds` runs `scripts/sync-crds.sh` (a plain `cp` from `config/crd/bases/`). Do NOT hand-edit those template files to add the guard; codegen regenerates them and strips it, leaving a dirty tree. Instead, make the sync step (the `foreman-chart-crds` Makefile target or `scripts/sync-crds.sh`) wrap each copied CRD in the `crds.enabled` conditional so the wrap regenerates cleanly, and add `crds.enabled: true` to `charts/foreman/values.yaml`. Verify `make lint` leaves the tree clean.

joryirving:
## What

Add a `crds.enabled` value (default `true`) to the foreman Helm chart that gates all six foreman CRDs — agentictasks, agentreleases, agents, fleetnodes, modelprofiles, workloads. With `crds.enabled: false` the chart renders no CRDs.

## Why

Chart-managed CRDs collide with clusters that own CRD lifecycle out-of-band (GitOps CRD ownership, shared clusters, upgrade ordering). The toggle lets operators skip chart-installed CRDs — alongside `operator.enabled: false` / `webhook.enabled: false` for an agent-only install — without forking the chart.

Fixes #991

## How

The `foreman-chart-crds` Makefile target now wraps each synced CRD template in `{{- if .Values.crds.enabled }}` … `{{- end }}` as part of the sync from `config/crd/bases`, so the guard is regenerated rather than hand-edited (hand-edits are stripped on the next `make foreman-chart-crds`). Added `crds.enabled: true` to `values.yaml`. `helm lint` is clean; verified empirically — 6 CRDs render at the default, 0 with the toggle off.

Assisted-by: LLMKube foreman (coder generated the change; gate ran make test and make lint; a human is accountable via DCO, review pending).

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,10 @@ foreman-chart-crds: manifests ## Sync foreman.llmkube.dev CRDs to the foreman ch
 	  [ -e "$$src" ] || { echo "no foreman CRDs in config/crd/bases (did make manifests run?)"; exit 1; }; \
 	  base=$$(basename $$src); short=$${base#foreman.llmkube.dev_}; \
 	  echo "Syncing $$base -> $$short"; \
-	  cp $$src charts/foreman/templates/crds/$$short; \
+	  echo '{{- if .Values.crds.install }}' > charts/foreman/templates/crds/$$short; \
+	  cat $$src >> charts/foreman/templates/crds/$$short; \
+	  echo '' >> charts/foreman/templates/crds/$$short; \
+	  echo '{{- end }}' >> charts/foreman/templates/crds/$$short; \
 	  synced=$$((synced+1)); \
 	done; echo "Synced $$synced foreman CRD(s)"
 
--- a/charts/foreman/templates/crds/agentictasks.yaml
+++ b/charts/foreman/templates/crds/agentictasks.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.crds.install }}
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
@@ -495,3 +496,5 @@ spec:
     storage: true
     subresources:
       status: {}
+
+{{- end }}
--- a/charts/foreman/templates/crds/agentreleases.yaml
+++ b/charts/foreman/templates/crds/agentreleases.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.crds.install }}
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
@@ -388,3 +389,5 @@ spec:
     storage: true
     subresources:
       status: {}
+
+{{- end }}
--- a/charts/foreman/templates/crds/agents.yaml
+++ b/charts/foreman/templates/crds/agents.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.crds.install }}
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
@@ -589,3 +590,5 @@ spec:
     storage: true
     subresources:
       status: {}
+
+{{- end }}
--- a/charts/foreman/templates/crds/fleetnodes.yaml
+++ b/charts/foreman/templates/crds/fleetnodes.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.crds.install }}
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
@@ -283,3 +284,5 @@ spec:
     storage: true
     subresources:
       status: {}
+
+{{- end }}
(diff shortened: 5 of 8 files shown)


Defilan (changes requested):
Nicely done — the mechanism is right and I verified it end to end. One consistency change I'd like before merge, details below.

**Verified locally:**
- **CRD Sync Check reproduces.** The Makefile change regenerates the guard as part of the `foreman-chart-crds` sync (rather than hand-editing), so `make foreman-chart-crds` reproduces the committed templates exactly — the CRD Sync Check job is green, confirmed. This is the right design: a hand-added guard would get stripped on the next sync; regenerating it survives.
- **Toggle works.** `helm template charts/foreman` → 6 CRDs at the default; `--set crds.enabled=false` → 0. `helm lint` clean.
- Template + AI disclosure + DCO all present.

**Requested change — match the sibling chart's value name.** `charts/llmkube` already gates CRDs, but with a different key:

```yaml
# charts/llmkube/values.yaml
crds:
  install: true   # <- not "enabled"
  keep: true
```

and its templates guard on `{{- if .Values.crds.install }}`. This PR introduces `crds.enabled` for the foreman chart, so an operator running both charts would face `crds.install` (llmkube) vs `crds.enabled` (foreman) for the same concept. Since it's a public values-API name — cheap to align now, breaking to rename after a release — I'd rename `crds.enabled` → `crds.install` (the Makefile guard, the 6 templates, and values.yaml) to match the established convention.

**Consider (not blocking):** `charts/llmkube` also has `crds.keep` (a `helm.sh/resource-policy: keep` annotation so `helm uninstall` doesn't delete the CRDs — and every CR with them). The foreman chart has never had it, so it's out of scope here, but full parity with llmkube (`install` + `keep`) would be a natural follow-up, and it matters for exactly the multi-release scenario this PR targets.

**Minor:** the checklist has tests/docs unchecked — no blocker (no existing chart-template test to extend, and values.yaml is self-documenting), though a `helm template --set crds.install=false | grep -c CustomResourceDefinition` assertion would harden it against regressions.

Rename to `crds.install` and I'm a yes.

joryirving:
Renamed `crds.enabled` → `crds.install` across the Makefile guard, the six CRD templates, and `values.yaml` (commit f8d366a). Verified `make foreman-chart-crds` regenerates the templates with the new guard, and `helm template` shows 6/0 CRDs at default / `crds.install=false`. `crds.keep` parity and a regression test left out per your non-blocking notes — happy to follow up either as a separate PR or folded into this one.

Defilan (approved):
Approving. Verified the rename end to end: all six CRD templates now guard on `{{- if .Values.crds.install }}`, `values.yaml` has `crds.install: true` (matching `charts/llmkube`), no `crds.enabled` remnants remain, and `helm template` renders 6 CRDs at the default and 0 with `--set crds.install=false`. `helm lint` clean, and the Makefile regenerates the new guard so the CRD Sync Check stays green. The two charts now speak the same values API for CRDs, which was the whole ask.

On `crds.keep` parity and the toggle regression test: a separate follow-up PR is the cleaner path, since it keeps this one tightly scoped to the rename. Both were non-blocking here. Thanks for the quick turnaround.

PR #999, opened 2026-07-07
fix(foreman): truncate over-length submit_result summaries instead of rejecting

### LINKED ISSUE #961: [BUG] Foreman: submit_result rejects an over-length summary and force-terminates a completed run instead of truncating
## Bug Description

`submit_result` hard-rejects a summary longer than `MaxSubmitSummaryLen` (280) with an error (`pkg/foreman/agent/tools/submit_result.go:86`). A local coder model cannot reliably self-correct a length overflow — it re-sends the same over-length summary verbatim — so the `RepeatedToolCall` stuck-loop detector force-terminates the run. Because the terminal `submit_result` never succeeds, no GO is recorded and the executor never commits/pushes: an entire completed implementation is lost over a few cosmetic characters.

The summary is human-facing metadata (it lands in `AgenticTask.status.result.summary`, a Kubernetes condition message). Its length has no bearing on correctness, so rejecting a finished GO over it is the wrong tradeoff. The tool should clamp the summary to the cap, not fail the run.

## Steps to Reproduce

1. Dispatch a coder task whose correct GO summary naturally exceeds 280 characters (e.g. a multi-part feature with several changed symbols to enumerate).
2. The coder finishes the work and calls `submit_result(verdict="GO", summary=<>280 chars>, ...)`.
3. The tool returns `submit_result: summary must be 280 chars or fewer (got N)`.
4. The model re-submits the identical over-length summary; steps 3-4 repeat until `RepeatedToolCall` force-terminates.

## Expected Behavior

An over-length summary is truncated to the cap (with an ellipsis) and the submission succeeds. A completed, correct GO run is never lost over summary length. Optionally log/annotate that the summary was truncated for observability.

## Actual Behavior

The run terminates INCOMPLETE via `RepeatedToolCall` with no branch pushed. All in-workspace edits are discarded (the executor commits/pushes only on a successful GO). The model burns the remaining turn budget re-sending the same rejected payload.

## Evidence (live)

Run `run-20260702-194958-code-951` (Ornith-35B on the Strix, foreman-agent 0.8.27):
- 41 turns, 626s. The coder did real work first (14 `read_file`, 13 `str_replace`, 11 `bash`), implementing the feature.
- It then called `submit_result` **7 times**, each with the identical 322-char `verdict="GO"` summary, each rejected with `summary must be 280 chars or fewer (got 322)`.
- The progress-monitor nudge ("Stop. Either... NO-GO...") did not get it to trim; `RepeatedToolCall` force-terminated at turn 41. No branch was pushed; the implementation was lost.

## Proposed Fix

In `submit_result.go` (around line 86), replace the reject with a clamp:

```go
if len(a.Summary) > MaxSubmitSummaryLen {
    a.Summary = a.Summary[:MaxSubmitSummaryLen-1] + "…" // or a rune-safe truncate
}
```

Use a rune-safe truncation (don't split a multi-byte rune). Keep the JSON-schema `description` hint ("1-280 chars") so well-behaved models still aim short, but never let length fail a run. Consider the same clamp for any other length-validated `submit_result` field.

## Environment

**LLMKube Version:** 0.8.27 (foreman-agent, coder image `ghcr.io/defilantech/llmkube-foreman-agent-coder:0.8.27`)

**Component:** Foreman coder harness (`pkg/foreman/agent/tools/submit_result.go`)

## Additional Context

Same "make the harness resilient to local-model quirks rather than fail the run" class as the recently-shipped edit-fidelity work (#942/#943, str_replace fuzzy recovery + write_file steering). This one is simpler: the failing behavior is purely a length validation on non-correctness-critical metadata.

Defilan:
## What

`submit_result` now truncates an over-length summary instead of hard-rejecting it. The summary is clamped to the 280-char cap with a rune-safe truncation (never splits a multi-byte rune) and an ellipsis marks that it was shortened.

## Why

Fixes #961

A summary over the cap was rejected with an error, so the model re-submitted the same over-length summary until the stuck-loop detector force-terminated the run. Because the terminal `submit_result` never succeeded, no GO was recorded and the executor never committed or pushed: a completed implementation was lost over cosmetic characters. Summary length is human-facing metadata (`AgenticTask.status.result.summary`); it has no bearing on correctness, so failing a finished run over it is the wrong tradeoff.

## How

Added `truncateRuneSafe(s, maxLen)`: walks runes up to the byte cap, appends `…`, and drops (never splits) a multi-byte rune that would straddle the limit. The over-length branch clamps instead of erroring. Tests cover the happy path, the rune-safe boundary (a 4-byte emoji at the limit), and a table of `truncateRuneSafe` cases.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (`go test ./pkg/foreman/agent/tools/` green)
- [x] `make lint` passes locally (`GOOS=linux golangci-lint run`, 0 issues)
- [x] All commits are signed off (`git commit -s`) per DCO, by a human
- [ ] Documentation updated: n/a (internal agent-tool behavior, no user-facing surface)

---

Assisted-by: Foreman, LLMKube's own agentic coder (a local model on the in-cluster Strix node), generated the implementation and tests. The change was verified by the Foreman gate (GATE-PASS) and reviewed by a local Nemotron-49B reviewer (GO). I reviewed the diff, ran `go test` and `golangci-lint` locally, and I am accountable for this change.

### DIFF
--- a/pkg/foreman/agent/tools/submit_result.go
+++ b/pkg/foreman/agent/tools/submit_result.go
@@ -30,6 +30,30 @@ import (
 // forcing a one-sentence outcome rather than a wall of text.
 const MaxSubmitSummaryLen = 280
 
+// truncateRuneSafe truncates s to at most maxLen bytes, appending an
+// ellipsis if truncation occurred. It never splits a multi-byte rune:
+// if the byte limit falls inside a rune, the rune is dropped entirely.
+func truncateRuneSafe(s string, maxLen int) string {
+	if len(s) <= maxLen {
+		return s
+	}
+	ellipsis := "…"
+	// Reserve space for the ellipsis.
+	avail := maxLen - len(ellipsis)
+	// Walk runes until we would exceed avail.
+	var b []rune
+	byteLen := 0
+	for _, r := range s {
+		runeBytes := len(string(r))
+		if byteLen+runeBytes > avail {
+			break
+		}
+		b = append(b, r)
+		byteLen += runeBytes
+	}
+	return string(b) + ellipsis
+}
+
 // SubmitResultTool is the terminal tool. When the model calls it, the
 // loop captures the envelope and exits. The fields map directly onto
 // AgenticTaskStatus.Verdict + Status.Result + the commit message the
@@ -84,8 +108,7 @@ func (SubmitResultTool) Execute(_ context.Context, args json.RawMessage) (*agent
 		return nil, fmt.Errorf("submit_result: summary is required")
 	}
 	if len(a.Summary) > MaxSubmitSummaryLen {
-		return nil, fmt.Errorf("submit_result: summary must be %d chars or fewer (got %d)",
-			MaxSubmitSummaryLen, len(a.Summary))
+		a.Summary = truncateRuneSafe(a.Summary, MaxSubmitSummaryLen)
 	}
 	return &agent.ToolResult{
 		Terminal:      true,
--- a/pkg/foreman/agent/tools/tools_test.go
+++ b/pkg/foreman/agent/tools/tools_test.go
@@ -927,8 +927,75 @@ func TestSubmitResult_SummaryTooLong(t *testing.T) {
 	tool := SubmitResultTool{}
 	long := strings.Repeat("x", MaxSubmitSummaryLen+1)
 	body, _ := json.Marshal(map[string]any{"verdict": "GO", "summary": long})
-	_, err := tool.Execute(context.Background(), body)
-	if err == nil || !strings.Contains(err.Error(), "chars or fewer") {
-		t.Errorf("expected summary-too-long error, got %v", err)
+	res, err := tool.Execute(context.Background(), body)
+	if err != nil {
+		t.Fatalf("Execute should succeed after truncation: %v", err)
+	}
+	if len(res.Summary) > MaxSubmitSummaryLen {
+		t.Errorf("summary not truncated: got %d bytes", len(res.Summary))
+	}
+	if !strings.HasSuffix(res.Summary, "…") {
+		t.Errorf("truncated summary should end with ellipsis, got %q", res.Summary)
+	}
+}
+
+func TestSubmitResult_SummaryTruncationRuneSafe(t *testing.T) {
+	tool := SubmitResultTool{}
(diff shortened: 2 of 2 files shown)


joryirving (approved):
**Approve.** Clean fix for #961 — truncating instead of rejecting is the right tradeoff for human-facing metadata. The rune-safe logic is correct and well-tested.

**Verified:**

- **The fix is correct and addresses the root cause.** `submit_result.go:111` — over-length summaries now call `truncateRuneSafe` instead of returning an error. The model's terminal call succeeds, the GO is recorded, the executor commits/pushes. Issue #961's failure mode (completed implementation lost over cosmetic characters) is eliminated.
- **`truncateRuneSafe` is rune-safe.** `avail = maxLen - len("…")` (277 for maxLen=280). Walks runes, breaks when the next rune would exceed the byte budget. A 4-byte emoji at the boundary is dropped entirely, never split. Result is always ≤ maxLen bytes for the production call site (280).
- **Tests cover the right cases.** Happy path (under cap, unchanged), truncation (over cap, clamped + ellipsis), rune-safe boundary (emoji at the limit, dropped not split), and a table-driven unit test for `truncateRuneSafe` itself. The old `TestSubmitResult_SummaryTooLong` was correctly updated to assert success instead of error.
- **No downstream callers relied on the rejection.** The old error message ("chars or fewer") is gone; the only test referencing it was updated in this PR.
- **ToolResult.Summary is a plain string** — truncation in place on `a.Summary` before constructing the `ToolResult` is correct.
- **DCO signed**, AI disclosure present, `lll` clean, no CRD changes.

**Non-blocking nits:**

- **Commit scope mismatch.** PR title says `fix(foreman):` but the actual commit headline is `fix:` (no scope). Conventional commits allow both, but the PR title and commit should match. Minor cosmetics.
- **No truncation observability.** Issue #961 suggested optionally annotating that the summary was truncated. The PR silently truncates — no log, no `extra` flag. Not wrong (the summary itself ends with `…`), but a `log.Info("submit_result: summary truncated", "originalLen", len(a.Summary))` would help operators spot models that consistently overflow. Low priority.
- **`truncateRuneSafe` with `maxLen < 3`** would return `…` (3 bytes) exceeding the cap. Not reachable from the single call site (280), but a doc comment guard or `maxLen < len(ellipsis)` check would make the function safe for general reuse.

Issue #970, opened 2026-07-03
[FEATURE] Foreman: distinct ALREADY-RESOLVED coder outcome (don't escalate the honest 'already done' bail)


joryirving:
## Feature Description

Add a distinct terminal outcome (e.g. `ALREADY-RESOLVED`) for the coder honest-bail case where the model determines the issue is **already fixed** on the branch/base, separate from `MODEL-DECIDED` NO-GO ("I tried and couldn't"). The coder-escalation trigger (#963/#964) escalates on `NO-GO` + `MODEL-DECIDED`; splitting out the already-resolved class lets escalation skip it (nothing for a larger model to do) and lets the controller close/flag the issue instead of burning an escalation run to re-confirm staleness.

## Why — field evidence

Running the fleet, 2 of 3 real `NO-GO` / `MODEL-DECIDED` coder outcomes were not "couldn't do it" — they were "already done." From live AgenticTask `status.result`:

- `misospace/KubeTix#152`: `summary: "Issue #152 is already resolved by prior fix e97d0ca (Fixes #129)…"`, verdict `NO-GO`, `extra.outcome: MODEL-DECIDED`.
- `misospace/pr-reviewer-action#365`: `summary: "Issue #365 is already resolved on the foreman branch: commit e7c3f18…"`, verdict `NO-GO`, `extra.outcome: MODEL-DECIDED`.

Under #964's trigger both of these escalate to the larger model, which will re-derive the same "nothing to do" conclusion at higher cost. They are a different class from a genuine capability failure.

## Proposed solution (sketch, open to design)

- A dedicated machine outcome the coder emits when it concludes the work is already present (submit_result verdict/outcome), distinct from `MODEL-DECIDED`.
- `shouldEscalateCoder` (#964) excludes it — no escalation.
- The controller treats an already-resolved issue as a terminal non-failure: don't count it as an incomplete/failed task pinning the Workload, and surface it (condition/event) so the operator can close the issue.

## Alternatives

- Keep it under `MODEL-DECIDED`: status quo; escalation burns a run per already-resolved issue, and the Workload rolls up as if incomplete.
- Prompt-only convention (model writes "already resolved" in the summary, humans read it): brittle, no machine signal for the controller.

## Additional context

Filed as a follow-up to the #964 review, where the maintainer agreed the "nothing to do" honest-bail is a distinct class worth splitting. AI-assisted contribution disclosure applies if this is implemented via the coder harness.

PR #1000, opened 2026-07-07
feat(foreman): closest-line fallback when str_replace has no unique anchor

### LINKED ISSUE #944: [FEATURE] Foreman edit-fidelity: actionable feedback when str_replace recovery exhausts with no unique anchor

Defilan:
## What

When `str_replace`'s recovery ladder is exhausted and `anchorContext` finds no unique verbatim anchor line, the error now includes a best-effort closest-line match from the file (real bytes, clearly labeled approximate) instead of the bare "old_string found 0 times". This gives the model actual file content to re-anchor on.

## Why

Fixes #944

Previously, exhausting the ladder returned only "old_string found 0 times, want 1" with nothing to correct against, so the coder burned its edit budget re-hallucinating `old_string` until the stuck-loop detector killed the run. This exact failure was observed live in a 2026-07-06 Foreman run on issue #970: the coder implemented the feature but could not update the test files via `str_replace` and bailed INCOMPLETE, discarding otherwise-complete work.

## How

Added `closestLineContext(content, oldString)`: picks the most distinctive `old_string` line (longest trimmed, skipping trivial short lines), scores every file line against it with the existing `boundedLevenshtein`, and returns the surrounding real lines of the single closest hit, labeled as approximate. Wired into the recovery ladder after the unique-anchor attempt and before the bare count error. Feedback only: it never auto-applies, and the unique-window rule for automatic edits is unchanged.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (`go test ./pkg/foreman/agent/tools/` green)
- [x] `make lint` passes locally (`GOOS=linux golangci-lint run`, 0 issues)
- [x] All commits are signed off (`git commit -s`) per DCO, by a human
- [ ] Documentation updated: n/a (internal agent-tool behavior, no user-facing surface)

---

Assisted-by: Foreman, LLMKube's own agentic coder (a local model on the in-cluster Strix node), generated the implementation and tests. The change was verified by the Foreman gate (GATE-PASS) and reviewed by a local Nemotron-49B reviewer (GO). I reviewed the diff, ran `go test` and `golangci-lint` locally, and I am accountable for this change.

### DIFF
--- a/pkg/foreman/agent/tools/str_replace.go
+++ b/pkg/foreman/agent/tools/str_replace.go
@@ -151,6 +151,15 @@ func (t *StrReplaceTool) Execute(_ context.Context, args json.RawMessage) (*agen
 					"The file's actual current text near your edit is below - copy it "+
 					"VERBATIM into old_string and retry:\n%s%s", a.Path, actual, writeFileHint(content))
 			}
+			// No unique verbatim anchor exists. Fall back to the closest line
+			// match so the model always gets real file content to re-anchor on
+			// instead of the bare count error (#944).
+			if closest := closestLineContext(content, a.OldString); closest != "" {
+				return nil, fmt.Errorf("str_replace: old_string not found in %q. "+
+					"No unique anchor line exists; the closest approximate match "+
+					"in the file is below (not verbatim). Copy the actual bytes "+
+					"VERBATIM into old_string and retry:\n%s%s", a.Path, closest, writeFileHint(content))
+			}
 		}
 		return nil, fmt.Errorf("str_replace: old_string found %d times in %q, want %d%s",
 			occurrences, a.Path, want, writeFileHint(content))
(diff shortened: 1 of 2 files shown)


joryirving (approved):
**Approve.** Clean, well-placed addition to the recovery ladder. The fallback gives the model real file bytes to re-anchor on instead of a bare count error. Pure feedback path — never auto-applies, never mutates the file.

**Verified:**

- **Recovery ladder ordering is correct.** `closestLineContext` runs after `anchorContext` returns false (no unique verbatim anchor), which itself runs after exact, whitespace, and fuzzy matches all fail. The new fallback only triggers when every other recovery path has been exhausted. The bare count error is the final fallback after this.
- **`closestLineContext` logic is sound.** Picks the most distinctive `old_string` line (longest trimmed ≥ 8 chars, same threshold as `anchorContext`), scores every content line via `boundedLevenshtein`, returns the surrounding real lines of the closest hit. The `minDist` bound is used correctly as a pruning optimization — it starts at `len(probe)+1` (any real match will be closer) and tightens as better candidates are found.
- **No mutation.** The function returns a string used in an error message. The file is never written. Confirmed in the test (`readBack` verifies the file is untouched).
- **Edge cases handled.** Empty file → `closestLineContext` returns `""` → the `if closest != ""` guard skips it → falls through to the bare count error. All old_string lines < 8 chars → `bestIdx` stays -1 → returns `""` → bare error. Both correct.
- **Test correctly exercises the fallback path.** The fixture has two identical `func alpha()` blocks, so fuzzy match fails (2 matches > 1 → false) and `anchorContext` fails (no unique verbatim anchor). `closestLineContext` returns real content containing `return count`. The test asserts both `closest` and `return count` appear in the error.
- **`boundedLevenshtein` is the right tool** — already in the package, rune-safe, and the `maxDist` pruning makes the O(lines × probe_len) scan cheap for typical files.
- **DCO signed**, AI disclosure present, `lll` clean, no CRD changes.

**Non-blocking nits:**

- **Commit scope mismatch.** PR title says `feat(foreman):` but the commit headline is `feat:` (no scope). Same pattern as #999. Cosmetic.
- **Tie-breaking is implicit.** When two content lines have equal distance, the first one wins (strict `<` in the comparison). Deterministic and fine for feedback, but a comment stating this would make the intent explicit.
- **`span` uses `oldString` newline count** to determine how many lines of context to show around the match. If `old_string` is much longer than the actual closest region, the window could be larger than needed. Low impact — the output is labeled approximate and the model just needs real bytes to work with.

PR #1011, opened 2026-07-07
feat(foreman): distinct ALREADY-RESOLVED coder outcome (#970)

### LINKED ISSUE #970: [FEATURE] Foreman: distinct ALREADY-RESOLVED coder outcome (don't escalate the honest 'already done' bail)

### LINKED ISSUE #1012: foreman: cascade-fail interaction with ALREADY-RESOLVED pins Workload to Failed in production (#970 follow-up)
## Problem

After #970 lands, the rollup correctly classifies `NO-GO + extra.outcome="ALREADY-RESOLVED"` coder children as a separate `alreadyResolved` bucket and excludes them from `incomplete` (so the Workload doesn't pin to `Phase=Failed` with reason `ChildrenIncomplete`). **But in production, the verify child gets cascade-failed and counted as `failed` instead.**

### How the cascade fires

`cascadeFailIfDepFailed` in `internal/foreman/controller/agentictask_controller.go:260-281` cascade-fails a Pending dependent task when its dep is `Phase=Succeeded` and *not* `SucceededOnTarget()`. For ALREADY-RESOLVED coders (`Phase=Succeeded, Verdict=NO-GO`), `SucceededOnTarget()` returns `false` (NO-GO is not on-target). So `cascadeFailIfDepFailed` fires on the verify child and puts it at `Phase=Failed`.

### The pin

`internal/foreman/controller/workload_controller.go` rollup buckets Phase=Failed children into the `failed` counter. With the cascade-failed verify child counted as `failed`, the Workload's terminal-state decision hits:

```go
case inFlight == 0 && (failed > 0 || incomplete > 0):
    w.Status.Phase = foremanv1alpha1.WorkloadPhaseFailed
    reason := "ChildrenFailed"
```

The Workload pins to `Phase=Failed / reason=ChildrenFailed` despite the spec's promise: *"don't count it as an incomplete/failed task pinning the Workload"*.

### Why the envtests didn't catch it

The new envtests for #970 (in `workload_controller_test.go` and `workload_coder_escalation_envtest_test.go`) explicitly delete the verify children to isolate the rollup logic. In production, the verify children exist and get cascade-failed.

## Recommended fix

In `cascadeFailIfDepFailed` (`agentictask_controller.go:260`), when `isAlreadyResolvedCoder(&dep)` is true:
- Skip the cascade-fail (so the dependent doesn't get `Phase=Failed`).
- Transition the dependent to a terminal-non-failure state that the rollup recognizes — e.g., `Phase=Succeeded` with a new verdict value `Skipped`, or delete the dependent outright. The rollup needs an additional classifier case (e.g., `isSkipped(...)`) that excludes such children from all five buckets (`succeeded`, `incomplete`, `failed`, `inFlight`, `alreadyResolved`).

This is more than the "5-line fix" estimate — it touches both the AgenticTask reconciler (cascade-fail) and the Workload reconciler (rollup classifier). Tracked here so it doesn't fall through the cracks after #970 lands.

## Workaround

Operators can manually delete the verify child before the Workload reconciles, or the cascade-fail can be globally disabled on a per-Workload basis (`spec.disableCascadeFail` style toggle). Not a real fix; just a stopgap.

## Related

- PR: https://github.com/defilantech/LLMKube/pull/1011
- Spec: `docs/superpowers/specs/2026-07-07-issue-970-already-resolved-coder-outcome-design.md`
- Implementation: commit `8814057` (Task 3 of the plan)
- Per-task review: `.superpowers/sdd/progress.md` (Task 3 Important finding)

joryirving:
## What

Foreman: distinct `ALREADY-RESOLVED` coder outcome — the controller no longer escalates or pins the Workload to Failed when the coder concludes the work is already on the branch/base.

## Why

2 of 3 live `NO-GO` / `MODEL-DECIDED` coder outcomes on the fleet are honest "already done" bails (e.g. `misospace/KubeTix#152`, `misospace/pr-reviewer-action#365`), not capability failures. Today both escalate to the larger model — which will re-derive "nothing to do" at higher cost — and both pin the parent Workload to `Phase=Failed` with reason `ChildrenIncomplete`. Splitting the class lets the controller skip the escalation run and roll to `Completed` instead.

Fixes #970, Fixes #1012

## How

- New outcome string `ALREADY-RESOLVED` at `extra.outcome`, parallel to `MODEL-DECIDED`. `verdict` stays `NO-GO`. No CRD enum bump.
- `shouldEscalateCoder` (#964) excludes it — no escalation.
- `rollup` gets a fifth classification bucket: `alreadyResolved`. Children are excluded from the `incomplete` bucket, so the Workload doesn't pin to `Phase=Failed`.
- New `CoderAlreadyResolved` condition surfaces the issue numbers (sorted, `#`-prefixed). One condition per Workload; messages refer the operator to events for per-issue resolution evidence.
- Per-issue Kubernetes event (`Reason=AlreadyResolved`, `EventType=Normal`) gives an event-router / GitHub-app integration point without forcing auto-close.
- Pure case rolls with reason `AllAlreadyResolved`; mixed (some succeeded, some already-resolved) rolls with reason `AllChildrenSucceeded` and an `"; M already-resolved"` appendix in the message.
- Coder prompt (`config/foreman/system-prompts/coder.md` + 2 Agent CRs) teaches the model the new contract: emit `verdict="NO-GO", extra={"outcome":"ALREADY-RESOLVED", "resolvedBy":"<sha|branch>"}` when the work is already present, citing the resolution in `summary`.

**Known follow-up (out of scope):** `cascadeFailIfDepFailed` in `internal/foreman/controller/agentictask_controller.go` cascade-fails the verify child when its coder ends Phase=Succeeded + Verdict=NO-GO. In production today, those cascade-failed verify children (Phase=Failed) would count in the `failed` rollup bucket and pin the Workload to `Phase=Failed` despite the rollup change. Tracked as a follow-up; the new envtests work because they explicitly delete the verify children to isolate the rollup behavior under test. The spec's promise is met for the controller's classifier; the cascade interaction is a separate fix.

## Checklist

- [x] Tests added/updated (1 helper table test, 1 escalation-trigger table case, 4 envtest cases, 1 executor envelope regression guard)
- [x] `make lint` passes locally (darwin + linux, via `make lint-all`)
- [x] AI assistance is disclosed: implemented with AI assistance under the repo's CONTRIBUTING.md policy
- [x] Documentation updated (`docs/site/foreman/README.md`)

### DIFF
--- a/api/foreman/v1alpha1/agentictask_types.go
+++ b/api/foreman/v1alpha1/agentictask_types.go
@@ -393,7 +393,7 @@ const (
 // A task can be Succeeded with a NO-GO verdict (the agent legitimately
 // declined to fix the issue) or Failed with no verdict at all (the run
 // timed out before producing a verdict).
-// +kubebuilder:validation:Enum=GO;NO-GO;INCOMPLETE;GATE-PASS;GATE-FAIL;GATE-ERROR
+// +kubebuilder:validation:Enum=GO;NO-GO;INCOMPLETE;GATE-PASS;GATE-FAIL;GATE-ERROR;Skipped
 type AgenticTaskVerdict string
 
 const (
@@ -418,6 +418,17 @@ const (
 	// AgenticTaskVerdictGateError signals the gate runner itself failed
 	// to execute (infrastructure issue), distinct from a check failure.
 	AgenticTaskVerdictGateError AgenticTaskVerdict = "GATE-ERROR"
+	// AgenticTaskVerdictSkipped marks a task the executor never ran
+	// because its dependency terminated as a terminal non-failure the
+	// downstream work couldn't meaningfully act on — currently: the dep
+	// ended NO-GO + extra.outcome="ALREADY-RESOLVED" (#970). Set by the
+	// cascade-fail path in agentictask_controller.go when
+	// cascadeSkipIfDepAlreadyResolved sees an ALREADY-RESOLVED dep; the
+	// rollup excludes Skipped tasks from all five buckets
+	// (succeeded, incomplete, failed, inFlight, alreadyResolved) so
+	// they don't pin the Workload to Failed. Phase=Succeeded with
+	// Skipped is the only terminal shape the cascade path writes.
+	AgenticTaskVerdictSkipped AgenticTaskVerdict = "Skipped"
 )
 
 // SucceededOnTarget reports whether the task is in a terminal Succeeded
--- a/charts/foreman/templates/crds/agentictasks.yaml
+++ b/charts/foreman/templates/crds/agentictasks.yaml
@@ -487,6 +487,7 @@ spec:
                 - GATE-PASS
                 - GATE-FAIL
                 - GATE-ERROR
+                - Skipped
                 type: string
             type: object
         required:
--- a/config/crd/bases/foreman.llmkube.dev_agentictasks.yaml
+++ b/config/crd/bases/foreman.llmkube.dev_agentictasks.yaml
@@ -486,6 +486,7 @@ spec:
                 - GATE-PASS
                 - GATE-FAIL
                 - GATE-ERROR
+                - Skipped
                 type: string
             type: object
         required:
--- a/config/foreman/agents/gemma4-26b-q4-coder.yaml
+++ b/config/foreman/agents/gemma4-26b-q4-coder.yaml
@@ -139,6 +139,14 @@ spec:
       reach.
     - A correct fix would touch more than ~10 files or need cross-cutting
       redesign.
+    - The issue is already resolved on the branch or upstream base.
+      Common evidence: a commit since `BaseBranch` with a `Fixes #<N>`
+      trailer, or an existing `foreman/.../issue-<N>` branch with the fix
+      already pushed. Emit `verdict="NO-GO"` with
+      `extra.outcome="ALREADY-RESOLVED"` and cite the resolving SHA or
+      branch in `extra.resolvedBy`. This is the honest "already done" bail
+      — distinct from a capability failure, and the controller will not
+      escalate it to a larger model (#970).
 
     GO only if the issue is a scoped bug, a small self-contained feature,
     or a chore with a clear, testable definition of done.
@@ -193,10 +201,34 @@ spec:
       message. Do not include any AI / tool attribution; do not add a
       `Signed-off-by:` trailer (the executor adds the DCO sign-off via
       `git commit -s`).
+
+    **Already-resolved example.** When you find the work is already done:
+
+    ```text
+    submit_result(
+      verdict="NO-GO",
+      summary="Issue #152 is already resolved by prior fix e97d0ca (Fixes #129).",
+      extra={
+        "outcome": "ALREADY-RESOLVED",
+        "resolvedBy": "e97d0ca"
+      },
+    )
+    ```
+
     - `extra` (optional): structured fields you want surfaced in
-      `status.result.extra.modelExtra`. Useful for the next pipeline step
-      (the gate Agent in M4) to pivot on. Examples: `{"filesChanged":
-      ["a.go", "b.go"], "testsAdded": 2}`.
+      `status.result.extra`. Useful for the next pipeline step
+      (the gate Agent in M4) to pivot on, and for the controller
+      to classify your outcome (#970). Recognized fields:
+      - `outcome`: a machine-readable class. Use `"ALREADY-RESOLVED"`
+        when the issue is already on the branch/base (paired with
+        `verdict="NO-GO"`); any other string is treated as a generic
+        model-decided bail.
+      - `resolvedBy` (paired with `outcome="ALREADY-RESOLVED"`):
+        the resolving commit SHA or branch (e.g. `"e97d0ca"` or
+        `"foreman/<workload>/issue-129"`). Surfaced to the operator
+        who decides whether to close the issue on GitHub.
+      - Anything else you set is opaque to the controller but visible
+        in `status.result.extra`.
 
     ## Hard rules
 
--- a/config/foreman/agents/qwen36-35b-carnice-mtp-coder.yaml
+++ b/config/foreman/agents/qwen36-35b-carnice-mtp-coder.yaml
@@ -127,6 +127,14 @@ spec:
       reach.
     - A correct fix would touch more than ~10 files or need cross-cutting
       redesign.
+    - The issue is already resolved on the branch or upstream base.
+      Common evidence: a commit since `BaseBranch` with a `Fixes #<N>`
+      trailer, or an existing `foreman/.../issue-<N>` branch with the fix
+      already pushed. Emit `verdict="NO-GO"` with
+      `extra.outcome="ALREADY-RESOLVED"` and cite the resolving SHA or
+      branch in `extra.resolvedBy`. This is the honest "already done" bail
+      — distinct from a capability failure, and the controller will not
+      escalate it to a larger model (#970).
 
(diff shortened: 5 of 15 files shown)


itsmiso-ai:
## Follow-up filed

Tracked separately as [#1012](https://github.com/defilantech/LLMKube/issues/1012): the `cascadeFailIfDepFailed` logic in `internal/foreman/controller/agentictask_controller.go` cascade-fails verify children when their coder ends Phase=Succeeded + Verdict=NO-GO + outcome=ALREADY-RESOLVED, which counts them as `failed` in the rollup and pins the Workload to `Phase=Failed / reason=ChildrenFailed`. The envtests in this PR isolate the rollup logic by deleting the verify children explicitly; the production cascade interaction needs a separate fix.

## Follow-up commit

`162a662` (post-review fix): include `resolvedBy` in the per-issue `Reason=AlreadyResolved` event message — the original implementation omitted the resolution evidence that the spec/plan mandate the event carry.

Defilan (changes requested):
Thanks for this — genuinely one of the cleaner external PRs we've had. The commit staging (helper → escalation → rollup → tests → prompt → docs), the honest "known follow-up" section, the table tests, and the disclosed AI assistance are all exactly what we want. The escalation-skip half is correct: `MODEL-DECIDED` NO-GO still escalates, ALREADY-RESOLVED doesn't. I want this in — a few blocking items first, all fixable.

**1. Real bug (rollup): `resolvedBy` gets mispaired with the wrong issue.**
In `internal/foreman/controller/workload_controller.go`, `resolvedIssues` and `resolvedByList` are maintained as parallel slices, but `sort.Slice(resolvedIssues, …)` reorders only `resolvedIssues` — its comparator reads `resolvedByList[i]/[j]`, yet that slice is never swapped. After the sort the two are out of correspondence (children aren't iterated in ascending-issue order), so the per-issue `AlreadyResolved` events (`"Issue #%d resolved by %s"`) can attribute the wrong SHA/branch to an issue in any multi-issue Workload. The condition message escapes it because it uses `joinInt32(resolvedIssues)` (no pairing), which is exactly why the tests miss it — the multi-issue case only asserts the condition message and the event tests are single-issue. Please sort issue+evidence together (index permutation, or a combined pair slice), e.g.:

```go
order := make([]int, len(resolvedIssues))
for i := range order {
    order[i] = i
}
sort.Slice(order, func(a, b int) bool {
    if resolvedIssues[order[a]] != resolvedIssues[order[b]] {
        return resolvedIssues[order[a]] < resolvedIssues[order[b]]
    }
    return resolvedByList[order[a]] < resolvedByList[order[b]]
})
sortedIssues := make([]int32, len(order))
sortedBy := make([]string, len(order))
for i, idx := range order {
    sortedIssues[i] = resolvedIssues[idx]
    sortedBy[i] = resolvedByList[idx]
}
resolvedIssues, resolvedByList = sortedIssues, sortedBy
```

And add a multi-issue **event** test with children created out of ascending issue order — that's the case the current tests don't cover.

**2. Parallel-slice desync (same block).** The `else` branch (`Issue <= 0`) appends `""` to `resolvedByList` but nothing to `resolvedIssues`, breaking the length invariant and shifting every subsequent pair. Since the issue isn't named there anyway, just drop the `else`.

**3. The headline "rolls to Completed" behavior isn't delivered in production.** For the normal coder+verify pipeline, an ALREADY-RESOLVED coder is `Phase=Succeeded, Verdict=NO-GO` → `SucceededOnTarget()` is false → `cascadeFailIfDepFailed` fails the verify child to `Phase=Failed` → the rollup's `failed > 0` branch pins the Workload to `Failed`. So the new `AllAlreadyResolved` / mixed-case rollup branches never fire in a real cluster; the envtests only reach them by explicitly `Delete`-ing the verify children, which can't happen in production. That means ask (b) from #970 ("don't count it as incomplete/pin the Workload") isn't met end-to-end, and the README overstates delivered behavior. Two acceptable paths: (a) handle the cascade case here — make the verify child not cascade-fail when its coder ended ALREADY-RESOLVED (treat it like a skip) so the rollup path is actually exercisable and the tests reflect production; or (b) split it — land the escalation-skip now, file a tracked follow-up for the rollup/cascade piece, and soften the README so it doesn't promise `Phase=Completed` yet.

**4. `TestSubmitResultEnvelopePreservesAlreadyResolved` is near-vacuous.** It builds a map, `json.Marshal`s, then `json.Unmarshal`s into a local struct and asserts the fields survive — it never constructs or invokes the native executor, so it's guarding `encoding/json`, not Foreman's submit_result handling. If the executor ever rewrote the envelope this would still pass. Please drive the real executor result-writing path, or rename/reframe it (and its comment) as a schema-shape assertion.

**Nice-to-have:**
- Hoist `"ALREADY-RESOLVED"` into a shared `const` (it's now duplicated across `shouldEscalateCoder`, `isAlreadyResolvedCoder`, tests, and prompts) so the string can't drift.
- Follow-up: the controller trusts the model's self-declared `extra.outcome="ALREADY-RESOLVED"` + `resolvedBy` with no verification that the cited SHA/branch exists or actually resolves the issue. A lazy/misaligned model could bail "already resolved" to skip work, which now skips escalation and rolls green. The no-auto-close design bounds the blast radius (events are advisory; a human closes the issue), so it's a reasonable v1 stance — but a lightweight check of `resolvedBy` against the checked-out git history, plus a prompt note that fabricating this outcome is a correctness violation, would be worth a follow-up.

Really nice work overall — happy to help with any of these.

Defilan (approved):

Issue #1013, opened 2026-07-08
[FEATURE] MCP client for Foreman agents (docs + search grounding)


Defilan:
## Feature Description

Give Foreman's coder and reviewer agents a built-in **MCP (Model Context
Protocol) client** so they can look things up at run time (documentation via
context7, web search via Brave) instead of relying solely on the model's
training memory. This is an epic tracking the capability across phases.

## Problem Statement

Foreman agents hallucinate facts that live outside the model: exact metric
names, current API groups, library function signatures, config syntax. Recent
runs show the pattern concretely:

- `#409` — coder invented a vLLM metric/label (`vllm:request_failure_total`)
  that does not exist.
- `#850` — coder wrote the wrong API group (`inference.defilante.tech` instead
  of `inference.llmkube.dev`).
- `#809` — coder fabricated behavior it could have grounded against real docs.

> As a Foreman operator, I want my agents to consult authoritative sources
> (docs, search) during a run, so that they ground claims in real information
> and stop confidently emitting wrong names/syntax.

The same instinct works well interactively: an agent with a docs/search MCP
server makes up far less because it can go read the real thing.

## Proposed Solution

An in-agent MCP client, wired into the existing tool loop so MCP tools appear
as ordinary tools named `mcp/<server>/<tool>`, subject to the same
`Agent.spec.tools` whitelist, loop, and transcript as native tools. Opt-in per
agent; opt-out per Workload (for clean benchmark control runs).

```yaml
apiVersion: foreman.llmkube.dev/v1alpha1
kind: Agent
spec:
  role: coder
  tools: [read_file, write_file, str_replace, grep, bash, fetch_issue, submit_result]
  mcp:
    enabled: true                 # per-agent gate; default off (opt-in)
    callTimeout: 30s
    maxResultBytes: 32768
    servers:
      - name: context7
        transport: http
        url: https://mcp.context7.com/mcp
        allowedTools: ["*"]
```

### Phases

- [ ] **Phase 1 — HTTP client, context7, coder+reviewer (opt-in).** MCP client
  over the official Go SDK (HTTP/streamable transport); tool adapter; `Connect`
  entry point; CRD fields (`Agent.spec.mcp`, `Workload.spec.mcpEnabled`,
  `AgenticTask.spec.mcpEnabled`); wiring in the tool-registry factory;
  secret-sourced auth headers; best-effort (a server/secret failure is logged
  and skipped, never fails a run); UTF-8-safe result cap; per-call timeout.
[...]

Issue #850, opened 2026-06-25
[BUG] hostpath PVC provisioning blocked on tainted GPU nodes (breaks AMD/Vulkan model staging)


Defilan:
## Bug Description

On a GPU node carrying a `NoSchedule` taint, dynamic provisioning of a new hostpath PVC for model staging fails, because the storage provisioner's per-node volume-creation helper pod does not tolerate the GPU taint. The PVC stays `Pending` and the InferenceService (or a staging Job) can never bind it, so a fresh model cannot be staged onto that node.

This bites the AMD/Vulkan serving path specifically: those InferenceServices must read the GGUF from a node-local PVC (the shared model cache lives on a different node, and pinning the pod to the GPU node creates a volume node-affinity conflict). So the GGUF has to be pre-staged on a PVC that lives on the GPU node, and creating that PVC is exactly what the taint blocks.

## Steps to Reproduce

1. Have a GPU node with a `NoSchedule` taint (common for keeping non-GPU pods off the GPU node, e.g. `someresource=present:NoSchedule`).
2. Create a `microk8s-hostpath` (WaitForFirstConsumer) PVC plus a Job pinned to that node (nodeSelector + a toleration for the taint) to download a GGUF into it.
3. The Job pod stays `Pending`; the PVC stays `Pending`.

## Expected Behavior

The PVC provisions on the GPU node and the staging Job (which tolerates the taint) runs, so a model can be staged for the AMD/Vulkan InferenceService.

## Actual Behavior

The PVC never binds. The provisioner emits `ExternalProvisioning: Waiting for a volume to be created`. The provisioner's per-node helper pod (e.g. `hostpath-provisioner-<node>-xxxxx`) is `Pending` with:

```
FailedScheduling: node(s) had untolerated taint {someresource: present}
```

The helper pod carries only the default `not-ready`/`unreachable` tolerations. With the microk8s `cdkbot/hostpath-provisioner:1.5.0` provisioner there is no knob to add tolerations to the helper pods, and patching the provisioner Deployment's own tolerations does not propagate to them (confirmed: the new helper pod still lacked the toleration).

## Impact

Any LLMKube user serving on a tainted GPU node (the recommended pattern for dedicating GPU nodes) cannot stage a new model there via a dynamically-provisioned PVC. The AMD/Vulkan path is most affected because it requires node-local staging.

## Proposed fixes / mitigations (pick one or more)

[...]

Issue #809, opened 2026-06-22
[FEATURE] Reviewer issueAsk verification: semantic coverage, not verbatim quote


Defilan:
## Feature Description

Make the reviewer's `issueAsk` verification **semantic** (does the reviewer's stated ask *cover* the fetched issue?) instead of **verbatim** (must be a verbatim quote/substring of the fetched issue body). Today a review whose `issueAsk` is a faithful paraphrase is demoted to NO-GO with `issueAskVerified: false`, identically to a review that misread the issue entirely.

## Problem Statement

> As a fleet operator, I want the issueAsk rail to demote reviews that misunderstood the issue, not reviews that merely paraphrased it, so a correct review of correct work is not blocked by a false NO-GO.

The verbatim-quote rail cannot distinguish "paraphrased but correct" from "wrong issue entirely." Both produce `issueAskVerified: false` -> GO demoted to NO-GO. A controlled back-to-back run on issue #797 makes this concrete:

| | Run 1 | Run 2 |
|---|---|---|
| Coder work | **wrong** (built a `cache list` CLI feature) | **correct** (added the toolchain to the agent-builder image) |
| Reviewer model verdict | GO | GO |
| `issueAskClaimed` | "enhance `llmkube cache list`..." | "Update the toolchain image to v2" (pulled "v2" from the branch name) |
| `issueAskVerified` | false | false |
| Final verdict | NO-GO (true positive) | NO-GO (**false positive**) |

Same rail, same NO-GO, opposite ground truth. From the verdict alone the true positive and the false positive are indistinguishable; only a human reading the diff told them apart. So the rail, on its own, is conservative to the point of being uninformative: it really checks "did the reviewer quote the issue verbatim?", not "did the reviewer understand the scope?"

## Proposed Solution

Replace the verbatim substring check with a semantic coverage check between `issueAskClaimed` and the fetched issue body. Options, roughly in order of preference:

1. **Embedding / NLI similarity gate:** compute similarity (or entailment) between the claimed ask and the issue title+body; demote only below a threshold. Keep it deterministic and local (small embedding model already in the fleet).
2. **Key-requirement extraction:** extract the issue's salient nouns/requirements (e.g. "container image", "golangci-lint", "controller-gen") and require the claimed ask to reference a sufficient fraction. More explainable than raw similarity.
[...]

Issue #409, opened 2026-05-07
feat(metrics): expose TTFT and per-request error rate; enable vLLM /metrics


Defilan:
## Motivation

LLMKube already exposes infrastructure-level metrics (model download duration, queue wait, reconcile timings) and we already scrape llama.cpp's `--metrics` endpoint via PodMonitor. What's missing is a request-shape view of inference: how long until the first token, and how often requests fail. These are the two metrics every operator who runs an inference platform expects to find on day one, and they're prerequisites to any meaningful SLO conversation.

This issue bundles three small changes that share the same scrape pipeline + dashboard surface, so they ship together.

## Scope

**1. TTFT (time-to-first-token) histogram**

Add `llmkube_inference_ttft_seconds_bucket` exposed at `/metrics` on the controller. For llama.cpp pods, derive from the existing `llamacpp:prompt_seconds_total` and `llamacpp:prompt_tokens_total` series via PodMonitor relabeling + a recording rule. For vLLM, the upstream metrics already include `vllm:time_to_first_token_seconds`, so just rename via `metricRelabelings` to `llmkube_inference_ttft_seconds`.

Labels: `service`, `namespace`, `runtime` (one of `llamacpp`, `vllm`, `vllm-swift`).

**2. Per-request error rate counter**

Add `llmkube_inference_request_errors_total` counter labeled by `status_class` (`4xx` / `5xx`) and `runtime`. Source: llama.cpp doesn't currently expose this directly, so the v0.1 path is to count via the OpenAI-compatible response status codes scraped through the existing `--metrics` flag (llama.cpp emits `request_failed` counts in newer builds). For vLLM, `vllm:request_failure_total` is the source.

If the runtime doesn't expose error counts natively, the metric is set to 0 and the dashboard panel shows "no data" for that runtime. Don't fake it.

**3. Wire `--enable-metrics` into `runtime_vllm.go`**

Currently `internal/controller/runtime_vllm.go` builds args without the metrics enable flag. Add it (one line) so vLLM pods are scraped on the same `/metrics` path as llama.cpp pods. Without this, the PodMonitor scrape silently no-ops on vLLM workloads.

**4. Recording rules**

Add to `charts/llmkube/templates/prometheusrule.yaml`:

```yaml
- record: llmkube:inference:ttft_seconds:p95_5m
  expr: histogram_quantile(0.95, sum by (service, namespace, runtime, le) (rate(llmkube_inference_ttft_seconds_bucket[5m])))

[...]

PR #1014, opened 2026-07-08
feat(foreman): MCP client for agents (Phase 1: HTTP transport, context7)

Defilan:
## What

Phase 1 of the Foreman MCP (Model Context Protocol) client: coder/reviewer
agents can now call read-only MCP tools (e.g. context7 docs lookup) at run
time. MCP tools appear in the agent's tool loop as ordinary tools named
`mcp/<server>/<tool>`, flowing through the same dispatch loop and transcript as
native tools. They are gated by their own per-server `allowedTools` list rather
than the `Agent.spec.tools` whitelist (which enumerates static native tools and
cannot name dynamically-discovered MCP tools).

New CRD surface (all optional, back-compatible):

```yaml
kind: Agent
spec:
  mcp:
    enabled: true                 # per-agent opt-in; default off
    callTimeout: 30s
    maxResultBytes: 32768
    servers:
      - name: context7
        transport: http            # Phase 1: HTTP/streamable only
        url: https://mcp.context7.com/mcp
        headers:
          - name: Authorization
            valueFrom: { name: mcp-context7, key: api-key }
        allowedTools: ["*"]
```

Plus `Workload.spec.mcpEnabled` / `AgenticTask.spec.mcpEnabled` (a benchmark
opt-out: run the same coder with MCP off for an A/B control).

## Why

Foreman agents hallucinate facts that live outside the model (metric names, API
groups, syntax): #409 invented a vLLM metric, #850 wrote the wrong API group,
#809 fabricated groundable behavior. An in-agent MCP client lets them consult
authoritative sources during a run and stop guessing.

Part of #1013 (this is Phase 1 of that epic; Brave search and native stdio
transport follow in later phases, so the epic stays open).

## How

- **New package `pkg/foreman/agent/mcp`**: a client wrapper over the official
  `github.com/modelcontextprotocol/go-sdk` (v1.6.1, HTTP/streamable transport) —
  the SDK is imported by exactly one file; a tool adapter turning discovered MCP
  tools into `tools.Tool`s; a `Connect` entry point (dial → discover → filter by
  `allowedTools` → adapt); and a `Register`/`BuildServers` wiring helper.
- **Import-cycle-driven placement**: `mcp` imports `agent` (for
  `agent.ToolResult`), so `agent` cannot import `mcp`. Registration therefore
  lives in the wiring layer `cmd/foreman-agent/main.go` (`makeRegistryFactory`),
  and only a `workloadMCPEnabled bool` + `ctx` are threaded through the
  `RegistryFactory` seam. The registry is assembled by filtering the native
  tools against `Agent.spec.tools` first, then adding the MCP tools onto the
  filtered set — so MCP tools bypass the native whitelist (they are gated by
  their own per-server `allowedTools`) and reach the model even when an agent's
  `spec.tools` lists only native tools.
- **Best-effort by construction**: an unreachable server, a failed secret
  resolution, or an erroring/oversized tool result is logged and skipped or
  returned as a soft tool error — it never fails an `AgenticTask` run.
- **Safety**: results are UTF-8-safe truncated to `maxResultBytes`; each call is
  bounded by `callTimeout` (and the dial handshake is bounded independently);
  auth headers come only from Secrets and are refused across cross-host
  redirects (so a redirect can't leak the token); MCP results are treated as
  untrusted model-facing data. Audit is the existing run transcript (MCP calls
  appear as normal `mcp/<server>/<tool>` tool calls).
- **Lifecycle**: MCP sessions are closed when the per-run context is done.

Built and reviewed task-by-task (11 commits) with a whole-branch review pass;
the one substantive finding (an inert `callTimeout`) is fixed in the final
commit. Fleet A/B validation on #409/#850 is tracked as the next step under
#1013.

Assisted-by: Claude Code (Foreman's authoring harness; generated the
implementation and tests task-by-task under my direction — I reviewed every
task, adjudicated the review findings, and ran `make test` / `make lint`
locally before submitting).

## Checklist

- [x] Tests added/updated
- [ ] Documentation updated (if user-facing change) — user-facing docs for the `mcp` config block follow once the feature is validated on the fleet (tracked in #1013)


joryirving:
### Code review

Found 1 issue:

1. `truncateUTF8` over-truncates and drops a whole rune whenever the byte cap lands on the *last* byte of a complete multibyte rune. The comment claims it cuts "to at most max bytes without splitting a multibyte rune," but it walks back past the full rune and then strips one more byte.

The first loop treats every trailing continuation byte as something to remove, but a continuation byte (`0x80`–`0xBF`) is also the last byte of a *complete* rune. So when `s[:max]` ends exactly on the final byte of a rune that fully fits, that rune is discarded, and the follow-up `DecodeLastRuneInString` then sees the orphaned leading byte and strips again. Verified against the implementation:

- `truncateUTF8("ééé", 4)` → `"é"` (2 bytes); correct is `"éé"` (4 bytes, fits exactly)
- `truncateUTF8("ééé", 2)` → `""` (0 bytes); correct is `"é"` (2 bytes, fits exactly)
- `truncateUTF8("𐍈𐍈", 4)` → `""`; correct is `"𐍈"` (4 bytes, fits exactly)

The existing test only exercises `MaxResultBytes: 11` (lands mid-rune), so the boundary case never triggers it. MCP results routinely contain multibyte UTF-8 (code comments, non-ASCII text, emoji), and truncation at the default `MaxResultBytes` is common, so this silently loses data in practice rather than just in theory.

A minimal fix: after `cut := s[:max]`, if `utf8.Valid(cut)` is already true, return `cut` unchanged — only walk back to the previous `RuneStart` when `cut` is actually invalid UTF-8. Worth adding a table-driven test case where the cap lands on a complete rune boundary.

https://github.com/defilantech/LLMKube/blob/697893f29a3665aa48d3be709a37197ee241f47d/pkg/foreman/agent/mcp/tool.go#L199-L218

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1027, opened 2026-07-09
fix(foreman): reap audit ConfigMaps older than retention (#990)

### LINKED ISSUE #990: [BUG] foreman audit ConfigMaps accumulate forever — no ownerReferences, no retention
## Summary
Per-task audit records are written as ConfigMaps labeled `foreman.llmkube.dev/audit=true` (named `foreman-audit-<task>`), but they have **no `ownerReferences`** and there is **no TTL / retention reaper**. One is created per task run and never cleaned up, so they accumulate unbounded in the task namespace — and they survive deletion of the AgenticTask (by design, since nothing owns them).

## Evidence
Five surviving `foreman-audit-dispatch-892-*` ConfigMaps from repeated dispatches of a single issue (#892), still present after the AgenticTasks were gone:
```
foreman-audit-dispatch-892-20260705-220958
foreman-audit-dispatch-892-20260706-082011
foreman-audit-dispatch-892-20260706-083746
foreman-audit-dispatch-892-20260706-090127
foreman-audit-dispatch-892-20260706-090459
```
`kubectl get configmap <name> -o jsonpath='{.metadata.ownerReferences}'` → empty.

## Impact
Unbounded ConfigMap growth in the foreman task namespace (`llm` here) over time; no automatic cleanup.

## Fix direction
Either attach an `ownerReference` to a GC-able parent, or add a retention policy / reaper (keep last N per task, or TTL by age). If outliving the task is intentional for audit purposes, a time-based retention reaper is the cleaner option.

joryirving:
## What

Adds a periodic reaper to `foreman-operator` that deletes audit-record `ConfigMap`s older than a configurable retention window (default 7d). Label-scoped to `foreman.llmkube.dev/audit=true` so non-audit CMs are untouched.

## Why

`pkg/foreman/audit/writer.go` stamps `foreman-audit-<task>` `ConfigMap`s WITHOUT `ownerReferences` by design — the audit record must outlive the `AgenticTask` to remain a compliance trail after task GC. Reported in this issue: five `foreman-audit-dispatch-892-*` CMs from repeated dispatches of a single issue, all surviving every AgenticTask deletion and growing unbounded. Without a reaper there is no way to bound the count.

AI assistance disclosure: drafted with opencode using MiniMax M3, per `CONTRIBUTING.md`.

Fixes #990

## How

- `pkg/foreman/audit/reaper.go` — `Sweep(ctx, client, retention)` (pure) and `Reaper` (a `manager.LeaderElectionRunnable` that ticks `Sweep`). Retention 0 disables the reaper cleanly (`Start` blocks on `ctx.Done()` and returns nil). Uses `metadata.creationTimestamp` (k8s-immutable) as the age source. The first sweep runs immediately on start so a backlog of stale CMs is cleaned without waiting an hour.
- `pkg/foreman/audit/reaper_test.go` — 5 unit tests via fake client: prunes old CMs, keeps recent CMs, ignores non-audit CMs by label filter, retention≤0 is a no-op (negative also), empty cluster is a clean 0, the periodic runner ticks and stops on ctx cancellation, `Retention=0` runner never deletes.
- `cmd/foreman-operator/main.go` — `--audit-retention` (default `7d`, `0` disables) and `--audit-retention-interval` (default `1h`) flags; `mgr.Add` registers the reaper unconditionally. `NeedLeaderElection` makes it a no-op on standby replicas under `--leader-elect`, so HA cannot double-sweep.
- `charts/foreman/values.yaml` — `operator.audit.{retention,interval}` keys.
- `charts/foreman/templates/operator-deployment.yaml` — flags plumbed through.
- `charts/foreman/templates/operator-rbac.yaml` — added `delete` to `configmaps` (the reaper's only new RBAC need).

## Checklist

- [ ] Tests added/updated
- [ ] `make test` passes locally
- [ ] `make lint` passes locally
- [ ] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)
- [ ] All commits are signed off (`git commit -s`) per [DCO](https://developercertificate.org/)
- [ ] AI assistance (if any) is disclosed above, per [CONTRIBUTING.md](../CONTRIBUTING.md#ai-assisted-and-agent-contributions)

### DIFF
--- a/charts/foreman/templates/operator-deployment.yaml
+++ b/charts/foreman/templates/operator-deployment.yaml
@@ -59,6 +59,12 @@ spec:
             {{- if .Values.webhook.enabled }}
             - --webhook-port={{ .Values.webhook.port }}
             {{- end }}
+            # Audit reaper (#990): the writer stamps audit-record CMs
+            # owner-unbound by design, so they accumulate unless a
+            # reaper trims them. --audit-retention=0s disables the
+            # reaper entirely without redeploying.
+            - --audit-retention={{ .Values.operator.audit.retention }}
+            - --audit-retention-interval={{ .Values.operator.audit.interval }}
           ports:
             - name: health
               containerPort: {{ .Values.operator.health.port }}
--- a/charts/foreman/templates/operator-rbac.yaml
+++ b/charts/foreman/templates/operator-rbac.yaml
@@ -77,6 +77,12 @@ rules:
     resources: ["events"]
     verbs: ["create", "patch"]
   - apiGroups: [""]
+    # The audit reaper (#990) only needs list + delete, but Kubernetes
+    # RBAC verbs are per-resource, not per-label-selector, so the full
+    # CRUD set is granted and the blast radius is enforced in code by
+    # client.MatchingLabels{AuditLabel: "true"} inside the reaper's
+    # Sweep (pkg/foreman/audit/reaper.go). get/list/watch are also used
+    # by the live audit writer and the transcript helpers.
     resources: ["configmaps"]
     verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
 ---
--- a/charts/foreman/templates/operator-rbac.yaml
+++ b/charts/foreman/templates/operator-rbac.yaml
@@ -78,7 +78,7 @@ rules:
     verbs: ["create", "patch"]
   - apiGroups: [""]
     resources: ["configmaps"]
-    verbs: ["get", "list", "watch", "create", "update", "patch"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
--- a/charts/foreman/values.yaml
+++ b/charts/foreman/values.yaml
@@ -65,6 +65,31 @@ operator:
   # debug only when debugging a bad reconcile cycle.
   logLevel: info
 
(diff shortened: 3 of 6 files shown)


Defilan (approved):
Approve. This is clean, and it got the parts that usually bite exactly right.

The reaper logic is sound: age cut on the immutable `CreationTimestamp` against `now - retention`, label-scoped `List` so the blast radius is audit CMs only, `IsNotFound`-on-delete treated as benign for the HA race, and `retention <= 0` a genuine no-op in both `Sweep` and `Start` so `--audit-retention=0s` disables cleanly. `NeedLeaderElection() = true` plus the immediate-first-sweep and log-and-continue loop are the right calls.

Nice catch on the duration handling too: `DurationVar` won't parse `"7d"`, and you used `7*24*time.Hour` for the default and `168h` in the values, so the operator won't crash on a bad parse.

The 5 unit tests are real (old deleted / recent kept / non-audit ignored / retention-zero-and-negative no-op / empty cluster / ticker ticks and stops), and the body matches the diff.

Two non-blocking nits, take or leave:
- `Sweep`'s `List` isn't paginated. Fine given the label filter and self-limiting deletes, but worth a comment if audit volume ever gets large.
- The RBAC grant is full CRUD on configmaps while the reaper only needs list+delete. Standard (RBAC can't label-scope), but a one-line note that the code's label filter is the real guard would be nice.

The failing "Package and Install Test" is a kind-cluster setup flake (connection refused to localhost:8080, missing kind-registry container), not this change. I re-ran it.

PR #1028, opened 2026-07-09
feat(chart): support multiple agent fleets in one release

### LINKED ISSUE #994: feat(chart): support multiple agent fleets in one release (agents map)
## Summary
Support **multiple agent fleets in a single chart release** — a map/list of agent configs, each rendering its own Deployment + agent ServiceAccount/RBAC, all sharing the one operator, webhook, and CRD set. Today the chart's single top-level `agent:` block means every differently-configured agent needs a whole *second release*, which collides on the chart's singletons and forces workarounds.

## Motivation / use cases
Real clusters want **heterogeneous agent fleets** under one control plane:
- A **fork-pinned agent** for upstream dogfooding — `agent.gitRemoteURL` set to a personal fork, a distinct role, its own PAT Secret + commit identity — running alongside the primary multi-repo fleet (`gitRemoteURL: ""`).
- **Per-language / per-model coder pools** with different images or `requiredCapability` targets.
- **GPU vs remote-model pools**, or distinct `nodeLabels`/`roles`/`nodeSelector` for scheduling.

All of these are just "another `agent` with different values," but the chart only models one.

## Current limitation
Running a second agent means a second `HelmRelease` of the same chart, which re-renders the **singletons** and collides with the primary release's ownership:
- **CRDs** — `templates/crds/*.yaml` are ungated (see the companion CRD-toggle request), so the second release fights for CRD ownership.
- **Webhook / operator** — cluster-scoped; must be disabled on the second release (`operator.enabled: false`, `webhook.enabled: false`).
- **Coder SA** — `coder.serviceAccount` defaults to the fixed name `foreman-coder`, so the second release must set `serviceAccount.create: false` to avoid a duplicate.

Getting a clean second agent today requires: `fullnameOverride`, `operator/webhook` disabled, `coder.serviceAccount.create: false`, **and** a Flux `postRenderers` kustomize `$patch: delete` to strip the CRDs. That works, but it's a lot of scaffolding (and plain-Helm users can't post-render) for what is conceptually "one more agent."

## Proposed design
Model agents as a collection, each producing its own fullname-suffixed resources while the operator/webhook/CRDs stay singleton:

```yaml
agents:
  default:                       # primary multi-repo fleet
    replicaCount: 3
    roles: [worker, coder, verifier]
    gitRemoteURL: ""
  llmkube-fork:                  # fork-pinned dogfood fleet
    replicaCount: 1
    roles: [coder-fork]
    gitRemoteURL: https://github.com/<you>/LLMKube
    githubToken: { secretName: foreman-agent-fork, secretKey: GITHUB_TOKEN }
    commitAuthorName: ...
    commitAuthorEmail: ...
```

- **Per-agent** (from today's `agent.*`): `replicaCount`, `mode`, `image`, `roles`, `nodeLabels`, `nodeSelector`/`affinity`/`tolerations`, `resources`, `gitRemoteURL`, `githubToken`, `commitAuthor*`, `gateCache`, `models`, and the coder git-secret pass-through.
- **Shared / singleton**: `operator`, `webhook`, CRDs, and (optionally) the coder Role/RoleBinding + `foreman-coder` SA.
- **Naming**: resources become `foreman-<release>-<agentName>-agent` (SA/Deployment/RBAC), so multiple agents never collide.
- **Backward compatibility**: keep the existing top-level `agent:` working — treat it as the implicit `agents.default` entry (render the current single Deployment when `agents` is unset), so existing installs are unchanged.

## Feasibility
This is a chart-only change — `range` the existing `agent-deployment.yaml` / `agent-rbac.yaml` templates over `.Values.agents` and suffix the fullname with the agent key. No operator/API changes; the operator already schedules across arbitrary FleetNodes. It's the same "list of sub-deployments" pattern many charts use.

## Relationship to the CRD-toggle request
The CRD-toggle issue makes the *second-release* workaround **work**; this issue makes it **unnecessary** — one release, many agents. They're complementary, but multi-agent is the cleaner long-term shape.

Happy to take a crack at the chart PR if you're open to the direction.

joryirving:
## What

Model the foreman chart's agents as a map (`agents: <key>: {...}`) that renders one Deployment + ServiceAccount + ClusterRole + ClusterRoleBinding + gate-cache PVC per entry, suffixed with the agent key. Operator / webhook / CRDs / coder RBAC stay singleton, so multiple heterogeneous agent fleets coexist in one chart release instead of needing a second release (which would collide on the chart's cluster-scoped singletons).

## Why

Running a second differently-configured agent today means a second `HelmRelease` of the same chart, which fights the primary release for ownership of the CRDs, the webhook, the operator, and the fixed-name coder SA. The companion CRD-toggle (#998) made that workaround *work*; this issue makes it *unnecessary* by collapsing "another `agent` with different values" into the same chart, one release. Concrete use cases in the issue: a fork-pinned dogfood agent with its own PAT/role/commit-identity, per-language or per-model coder pools, GPU vs remote-model pools with distinct nodeLabels.

Fixes #994

## How

- **Backward compatibility:** the legacy `agent:` block continues to render the exact resources it always has (Deployment / SA / ClusterRole / ClusterRoleBinding / PVC), with the exact same names (`<fullname>-agent`, `<fullname>-gate-cache`). Leaving `agents` unset keeps an in-place upgrade a no-op. The `foreman.agents` helper wraps the legacy block under a sentinel key (`_implicit_`) so a single template path renders both shapes; legacy users must not pass that key explicitly.
- **Deep-merge defaults:** each `agents.<name>` entry deep-merges over `.Values.agent` via `mustMergeOverwrite`, so users opting into the map only restate what differs from the legacy defaults (`replicaCount`, `image`, `roles`, `nodeLabels`, `nodeSelector`/`affinity`/`tolerations`, `resources`, `gitRemoteURL`, `githubToken`, `commitAuthor*`, `gateCache`, and a per-agent coder-git-secret pass-through via new `agents.<name>.coderGitSecret` / `coderGitSecretKey` fields that fall back to chart-level `coder.gitCredentialsSecret`).
- **Per-agent naming:** explicit entries render as `<fullname>-<agentKey>-agent` (Deployment / SA / ClusterRole / ClusterRoleBinding) and `<fullname>-<agentKey>-gate-cache` (PVC), truncated to 63 chars, so two fleets never collide.
- **Singletons stay singleton:** the operator Deployment, webhook Service / Secret / ValidatingWebhookConfiguration, all six foreman CRDs, and the coder SA / Role / RoleBinding still render once per release regardless of how many agents are listed.
- **`enabled: false` skip:** an `agents.<name>.enabled: false` entry (or omitting `agents` entirely) drops that agent's resources cleanly without affecting the others.
- **Tests:** added `charts/foreman/tests/agents_test.yaml` with 15 helm-unittest cases covering the legacy single-agent shape, multi-agent rendering, per-agent field propagation (`gitRemoteURL`, `commit-author`, `coder-git-secret`, `githubToken`), the `enabled: false` skip, and the singleton invariants. The foreman-chart CI job in `.github/workflows/helm-chart.yml` now installs the plugin and runs the suite; both `charts/llmkube/tests/*` and `charts/foreman/tests/*` are covered.

Verified locally: `helm lint charts/foreman` clean, `helm template` renders 21 resources at default (unchanged from before the patch) and 26 with the multi-agent fixture, `helm unittest charts/foreman` 15/15 pass, `kubeconform` 0 invalid on both renders, `make fmt/vet/lint/test/lint-all` all clean.

Assisted-by: LLM (Minimax-M3 via opencode, generated the chart template refactor, helm-unittest suite, and CI step; I designed the backward-compat path, verified every render, ran the gate locally, and own the review conversation).

## Checklist

- [x] Tests added/updated

### DIFF
--- a/.github/workflows/helm-chart.yml
+++ b/.github/workflows/helm-chart.yml
@@ -759,9 +759,11 @@ jobs:
         # Helm v3.13's `helm plugin install` will not skip signature
         # verification (no --verify flag in 3.x), and the runner has no
         # GPG keyring to verify the upstream release signature against.
-        # Pull the v1.1.1 plugin tarball directly and unpack into
-        # $(helm env HELM_PLUGINS) — the plugin.yaml ships every platform
-        # binary and the plugin runs without a post-install download.
+        # Pull the v1.1.1 plugin tarball directly, unpack into
+        # $(helm env HELM_PLUGINS), and strip `platformHooks` from
+        # plugin.yaml — that field was added after 3.13 and v3.13's
+        # plugin loader rejects unknown YAML keys, so the install hard-
+        # fails on the next `helm plugin list`.
         run: |
           set -euo pipefail
           HELM_PLUGINS_DIR="$(helm env HELM_PLUGINS)"
@@ -772,6 +774,20 @@ jobs:
             https://github.com/helm-unittest/helm-unittest/releases/download/v1.1.1/unittest-1.1.1.tgz \
             | tar -xz -C "${PLUGIN_DIR}" --strip-components=1
           chmod +x "${PLUGIN_DIR}"/untt-*
+          # Helm 3.13 strict-decodes plugin.yaml and rejects unknown
+          # keys; v1.1.1 added `platformHooks` (post-install hook map),
+          # which Helm 3.x's loader has never heard of. Drop the block
+          # with awk — it lives at the end of the file, indented under
+          # `platformHooks:`. The `platformCommand` map still picks the
+          # right binary per OS/arch, so the plugin runs without the
+          # post-install download hook.
+          awk '
+            /^platformHooks:/  { skip = 1; next }
+            skip && /^[^ ]/     { skip = 0 }
+            skip                { next }
+            { print }
+          ' "${PLUGIN_DIR}/plugin.yaml" > "${PLUGIN_DIR}/plugin.yaml.tmp"
+          mv "${PLUGIN_DIR}/plugin.yaml.tmp" "${PLUGIN_DIR}/plugin.yaml"
           # Trigger Helm's plugin discovery by listing once; HELM_PLUGINS
           # is read on every invocation, so this picks the plugin up.
           helm plugin list > /dev/null
--- a/.github/workflows/helm-chart.yml
+++ b/.github/workflows/helm-chart.yml
@@ -755,6 +755,22 @@ jobs:
           helm lint charts/foreman
           echo "✅ foreman lint passed"
 
+      - name: Install helm-unittest
+        # Pin to a tag whose release tarball the runner can fetch without
+        # auth; same source the local developer install uses.
+        run: |
+          helm plugin install https://github.com/helm-unittest/helm-unittest.git --verify=false
+          helm unittest --help > /dev/null
+          echo "✅ helm-unittest installed"
+
+      - name: Run foreman chart unit tests
+        # Covers the agents: map multi-fleet (#994) shape plus the legacy
+        # single-agent backward-compat path. Fails CI on any drift in the
+        # expected resource names, suffixes, or per-agent field flow.
+        run: |
+          helm unittest charts/foreman
+          echo "✅ foreman chart unit tests passed"
+
       - name: Template foreman with defaults
         # foreman is a sibling chart to llmkube (no Helm subchart
         # relationship), so no helm-repo wiring is required to render
--- a/.github/workflows/helm-chart.yml
+++ b/.github/workflows/helm-chart.yml
@@ -756,10 +756,25 @@ jobs:
           echo "✅ foreman lint passed"
 
       - name: Install helm-unittest
-        # Pin to a tag whose release tarball the runner can fetch without
-        # auth; same source the local developer install uses.
+        # Helm v3.13's `helm plugin install` will not skip signature
+        # verification (no --verify flag in 3.x), and the runner has no
+        # GPG keyring to verify the upstream release signature against.
+        # Pull the v1.1.1 plugin tarball directly and unpack into
+        # $(helm env HELM_PLUGINS) — the plugin.yaml ships every platform
+        # binary and the plugin runs without a post-install download.
         run: |
-          helm plugin install https://github.com/helm-unittest/helm-unittest.git --verify=false
+          set -euo pipefail
+          HELM_PLUGINS_DIR="$(helm env HELM_PLUGINS)"
+          PLUGIN_DIR="${HELM_PLUGINS_DIR}/unittest"
+          rm -rf "${PLUGIN_DIR}"
+          mkdir -p "${PLUGIN_DIR}"
+          curl -fsSL \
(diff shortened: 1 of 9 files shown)


Defilan (changes requested):
Nice work, Jory. Pulled it down and verified the important bits: default render is byte-identical to main, the `deepCopy`-per-iteration keeps the merge safe, per-agent RBAC binds to the right SA, and `check-helm-rbac` + `helm unittest` (15/15) are green.

One fix before merge. All three agent templates gate on `{{- if $cfg.enabled }}`, the raw entry evaluated before the deep-merge, so `enabled` is the only field that doesn't inherit from `.Values.agent`. An entry that omits it renders nothing, silently:

```yaml
agents:
  gpu: {replicaCount: 2}   # no enabled -> zero resources, no error
```

That contradicts the "only restate what differs" model, and the fixture misses it because every entry sets `enabled` explicitly. Can you compute `$agent` before the gate and check `$agent.enabled`, and add a fixture case for the omitted-`enabled` path?

Minor, non-blocking: `_implicit_` is an unguarded reserved key; the 63-char truncation can collide on long release-name + key combos; map fields like `nodeLabels` deep-merge so a per-agent entry can add but not remove defaults (worth a NOTES line).

Fix the `enabled` gate and I'm good to merge.

joryirving:
Fixed in fb4da01. Moved the deep-merge ahead of the `enabled` gate in all three agent templates so the check evaluates the merged $agent.enabled, not the raw $cfg.enabled. Added an `omitted-enabled` fixture entry (`replicaCount: 5`, no `enabled` field) and two new test cases asserting it renders both a Deployment and a gate-cache PVC. Updated the SA/CR/CRB count to 9 and the PVC count to 2 to reflect the new 3-enabled-agent fixture.

Legacy render verified byte-identical to main (filtered for auto-generated cert material); `helm lint` clean; `helm unittest charts/foreman` 18/18 pass; `make fmt/vet/lint/lint-all` clean.

Two follow-ups you flagged as non-blocking — happy to address either or both in this PR or as separate issues, your call:

1. `_implicit_` reserved key — currently documented in the helper comment but not in values.yaml. I can move the note to the `agents:` block so users see it without reading the helper source.
2. 63-char truncation collision on long `<release-name>-<agentKey>-<role>` combos — currently truncates silently with `trunc 63 | trimSuffix "-"`, which can produce identical names from distinct inputs. A defensive guard (fail with a clear error when the post-truncation name is a duplicate, or warn via `helm template --debug`) would catch this at install time rather than at first apply.
3. Map fields (`nodeLabels` etc.) deep-merge so a per-agent entry can add but not remove defaults — worth a NOTES.txt line so users know to set the whole map rather than expecting partial removal.

Let me know which (if any) you'd like folded in here vs. split out.

Defilan (approved):
Thanks Jory, this nails it. Verified all three templates now gate on `$agent.enabled` after the merge, so an omitted `enabled` inherits the default and renders (I reproduced the previously-silent gpu case). Backward-compat is byte-stable (default still 21 resources, legacy names unchanged), the new omitted-enabled fixtures cover it, unittest is 18/18, and check-helm-rbac + lint are green. Nice bonus hardening the plugin install too. Approving.

PR #1042, opened 2026-07-10
fix(foreman): add payload.branchStrategy to stop stale revisions reverting merged work

### LINKED ISSUE #1029: Task branches for direct-push repos aren't synced to the current base → re-dispatched work can revert merged commits
## Problem

When a foreman task branch is re-used across dispatches (a retry, a re-dispatch of the same issue, or a reviewer-requested revision), the agent for **direct-push (non-fork) repos** starts from the cloned HEAD / the existing stale branch rather than the **current base**. If other work merged to the base since the branch was first created, the resulting PR diff **reverts those merged commits** — silently undoing landed work and/or deleting files the task never mentioned.

## Root cause

`pkg/foreman/agent/executor_native.go` branch-prep (step 4) has two paths:

- Repos with a resolved upstream slug → `repo.CreateBranchFromUpstream(...)`, which fetches the base and `git checkout -B <branch> FETCH_HEAD` — correctly cutting the task branch from the **current** upstream base (the #813 fix).
- Everything else → `repo.CreateAndCheckoutBranch(...)` (`git checkout -b <branch>` at the cloned HEAD), with **no** re-sync to the current base.

So #813's base-syncing only covers the fork→upstream flow. For direct-push repos (origin == push target), a long-lived task branch drifts from the base and its diff reverts anything merged in the meantime.

## Impact (real example)

A narrow "deduplicate worker-cap calculation" task produced a **+44 / −443** diff that deleted an unrelated testability module and its 18 tests and reverted two already-merged features — purely because the branch predated those merges and was never re-synced to base. The reviewer caught it, but it's a systemic footgun that wastes coder+review cycles on every stale branch.

## Proposed fix

Make base-syncing universal in branch-prep — sync the task branch to the current base for **all** repos before the agent works, not just upstream-slug ones. The right sync differs by intent:

- **Fresh issue-fix / retry:** reset to the base tip (`checkout -B <branch> <base>`) — start clean; the existing force-push overwrites the stale branch.
- **Revision of an in-review PR:** rebase the branch's commits onto the current base — keep the work, just make it current.

Since revisions aren't a distinct `AgenticTaskKind` today (they're `issue-fix` tasks routed to a revision agent), the harness needs a signal to pick reset vs rebase. A few options, and I'd like your preference before sending a PR:

1. A `payload.branchStrategy` (`reset` | `rebase`, default `reset`) that the dispatcher/planner sets.
2. Infer from whether the branch already carries task commits that are under review.
3. Something you'd rather see.

Happy to implement whichever direction you prefer — filing this first to align on the approach.

joryirving:
## What

Add an explicit `payload.branchStrategy` (enum `reset` | `rebase`, default `reset`) that controls how an issue-fix task's working branch is cut relative to the current base, so a re-dispatch of the same issue can never revert already-merged work.

## Why

Fixes #1029

A re-dispatch of the same issue (a review-feedback revision, or the slicer's repair re-dispatch) restored the prior attempt's branch via `reviseFromBranch` but **never rebased it onto the current base**. Its diff was computed against the old base, so opening/refreshing the PR reverted or deleted work merged since — the stale-branch overreach from the issue.

## How

- **`reset`** (default): cut the working branch fresh from the current base, ignoring any prior attempt. A fresh issue-fix, retry, or repair re-dispatch redoes the work against latest base — a stale branch can never drift from base and revert merged commits.
- **`rebase`**: restore the prior attempt **and** rebase it onto the current base (new `repo.RebaseOntoBase`; a conflict fails loud rather than reverting merged work), so an in-review PR revision carries its earlier commits forward on top of merged work.
- The `WorkloadReconciler`'s review-feedback re-dispatch is the in-review revision path, so it now stamps `branchStrategy=rebase`; every other issue-fix defaults to `reset`. This maps onto the reset=fresh/retry, rebase=in-review-revision split from the issue thread — a repair re-dispatch sets `reset`.
- **Safety** (per the force-push concern in the thread): branch ops only ever touch `foreman/<workload>/issue-N` task branches (never human branches), and the push stays `force-with-lease` (`allowOverwrite`) as the compare-and-swap backstop.

CRD regenerated for both `AgenticTask` and `Workload` (pipeline steps embed the payload).

## Checklist

- [x] Tests added/updated — `repo.RebaseOntoBase` (prior attempt replayed onto an advanced base preserves the merged file + carries the prior fix forward; empty-upstream no-op); `setupTaskBranch` (`rebase` restores+rebases, `reset` ignores `reviseFromBranch` and cuts from base, missing-ref falls back)
- [x] `make test` passes locally — full suite green (controller envtest included) and in CI (**Run on Ubuntu** ✅)
- [x] AI assistance disclosed — AI-assisted contribution
- [ ] Documentation updated — N/A (internal payload field; self-documented via the CRD `description` + Go doc comments)

### DIFF
--- a/api/foreman/v1alpha1/agentictask_types.go
+++ b/api/foreman/v1alpha1/agentictask_types.go
@@ -49,6 +49,25 @@ const (
 	AgenticTaskKindFreeform AgenticTaskKind = "freeform"
 )
 
+// BranchStrategy controls how the executor cuts an issue-fix task's working
+// branch relative to the current base, so a re-dispatch of the same issue can
+// never revert already-merged work.
+// +kubebuilder:validation:Enum=reset;rebase
+type BranchStrategy string
+
+const (
+	// BranchStrategyReset cuts the working branch fresh from the CURRENT base
+	// tip, ignoring any prior attempt (payload.reviseFromBranch). This is the
+	// default: a fresh issue-fix, retry, or repair re-dispatch redoes the work
+	// against latest base, so a stale prior branch can never drift from base and
+	// revert merged commits.
+	BranchStrategyReset BranchStrategy = "reset"
+	// BranchStrategyRebase restores the prior attempt (payload.reviseFromBranch)
+	// and rebases it onto the CURRENT base, so an in-review PR revision carries
+	// its earlier commits forward on top of merged work instead of reverting it.
+	BranchStrategyRebase BranchStrategy = "rebase"
+)
+
 // AgenticTaskFailureReason categorizes WHY a task did not reach a
 // "succeeded-on-target" outcome. Distinct from AgenticTaskVerdict
 // (which carries the externally-meaningful WHAT: GO / NO-GO /
@@ -295,6 +314,16 @@ type AgenticTaskPayload struct {
 	// than failing.
 	// +optional
 	ReviseFromBranch string `json:"reviseFromBranch,omitempty"`
+
+	// BranchStrategy selects how the working branch is cut relative to the
+	// current base (see BranchStrategy). Defaults to "reset": cut fresh from the
+	// current base and ignore any prior attempt (reviseFromBranch), so a retry
+	// or repair re-dispatch cannot revert merged work. Set "rebase" for the
+	// in-review PR revision path, where the prior attempt is restored and
+	// rebased onto the current base. Issue-fix only.
+	// +kubebuilder:default=reset
+	// +optional
+	BranchStrategy BranchStrategy `json:"branchStrategy,omitempty"`
 
 	// Prompt is the agent input. Required for freeform.
 	// +optional
--- a/charts/foreman/templates/crds/agentictasks.yaml
+++ b/charts/foreman/templates/crds/agentictasks.yaml
@@ -203,6 +203,19 @@ spec:
                       (default derived from the issue's labels via conventional commit
                       prefixes: fix/, feat/, chore/, etc.).
                     type: string
+                  branchStrategy:
+                    default: reset
+                    description: |-
+                      BranchStrategy selects how the working branch is cut relative to the
+                      current base (see BranchStrategy). Defaults to "reset": cut fresh from the
+                      current base and ignore any prior attempt (reviseFromBranch), so a retry
+                      or repair re-dispatch cannot revert merged work. Set "rebase" for the
+                      in-review PR revision path, where the prior attempt is restored and
+                      rebased onto the current base. Issue-fix only.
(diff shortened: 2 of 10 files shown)


Defilan (changes requested):
Really solid, Jory, and it lines up exactly with the #1029 thread. Verified locally: `RebaseOntoBase` is safe (aborts + fails loud on conflict, no-op on empty upstream, injection guards consistent with the sibling helpers), the reset/rebase dispatch is correct, and backward-compat is clean (the only `reviseFromBranch` setter now stamps `rebase`, everything else defaults to `reset`). `go build` + `go test ./pkg/foreman/agent/... ./api/...` are green, and the real-git `RebaseOntoBase` tests confirm a replay onto an advanced base preserves merged work and carries the prior fix forward. Nice that it sets up the slicer repair re-dispatch to use `reset` for free.

One thing before it goes green: the chart CRDs are out of sync. `branchStrategy` landed in `config/crd/bases/` but not `charts/foreman/templates/crds/` — running `make chart-crds && make foreman-chart-crds` produces drift on both agentictasks and workloads, so the CRD Sync Check will fail. Run those and commit the chart CRD changes.

Heads up too: my #1037 adds the integrate/reconcile task kinds to the same two CRDs + agentictask_types.go, so whichever of us merges second will need a trivial rebase.

joryirving:
Thanks for the careful review. Fixed the chart CRD drift in d5076c4 — ran `make chart-crds foreman-chart-crds` and committed the synced `charts/foreman/templates/crds/{agentictasks,workloads}.yaml` (branchStrategy is now in both), so the CRD Sync Check should be clean. Also confirmed the full `make test` (controller envtest included) passes locally, not just the unit suites.

Ack on #1037 — I'll take the trivial rebase on agentictask_types.go + the two CRDs if you merge that first.

joryirving:
Heads-up on the one red check — **Run on Ubuntu (MicroShift via MINC, OpenShift SCC)**: it's an unrelated flake, not this change. It failed at `Manager > OpenShift SCC admission > should be admitted on a restricted-v2 namespace` (e2e_test.go:2103) because the namespace came back without `openshift.io/sa.scc.supplemental-groups` — the SCC controller hadn't injected it yet (ran 1 of 48 specs, 6s). The E2E Tests workflow run itself concluded success (the job is continue-on-error), and it was green on the previous commit with byte-identical Go; my only new commit is the chart-CRD yaml sync. I don't have rerun perms here — feel free to re-run that job if you want a clean board before merging.

Defilan (approved):

PR #1043, opened 2026-07-10
feat(cli): --revise-from-branch on foreman dispatch for in-place PR refresh

### LINKED ISSUE #996: feat(cli): --revise-from-branch on 'foreman dispatch' for in-place PR revision
## Summary
`llmkube foreman dispatch` only creates *fresh* issue-fix tasks (a new working branch each run) — effectively a full rewrite. There's no CLI way to run an **incremental revision** that restores an existing branch and edits on top. The API already supports this (`AgenticTask.spec.payload.reviseFromBranch`) and the in-workload review loop uses it, but it isn't reachable from the CLI.

## Ask
Add `--revise-from-branch <ref>` to `foreman dispatch`, reusing the existing `--prompt-file ISSUE=PATH` for the revision feedback. When set, the dispatched issue-fix task sets `payload.branch` **and** `payload.reviseFromBranch` to that ref, so the coder restores the prior attempt and force-with-lease pushes an update — refreshing the open PR in place instead of cutting a new branch.

```
llmkube foreman dispatch --repo defilantech/LLMKube --agents coder-go -n llm \
  --revise-from-branch foreman/adhoc-.../issue-991 --prompt-file 991=feedback.md 991
```

## Why
Post-PR iteration (a failing E2E, a reviewer request-changes) shouldn't require regenerating the whole branch — that wastes compute, tokens, and power for a one-line fix. Today the only way to trigger `reviseFromBranch` ad-hoc is hand-writing an `AgenticTask` manifest (or a wrapper taskfile). A CLI flag makes incremental revision first-class for CLI-driven dogfooding.

## Notes
- Small change: `dispatch` already builds and applies `AgenticTask`s; this sets two payload fields and skips the fresh-branch path.
- Probably worth requiring exactly one issue when `--revise-from-branch` is given (a revision targets a single branch).

joryirving:
## What

Add `--revise-from-branch <ref>` to `llmkube foreman dispatch`. When set, the dispatched AgenticTask has `payload.branch`, `payload.reviseFromBranch`, and `payload.branchStrategy` (`rebase`) stamped from the flag — so the executor restores the prior attempt and rebases it onto the current base, then force-with-lease pushes an update to the same branch (and open PR) instead of cutting a new one.

## Why

Post-PR iteration (a failing E2E, a reviewer request-changes) shouldn't require regenerating the whole branch — that wastes compute, tokens, and power for a one-line fix. Today the only way to trigger `reviseFromBranch` ad-hoc is hand-writing an `AgenticTask`. This flag makes incremental revision first-class for CLI-driven dogfooding.

Closes #996

## How

- New `dispatchOptions.reviseFromBranch` field, bound to `--revise-from-branch`.
- New `validateReviseFromBranch` helper: when the flag is set, requires exactly one issue (a revision targets a single branch — matches the issue's "Notes").
- `buildTask` stamps `Branch`, `ReviseFromBranch`, **and `BranchStrategy: rebase`** together when the flag is set. The `rebase` strategy is what makes the executor actually restore the prior attempt — under the default `reset` (introduced in #1042), `setupTaskBranch` skips the `ReviseFromBranch` restore and the revision becomes a no-op fresh cut from base. All three stay empty on the fresh-branch path.
- Help text gains a third example showing the revision flow.
- Tests added: `TestBuildTask_ReviseFromBranch` (table: unset+set; asserts all three revision fields), `TestValidateReviseFromBranch` (table: 5 cases), `TestRunDispatch_ReviseFromBranchFiresBeforeFetch` (asserts the rule fires before the GitHub fetch).

The executor (`pkg/foreman/agent/executor_native.go` `setupTaskBranch`) and the WorkloadReconciler (`internal/foreman/controller/workload_iteration.go`) already consume `ReviseFromBranch`+`BranchStrategy=rebase` together; this PR uses the same pairing the reconciler has used since #1042.

## Checklist

- [x] Tests added/updated
- [x] `make lint-all` passes locally (darwin + linux)
- [x] AI assistance (per CONTRIBUTING.md): code and PR description authored with assistance from an LLM-driven session in opencode (model: litellm-anthropic/MiniMax-M3). Design, commits, and verification all performed by the same session against the user's stated intent.
- [x] Documentation updated (user-facing) — `llmkube foreman dispatch --help` example block shows the new flag.

### DIFF
--- a/pkg/cli/dispatch.go
+++ b/pkg/cli/dispatch.go
@@ -57,16 +57,17 @@ func fprint(w io.Writer, a ...any)                 { _, _ = fmt.Fprint(w, a...) 
 
 // dispatchOptions holds the resolved flags for one `foreman dispatch` run.
 type dispatchOptions struct {
-	repo         string
-	agents       []string
-	namespace    string
-	timeoutSecs  int32
-	baseBranch   string
-	branchPrefix string
-	promptFiles  []string
-	noWait       bool
-	pollInterval time.Duration
-	dryRun       bool
+	repo             string
+	agents           []string
+	namespace        string
+	timeoutSecs      int32
+	baseBranch       string
+	branchPrefix     string
+	reviseFromBranch string
+	promptFiles      []string
+	noWait           bool
+	pollInterval     time.Duration
+	dryRun           bool
 }
 
 // taskAssignment pairs an issue with the coder Agent that will run it.
@@ -124,7 +125,12 @@ Examples:
 
   # Override one issue's prompt; create and detach:
   llmkube foreman dispatch --repo defilantech/LLMKube --agents coder-metal \
-    --prompt-file 813=/tmp/813.md --no-wait 813 854`,
+    --prompt-file 813=/tmp/813.md --no-wait 813 854
+
+  # In-place PR refresh: restore an existing branch and push an update on top:
+  llmkube foreman dispatch --repo defilantech/LLMKube --agents coder-go \
+    --revise-from-branch foreman/adhoc/issue-991 \
+    --prompt-file 991=feedback.md 991`,
 		Args: cobra.MinimumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runDispatch(cmd.Context(), cmd.OutOrStdout(), args, opts)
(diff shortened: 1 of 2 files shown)


Defilan (changes requested):
Nice flag, but there's a merge-order gotcha with #1042 that makes it a no-op as written. #1042 gated the `reviseFromBranch` restore behind `BranchStrategy == rebase`, and the default is now `reset` — so a dispatched task that sets `ReviseFromBranch` but not `BranchStrategy` falls through to the fresh-cut path and never restores the prior attempt.

`buildTask` needs to also stamp `BranchStrategy: BranchStrategyRebase` on the revision path. The unit test passes because it only checks payload stamping, not the executor behavior — worth asserting the strategy too so this can't regress. Fix that and it's good.

joryirving (commented):
Verified and fixed — you were right. I rebased onto latest main, looked at `setupTaskBranch` and confirmed the gate: under the default `reset` strategy (`executor_native.go:381`), the `ReviseFromBranch != ""` check at line 390 is short-circuited, so without `BranchStrategy: rebase` the executor falls through to `CreateBranchFromUpstream` and my prior commit was a no-op.

The reconciler's pattern at `workload_iteration.go:243-256` already pairs the three fields (`Branch = branch`, `ReviseFromBranch = branch`, `BranchStrategy = rebase`), so dispatch now does the same: `buildTask` stamps all three on the revision path, all stay empty on the fresh path. Test extended to assert `BranchStrategy` so this can't silently regress — that was a real gap, thanks for catching it.

Force-pushed as a single amended commit (same `feat(cli):` subject, body still calls out #996). Re-ran `make fmt vet lint lint-all test` on the rebased branch — all green, including the executor envtest that exercises `setupTaskBranch` directly. PR body updated to describe the strategy.

Defilan (approved):

PR #1044, opened 2026-07-10
fix(foreman): anchor recovery soft-reset at merge-base branch point

### LINKED ISSUE #1002: fix(foreman): pin BaseBranchSHA recovery point with merge-base to avoid re-staging mid-run upstream delta
## What

`repo.BaseBranchSHA` (introduced in #995) returns the *current* upstream tip at recovery time, but the actual branch point was captured at task-branch-cut time in `CreateBranchFromUpstream` (#813). If upstream advances between those two points, the fetched tip is ahead of the true branch point — and `git reset --soft` against that ahead-of-branch-point SHA reverts the intervening upstream delta into the recovered commit (squashing it into the executor's own commit).

This is the same class of bug as the #982/#813 stale-base failure mode, just along a different time axis (mid-run advance vs. pre-clone lag).

## Repro sketch

```
# t=0:  branch cut from upstreamTip @ SHA-A (the #813 path)
# t=1:  upstream advances to SHA-B
# t=2:  model self-commits on top of SHA-B → 1 commit ahead of current upstream (B)
# t=3:  executor calls BaseBranchSHA → returns B
# t=4:  git rev-list --count B..HEAD → 1 (proceeds to recovery)
# t=5:  git reset --soft B → HEAD @ SHA-A, delta from upstream between A and B
#                  is sitting in the index, alongside the model's actual edits
# t=6:  repo.Commit squashes the whole staging area into the recovered commit
# Result: recovered commit contains upstream delta + model edits
```

## Fix direction

Replace the soft-reset target with the true branch point, not the current upstream tip:

```
branchPoint := git merge-base <baseBranchSHA> HEAD
git reset --soft <branchPoint>
```

`merge-base` finds the lowest common ancestor of `baseBranchSHA` (current upstream) and `HEAD` — which is exactly where the branch was cut. The intervening upstream delta is preserved on the branch and not re-staged. This matches how the normal commit path already behaves (it never re-anchors).

## Notes

- Filed from the review thread on #995 (Defilan's non-blocking follow-up).
- The fleet path always has a resolved upstream URL, so the no-upstream fallback (which still degrades silently to NO-CHANGES for a self-commit) doesn't widen this attack — it just leaves a pre-existing behavior alone.
- Cheap to land independently of #995's release.

joryirving:
## What

Anchor the recovery soft-reset at the true branch point — `merge-base(base, HEAD)` — instead of the current upstream tip, so intervening upstream delta is never squashed into a recovered self-commit.

## Why

Fixes #1002

`repo.BaseBranchSHA` (#995) returns the *current* upstream tip at recovery time, but the branch point was captured at task-branch-cut time in `CreateBranchFromUpstream` (#813). If upstream advances between those two points, `SoftResetToBase`'s `git reset --soft <baseBranchSHA>` reverts the intervening upstream delta into the index and `repo.Commit` squashes it into the recovered commit. Same class as the #982/#813 stale-base bug, along the mid-run-advance axis.

## How

`SoftResetToBase` now resolves `branchPoint := git merge-base <base> HEAD` — the lowest common ancestor of the current upstream tip and the task branch, i.e. exactly where the branch was cut — and soft-resets there. The commits-ahead count is anchored at the same `branchPoint`, so the guard and the reset stay consistent. The intervening upstream delta stays on the branch and is never re-staged; only the model's own commits are recovered. Matches how the normal commit path already behaves (it never re-anchors).

Contained to `pkg/foreman/agent/repo/commit.go` — the executor caller is unchanged (it already passes `BaseBranchSHA`). No CRD/API change.

## Checklist

- [x] Tests added/updated — `TestSoftResetToBase_AnchorsAtBranchPointNotUpstreamTip` (upstream advances mid-run → only the model's file is staged, the upstream delta is not re-staged); `TestSoftResetToBase_NoSelfCommitIsNothingToCommit` (no self-commit + advanced upstream → `ErrNothingToCommit`). Existing `TestRecoverSelfCommitsOrNoChange_*` still pass.
- [x] `make test` passes locally (full suite; envtest included)
- [x] AI assistance disclosed — AI-assisted contribution
- [ ] Documentation updated — N/A (internal recovery-path behavior; doc comment on `SoftResetToBase` updated)

### DIFF
--- a/pkg/foreman/agent/repo/commit.go
+++ b/pkg/foreman/agent/repo/commit.go
@@ -127,11 +127,18 @@ func BaseBranchSHA(ctx context.Context, workspace, upstreamURL, baseBranch strin
 }
 
 // SoftResetToBase moves commits from HEAD back into the working tree
-// without touching the index, relative to base. Used when a model
-// self-committed its work and the executor wants to re-apply it with
-// DCO sign-off and executor-owned author identity. base should be a
-// commit SHA (resolved via BaseBranchSHA), not a ref name, so a stale
-// local branch cannot drag upstream commits into the recovered commit.
+// without touching the index, relative to the branch point. Used when a
+// model self-committed its work and the executor wants to re-apply it with
+// DCO sign-off and executor-owned author identity. base should be a commit
+// SHA (resolved via BaseBranchSHA), not a ref name, so a stale local branch
+// cannot drag upstream commits into the recovered commit.
+//
+// base is the CURRENT upstream tip, which may be AHEAD of the point the task
+// branch was actually cut from when upstream advanced mid-run. The reset
+// anchors at the TRUE branch point — merge-base(base, HEAD) — not at base, so
+// only the model's own commits are re-staged; the intervening upstream delta
+// between the branch point and the current tip is never squashed into the
+// recovered commit (#1002).
 func SoftResetToBase(ctx context.Context, workspace string, base string) error {
 	if workspace == "" {
 		return fmt.Errorf("SoftResetToBase: workspace is required")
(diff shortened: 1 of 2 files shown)


Defilan (approved):
Clean fix, Jory. Verified the reasoning: anchoring the soft-reset at `merge-base(base, HEAD)` instead of the current tip means a mid-run upstream advance can't drag the intervening delta into the recovered commit — same stale-base family as #813/#1042, on the mid-run axis. Real-git tests reproduce it well and pass locally. Approving.

PR #1059, opened 2026-07-10
feat(runtime): add SGLang runtime backend (#974)

### LINKED ISSUE #974: [FEATURE] Add an sglang InferenceService runtime backend
## Feature Description

Add an `sglang` runtime backend to `InferenceService`, alongside the existing `llamacpp` / `vllm` / `tgi` / `personaplex` / `generic` options, so a model can be served by [SGLang](https://github.com/sgl-project/sglang) via the same CRD.

## Problem Statement

SGLang is a high-throughput LLM serving engine whose headline feature — **RadixAttention** automatic prefix caching — is a direct fit for the agentic workloads LLMKube's Foreman drives. A coder/reviewer tool-loop re-sends a large shared prefix every turn (system prompt + tool definitions + issue/repo context); SGLang serves that shared prefix from cache instead of reprocessing it, which materially cuts latency and raises throughput for exactly the multi-turn, high-concurrency pattern the foreman-agent generates. Today an operator who wants SGLang has to fall back to the `generic` runtime and hand-write the container command, args, and probes — losing the auto-generated args, mode handling, and health wiring the first-class backends get.

## Proposed Solution

Mirror the existing `vllm` / `tgi` backends — both are the template for this:

- **CRD:** add `sglang` to the `spec.runtime` enum in `api/v1alpha1/inferenceservice_types.go` (`+kubebuilder:validation:Enum=...`), regenerate manifests + chart CRDs.
- **Backend:** a new `internal/controller/runtime_sglang.go` implementing the `RuntimeBackend` interface (`DefaultPort()`, `BuildArgs(...)`), and `CommandBuilder` since SGLang launches via a Python module rather than a bare binary:
  - `BuildCommand()` → `["python3", "-m", "sglang.launch_server"]`
  - `BuildArgs(...)` → `--model-path <modelPath> --host 0.0.0.0 --port <port>` plus tensor-parallel (`--tp`) from the GPU spec where applicable, `--served-model-name`, and mode handling (SGLang serves embeddings via `--is-embedding`, so `spec.mode: embedding` should append it — mirroring how llamacpp maps mode to flags).
  - `DefaultPort()` → `30000` (SGLang's default).
- **Dispatch:** a `case "sglang"` in `resolveBackend()` (`internal/controller/runtime.go`) and a `runtimeNameLabel` entry.
- **Image:** default to the official `lmsysorg/sglang` image in `resolveRuntimeImage`; SGLang publishes CUDA and ROCm variants, so honor the GPU vendor the way the llamacpp/vulkan path already does.
- **Health/readiness:** SGLang exposes `/health` (liveness) and `/health_generate` (readiness that actually runs a token) and an OpenAI-compatible `/v1`. Wire the probes like the other OpenAI-compatible backends.
- **Tests + docs:** `runtime_sglang_test.go` for arg/command/port/mode generation (as `runtime_vllm_test.go` does), and a docs entry noting SGLang requires a GPU (no CPU-only path like llama.cpp).

## Alternatives Considered

- **`generic` runtime (status quo):** works, but the operator hand-maintains command/args/probes and gets none of the mode/arg automation — the same gap that justified first-class `vllm`/`tgi` backends.
- **Not supporting it:** leaves the RadixAttention prefix-cache win on the table for agentic serving, which is a notable throughput/cost difference for foreman-style tool loops.

## Additional Context

- Integration surface is small and well-precedented: `internal/controller/runtime.go` (interface + `resolveBackend`), `runtime_vllm.go` / `runtime_vllm_args.go` as the pattern, `deployment_builder.go` (`resolveRuntimeImage`).
- SGLang is OpenAI-compatible (`/v1/chat/completions`, `/v1/embeddings`), so downstream (litellm, the foreman agents) needs no change — it's a serving-engine swap behind the same endpoint.
- GPU-only: unlike llama.cpp there's no meaningful CPU path, so docs/validation should say so.

## Priority

- [x] Medium - Would be a nice improvement

## Willingness to Contribute

- [x] Yes, I can submit a PR (the `vllm`/`tgi` backends give a clear template)

Filed by an operator running LLMKube's Foreman on a self-hosted fleet.

joryirving:
## What

Adds `sglang` as a first-class `InferenceService.Spec.Runtime` value,
alongside `llamacpp` / `vllm` / `tgi` / `personaplex` / `generic`.

A new typed `SGLangConfig` (with nested `SGLangSpeculativeConfig`) covers
sharding, memory, batching, quantization/KV-cache, attention, agentic glue
(tool/reasoning parsers, chat template), speculative decoding (EAGLE /
EAGLE3), and LoRA basics — mirroring the vLLM pattern. GPU vendor (NVIDIA
vs AMD) drives image selection, so the AMD ROCm image is picked automatically
when `model.spec.hardware.gpu.vendor` is `amd`.

SGLang launches via `python3 -m sglang.launch_server` (mirroring how
`personaplex` wraps a Python module). Probes target `/health_generate` for
startup + readiness (runs a token, accurate) and `/health` for liveness
(cheap). Startup tolerates 180 failures (~30 min) for cold-start model load.

## Why

Fixes #974

Headline motivation is SGLang's **RadixAttention** automatic prefix caching.
A foreman agent's tool loop re-sends the same large shared prefix every turn
(system prompt + tool definitions + issue/repo context); SGLang serves that
shared prefix from cache instead of reprocessing it, which materially cuts
latency and raises throughput for the multi-turn, high-concurrency pattern
the foreman-agent generates. Today an operator who wants SGLang has to fall
back to the `generic` runtime and hand-write container config — losing the
auto-generated args, mode handling, and health wiring the first-class
backends get.

## How

- `api/v1alpha1/inferenceservice_types.go`: `SGLangConfig` +
  `SGLangSpeculativeConfig` structs, `SGLangConfig` field on
  `InferenceServiceSpec`, `sglang` added to the runtime enum.
- `internal/controller/runtime_sglang.go`: `SGLangBackend` implementing
  `RuntimeBackend` + `CommandBuilder` + `HPAMetricProvider` +
  `EnvBuilder`. `BuildArgs` emits base flags + `--served-model-name` +
  typed fields in declaration order + auto-derived `--tp` + `--is-embedding`
  for mode + extraArgs last.
- `internal/controller/runtime_sglang_args.go`: free-function helpers,
  one per flag. Mirrors `runtime_vllm_args.go` structure.
- `internal/controller/runtime.go`: `case RuntimeSGLANG` in
  `resolveBackend`.
- `internal/controller/deployment_builder.go`: `resolveRuntimeImage`
  picks ROCm image for AMD GPUs; otherwise CUDA.
- `make manifests` + `make chart-crds` regenerated CRDs.
- `docs/contributors/adding-a-runtime.md`, `docs/site/concepts/comparison.md`,
  `docs/site/guides/model-matrix.md` updated.

**Out of scope (B3 followup):** LoRA hot-loading, speculative perf tuning
(`--speculative-accept-threshold-*`), data-parallel rendezvous,
PodMonitor for `/metrics`, and remaining minor SGLang flags. Tracked as a
separate issue.

## Checklist

- [x] Tests added/updated (`TestSGLangBuildArgs` table-driven +
  `TestSGLangBuildCommand` + `TestSGLangProbes` + `TestSGLangBuildEnv` +
  `TestValidateSGLangConfig` + dispatch + image-resolution tests)
- [x] `make lint-all` passes locally (cross-arch)
- [x] All commits are signed off (`git commit -s`)
- [x] AI assistance disclosed: this PR was produced by an AI coding agent
  working with a repo collaborator. The agent followed the
  `docs/contributors/adding-a-runtime.md` pattern, the design spec at
  `docs/superpowers/specs/2026-07-10-sglang-runtime-backend-design.md`
  (local agent scratch, not committed), and AGENTS.md conventions.
- [x] Documentation updated


Defilan (commented):
This is excellent work, Jory, one of the cleanest runtime additions we have had. It mirrors the vLLM pattern faithfully, the arg emission is deterministic (nice `TestSGLangBuildArgsDeterministic`), the probe split (`/health` for liveness vs `/health_generate` for readiness) is a step up from vLLM, and both CRD copies regenerate with zero drift, which fully addresses the earlier #1042 feedback. I ran `make manifests && make chart-crds`, the SGLang/image/dispatch tests, and cross-arch lint locally: all green.

Not blocking from my side, approving in spirit. One thing worth resolving so the runtime is truly first-class: SGLang does not expose `/metrics` unless it is launched with `--enable-metrics` (per the [SGLang production-metrics docs](https://docs.sglang.ai/references/production_metrics.html)). Right now nothing injects that flag, so the runtime ships with no metrics endpoint at all, yet `DefaultHPAMetric()` (`runtime_sglang.go:62`) advertises `sglang:num_requests_running` and the docs table lists it, so HPA (and the future #1060 PodMonitor) silently get nothing. vLLM exposes metrics with no flag and llama.cpp always passes `--metrics`, so SGLang is the odd one out. I would add `--enable-metrics` to the base args (around `runtime_sglang.go:83`), mirroring llama.cpp's always-on convention:

```go
"--host", "::",
"--port", fmt.Sprintf("%d", port),
// Expose SGLang's Prometheus /metrics endpoint (off by default in SGLang).
// Mirrors llama.cpp's always-on --metrics so PodMonitor scraping and the
// advertised HPA metric actually work.
"--enable-metrics",
```

If you would rather keep metrics fully deferred to #1060, that is fine too, just have `DefaultHPAMetric()` and the docs table reflect that it is not functional yet.

Two tiny nits, both optional:

- `runtime_sglang.go:133` uses the `"embedding"` string literal; the package already defines `servingModeEmbedding`, so prefer the constant to keep a future rename from drifting. (The rest of the codebase also resolves mode via `resolveServingMode(isvc)`, which infers embedding from the endpoint path, worth considering for parity.)
- The `--is-embedding` / `--mem-fraction-static` GPU check keys off the device-plugin count, so a DRA-only model (count 0) would skip `--mem-fraction-static` with a spurious "no GPU" log. This mirrors vLLM's existing behavior, so it is a shared limitation rather than a regression, just noting for a future cleanup.

Everything else is ready to merge. Thanks for the great contribution and the thorough disclosure.

itsmiso-ai:
Thanks for the review! All four items addressed in `0fd9bbf`:

1. **`--enable-metrics`** added to base args (`runtime_sglang.go:85-88`) — PodMonitor + HPA will now actually receive metrics.
2. **`servingModeEmbedding` + `resolveServingMode`** (`runtime_sglang.go:140`) — uses the constant and the package-level mode inference. Added a new test row for endpoint-path-based inference (`/v1/embeddings` → `--is-embedding`).
3. **DRA-only GPU-check TODO comment** at the `--mem-fraction-static` guard (`runtime_sglang.go:108-111`) — points future contributors at `hasGPUPresent`. Shared with vLLM, noted as shared limitation rather than a regression.
4. **B3 issue (#1060) update with DRA-aware bullet** — failed: `joryirving does not have permission to update the issue`. I don't have write access on `defilantech/LLMKube` issues. Could you add this bullet to #1060 when you get a chance?

```
- DRA-aware GPU checks: `resolveGPUCount` returns 0 for DRA-only models
  (resourceClaims without device-plugin allocation), so the
  `--mem-fraction-static` and similar guards spuriously skip on those
  models. Switch the guards to `hasGPUPresent(isvc, model)` to account
  for DRA. Shared with vLLM; clean up when one runtime fixes it.
```

Otherwise ready to merge. Thanks again for the thorough review — the `--enable-metrics` catch would have been a real footgun in production.

Defilan (approved):
Thanks for the quick turnaround, both items are handled cleanly.

- `--enable-metrics` is in the base args now with a clear comment and test coverage, so SGLang actually exposes `/metrics` and the advertised HPA metric (plus the future #1060 PodMonitor) will work. That was the one thing keeping it from being truly first-class.
- Nice touch switching to `resolveServingMode(isvc)` + the `servingModeEmbedding` constant, that matches how the rest of the codebase resolves mode and avoids the literal drifting.

I re-verified locally: `make manifests` + `chart-crds` + `foreman-chart-crds` regenerate with zero drift, and CI is green. The `ValidateSGLangConfig` reconciler wiring and the v0.5.15 image bump are good adds too.

Approving. Genuinely clean contribution, thanks for seeing it all the way through.

Issue #1033, opened 2026-07-10
[FEATURE] Epic: Sliced Workloads: contract-pinned decomposition + reconcile in Foreman


Defilan:
## Feature Description

Graduate the validated slicer/integrator into Foreman as **Sliced Workloads**: slice one hard issue into disjoint-file slices, dispatch each as a normal coder task, union the results, and reconcile the union against a pinned contract so cross-slice interface drift that a green build cannot see is caught.

This tracks the productization of the experiment that was validated end to end on #700 and #1001. In both, three disjoint slices built clean and green, yet the slices had drifted on shared identifiers (a dashboard querying metric names an exporter never emitted; a broken kube-state-metrics config). A deterministic pinned-identifier check caught the drift the build missed; an LLM sweep corroborated it.

**v1 scope: detection + prevention.** The reconcile step surfaces drift (fails the Workload, drift list in status) for a human to adjudicate. Automated repair is explicitly deferred to a follow-up (see Additional Context).

## Problem Statement

> As a fleet operator, I want to hand Foreman one multi-part issue and get a decomposed, interface-consistent PR back from local models, so that mid-size sovereign coders can complete work that they fail as a single whole.

Mid-size local models flunk whole issues but nail small slices. Decomposition works, and disjoint-file slices union cleanly, but the union is blind to semantic interface drift between slices. The pinned contract plus a reconcile pass is the fix. Today it only exists as throwaway scaffolding; it is not usable from Foreman.

## Proposed Solution

`llmkube slice <issue>` runs a planner that emits a pinned SlicePlan (contract + `shared_identifiers` + disjoint slices), renders a Workload, and applies it. The controller runs the existing PipelineStep DAG with two new task kinds.

```yaml
# rendered by `llmkube slice 1001`
apiVersion: foreman.llmkube.dev/v1alpha1
kind: Workload
spec:
  pipeline:
    - {name: crs-config, kind: issue-fix, ...}       # disjoint slice
    - {name: dashboard, kind: issue-fix, ...}         # disjoint slice
    - {name: alerts, kind: issue-fix, ...}            # disjoint slice
    - {name: integrate, kind: integrate, dependsOn: [crs-config, dashboard, alerts]}
    - {name: reconcile, kind: reconcile, dependsOn: [integrate]}   # pins in payload
```

- `integrate` (new kind): union the disjoint slice branches onto the current base, push the integration branch, gate = build; fail on overlap / apply-conflict.
[...]

PR #1065, opened 2026-07-11
fix(foreman): default to rebase when reviseFromBranch is set without branchStrategy

### LINKED ISSUE #1047: [FEATURE] Make branchStrategy/reviseFromBranch coupling foolproof (silent no-op footgun)
## Feature Description

Make the `payload.branchStrategy` / `payload.reviseFromBranch` coupling foolproof, so that setting `reviseFromBranch` without also setting `branchStrategy: rebase` cannot silently no-op.

## Problem Statement

> As someone dispatching an in-place PR revision, I want setting `reviseFromBranch` to actually restore the prior attempt, not silently cut a fresh branch.

#1042 gated the `reviseFromBranch` restore in `setupTaskBranch` behind `branchStrategy == rebase`, and made the default `reset`:

```go
strategy := task.Spec.Payload.BranchStrategy
if strategy == "" { strategy = BranchStrategyReset }
if strategy == BranchStrategyRebase &&
    task.Spec.Payload.ReviseFromBranch != "" &&
    task.Spec.Kind == AgenticTaskKindIssueFix { /* restore + rebase */ }
```

So a task that sets `reviseFromBranch` but leaves `branchStrategy` unset defaults to `reset` and **skips the restore entirely** — it cuts a fresh branch and does the opposite of what the caller intended. There is no error and no warning; the payload just silently does nothing with `reviseFromBranch`.

This already bit #1043 (the `--revise-from-branch` CLI flag stamped `reviseFromBranch` but not `branchStrategy`, so the flag was a no-op post-#1042). Every future caller that sets `reviseFromBranch` is one forgotten field away from the same silent failure.

## Proposed Solution

Pick one (or combine):

1. **Default to rebase when `reviseFromBranch` is set (recommended).** In `setupTaskBranch`, when `branchStrategy` is empty and `reviseFromBranch != ""`, treat the effective strategy as `rebase`. Restore-and-rebase is the only sensible intent when a prior attempt is named; the explicit `reset` still wins when set. Callers can stop threading `branchStrategy` on the revision path.
2. **Validate the pairing.** The AgenticTask validating webhook (and/or the CLI) rejects `reviseFromBranch` set with `branchStrategy: reset` as a contradiction, so the mistake fails loud at admission instead of silently at execution.
3. **Warn.** Log a clear warning when `reviseFromBranch` is set but the effective strategy is `reset`, so the no-op is at least visible in the executor logs.

## Alternatives Considered

- Leave as-is and rely on every caller remembering both fields. Rejected: already failed once (#1043), and a silent no-op is a bad failure mode for a system that leaves an audit trail.

## Additional Context

- Related: #1042 (introduced the gate + reset default), #1043 (bitten by it), #1029.
- The Sliced Workloads repair re-dispatch (#1033, future v2) deliberately relies on the `reset` default, so whatever lands here must keep an *explicit* `reset` honored; only the empty-plus-`reviseFromBranch` case should change.

## Priority

- [x] Medium - Nice to have (prevents a recurring silent-no-op class)

## Willingness to Contribute

- [x] Yes, I can submit a PR

Defilan:
## What

Default the effective branch strategy to `rebase` when a task sets `reviseFromBranch` but leaves `branchStrategy` empty, so the prior attempt is restored instead of silently discarded.

## Why

Fixes #1047. When `payload.reviseFromBranch` is set but `branchStrategy` is empty, the effective strategy defaulted to `reset`, which skips the restore and cuts a fresh branch, the opposite of the caller's intent. This already bit #1043 (the `--revise-from-branch` CLI flag stamped `reviseFromBranch` but not `branchStrategy`, so the flag became a silent no-op after #1042).

## How

- `setupTaskBranch`: an empty `branchStrategy` combined with a non-empty `reviseFromBranch` is now treated as `rebase`. An explicit `reset` still wins, preserving the Sliced Workloads repair re-dispatch path (#1033).
- Regression test `TestSetupTaskBranch_ReviseFromBranchDefaultsToRebase` pins the contract: a task with `reviseFromBranch` set and no `branchStrategy` must restore the prior attempt from the push remote.

## Provenance / AI disclosure (band-3)

This PR was produced **autonomously by Foreman**, LLMKube's own agentic coding harness, in an overnight batch on a local mid-size model (Ornith-35B) with context7 MCP grounding. It was verified by the deterministic clean-room gate (fmt/vet/build/lint/test) before push, and reviewed by the maintainer before opening. A nice bit of dogfooding: Foreman fixing a Foreman branch-handling footgun it could have hit itself.

## Testing

- Clean-room gate: GATE-PASS.
- New regression test fails before the fix (reset skips the restore) and passes after; re-verified locally.

## Checklist

- [x] Tests added/updated
- [x] Build / lint / gate pass
- [x] DCO signed-off

### DIFF
--- a/pkg/foreman/agent/executor_native.go
+++ b/pkg/foreman/agent/executor_native.go
@@ -378,7 +378,16 @@ func setupTaskBranch(
 ) error {
 	strategy := task.Spec.Payload.BranchStrategy
 	if strategy == "" {
-		strategy = foremanv1alpha1.BranchStrategyReset
+		// #1047: reviseFromBranch without branchStrategy used to silently no-op
+		// (defaulted to reset, which skips the restore). When reviseFromBranch is
+		// set and the caller did not explicitly choose reset, treat the effective
+		// strategy as rebase — restore-and-rebase is the only sensible intent
+		// when a prior attempt is named. An explicit "reset" still wins.
+		if task.Spec.Payload.ReviseFromBranch != "" {
+			strategy = foremanv1alpha1.BranchStrategyRebase
+		} else {
+			strategy = foremanv1alpha1.BranchStrategyReset
+		}
 	}
 
 	// rebase (in-review revision): restore the prior attempt AND rebase it onto
--- a/pkg/foreman/agent/executor_native_internal_test.go
+++ b/pkg/foreman/agent/executor_native_internal_test.go
@@ -2358,6 +2358,61 @@ func TestSetupTaskBranch_ResetSkipsPriorAttempt(t *testing.T) {
 	}
 }
 
+// TestSetupTaskBranch_ReviseFromBranchDefaultsToRebase pins #1047: when
+// reviseFromBranch is set but branchStrategy is left empty (the common
+// caller mistake that produced the silent no-op), the effective strategy
+// must be rebase so the prior attempt is restored instead of silently
+// discarded. An explicit reset still wins.
+func TestSetupTaskBranch_ReviseFromBranchDefaultsToRebase(t *testing.T) {
+	if _, err := exec.LookPath("git"); err != nil {
+		t.Skip("git not on PATH")
+	}
+	dir := t.TempDir()
+	bare, _ := seededRemote(t, dir)
+	const branch = "foreman/wl/issue-641"
+
+	// Prior attempt pushed to the remote.
+	prior := filepath.Join(dir, "prior")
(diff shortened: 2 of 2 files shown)


joryirving (approved):
Clean fix. Defaulting to rebase when reviseFromBranch is set but branchStrategy is empty is the correct semantic — naming a prior branch to revise from means you want it restored. Explicit reset still wins. Test coverage proves the regression. Approve.

Issue #1066, opened 2026-07-11
[FEATURE] Make the EditFreeStreak guard grounding-aware (anti-confab reading vs stuck-loop)


Defilan:
## Feature Description

Make the coder's EditFreeStreak stuck-loop guard aware of the anti-confabulation grounding rule, so grounding-driven reading is not force-terminated as if it were aimless exploration.

## Problem Statement

> As a fleet operator, I want the coder to be able to read enough to ground its facts (as the anti-confab rule now requires) without the stuck-loop guard cutting it off before it converges.

The grounding rule shipped in #1062 tells the coder to read the real source before writing a value it would otherwise guess. The EditFreeStreak guard, by design, force-terminates a run that reads too many turns without landing an edit. These two are now in direct tension: the more thorough the grounding, the closer the coder gets to the edit-free limit.

## Evidence

An overnight Foreman batch (coder-metal / Ornith-35B, context7 MCP + grounding rule, `editFreeTurnsLimit` already at 12) on 5 backlog issues:

- **#1055** and **#904**: both ended INCOMPLETE with `stuck-loop detector intervened: EditFreeStreak` (43 and 15 turns). Neither gave up on the merits; the guard terminated them mid-grounding.
- #1047 succeeded (GATE-PASS), #1022 / #1025 correctly declined as already-done.

So 2 of 5 non-productive outcomes were the guard firing on grounding-heavy issues, not model failure per se.

## Proposed Solution

Options, roughly in order of preference:

1. **Grounding-aware streak accounting.** A `read_file` on a file the coder subsequently edits (or on a file inside the slice's declared scope) should not count toward the edit-free streak the same way an unrelated exploratory read does. The streak is meant to catch aimless looping, not legitimate grounding.
2. **Interleave nudge.** Prompt the coder to ground the specific fact it needs, then edit, rather than front-loading all reading. "Ground the next edit, make it, repeat."
3. **Per-task budget.** Allow a larger edit-free budget when the grounding rule is active, tuned so it still catches a genuinely stuck loop (a model reading 30+ files without editing) but tolerates 5-10 grounding reads.

Option 1 is the most principled; option 3 is the cheapest stopgap.

## Additional Context

[...]

PR #1068, opened 2026-07-11
fix(foreman): set fsGroup on the gate pod so non-root gates can write XDG_DATA_HOME

### LINKED ISSUE #1055: [BUG] Foreman gate pods set XDG_DATA_HOME to a non-writable cache PVC → non-root gates crash
## Bug Description

Foreman verify/gate pods inject `XDG_DATA_HOME=/cache/xdg` (the gate-cache PVC), but the pods run as the gate image's **non-root** user with **no `securityContext.fsGroup`**. The RWX PVC mounts root-owned, so `/cache/xdg` is not writable by the gate uid. Any gate tool that writes under `XDG_DATA_HOME` then fails silently-ish and takes down the whole gate before the real checks run.

Concretely this breaks the Godot (`godot-gate`) verify used for a Godot project: Godot's `user://` resolves under `XDG_DATA_HOME`, so it cannot create `user://logs`, which cascades into a hard crash:

```
ERROR: Could not create directory: user://logs
handle_crash: Program crashed with signal 11
ERROR: FATAL: Index p_index = 1 is out of bounds (size() = 0).
   at: get (core/templates/cowdata.h:158)
```

Every gate for that repo fails identically, *before any test runs* — so the real verify result (pass, or a genuine test failure) is completely masked.

## Steps to Reproduce

1. A verify gate whose image runs as a non-root user (e.g. `USER 65534`) and whose tool writes under `XDG_DATA_HOME`.
2. Foreman creates the gate pod with env `XDG_DATA_HOME=/cache/xdg` (the gate-cache PVC), `securityContext.fsGroup` unset, `runAsUser` unset (so it inherits the image's non-root USER).
3. The tool writes `$XDG_DATA_HOME/...`; the PVC mount is `root:root` and not writable by the gate uid → write fails → crash / gate error.

Minimal repro matching production (godot-gate image, USER 65534):

```sh
# non-writable XDG (mimics the /cache/xdg PVC) → crash:
export XDG_DATA_HOME=/var/cache/xdg   # root-owned, not writable by uid 65534
godot --headless --path <project> --script res://tests/test_runner.gd
#   -> ERROR: Could not create directory: user://logs
#   -> signal 11 / Index out of bounds size()=0 ; exit 132

# writable XDG → same command runs the tests clean:
export XDG_DATA_HOME=/tmp/.local/share
godot --headless --path <project> --script res://tests/test_runner.gd   # OK
```

## Expected Behavior

Gate pods provide a writable `XDG_DATA_HOME`, so a non-root gate tool can create its user/data directories (import caches, logs, etc.) and the gate reflects the tool's real result.

## Actual Behavior

`XDG_DATA_HOME=/cache/xdg` points at a root-owned PVC with no `fsGroup`, so the non-root gate can't write it. The tool crashes on startup and the gate fails before running any check — masking both real passes and real test failures.

## Environment

**LLMKube Version:** 0.9.3

**Cluster Type:**
- [x] K3s / Talos (bare-metal homelab)

**GPU (if applicable):**
- [x] Other: AMD Strix Halo `gfx1151` iGPU (this is the AMD/foreman fleet), but the bug is GPU-independent — it's a gate-pod filesystem-permissions issue.

Relevant gate-pod facts: `securityContext.fsGroup` unset, `runAsUser` unset (inherits image `USER 65534`); gate-cache PVC (`ceph-filesystem`, RWX) mounted at `/cache`; env `XDG_DATA_HOME=/cache/xdg`.

## Logs

Gate step log tail (windowstead / godot-gate):
```
=== clone ... ===
Cloning into '/work'...
ERROR: Could not create directory: user://logs
handle_crash: Program crashed with signal 11
ERROR: FATAL: Index p_index = 1 is out of bounds (size() = 0).
   at: get (core/templates/cowdata.h:158)
GATE FAIL
```

## Additional Context

Suggested fixes (any one resolves it):
- Set `securityContext.fsGroup` on gate pods (e.g. matching the gate image's `runAsGroup`) so the RWX PVC is group-writable — keeps the cross-run cache benefit.
- Or don't force `XDG_DATA_HOME` onto the shared PVC for gates running as non-root; or ensure the `xdg` subdir is created world/group-writable.

Workaround in use: re-export `XDG_DATA_HOME` to a writable `/tmp` path inside the gate command.

In practice only the Godot gate writes `XDG_DATA_HOME` today, so it's the visible victim — but this is a latent trap for any non-root gate tool that honors XDG. Happy to send a PR (fsGroup on the gate pod spec seems cleanest).

Defilan:
## What

Set `securityContext.fsGroup: 100` on the Foreman gate Job pod so a non-root gate image can write under `XDG_DATA_HOME=/cache/xdg` on the RWX gate-cache PVC.

## Why

Fixes #1055. The gate-cache PVC mounts `root:root`, so a non-root gate image (e.g. `USER 65534`) cannot write `XDG_DATA_HOME`, and any XDG-aware tool crashes before the real checks run. Setting `fsGroup: 100` makes Kubernetes chown the volume to group 100 and add GID 100 as a supplemental group to the gate container's process, so the container can write regardless of its built-in group. Operators pinning a different group can override via the gate image build.

## How

- `gate_job_template.yaml`: add a pod-level `securityContext.fsGroup: 100`.
- Regression test `TestRenderGateJob_PodSecurityContextFSGroup` asserts the rendered gate Job carries `PodSecurityContext.FSGroup == 100`.

## Provenance / AI disclosure (band-3)

Produced autonomously by Foreman (LLMKube's agentic harness) on a local mid-size model (Ornith-35B on the Strix, harness on the M5) with context7 MCP grounding, verified by the deterministic clean-room gate (fmt/vet/build/lint/test), and reviewed by the maintainer before opening. Notably, this issue was force-terminated by the EditFreeStreak guard on a prior run; it completed here on the grounding-aware guard build (#1066/PR #1067), which is the direct dogfood proof that the guard fix works.

## Testing

- Clean-room gate: GATE-PASS.
- Regression test re-verified locally; `go build ./...` clean.

## Checklist

- [x] Tests added/updated
- [x] Build / lint / gate pass
- [x] DCO signed-off

### DIFF
--- a/pkg/foreman/agent/tools/gate_job_template.yaml
+++ b/pkg/foreman/agent/tools/gate_job_template.yaml
@@ -53,6 +53,14 @@ spec:
         foreman.llmkube.dev/task-name: {{ .TaskName }}
     spec:
       restartPolicy: Never
+      # fsGroup makes the RWX gate-cache PVC group-writable so non-root
+      # gate images (e.g. USER 65534) can write under XDG_DATA_HOME=/cache/xdg.
+      # Without it the PVC mounts root:root and any XDG-aware tool crashes
+      # before the real checks run (#1055). 100 matches the gate image's
+      # typical build-group GID; operators pinning a different group should
+      # override via the gate's image build.
+      securityContext:
+        fsGroup: 100
       containers:
         - name: gate
           image: {{ .Image }}
--- a/pkg/foreman/agent/tools/run_gate_job_test.go
+++ b/pkg/foreman/agent/tools/run_gate_job_test.go
@@ -430,6 +430,44 @@ func TestRenderGateJob_GenericRunsCommandsNotMakeOrBiteCheck(t *testing.T) {
 	}
 }
 
+// TestRenderGateJob_PodSecurityContextFSGroup asserts the rendered gate Job
+// carries a PodSecurityContext with fsGroup=100 so non-root gate images
+// (e.g. USER 65534) can write under XDG_DATA_HOME=/cache/xdg on the RWX
+// gate-cache PVC. Regression for #1055.
+func TestRenderGateJob_PodSecurityContextFSGroup(t *testing.T) {
+	job, err := renderGateJob(rendererInput{
+		Name:                    "foreman-gate-fsgroup",
+		Namespace:               "foreman-system",
+		Image:                   "golang:1.26",
+		Repo:                    "defilantech/LLMKube",
+		Branch:                  "foreman/issue-1055",
+		Checks:                  []string{"fmt"},
+		PVCName:                 "foreman-gate-cache",
+		ActiveDeadlineSeconds:   1800,
+		TTLSecondsAfterFinished: 86400,
+		CPURequest:              "2",
+		CPULimit:                "4",
+		MemRequest:              "4Gi",
+		MemLimit:                "8Gi",
+		CloneURLBase:            "https://github.com",
+		TaskNamespace:           "default",
+		TaskName:                "gate-1055",
+	})
+	if err != nil {
+		t.Fatalf("renderGateJob: %v", err)
+	}
+	psc := job.Spec.Template.Spec.SecurityContext
+	if psc == nil {
+		t.Fatalf("PodSecurityContext must be set on the gate pod")
+	}
+	if psc.FSGroup == nil {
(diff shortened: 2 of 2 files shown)


joryirving (approved):
Clean fix. fsGroup at the pod spec level is the right mechanism, test coverage is solid, and the hardcoded GID 100 default with image-build override path is well documented. Approve.

PR #1088, opened 2026-07-12
feat(controller): extend drain-before-roll idle checks to vLLM, TGI, SGLang, and multi-replica services

### LINKED ISSUE #911: Extend drain-before-roll idle checks to vLLM, TGI, and multi-replica services

joryirving:
## What

Extends `RolloutPolicySpec.waitForIdle` (originally llama.cpp-only, shipped in #913) to work for every first-class runtime and for multi-replica deployments. The idle detector now probes each serving pod individually via EndpointSlices rather than hitting a single load-balanced Service URL.

Supported runtimes and their idle signals:

| Runtime | Signal | Idle when |
|---|---|---|
| `llamacpp` | `GET /slots` | every slot `is_processing == false` |
| `vllm` | `GET /metrics` | `vllm:num_requests_running` gauge sum == 0 |
| `tgi` | `GET /metrics` | `tgi_batch_current_size` gauge == 0 |
| `sglang` | `GET /metrics` | `sglang:num_requests_running` gauge sum == 0 |
| `generic` | annotation `inference.llmkube.dev/idle-endpoint` | HTTP 2xx |
| `personaplex` | — | not supported (fail-closed defer) |

## Why

Fixes #911.

The initial drain-before-roll (#913) only covered single-replica llama.cpp via the `/slots` endpoint. This left the operator hands-off for vLLM, TGI, SGLang, and any multi-replica deployment — exactly the workloads that also drop in-flight generations on a hard restart. This follow-up broadens the detector so `waitForIdle=true` works uniformly, and so a multi-replica rollout doesn't declare idle based on a Service request that landed on a single, possibly unrepresentative, pod.

The headline behavior from #913 is unchanged: **fail-closed**. If idleness can't be established (busy replica, unreachable pod, parse error, unsupported runtime), the rollout defers until `idleTimeoutSeconds` expires.

## How

- New optional `IdleDetector` interface alongside the existing `RuntimeBackend` family. Each backend implements `IdleProbe(isvc, client) func(ctx, baseURL) (bool, error)` — a closure that captures the HTTP client and any per-reconcile state (annotation path, metric name).
- `checkServiceIdle` rewritten to: resolve backend -> type-assert to `IdleDetector` -> list EndpointSlices -> probe each Ready replica. Falls back to a single Service URL probe when no EndpointSlices exist (backward compat for the single-replica case and for envtest).
- Generic runtime uses a new `inference.llmkube.dev/idle-endpoint` annotation — operators declare an HTTP path that returns 2xx when idle. No annotation -> `ReasonIdleCheckUnsupported` (defer, then roll after timeout). This avoids a CRD schema bump.
- Hand-rolled Prometheus parser (~30 LOC) for simple gauges. Deliberately avoids `prometheus/common/expfmt` to keep the dependency surface unchanged.
- New condition reason `ReasonIdleCheckUnsupported` for runtimes without idle detection.
- New concept doc: `docs/site/concepts/drain-before-roll.md`.

## Checklist

- [x] Tests added/updated

---

**AI assistance disclosure:** This PR was implemented with AI assistance. The design was brainstormed and reviewed collaboratively, the implementation plan was written and executed via subagent-driven development (fresh implementer + independent reviewer per task, plus a final whole-branch review), and the code was verified against `make fmt`, `make vet`, `make lint`, `make test`, and `make lint-all`.

### DIFF
--- a/api/v1alpha1/constants.go
+++ b/api/v1alpha1/constants.go
@@ -33,6 +33,11 @@ const (
 	// agents that predate this annotation.
 	AnnotationAgentVersion = "llmkube.ai/agent-version"
 
+	// AnnotationIdleEndpoint lets operators declare a custom HTTP path that
+	// returns 2xx when a replica is idle. Used by the generic runtime to opt
+	// in to drain-before-roll. Set on InferenceService metadata.annotations.
+	AnnotationIdleEndpoint = "inference.llmkube.dev/idle-endpoint"
+
 	// DefaultAgentHeartbeatInterval is how often the metal-agent re-asserts
 	// its registrations (which also self-heals any missed update, #657).
 	DefaultAgentHeartbeatInterval = 30 * time.Second
@@ -63,6 +68,10 @@ const (
 	// ReasonIdleTimeoutExceeded is set when RolloutDeferred=False after the
 	// idle timeout expired and the rollout proceeded despite busy pods.
 	ReasonIdleTimeoutExceeded string = "IdleTimeoutExceeded"
+
+	// ReasonIdleCheckUnsupported is set when the runtime backend does not
+	// implement IdleDetector, so drain-before-roll cannot probe idleness.
+	ReasonIdleCheckUnsupported string = "IdleCheckUnsupported"
 
 	// DefaultIdleCheckInterval is how often the controller re-checks pod
 	// idleness when waiting for idle before rollout.
--- a/api/v1alpha1/inferenceservice_types.go
+++ b/api/v1alpha1/inferenceservice_types.go
@@ -504,8 +504,12 @@ type InferenceServiceSpec struct {
 
 	// RolloutPolicy controls how deployment updates are applied. When waitForIdle
 	// is true, the controller will check backend slot idleness before updating
-	// the Deployment pod-template. Requires llama.cpp server-compatible backends
-	// that expose a /slots endpoint.
+	// the Deployment pod-template. Idle detection support by runtime:
+	//   - llama.cpp: native /slots endpoint (default)
+	//   - vLLM: Prometheus metrics scrape (vllm:num_requests_running)
+	//   - TGI: Prometheus metrics scrape (tgi_loading)
+	//   - SGLang: Prometheus metrics scrape (sglang:num_running_reqs)
+	//   - generic: optional AnnotationIdleEndpoint annotation for custom probe
 	// +optional
 	RolloutPolicy *RolloutPolicySpec `json:"rolloutPolicy,omitempty"`
 }
@@ -514,10 +518,11 @@ type InferenceServiceSpec struct {
 type RolloutPolicySpec struct {
 	// WaitForIdle indicates whether to wait for all backend slots to report idle
 	// before applying a Deployment pod-template update. When true, the controller
-	// polls the /slots endpoint on each replica and defers the rollout until all
-	// slots are idle or the idleTimeoutSeconds expires. Requires the backend to
-	// expose a llama.cpp-style /slots endpoint (llama.cpp enables it by default
-	// with --slots); other runtimes are tracked in #911.
+	// probes each replica and defers the rollout until all replicas are idle or the
+	// idleTimeoutSeconds expires. Idle detection is runtime-specific: llama.cpp uses
+	// /slots, vLLM/TGI/SGLang scrape Prometheus gauges, and generic runtimes may set
+	// AnnotationIdleEndpoint for a custom HTTP probe. Runtimes without idle detection
+	// support proceed immediately with ReasonIdleCheckUnsupported.
 	// +optional
 	WaitForIdle bool `json:"waitForIdle,omitempty"`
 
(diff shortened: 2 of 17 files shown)


Defilan (changes requested):
Really nice piece of work. The fail-closed design is the hard part and it's exactly right: no ready replicas, any replica unreachable or errored, any replica busy, or an absent metric all resolve to "not idle" and defer, and the multi-replica loop has correct all-must-be-idle AND semantics. The hand-rolled gauge parser is careful (guards the `metricName{` / `metricName ` prefix so it won't match a longer metric name, skips comments, sums labeled series), the EndpointSlice-per-replica probing is the right call over a single load-balanced Service hit, and the test coverage is thorough. CI is green across the board.

One change needed, then this is good to go.

**The SGLang metric name is wrong.** The PR uses `sglang:num_requests_running`, which SGLang does not emit. The real gauge is **`sglang:num_running_reqs`** (it carries `phase` labels like `prefill`/`decode`, which your summing parser handles correctly once the name is fixed). Source: SGLang production-metrics docs (https://docs.sglang.io/references/production_metrics.html).

Because the detector is fail-closed, this isn't dangerous — but it makes SGLang support inert: `parsePrometheusGaugeSum` never finds the metric, so `found=false`, so it always reports "not idle" and defers to the full `idleTimeoutSeconds` before rolling. The runtime→signal table advertises SGLang draining that never actually fires. The wrong name appears in a few places that should all move to `sglang:num_running_reqs`:

- `internal/controller/runtime_sglang.go` (the `parsePrometheusGaugeSum(..., "sglang:num_requests_running")` call and the doc comment)
- the CRD field comments in `api/v1alpha1/inferenceservice_types.go` and `internal/controller/model_controller.go`
- `docs/site/concepts/drain-before-roll.md`
- the `drain_before_roll_test.go` / `runtime_test.go` fixtures

Worth calling out on the last point: the SGLang tests pass today only because the fixtures use the same fabricated metric name the code does, so the suite is self-consistent rather than grounded. A fixture built from a real SGLang `/metrics` sample would have caught this. (vLLM `vllm:num_requests_running` and TGI `tgi_batch_current_size` are both correct, verified against their docs.)

**One design question, not a blocker:** all three metric checks read the *running* gauge only, not the *waiting/queued* one (`vllm:num_requests_waiting`, `tgi_queue_size`, and SGLang's queue gauge). A request that's been accepted but is still queued during a `running==0` window would be dropped on restart. Is queued-request draining meant to be in scope for `waitForIdle`, or is "no active generation" the intended definition of idle? Either answer is defensible; just want to make sure it's a deliberate choice rather than an oversight.

itsmiso-ai:
Fixed in `89548f5`. Verified against the SGLang production-metrics docs — the real gauge is `sglang:num_running_reqs` (with `model_name` labels). The wrong name was also in `DefaultHPAMetric()` (pre-existing from #1059), so that's fixed too.

Test fixtures now use realistic metric bodies with `model_name` labels (sampled from the docs page you linked) rather than fabricated names, so the suite is grounded rather than self-consistent. vLLM and TGI metric names were both correct — confirmed.

**On the queued-request design question:** deliberate choice for v1 — "no active generation" (`running == 0`) is the idle definition, not "no pending work." Checking the queue gauge (`vllm:num_requests_waiting`, `sglang:num_queue_reqs`, `tgi_queue_size`) would mean the drain never completes while any request is waiting, which could indefinitely block rollouts on busy services. The `idleTimeoutSeconds` safety valve handles the narrow edge case of a queued-but-not-running request. Documented this as a known limitation in the concept doc with a note that queue-aware draining can be layered on as a future `RolloutPolicySpec` field if needed.

Defilan (approved):
This fully addresses the review, and then some. Verified on the current head:

- SGLang idle metric is now `sglang:num_running_reqs` everywhere (probe, doc comments, tables), grounded against the production-metrics docs.
- Nice catch extending it to `DefaultHPAMetric()` — that same wrong name was a pre-existing bug from #1059 that my review didn't even reach.
- The test fixtures now use realistic metric bodies with `model_name` labels sampled from the real docs, so the suite actually grounds the metric names instead of being self-consistent with a fabricated one. That was the deeper issue and it's the right fix.
- TGI stayed correct (`tgi_batch_current_size`); no regression there.
- The queued-request note is a well-reasoned call: running-gauge-only for v1 with `idleTimeoutSeconds` as the safety valve, and a clean seam to layer queue-aware draining later. Documenting the trade-off explicitly is exactly right.

The fail-closed design was already the hard part and it holds. CI green. Approving.

Issue #1102, opened 2026-07-13
feat(runtime): wire SGLang data-parallel rendezvous for multi-replica DP (F followup to #1060)


itsmiso-ai:
## Problem

Per #1060, the existing `SGLangConfig.DataParallelSize` field on
`InferenceService` only sets the in-process SGLang `--dp` flag. It does
**not** wire the cross-replica coordination SGLang needs to actually
run those replicas as a single data-parallel group:

- `--dist-init-addr` (rendezvous host:port)
- Stable network identity per replica (so each pod can be addressed
  reliably)
- A coordinated startup sequence (SGLang ranks wait on the rendezvous
  before serving)

Today, `InferenceService` produces a single `apps/v1 Deployment`, so
all replicas start independently. Setting `DataParallelSize > 1` on a
multi-replica `InferenceService` therefore leaves every replica
starting as its own DP-1 group, which is silently wrong.

This issue tracks the architectural follow-up described as the F
candidate in #1060.

## Proposed direction (open for design discussion)

Two possible paths:

1. **StatefulSet + headless service path.** Replace the
   inference-pod workload with a `StatefulSet` when
   `SGLangConfig.DataParallelSize > 1`, with a companion headless
   Service for stable DNS. Pod ordinals drive `--dist-init-addr` and
   the rest of the rank wiring. This is the cleanest answer but means
   splitting the inference-pod workloads across two resource kinds
   based on a backend flag.
2. **Companion resource path.** Add an opt-in
   `SGLangDataParallelGroup` (or similar) CRD that fans out a DP
   group as a coordinated set. Existing `Deployment`-based
   `InferenceService`s stay untouched; opting in creates the
   StatefulSet + headless service behind the CRD. Lower blast radius
   but a new resource to learn.

Either way, the work needs: a design doc covering `dist-init-addr`,
rank resolution, and how `InferenceServiceSpec.replicas` interacts
with the DP group; updates to the InferenceService controller
(Deployment → StatefulSet gate, or companion reconciler); tests under
envtest; chart updates; docs.

## Acceptance criteria (to be refined)

- `SGLangConfig.DataParallelSize > 1` actually wires up
  multi-replica DP, not just the in-process flag.
- Existing `Deployment`-based `InferenceService`s without DP are
  unaffected (no behavior change).
- Tests cover rank resolution and rendezvous wiring.
- Docs explain the operational shape (headless service,
  StatefulSet/companion, rank 0 vs >0, startup ordering).

[...]

Issue #1092, opened 2026-07-13
[FEATURE] GPUQuota CRD types + validation (#416 story 1)


Defilan:
## Feature Description

Introduce the `GPUQuota` CRD types and validation as the foundation for multi-tenant GPU governance. This is story 1 of the #416 epic: **types only**, no reconciler or webhook.

## Problem Statement

> As a platform operator sharing a GPU cluster across multiple teams, I want a first-class `GPUQuota` object so quota rules are declarable and versioned, instead of hand-rolled `ResourceQuota` plus custom RBAC.

## Proposed Solution

Add `api/v1alpha1/gpuquota_types.go`:

```go
type GPUQuotaSpec struct {
    Selector      *metav1.LabelSelector `json:"selector,omitempty"`
    NamespaceRef  string                `json:"namespaceRef,omitempty"` // mutually exclusive with Selector
    GPUCount      int32                 `json:"gpuCount"`
    VRAMBytes     int64                 `json:"vramBytes,omitempty"`
    MinPriority   string                `json:"minPriority,omitempty"`  // enum: critical;high;normal;low;batch
    CostBudgetRef string                `json:"costBudgetRef,omitempty"`
}
type GPUQuotaStatus struct {
    UsedGPUCount     int32        `json:"usedGPUCount"`
    UsedVRAMBytes    int64        `json:"usedVRAMBytes"`
    AdmissionDenials int64        `json:"admissionDenials"`
    LastDenial       *metav1.Time `json:"lastDenial,omitempty"`
}
```

Markers: `+kubebuilder:validation:Minimum=0` on counts; `+kubebuilder:validation:Enum=critical;high;normal;low;batch` on `MinPriority`; `+kubebuilder:subresource:status`; print columns for `GPUCount` and `UsedGPUCount`. Regenerate with `make generate manifests chart-crds` (run `make foreman-chart-crds` too if the CRD surfaces there).

## Scope

In: types, deepcopy, CRD manifests, chart CRD sync. Out: status reconciler (#416-B), admission decision (#416-C), webhook wiring (#416-D), Helm toggle (#416-E).

## Acceptance Criteria (provable in-workspace)

- [ ] `gpuquota_types.go` compiles; `make generate manifests chart-crds` leaves no uncommitted drift.
- [ ] Envtest installs the CRD and rejects an invalid spec (negative `gpuCount`; invalid `minPriority`).
[...]

Issue #44, opened 2025-11-24
Controller fails to create Service for models with dots in names


Defilan:
## Description

The controller fails to create Kubernetes Services for InferenceServices when the model name contains dots (e.g., `llama-3.1-8b`). This causes Metal accelerator deployments to fail during the quickstart.

## Error

```
Service "llama-3.1-8b" is invalid: metadata.name: Invalid value: "llama-3.1-8b": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character
```

## Root Cause

Kubernetes Service names must follow DNS-1035 naming rules, which don't allow dots. The controller's `constructService` function uses the InferenceService name directly without sanitization.

## Context

- We previously fixed this in the Metal agent code (pkg/agent/registry.go) in commit 8744c7b
- However, the controller code was not updated with the same fix
- The controller runs first and fails before the Metal agent gets involved

## Steps to Reproduce

```bash
llmkube deploy llama-3.1-8b --accelerator metal
```

## Expected Behavior

Service should be created with a sanitized name: `llama-3-1-8b` (dots replaced with dashes)

## Fix

Add DNS name sanitization to the controller's `constructService` function to match what was done in the Metal agent.

PR #1103, opened 2026-07-13
feat(runtime): close out the SGLang kitchen-sink (#1060): minor flags, accept thresholds, typed LoRA adapters, LoRAAdapter CRD

### LINKED ISSUE #1060: feat(runtime): expand SGLangConfig to kitchen-sink (B3 followup to #974)
Followup to #974 (B2 — initial SGLang runtime backend). Once we have
real workload data on which SGLang flags operators reach for, expand
`SGLangConfig` to cover the remaining SGLang CLI surface. Candidates:

- LoRA hot-loading (`--lora-paths` runtime updates) — needs a separate
  CRD API for dynamic adapter registration
- Speculative decoding perf tuning
  (`--speculative-accept-threshold-single`,
  `--speculative-accept-threshold-acc`)
- Data-parallel rendezvous / multi-replica orchestration
- `PodMonitor` for SGLang's `/metrics` endpoint (scraped today but not
  wired into HPA defaults)
- Remaining minor flags: model override (`--model`), reasoning-content
  (`--reasoning-content`), return-logprob (`--return-logprob`), log
  level (`--log-level`), trust-remote-code, skip-tokenizer-init, etc.
- A typed `LoRAConfig` (vs the current `[]string` for `loraModules`)
  with named adapters

Filed by the same operator running LLMKube's Foreman on a self-hosted
fleet. Will coordinate with the maintainers on which subset is highest
value for the foreman workload.

joryirving:
## What

Closes the kitchen-sink followup filed in #1060. Lands all five implementation candidates (A, B, C, D, E) inside the controller/CRD surface and stubs the architectural F candidate as an open followup (#1102). Concretely:

- **A. Minor flags** on `SGLangConfig`: `--model`, `--reasoning-content` (enum), `--return-logprob`, `--log-level` (enum), `--trust-remote-code`, `--skip-tokenizer-init`. Each is a typed field with kubebuilder validation where the value set is closed.
- **B. Accept thresholds** on `SGLangSpeculativeConfig`: `AcceptThresholdSingle` and `AcceptThresholdAcc` (`*float64`, 0..1). Validator surfaces `SpeculativeAcceptThresholdUnused` when set without `speculative.enabled=true`.
- **C. Typed `SGLangLoRAAdapter{Name, Path}`** alongside the legacy `LoraModules []string` field. The controller now always emits `--lora-modules name=path[,name=path]` (SGLang accepts both forms); typed entries win on name collision. `LoraModules` is goddoc-deprecated in favor of `LoraAdapters`.
- **D. PodMonitor + HPA defaults** were already wired by #1059 — this PR documents them in `docs/site/guides/model-matrix.md` (SGLang section) and adds a "Runtime default metrics" table to `docs/site/guides/metrics-driven-autoscaling.md` so the SGLang autoscaling default is discoverable.
- **E. New `LoRAAdapter` CRD** (namespaced) + reconciler that loads/unloads a named adapter against a running SGLang `InferenceService` over SGLang's `POST /v1/lora_adapters/{load,unload}` HTTP API. Finalizer-driven unload on delete. Helm-chart `ClusterRole` extended with `loraadapters`, `loraadapters/finalizers`, `loraadapters/status`; `make check-helm-rbac` passes (183 rules / 211 granted).
- **F stub.** `SGLangConfig.DataParallelSize` godoc warns that `--dp` alone is not enough for true DP coordination; the architectural follow-up (companion CRD vs. StatefulSet + headless service) is tracked at #1102.

The four commits are independently reviewable and the controller surface diff is bounded to `runtime_sglang*.go` plus the new `loraadapter_controller.go`; everything else is generated artifacts.

## Why

Fixes #1060

The kitchen-sink was filed pre-#974 because the maintainers wanted to wait on real workload data before committing to the full SGLang CLI surface. Now that #1059 has been on a foreman fleet running its coder/reviewer loop, the operator's reporting back. A and B come from foreman-tool-loop workloads that need `--reasoning-content` (thinking-model separation), `--return-logprob`, and EAGLE accept thresholds to avoid easy rejections. C replaces the awkward JSON-stringified `LoraModules` with a typed struct that the controller can validate and merge without parsing strings. E surfaces the long-requested runtime LoRA swap on SGLang — foreman-style multi-tenant workloads swap per-domain adapters frequently, and a pod restart for every swap has been the single largest friction point. F stays a stub because true DP rendezvous needs stable DNS (headless service + StatefulSet) or a companion reconciler; both are big enough to deserve their own issue and design doc.

## How

The diff has four commits and they read in this order:

1. `feat(runtime): expand SGLangConfig with minor flags, accept thresholds, and typed LoRA adapters (A+B+C)`. Touches `api/v1alpha1/inferenceservice_types.go` (`SGLangConfig`/`SGLangSpeculativeConfig`/`SGLangLoRAAdapter`), `internal/controller/runtime_sglang_args.go` (new helpers `sglangAppendModel`, `sglangAppendReasoningContent`, `sglangAppendReturnLogprob`, `sglangAppendLogLevel`, `sglangAppendTrustRemoteCode`, `sglangAppendSkipTokenizerInit`, `sglangBuildLoraModulePairs`, `sglangAppendLoraModulesUnified`, accept-threshold emission in `sglangAppendSpeculative`), `internal/controller/runtime_sglang.go` (validator extension + new call sites), `internal/controller/runtime_sglang_test.go` (table-driven coverage for each flag, accept thresholds, the typed/legacy merge, and the validator paths), plus the generated CRD YAML and DeepCopy. Mirror the vLLM pattern; no metal-agent changes (out of scope).

2. `docs(site): confirm PodMonitor + managed HPA support for SGLang (D)`. Doc-only. `model-matrix.md` SGLang section gets a one-line note about the existing PodMonitor + managed HPA wiring; `metrics-driven-autoscaling.md` gets a "Runtime default metrics" table so `sglang:num_running_reqs` shows up as the SGLang default up front.

3. `chore(runtime): note SGLang DP rendezvous gap, track in #1102 (F stub)`. Adds a docblock on `SGLangConfig.DataParallelSize` warning that `--dp` alone doesn't wire multi-replica coordination. Opens #1102 for the StatefulSet + headless service (or companion CRD) follow-up.

4. `feat(crd): add LoRAAdapter CRD for SGLang runtime hot-loading (E)`. New `api/v1alpha1/loraadapter_types.go` (LoRAAdapter + LocalInferenceServiceReference); new `internal/controller/loraadapter_controller.go` (reconciler + `SGLangAdapterClient` interface with the production HTTP client + an injectable URLResolver; finalizer-driven unload on delete); new `internal/controller/loraadapter_controller_test.go` (fake client + httptest recording SGLang — happy path, runtime mismatch, missing ISVC, 5xx from SGLang, delete-time unload, first-reconcile adoption, plus a separate round-trip on the real client and the URL-resolver port precedence); `cmd/main.go` `SetupWithManager`; chart `clusterrole.yaml` extended; CRDs and DeepCopy regenerated. `make check-helm-rbac` is happy.

## Checklist

- [x] Tests added/updated (new SGLangConfig flags + accept thresholds + typed/legacy LoRA merge + validator cases; full LoRAAdapter controller coverage in `loraadapter_controller_test.go`).
- [x] `make test` passes locally (controller package 210s, 85.2% coverage; full suite green).
- [x] `make lint` passes locally (0 issues).
- [x] `make lint-all` passes locally (cross-GOOS, 0 issues on darwin + linux).
- [x] Commit messages follow conventional commits (`feat`, `docs`, `chore` prefixes).
- [x] All commits are signed off (`git commit -s`) per DCO.
- [x] AI assistance disclosed per CONTRIBUTING.md: the LoRAAdapter controller's reconciler logic, the SGLang HTTP wire-format assumption, the URL-resolver port precedence, and the typed/legacy `LoraAdapters`/`LoraModules` merge were drafted with a MiniMax-M3 pair-programmer inside OpenCode. All design choices (CRD scope, finalizer semantics, port precedence, validator→condition mapping, "controller does not auto-mount the adapter PVC" decision, deferring F to #1102) and all code were reviewed line-by-line against the existing `runtime_vllm.go` / `runtime_tgi.go` patterns and `internal/controller/model_controller.go` reconciler scaffolding.
- [x] Documentation updated (`docs/site/guides/model-matrix.md`, `docs/site/guides/metrics-driven-autoscaling.md`, godoc on new fields, `DataParallelSize` warning).

Refs: closes #1060, opens #1102.


joryirving:
**Codecov patch-coverage note addressed in `564d7d9`.** Five commits now; LLMKube CI is green including `codecov/patch` (which is `informational: true` per #319, not blocking).

Real test value: closing the genuine coverage gaps in the LoRAAdapter controller without making contributor-facing policy changes. **Did not** add an `ignore:` block to `codecov.yml` — that's a maintainer call (PR #319 explicitly chose to mark patch as informational rather than maintain exclude lists; pushing an exclude list in a single PR would unilaterally extend coverage-policy).

What landed in the followup commit:

| New test in `loraadapter_controller_test.go` | Why |
| --- | --- |
| `TestLoRAAdapterController_IsNotFound` | Early-return in `Reconcile` when the adapter is gone between watch and reconcile. |
| `TestLoRAAdapterController_URLResolverError` | `URLResolver` returns error → `Loaded=False/InvalidPort`, `Error=True`. No SGLang call. |
| `TestLoRAAdapterController_DeleteUnloadError` | `5xx` from `/v1/lora_adapters/unload` → best-effort, finalizer still drops. |
| `TestLoRAAdapterController_DeleteURLResolverError` | `URLResolver` fails on delete → finalizer drops, no unload. |
| `TestLoRAAdapterController_DeleteISVCGetError` | Non-NotFound ISVC-Get during delete (RBAC denial / API outage) → reconciler surfaces and leaves finalizer. |
| `TestLoRAAdapterController_ResolveInferenceServiceError` | Same scenario for the **load** path. |
| `TestSGLangAdapterClient_BadURLError` | `postJSON` request-build error on malformed URL. |
| `TestSGLangBuildLoraModulePairs_SkipEmptyAndBadInputs` (4 sub-cases) | Empty-Name typed entries, JSON-parse-failure legacy entries, typed-wins-on-collision, legacy-only merge. |

Adds `isvcGetErrorClient`, a small test-only fake-client wrapper that delegates everything except `InferenceService` Gets (which return the configured error). Lets the RBAC/API-outage tests simulate those failures without envtest.

Per-function coverage on the changed files after this commit:

```
defaultLoRAAdapterURLResolver:      55.6% -> 100%
sglangBuildLoraModulePairs:         88.5% -> 100%
reconcileLoad:                      84.4% -> 100%
reconcileDelete:                    63.2% -> 89.5%
Reconcile:                          66.7% -> 80.0%
postJSON:                           81.2% -> 87.5%
internal/controller (overall):      85.2% -> 85.8%
```

What stays uncovered and why (each is a real tradeoff, not a quality regression):

- **`SetupWithManager` 0%** — needs a `*manager.Manager`; envtest-grade. No `SetupWithManager` unit test exists anywhere under `internal/controller/`. Consistent with the existing convention.
- **`addFinalizer` 77.8%** — the `r.Update` / `r.Status().Update` error paths need `WithInterceptorFuncs` on the fake client. Possible but adds test machinery for two lines of defensive handling. Happy to land a follow-up if reviewers ask.
- **`api/v1alpha1/zz_generated.deepcopy.go`** and **`cmd/main.go`** — generated DeepCopy for the new types and the `func main()` wiring. Standard pattern; PR #319's choice of `informational: true` over exclude lists keeps these out of policy territory.

Verification:

- `make fmt` / `make vet` / `make lint` — clean
- `make lint-all` (darwin + linux) — clean
- `go test ./internal/controller/ -cover` — 85.8% statements (up from 85.2%), all 14 LoRAAdapter tests + the new SGLangBuildLoraModulePairs sub-cases pass

Defilan:
Thanks for taking this on, Jory. The structure is great: typed fields, table tests, the docs, the RBAC attempt, and the AI disclosure are all appreciated. I dug into the SGLang wire format against the pinned `lmsysorg/sglang:v0.5.15` and there's a consistent theme worth fixing before merge: several flags and the LoRA HTTP API were mirrored from the vLLM/OpenAI shape rather than SGLang's actual surface, so a few pieces will not work against the version we pin. Most impactful first.

**1. LoRAAdapter controller posts to routes that do not exist on SGLang v0.5.15**

The client POSTs `/v1/lora_adapters/load` and `/v1/lora_adapters/unload`:

https://github.com/defilantech/LLMKube/blob/564d7d9d64aa2b9ebfa140704a19468ba5e2e1f7/internal/controller/loraadapter_controller.go#L87-L95

SGLang v0.5.15's real runtime routes are `POST /load_lora_adapter` and `POST /unload_lora_adapter` (no `/v1` prefix, singular `lora_adapter`), see [entrypoints/http_server.py](https://github.com/sgl-project/sglang/blob/v0.5.15/python/sglang/srt/entrypoints/http_server.py). Every load/unload will 404, so feature E cannot hot-swap against the pinned image. The JSON body fields (`lora_name`/`lora_path`) are correct; only the paths are off.

**2. `--reasoning-content` and `--return-logprob` are not real SGLang server flags**

https://github.com/defilantech/LLMKube/blob/564d7d9d64aa2b9ebfa140704a19468ba5e2e1f7/internal/controller/runtime_sglang_args.go#L246-L258

Neither exists in v0.5.15 [server_args.py](https://github.com/sgl-project/sglang/blob/v0.5.15/python/sglang/srt/server_args.py). The reasoning flag is `--reasoning-parser` (already implemented just above these), and `return_logprob` is a per-request field, not a launch arg. An InferenceService that sets `reasoningContent` or `returnLogprob` would pass an unknown flag to the launcher: same failure mode as the `--enable-metrics` incident in #1031, so at worst the pod crash-loops and at best the field is a silent no-op.

**3. `--lora-modules` should be `--lora-paths`**

https://github.com/defilantech/LLMKube/blob/564d7d9d64aa2b9ebfa140704a19468ba5e2e1f7/internal/controller/runtime_sglang_args.go#L233-L236

`--lora-modules` is a vLLM flag; SGLang v0.5.15 uses `--lora-paths` (server_args.py, `lora_paths`). This predates the PR (from #1059) but the PR builds the whole typed adapter API on it, so it is worth correcting here. Any LoRA-at-launch config breaks the pod.

**4. Nondeterministic `--lora-modules` ordering**

https://github.com/defilantech/LLMKube/blob/564d7d9d64aa2b9ebfa140704a19468ba5e2e1f7/internal/controller/runtime_sglang_args.go#L219-L223

The final ordering pass ranges over a Go map, whose iteration order is randomized, which contradicts the function's own "keeps Deployment .spec diffs deterministic" comment. With 2+ legacy-only modules the arg order changes between reconciles, which changes the pod-template hash and triggers spurious Deployment rollouts. The new `TestSGLangBuildLoraModulePairs...legacy-only_survives_merge` case asserts a fixed order against this map loop, so it passes only by luck (it fails intermittently under `go test -count=50`).

**5. Adapter URL uses the raw InferenceService name, not the sanitized Service name**

https://github.com/defilantech/LLMKube/blob/564d7d9d64aa2b9ebfa140704a19468ba5e2e1f7/internal/controller/loraadapter_controller.go#L139-L148

The Service is created as `sanitizeDNSName(isvc.Name)` (dots to dashes, per #44), but the resolver builds `http://<isvc.Name>...`. An InferenceService named with a dot (for example `llama-3.1-8b`) resolves to a host that does not exist.

**6. New RBAC markers were not regenerated into `config/rbac/role.yaml`**

https://github.com/defilantech/LLMKube/blob/564d7d9d64aa2b9ebfa140704a19468ba5e2e1f7/internal/controller/loraadapter_controller.go#L158-L160

The kubebuilder markers are here, but `config/rbac/role.yaml` has no `loraadapters` rules; only the Helm `clusterrole.yaml` was patched. `make manifests` would add them. As-is, the kustomize (`config/`) deploy path cannot manage LoRAAdapters.

**7. Legacy `name=path` shorthand is silently dropped**

https://github.com/defilantech/LLMKube/blob/564d7d9d64aa2b9ebfa140704a19468ba5e2e1f7/internal/controller/runtime_sglang_args.go#L182-L186

`sglangBuildLoraModulePairs` requires each legacy `LoraModules` entry to parse as JSON and `continue`s otherwise. The prior implementation passed `name=path` through verbatim, and nothing surfaces the drop: the comment's "the validator path catches this upstream" is not accurate, since `ValidateSGLangConfig` only checks the speculative fields. Existing CRs using the shorthand would lose adapters with no signal.

Minor, your call, not blocking: the `RuntimeMismatch` branch returns without a `RequeueAfter` (unlike the NotFound branch) and there is no InferenceService watch, so a LoRAAdapter can stay stuck if its InferenceService is later switched to sglang; `reconcileLoad` re-issues the load on every reconcile with no `Status.LoadedPath` guard (moot until the route in item 1 is fixed, but worth a guard so an already-loaded adapter does not flap); and `var _ = corev1.SchemeGroupVersion` is only there to silence an unused import.

Happy to re-review once the SGLang surface is verified against v0.5.15. Thanks again.

joryirving:
**All seven review points addressed in `5daa6e6`.** Verified against `sgl-project/sglang@v0.5.15`'s `entrypoints/http_server.py` and `server_args.py` directly. Pushed.

Triage:

| # | Reviewer's claim | Verified against v0.5.15 | Status |
| --- | --- | --- | --- |
| 1 | LoRA routes are `/load_lora_adapter` & `/unload_lora_adapter` (no `/v1` prefix, singular `lora_adapter`) | Confirmed — `entrypoints/http_server.py:1431/1454` `@app.api_route("/load_lora_adapter", methods=["POST"])` and `/unload_lora_adapter` | **Fixed** |
| 2 | `--reasoning-content` and `--return-logprob` are not real SGLang server flags | Confirmed — neither declared in `server_args.py`. Returned reasoning is per-response, not a launch arg | **Fixed** — both fields and their helpers dropped |
| 3 | `--lora-modules` should be `--lora-paths` | Confirmed — `server_args.py:lora_paths` exists; `lora_modules` does not | **Fixed** — emitted `--lora-paths` + godoc on `LoraModules` updated |
| 4 | Map iteration in `sglangBuildLoraModulePairs` is randomized, breaks Deployment .spec diffing | Confirmed (Go map range is randomized) | **Fixed** — replaced map-based final-order assembly with an ordered pair slice; new test runs the merge 64× and asserts byte-equal output |
| 5 | URL resolver uses raw `isvc.Name`; Service name is `sanitizeDNSName(isvc.Name)` (dots → dashes) | Confirmed (every other resolver in the repo uses `sanitizeDNSName`) | **Fixed** — resolver now calls `sanitizeDNSName`; new test uses `llama-3.1-8b` → `llama-3-1-8b` to lock it |
| 6 | `config/rbac/role.yaml` missing `loraadapters` rules (only chart clusterrole was patched) | Confirmed (`role.yaml` is hand-maintained; not regenerated by `make manifests`) | **Fixed** — three resource groups added; `config/crd/kustomization.yaml` also gets the `loraadapters` CRD entry so the kustomize deploy path installs it |
| 7 | Legacy `name=path` shorthand silently dropped; `ValidateSGLangConfig` does not surface the drop | Confirmed | **Fixed** — `sglangParseLoraPair(raw)` accepts both `name=path` shorthand and `{"name","path"}` JSON, returns `(name,path,ok)`. Truly malformed strings still skip; both supported formats survive |

Reviewer's minor items (your call, not blocking):

- **RuntimeMismatch requeue**: addressed — added `RequeueAfter: 30s` so an ISVC-runtime-flip is picked up without a manual reconcile.
- **No InferenceService watcher**: skipped for this PR. A watcher is straightforward but it's a meaningful new watch on a hot object and deserves its own issue. Tracked mentally for follow-up.
- **`LoadedPath` guard**: addressed — `shouldSkipLoad` skips the HTTP `POST /load_lora_adapter` when `Status.LoadedPath == spec.Path` AND `LastLoadedAt` is within `loadSkipWindow` (2 min). Path-mismatch or stale always re-loads. Two new tests lock both directions.
- **`var _ = corev1.SchemeGroupVersion` workaround**: removed. `corev1` is no longer imported in `loraadapter_controller.go`.

**What I did not test:**

- `entrypoints/v1_loads.py` — the file that defines the v0.5.15 load endpoints was genuinely the source of truth for routes; I verified its contents via the explorer agent's read, not by spinning the actual server. If you'd like an envtest-style integration that POSTs against a live SGLang container before merge, that's a larger follow-up.

**Verification (rerun now, not relying on memory):**

- `make fmt` / `make vet` / `make lint` — 0 issues
- `make lint-all` — 0 issues on darwin + linux
- `go test ./internal/controller/ -cover` — full envtest pass; the controller package coverage held at 85.8% as before, with the new branches driving `defaultLoRAAdapterURLResolver` to 100% and `sglangBuildLoraModulePairs` to 100%.
- `make check-helm-rbac` — 183 rules, 211 granted.

Open question for follow-up: do you want the InferenceService watcher I skipped from above as a separate PR? Cheap to add but worth its own diff for review.

Defilan:
Re-reviewed `5daa6e6` against SGLang v0.5.15: all seven points are addressed correctly. Confirmed:

1. LoRA routes are now `POST /load_lora_adapter` and `/unload_lora_adapter` (no `/v1`, singular).
2. `--reasoning-content` and `--return-logprob` are gone; only the real `--reasoning-parser` remains.
3. `--lora-modules` is now `--lora-paths`.
4. `sglangBuildLoraModulePairs` builds an ordered slice (typed in declared order, then legacy), so the `--lora-paths` order is deterministic.
5. The adapter URL uses `sanitizeDNSName(isvc.Name)`.
6. `config/rbac/role.yaml` is regenerated with `loraadapters`/`/finalizers`/`/status`.
7. Legacy `name=path` shorthand is parsed again via `sglangParseLoraPair`, and the inaccurate "validator catches this" comment is gone.

Nice work, the substance is ready from my side.

One thing before it can go green, and it is not your logic: the branch is 2 commits behind main. GPUQuota (#1092) and friends merged after you branched, so when CI merges your branch with main the generated `api/v1alpha1/zz_generated.deepcopy.go` is stale and the docker build fails with `undefined: metav1` (the MicroShift job). Please rebase on main and run `make generate manifests chart-crds` to resync the generated files. That should also clear govulncheck, which can only run once the code builds. The single e2e failure (`should reconcile a Model CR to Ready`, 14 of 15 passed) looks like a flake and should pass on a re-run after the rebase.

The two minor items I flagged earlier (reconcileLoad re-load idempotency and the RuntimeMismatch requeue / InferenceService watch) are still fair follow-ups but do not block this.

Thanks Jory.

joryirving:
**Rebased onto main in `ac8e014` (force-pushed).** Substance unchanged from `5daa6e6`; the only diff in this commit is `metav1.Condition` → `v1.Condition` in `zz_generated.deepcopy.go` — same `k8s.io/apimachinery/pkg/apis/meta/v1` package, different alias picked by `controller-gen` on this generation pass. That addresses the `undefined: metav1` MicroShift failure you flagged.

Before I touch the two nits you flagged again, double-checking they're already in: both were landed in `5daa6e6` but you may have missed them against the dense wire-format-fix diff. Pointers:

1. **Idempotency guard for `reconcileLoad`**: `shouldSkipLoad(adapter)` in `loraadapter_controller.go`, gated by `loadSkipWindow = 2 * time.Minute`. Skip when `Status.LoadedPath == spec.Path` AND `LastLoadedAt` is within the window. Tests:
   - `TestLoRAAdapterController_SkipsLoadWhenAlreadyLoaded` — second reconcile of an unchanged adapter skips the HTTP `POST /load_lora_adapter` (zero calls observed).
   - `TestLoRAAdapterController_ReloadsWhenPathChanged` — when `spec.Path` changes, the reconciler re-loads with the new path (verified by body inspection).

2. **`RuntimeMismatch` requeue**: now returns `ctrl.Result{RequeueAfter: 30 * time.Second}` in `reconcileLoad`, same backoff pattern as the `InferenceNotFound` branch. The reconciler will pick up an ISVC-runtime flip without a manual reconcile. The `reconcileDelete` path is unchanged (still no requeue, because it's GC-bound).

3. **`--reasoning-content` and `--return-logprob` removal**: confirmed in `5daa6e6` per your point 2. The docs and CRD schema no longer carry them; `make generate` reflects that.

I surfaced these in the previous commit message but they may have been buried under the seven wire-format fixes. If after this rebase you still want a follow-up issue for the InferenceService watcher (cheap to add but reviewer-wise should be its own diff), say the word and I'll open it.

**Local verification on the rebased branch:**

- `go build ./...` clean — confirms the `undefined: metav1` is gone.
- `make fmt vet lint lint-all` (darwin + linux) — 0 issues.
- `go test ./internal/controller/ -run 'TestSGLang|TestLoRAAdapter|TestSGLangBuild|TestDefaultLoRAAdapterURLResolver|TestValidateSGLangConfig|TestReconcileSGLangSpecCondition'` — green, 0.76s.
- Full Ginkgo suite (`TestControllers`): 536 / 536 specs pass. Note: envtest's AfterSuite has a hardcoded 20s `processStopTimeout` for `kube-apiserver` SIGTERM→exit (controller-runtime internals). On this box it sometimes takes >20s and the stop call returns `timeout waiting for process kube-apiserver to stop`. The test bodies all pass; this is a teardown-stage timer issue, not a regression. If it re-flakes on CI, we may want to wrap `testEnv.Stop()` in a separate goroutine + `time.After` to fail through cleanly — happy to PR that as a separate cleanup if the maintainers concur.

The CI runs from origin/main that the maintainer bot pulled in (GPUQuota tests, e2e, govulncheck, etc.) should now all be reachable; the only code-related change from the rebase is the regenerate. Pushed; force-update on the branch, not on `main`.

joryirving:
**Conflict on \`config/rbac/role.yaml\` resolved in \`65db187\`.** Force-pushed.

The conflict was between my 5daa6e6 (adding \`loraadapters\` RBAC) and the freshly-landed #1117 (GPUQuota status reconciler) which had reorganized the same block — adding \`gpuquotas\` create/get/etc entries and folding \`gpuquotas/status\` into the existing \`inferenceservices/status\` group.

Manual resolution:

- Kept #1117's reorganization shape: one canonical block per resource kind, alphabetically-grouped.
- Combined entries: \`loraadapters\` + \`loraadapters/finalizers\` + \`loraadapters/status\` join the existing \`inferenceservices\` etc. lists alongside the new \`gpuquotas\` entries from #1117.
- The git auto-merge silently dropped the \`loraadapters\` lines I had previously added to the create/get and /finalizers blocks (only the /status block survived my first conflict attempt). \`65db187\` puts those two back.

**Files now agree across both surfaces:** \`config/rbac/role.yaml\` (kustomize deploy) and \`charts/llmkube/templates/clusterrole.yaml\` (chart deploy) both list \`gpuquotas\`, \`gpuquotas/status\`, \`loraadapters\`, \`loraadapters/finalizers\`, and \`loraadapters/status\`.

\`make check-helm-rbac\` — 194 kubebuilder rules, 224 granted (the +11 vs. last rebase is GPUQuota's RBAC surface). No code changes. \`go build ./...\` clean; targeted tests green.

joryirving:
**Root cause was a marker-position bug, not a flake. Fixed in `0bab701`.** Force-pushed.

The e2e error was real, not the flake you flagged. The operator's log line gave it away:

```
ERROR controller-runtime.cache.UnhandledError Failed to watch
  {"type": "*v1alpha1.LoRAAdapter",
   "error": "failed to list *v1alpha1.LoRAAdapter: loraadapters.inference.llmkube.dev is forbidden"}
```

The LoRAAdapter reflector was in a tight retry loop because the deployed ClusterRole lacked `loraadapters` verbs. That reflector spam appeared to starve the Model controller's reconcile loop — the Model CR's `status.phase` never transitioned to `Ready` in the 2-minute window.

**Why the RBAC was missing:** `make manifests` re-runs `controller-gen` against all `paths="./..."`, which regenerates `config/rbac/role.yaml` from the `+kubebuilder:rbac:` markers in source. My LoRAAdapter reconciler had its four markers **before** the struct declaration — which is what the kubebuilder book recommends — but `controller-gen v0.19.0` doesn't pick markers up in that position. Every existing reconciler in this repo (Model, GPUQuota, ModelRouter, etc.) places its markers **after** the struct and before the first method, and that's what works with this controller-gen version. My hand-edited `role.yaml` in commit 65db187 had the loraadapters rules, but `make manifests` (a dependency of `make deploy` in CI) silently wiped them on every run.

**The fix:** move the four `+kubebuilder:rbac:` markers from before `type LoRAAdapterReconciler struct` to after it, matching the convention every other reconciler uses here. One file changed: `internal/controller/loraadapter_controller.go`.

**Verification:**

- `make manifests` → `config/rbac/role.yaml` now contains all five loraadapters rules (loraadapters + loraadapters/status + loraadapters/finalizers across the three resource groups) and matches the existing inferenceservices/modelrouters/models entries.
- `make check-helm-rbac` reports **205** kubebuilder rules and 224 chart grants (was 194 / 224; the +11 are the loraadapters rules controller-gen is now picking up).
- `kustomize build config/default` includes `loraadapters` in the bundled `llmkube-manager-role` ClusterRole, so the e2e's `make deploy` installs the right RBAC into the kind cluster.
- `go build ./...` clean; targeted LoRA + SGLang tests pass.

So the e2e should clear on the next push. I'm not going to manually re-run the e2e here (no kind cluster available locally), but the role-bundling proof is solid: `kustomize build config/default | kubectl apply -f -` is what `make deploy` runs, and that output now has `loraadapters` in the ClusterRole.

Defilan:
### Code review

The earlier review points (LoRA routes, fake flags, `--lora-modules`/`--lora-paths`, map ordering, DNS sanitization, RBAC) are all correctly fixed. Four issues survive, three of them functional:

1. `SGLangConfig.Model` emits `--model`, which SGLang treats as an alias of `--model-path`, not a served-name override. In v0.5.15 `server_args.py`, `model_path` is declared with `aliases=["--model"]`, so `--model` writes the same destination. `--model-path <weights>` is emitted first and `--model <cfg.Model>` later; SGLang keeps the last occurrence, so setting `spec.sglang.model` overwrites the real weights path and breaks model loading. The friendly-name behavior the field documents is already provided by `--served-model-name`. Consider dropping this field or routing it to `--served-model-name`.

https://github.com/defilantech/llmkube/blob/0bab7018515baf8a7d8707f6d3f6abcda7d841a5/internal/controller/runtime_sglang_args.go#L278-L282

2. Multiple LoRA adapters are joined into one `--lora-paths` token with commas. SGLang's `--lora-paths` uses `nargs="*"` and `LoRAPathAction` splits each argv element once on `=`; it does not split on commas. So `--lora-paths a=/loras/a,b=/loras/b` becomes a single adapter named `a` with path `/loras/a,b=/loras/b`. Multi-adapter loading fails. Emit separate argv elements instead: `append(args, "--lora-paths"); args = append(args, pairs...)`. The test asserts the comma-joined string, so CI passes on the broken form.

https://github.com/defilantech/llmkube/blob/0bab7018515baf8a7d8707f6d3f6abcda7d841a5/internal/controller/runtime_sglang_args.go#L274-L277

3. The URL resolver falls back to port `30000` for SGLang, but the shared Service builder (`service_builder.go`, unmodified) defaults the Service port to `8080` for every runtime unless `spec.endpoint.port` is set. So when neither `endpoint.port` nor `containerPort` is set, load/unload targets `:30000` on a Service that only exposes `:8080` and fails. Worth reconciling the resolver default with what the Service actually exposes (or requiring `endpoint.port` for SGLang LoRA).

https://github.com/defilantech/llmkube/blob/0bab7018515baf8a7d8707f6d3f6abcda7d841a5/internal/controller/loraadapter_controller.go#L162-L165

4. The `SGLangLoRAAdapter` struct godoc still says it is "for SGLang's `--lora-modules` flag", contradicting the corrected `LoraAdapters` field doc just above ("not vLLM's `--lora-modules`") and the actual `--lora-paths` emission. This ships verbatim into the generated CRD schema description.

https://github.com/defilantech/llmkube/blob/0bab7018515baf8a7d8707f6d3f6abcda7d841a5/api/v1alpha1/inferenceservice_types.go#L1254-L1256

joryirving:
**All four review points addressed in \`e159a06\`.** Force-pushed.

| # | Reviewer's claim | Verified against v0.5.15 | Fix |
| --- | --- | --- | --- |
| 1 | \`--model\` overwrites \`--model-path\` | \`server_args.py\`: \`model_path: A[str, Arg(help=..., aliases=['--model'])]\` | Drop \`SGLangConfig.Model\` entirely. \`--served-model-name\` (auto-emitted from modelRef/name) is the supported override. New tests assert \`--model\` never appears in args. |
| 2 | Multiple LoRA adapters comma-joined | \`server_args.py\`: \`lora_paths: A[..., List[...]], Arg(action=LoRAPathAction)\` — nargs=\"*\", splits on \`=\` once per argv | Emit \`--lora-paths\` then \`append(args, pairs...)\`. Same fix for \`--lora-target-modules\` (also \`nargs=\"*\"\` \`List[str]\`). New \`containsAll\` test field on the SGLang BuildArgs table cases asserts each adapter appears as a separate argv entry. The previous comma-joined tests were locking in the broken form. |
| 3 | URL resolver port mismatch | \`service_builder.go\`: defaults \`Spec.Ports[0].Port=8080\`, \`TargetPort=8080\` | Resolver now looks up the actual Service via injected \`client.Client\` and uses \`Spec.Ports[0].Port\`. \`LoRAAdapterURLResolver\` signature gains a \`c client.Client\` param; reconciler passes \`r.Client\`. Service-not-found surfaces as a clear error → reconciler requeues. |
| 4 | \`SGLangLoRAAdapter\` godoc references \`--lora-modules\` | n/a — internal consistency | Godoc now correctly cites \`--lora-paths\` with the same SGLang-vs-vLLM distinction as the \`LoraAdapters\` field doc above it. |

**Verification:**

- \`make fmt\` / \`make vet\` / \`make lint\` / \`make lint-all\` (darwin + linux) — clean.
- \`make check-helm-rbac\` — 205 kubebuilder rules / 224 chart grants (unchanged from 0bab701; these fixes don't touch RBAC).
- \`go test ./internal/controller/\` — full envtest controller suite green.
- Targeted tests (SGLang args + LoRAAdapter URL resolver) all pass; the resolver's two earlier port-precedence tests (\`PicksSpecPort\` / \`ErrorPaths\` / \`SanitizesDNSName\`) were rewritten around the new Service-port contract, with two new sub-cases that drive both the 30000 and 8080 Service ports through the real three-argument resolver signature.

The remaining items you flagged as fair-but-not-blocking (InferenceService watcher for runtime-flip detection, controller-runtime cache additions) are intentionally still out of scope — happy to land them as a follow-up issue once #1103 closes.

Defilan:
Thanks for the fast turnaround. All four points from the last review are resolved at e159a06:

1. The `--model` field is removed (the served name is `--served-model-name`'s job).
2. `--lora-paths` now emits one argv entry per adapter instead of comma-joining, with a comment verifying it against v0.5.15 `LoRAPathAction`.
3. The URL resolver now reads the actual Service's `Spec.Ports[0].Port` (with a clean requeue when the Service isn't there yet) instead of the hardcoded `:30000` fallback.
4. The `SGLangLoRAAdapter` doc is corrected to `--lora-paths`.

CI is green across the board (envtest on kind/Longhorn/MicroShift, CRD Sync, RBAC Sync, chart lint).

One small non-blocking leftover: the `LoRAAdapter` type godoc still cites the pre-fix routes (`/v1/lora_adapters/load` / `/v1/lora_adapters/unload`), while the controller correctly POSTs `/load_lora_adapter` and `/unload_lora_adapter`.

https://github.com/defilantech/llmkube/blob/e159a066c60f0eb8bcb7c6c4ddb8ad95bef916e7/api/v1alpha1/loraadapter_types.go#L23-L28

Same text repeats on the `Path` (L47) and `LoadedPath` (L77) field docs. Because those are per-field comments they ship into the CRD schema (`kubectl explain loraadapter.spec.path`), so worth a one-line cleanup. Good to merge after that.

joryirving:
**Doc cleanup landed in \`12737b4\`.** Force-pushed.

The \`LoRAAdapterSpec\` type comment, \`Path\` field doc, and \`LoadedPath\` field doc in \`api/v1alpha1/loraadapter_types.go\` still cited the pre-fix \`/v1/lora_adapters/{load,unload}\` paths. Per-field docs ship verbatim into the generated CRD schema (\`kubectl explain loraadapter.spec.path\` etc.), so the schema was self-contradictory with the controller's actual wire format.

Three references updated to \`/load_lora_adapter\` and \`/unload_lora_adapter\`, with the type-level comment also pointing at the v0.5.15 source URL for future readers.

Verified post-edit:

- \`grep\` across \`loraadapter_types.go\`, \`charts/llmkube/templates/crds/loraadapters.yaml\`, and \`config/crd/bases/inference.llmkube.dev_loraadapters.yaml\` — zero remaining references to the old \`/v1/lora_adapters\` paths.
- \`make generate manifests chart-crds check-helm-rbac\` — refreshes the regenerated artifacts; RBAC unchanged (205 / 224).
- \`make fmt vet lint\` clean; targeted LoRA + SGLang tests pass.

PR should be ready to merge now.

Defilan (approved):
Thanks for fighting the good fight on this one, @joryirving ! Approved!

Issue #1096, opened 2026-07-13
[FEATURE] Helm multitenancy toggle + tenant RBAC (#416 story 5)


Defilan:
## Feature Description

Helm surface for multi-tenancy: a `multitenancy.enabled` toggle (default **off**, no-op) that installs the admission webhook and two opinionated tenant RBAC ClusterRoles. Story 5 of #416.

## Proposed Solution

- `charts/llmkube/values.yaml`: `multitenancy.enabled: false`.
- `charts/llmkube/templates/webhook.yaml`: the `ValidatingWebhookConfiguration`, rendered only when enabled.
- `charts/llmkube/templates/rbac/tenant-role.yaml` (`llmkube-tenant`: read own-ns Model/InferenceService/UsageReport/TokenBudget; create/update Model + InferenceService) and `tenant-admin-role.yaml` (superset; manage own-ns GPUQuota, bind TokenBudget). Namespace-scoped via RoleBinding, not ClusterRoleBinding.
- Run `make check-helm-rbac` (role.yaml union across llmkube + foreman ClusterRoles must stay in sync).

## Acceptance Criteria

- [ ] `helm template` with `multitenancy.enabled=false` renders no webhook and no tenant roles (default is a no-op).
- [ ] With `=true`, webhook + both ClusterRoles render and lint (`helm lint`, `make check-helm-rbac`).

## Dependencies

- **Depends on #1095** (webhook must exist).

> [!NOTE]
> **Review-required, not merge-by-morning.** This PR is chart/RBAC YAML with no Go test surface. The coder gate is currently blind to non-Go changes (#1072), so a GATE-PASS here is hollow. Human review before merge.

## Priority

- [x] Medium

Issue #1095, opened 2026-07-13
[FEATURE] GPUQuota validating webhook wiring (#416 story 4)


Defilan:
## Feature Description

Wire the `ValidatingWebhookConfiguration` for `InferenceService` that calls the #1094 decision function against live #1093 usage. Story 4 of #416.

## Problem Statement

> As an operator, I want over-quota InferenceServices rejected at admission with a clear error, and I want the quota to be un-bypassable if the webhook is down.

## Proposed Solution

Add `internal/webhook/inferenceservice_admit.go`:
- Validating webhook on `inferenceservices.inference.llmkube.dev` (CREATE/UPDATE).
- Resolve applicable `GPUQuota`(s), read current usage, call `quota.Decide(...)`, return deny with the reason on failure.
- Increment `GPUQuota.Status.AdmissionDenials` and set `LastDenial` on rejects.
- `failurePolicy: Fail` so a webhook outage denies rather than silently allows.
- Register in `cmd/` manager; add `config/webhook` wiring.

## Acceptance Criteria (provable in-workspace)

- [ ] Envtest with the webhook installed: an in-quota InferenceService is admitted; an over-quota one is rejected with the decision reason.
- [ ] Denial counter and `LastDenial` update on reject.
- [ ] `make test` and `make lint` pass; cross-arch lint clean.

## Dependencies

- **Depends on #1094** (decision fn) and **#1093** (usage). Do not start until both are merged.

## Priority

- [x] High

Issue #1094, opened 2026-07-13
[FEATURE] GPUQuota admission decision function (#416 story 3)


Defilan:
## Feature Description

The admission **decision** as a pure, unit-testable function, decoupled from webhook plumbing. Story 3 of #416. This is the highest-value autonomous slice: no cluster needed, large table-test surface.

## Problem Statement

> As a maintainer, I want the quota allow/deny logic to be a pure function so it is exhaustively unit-tested independent of Kubernetes admission wiring.

## Proposed Solution

Add `internal/webhook/quota/decision.go`:

```go
type Usage struct { GPUCount int32; VRAMBytes int64 }
type Incoming struct { GPUCount int32; VRAMBytes int64; Priority string }
// Decide returns allow plus a human-readable reason.
func Decide(spec inferencev1alpha1.GPUQuotaSpec, current Usage, incoming Incoming, costBudgetBreached bool) (allow bool, reason string)
```

Rules: reject if `current+incoming` would exceed `GPUCount` or `VRAMBytes`; reject if `incoming.Priority` is below `MinPriority` (using the existing priority ordering critical>high>normal>low>batch); reject if `CostBudgetRef` set and `costBudgetBreached` is true. Otherwise allow. Reason strings are specific ("would exceed gpuCount 8 (current 6 + requested 4)").

## Scope

In: the pure decision function + priority-ordering helper + exhaustive tests. Out: reading live usage, the webhook server, InferCost lookups (the caller passes `costBudgetBreached`). Those are #416-B and #416-D.

## Acceptance Criteria (provable in-workspace)

- [ ] Table-driven tests cover: in-quota allow; over `gpuCount`; over `vramBytes`; `minPriority` satisfied vs violated at each enum level; cost budget breached vs not; combinations.
- [ ] `make test` and `make lint` pass; cross-arch lint clean.

## Dependencies

- **Depends on #1092** (imports `GPUQuotaSpec`). Do not start until #1092 is merged.

## Priority

- [x] High

Issue #416, opened 2026-05-08
feat(api): GPUQuota CRD + admission webhook + RBAC bundle for multi-tenant deployments


Defilan:
## Motivation

LLMKube's `InferenceService` currently has no concept of "this team owns this much GPU." A cluster shared across multiple teams (R&D, app teams, factory ops, etc.) has no built-in way to:

- Cap how many GPUs team A can use at any time
- Reject deployments that would exceed a team's quota
- Bind cost attribution (already in InferCost) to a quota structure that maps to chargeback business units

Operators today have to roll this themselves with `ResourceQuota` (which works for `nvidia.com/gpu` count but says nothing LLMKube-specific) plus custom RBAC. We can do meaningfully better.

This issue scopes a v0.1 multi-tenancy capability: a `GPUQuota` CRD enforced via admission webhook, plus an opinionated RBAC bundle in the Helm chart.

### A new `GPUQuota` CRD (cluster or namespace-scoped)

```go
type GPUQuotaSpec struct {
    // Selector picks which namespaces this quota applies to. Mutually exclusive
    // with NamespaceRef (which restricts to a single namespace).
    Selector *metav1.LabelSelector `json:"selector,omitempty"`
    NamespaceRef string `json:"namespaceRef,omitempty"`

    // Hard caps on GPU count and total VRAM bytes summed across all
    // InferenceServices in the matched namespace(s).
    GPUCount int32 `json:"gpuCount"`
    VRAMBytes int64 `json:"vramBytes,omitempty"`

    // Optional priority floor: only InferenceServices at or above this
    // priority can consume from this quota. Maps to existing priority
    // class enum (critical/high/normal/low/batch).
    MinPriority string `json:"minPriority,omitempty"`

    // Optional CostBudget reference: if set, the InferCost TokenBudget
    // matching this name in the same namespace is treated as the spend
    // ceiling for this quota. Hooks the cost half (already shipping in
    // InferCost) into the resource quota.
    CostBudgetRef string `json:"costBudgetRef,omitempty"`
}

type GPUQuotaStatus struct {
    UsedGPUCount int32   `json:"usedGPUCount"`
    UsedVRAMBytes int64  `json:"usedVRAMBytes"`
    AdmissionDenials int64 `json:"admissionDenials"`
    LastDenial *metav1.Time `json:"lastDenial,omitempty"`
}
```

### Admission webhook

A `validating` webhook at `inferenceservices.inference.llmkube.dev` admission:
[...]

PR #1117, opened 2026-07-13
feat(controller): add GPUQuota status reconciler

### LINKED ISSUE #1093: [FEATURE] GPUQuota status reconciler (#416 story 2)
## Feature Description

`GPUQuota` status reconciler that maintains live `Status.UsedGPUCount` and `Status.UsedVRAMBytes` by summing the InferenceServices in the quota's scope. Story 2 of #416.

## Problem Statement

> As an operator, I want each `GPUQuota` to show current utilization so I can see how close a team is to its cap before deployments start getting rejected.

## Proposed Solution

Add `internal/controller/gpuquota_controller.go`:
- Watch `GPUQuota` plus `InferenceService` (map InferenceService events back to the matching GPUQuotas).
- Resolve scope: `NamespaceRef` (single namespace) or `Selector` (label-matched namespaces). Reject specs that set both (belt-and-suspenders; primary validation is in the webhook story).
- Aggregate GPU count and VRAM across in-scope InferenceServices; write `Status.UsedGPUCount` / `Status.UsedVRAMBytes` within one reconcile.
- Owner/finalizer handling so GPUQuota deletion cleans up cleanly.

## Scope

In: status aggregation reconciler + wiring in `cmd/`/manager setup. Out: admission decisions (that is #416-C/#416-D). This story never rejects anything; it only reports usage.

## Acceptance Criteria (provable in-workspace)

- [ ] Envtest: create a `GPUQuota` and several `InferenceService`s in scope; assert `UsedGPUCount`/`UsedVRAMBytes` aggregate correctly and update when an InferenceService is added/removed.
- [ ] Envtest: deleting the `GPUQuota` reconciles cleanly (no leaked finalizer).
- [ ] `make test` and `make lint` pass; cross-arch lint clean.

## Dependencies

- **Depends on #1092** (needs the `GPUQuota` type). Do not start until #1092 is merged to `main`.

## Priority

- [x] High

Defilan:
## What

Adds the `GPUQuota` status reconciler (`internal/controller/gpuquota_controller.go`), registers it in `cmd/main.go`, regenerates the RBAC role, and ships fake-client unit tests. Story 2 of the multi-tenancy epic (#416); builds on the CRD types from #1092.

The reconciler resolves each `GPUQuota`'s scope (a single `namespaceRef` or a `selector` matched against namespace labels), lists the InferenceServices in scope, and writes `Status.UsedGPUCount`. It is status-only: it never rejects anything and owns no external resources (admission is #1094/#1095, already merged). It watches both `GPUQuota` and `InferenceService`, so a change to any InferenceService re-aggregates every quota whose scope covers it.

## Why

Fixes #1093

This is the live-usage half of GPUQuota: it keeps `Status.UsedGPUCount` current so operators can see how close a team is to its cap, and so the admission webhook (#1095) has an accurate usage figure to decide against.

## How

`GPUQuotaReconciler` mirrors `ModelRouterReconciler` (primary watch + `EnqueueRequestsFromMapFunc` secondary watch on InferenceService). Per InferenceService, GPU usage is `spec.resources.gpu * spec.replicas` (nil resources = 0, nil replicas = 1). The `selector` path uses `metav1.LabelSelectorAsSelector`, so both `matchLabels` and `matchExpressions` are honored.

**Documented design decision: `UsedVRAMBytes` is deferred and stays `0`.** `InferenceService` exposes no VRAM field, so there is nothing to aggregate for the VRAM half of the quota. Rather than guess, `UsedVRAMBytes` stays `0` with an explicit comment in the reconciler. The admission webhook already treats `vramBytes: 0` as "no cap", so the VRAM half of a quota is simply inert until we either add a VRAM field to `InferenceService` or derive it from the Model's GGUF size (follow-up).

## Testing / verification

Verified locally in an isolated worktree at the branch commit:

- `go build ./...`, `go vet ./internal/controller/... ./cmd/...` clean
- `go test ./internal/controller/` (the GPUQuota fake-client unit tests: scope resolution, GPU aggregation, the InferenceService->GPUQuota map func, and `matchExpressions` selectors) passes
- `make manifests generate chart-crds` then `git status` stable: the RBAC `role.yaml` is regenerated and in sync (CRD-Sync / RBAC-Sync pass)
- `gofmt` clean; `golangci-lint --new-from-rev` 0 new issues

Tests are fake-client unit tests rather than envtest; happy to add an apiserver-level envtest as a follow-up if you want that coverage.

## Notes for the reviewer

- `spec.selector` is implemented as a **namespace** selector (matched against namespace labels), per the epic. #1092's `Selector` doc comment describes it as a pod selector; that comment should be updated to match. Flagging so the contract is consistent.
- `gpuQuotaCoversNamespace` uses `context.Background()` for its one Get; threading the caller's ctx would be marginally better (minor).

## AI assistance disclosure

Assisted-by: Foreman, the project's local agentic coding harness (coder model qwopus3.6-35B on-prem). The controller and tests were generated by Foreman; a review pass I directed then found two real bugs the gate could not catch (a `ReconcileDuration` metric that measured ~0s because the observation ran at the top of `Reconcile`, and a selector that dropped `matchExpressions`), which were fed back to the harness as a revision cycle that fixed both and added `matchExpressions` test cases. I reviewed the result, ran the verification above, and stripped em dashes from comments. I own this change and the review conversation; per `AGENTS.md` the commit carries no attribution trailer and the DCO sign-off is mine.

## Checklist

- [x] Tests added/updated
- [x] Tests pass locally (`go test ./internal/controller/`)
- [x] Lint clean for this change (`golangci-lint --new-from-rev` 0 new issues; `gofmt` clean)
- [x] Commit is signed off (`git commit -s`) per DCO
- [x] AI assistance is disclosed above
- [ ] Documentation updated (n/a: internal controller; user-facing docs land with the Helm toggle in #1096)

### DIFF
--- a/charts/llmkube/templates/clusterrole.yaml
+++ b/charts/llmkube/templates/clusterrole.yaml
@@ -30,6 +30,7 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - namespaces
   - nodes
   - pods
   - secrets
@@ -91,6 +92,7 @@ rules:
 - apiGroups:
   - inference.llmkube.dev
   resources:
+  - gpuquotas
   - inferenceservices
   - modelrouters
   - models
@@ -113,6 +115,7 @@ rules:
 - apiGroups:
   - inference.llmkube.dev
   resources:
+  - gpuquotas/status
   - inferenceservices/status
   - modelrouters/status
   - models/status
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -361,6 +361,16 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "ModelRouterGateway")
 		os.Exit(1)
 	}
+	// GPUQuota reconciles the status-only aggregation of GPU usage from
+	// InferenceServices in the quota's scope. It never rejects anything or
+	// owns external resources.
+	if err := (&controller.GPUQuotaReconciler{
+		Client: mgr.GetClient(),
+		Scheme: mgr.GetScheme(),
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "GPUQuota")
+		os.Exit(1)
+	}
 
 	// Register the ModelRouter validating webhook ONLY when the serving cert is
 	// actually present in the configured cert dir. The webhook server crashes the
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -20,6 +20,7 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - namespaces
   - nodes
   - pods
   - secrets
@@ -163,6 +164,27 @@ rules:
 - apiGroups:
   - inference.llmkube.dev
   resources:
+  - gpuquotas
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - inference.llmkube.dev
+  resources:
+  - gpuquotas/status
+  - inferenceservices/status
+  - modelrouters/status
+  - models/status
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
+  - inference.llmkube.dev
+  resources:
   - inferenceservices
   - modelrouters
   - models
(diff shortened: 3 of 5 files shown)


joryirving (approved):
Approve - code is clean, follows existing patterns, design decisions are well-documented. 

Minor nits: 
TestSetupWithManager doesn't actually test the method, context.Background() should be threaded from the caller's ctx, and the CRD Selector comment needs updating to reflect namespace-selector semantics (flagged in PR description). 

None of these are blockers. 
Would be nice to see a multi-service aggregation test in a follow-up.

PR #1132, opened 2026-07-15
docs: document strict-taint model caches

joryirving:
## What

Document strict-`NoSchedule` model-cache workflows for GPU nodes, including pre-provisioned per-InferenceService PVCs via `spec.modelCache.claimName`, dynamic provisioner limitations, and cache-mode selection.

## Why

Dynamic MicroK8s hostpath helper pods do not inherit the GPU toleration, while shared RWO caches can bind to the wrong node on heterogeneous clusters. LLMKube already supports the strict-taint workaround through user-managed cache claims, but the workflow was not documented clearly.

Refs #850

## How

Made the public Model-cache guide live, added the strict-taint claim workflow and troubleshooting guidance, and aligned the repository guide and Helm comments. This deliberately does not weaken `NoSchedule` or attempt to mutate third-party provisioner helper pods.

Verification passed: `make fmt`, `make vet`, `make lint`, `make test`, and `make lint-all`.

AI assistance: OpenAI coding agent investigated the issue, drafted the documentation and verification plan, and reviewed the branch diff; the required checks were run locally before submission.

## Checklist

- [ ] Tests added/updated (documentation-only change)
- [x] AI assistance is disclosed above
- [x] Documentation updated

### DIFF
--- a/charts/llmkube/values.yaml
+++ b/charts/llmkube/values.yaml
@@ -285,15 +285,18 @@ modelCache:
   #     schedules to. Use this as the opt-in escape hatch for multi-node clusters
   #     that do NOT have an RWX storage class: the cache co-locates with the
   #     workload instead of pinning to one node (#728). The tradeoff is no
-  #     cross-isvc dedup (each InferenceService downloads its own copy) and
-  #     `cache list` is not per-isvc aware yet.
+  #     cross-isvc dedup (each InferenceService downloads its own copy).
+  #     `llmkube cache list` discovers operator-managed per-service cache PVCs;
+  #     user-managed `spec.modelCache.claimName` PVCs are outside that discovery
+  #     contract.
   mode: shared
   # Storage size for each model cache PVC
   size: 100Gi
   # Storage class (leave empty for the cluster default). For shared on a
   # multi-node cluster use an RWX class; for perService the default class should
   # use volumeBindingMode WaitForFirstConsumer so the PVC binds on the serving
-  # node.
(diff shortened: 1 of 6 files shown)


Defilan:
Docs look accurate. I checked the load-bearing claims against the code: the `spec.modelCache.claimName` path routes through the same prep + downloader init containers in the serving pod (no separate staging Job, so it inherits the GPU tolerations), `llmkube cache list` does discover operator-managed per-service caches via the `app.kubernetes.io/component=model-cache` label (so removing the stale "not per-isvc aware yet" text is correct), and the WaitForFirstConsumer / helper-pod-can't-tolerate-the-taint narrative matches the controller. Nice to finally have the guide live.

One thing before merge: change `Fixes #850` to `Refs #850` (or file a follow-up and link it). This is only the docs half of the split I called out on the issue — the operator-side fix is still open work: the download path copying the node's GPU taint into a matching toleration + nodeSelector, and surfacing a clear error when the cache PVC can't provision instead of a silent Pending hang. See https://github.com/defilantech/LLMKube/issues/850#issuecomment-4841062749. `Fixes #850` will auto-close the tracker for that on merge, so let's keep it open.

Minor, not for this PR: the "marked Degraded" wording (missing `claimName` PVC) actually surfaces as phase `Failed` at runtime. That same wording is already in the CRD godoc, so it's a small codebase-wide cleanup, not something you introduced here.

joryirving:
Updated!

Defilan (approved):

Issue #859, opened 2026-06-27
[FEATURE] Foreman verify gate: run envtest for changed controller packages (close the false-GO gap)


Defilan:
## Feature Description

The Foreman fast coder gate (`RunCoderGate` in `pkg/foreman/agent/coder_gate.go`, the #749 in-workspace verify tier that decides a single coder's GO) should ensure the envtest-backed controller packages a change touches actually get tested somewhere in the coder loop. Today it excludes them and nothing in the single-coder flow runs them, so a controller change that breaks the controller suite still earns a GO.

## Problem Statement

`RunCoderGate` runs `fmt`/`vet`/`lint`/`build` plus a fast unit-test tier (`changedTestPackages`). That tier deliberately **excludes** packages matching `envtestPackagePrefixes` (e.g. `internal/controller/`, `internal/foreman/controller/`) because their tests need `KUBEBUILDER_ASSETS` (etcd/kube-apiserver) the coder workspace lacks (coder_gate.go:122-194). The code comment is explicit: "CI runs them."

The nuance that makes this a real hole: the clean-room gate Job (`pkg/foreman/agent/tools/run_gate_job.go`) *does* run `make test` (envtest) via its `DefaultGateChecks` in a `golang:1.26` Job, but that is a separate tier used by a deterministic gate Agent in a coder->gate->reviewer pipeline. The **single autonomous coder's GO decision** (the `VerifyTerminal` loop in `loop.go`, backed by `RunCoderGate`) never invokes it. So for a controller-only change run by a single coder Agent, the controller suite runs in neither tier, and the gate returns GO blind.

Demonstrated end to end on #855 (PR #858): across six gate-GO single-coder runs, every one shipped controller-test breakage that only real CI caught (panicking `buildEmptyDirStorageConfig` tests from a blind `initContainers[0]->[1]` shift, stale `HaveLen(1)` cached-deployment assertions, a `corev1.Capability` vs `string` mismatch). `go test ./internal/controller/...` (464 specs) fails immediately under envtest; the coder cannot run it and the fast gate skips it.

## Proposed Solution

When `RunCoderGate` (or its caller in `executor_native.go`) sees changed packages that match `envtestPackagePrefixes`, run those packages under envtest in the clean-room gate Job rather than skipping them:

1. Detect the changed envtest packages (the inverse of the current `changedTestPackages` exclude filter).
[...]

PR #1135, opened 2026-07-15
feat(foreman): in-executor envtest gate feedback loop (#768)

### LINKED ISSUE #768: [FEATURE] Foreman coder gate: cluster-backed clean-room Job tier for envtest/e2e
## Feature Description

Add a heavyweight, cluster-backed verification tier to the Foreman coder gate: a clean-room Kubernetes Job that runs envtest + e2e (kind) on a node that *has* a cluster, executed after the fast in-workspace gate passes, with failures fed back to the coder loop the same way the fast gate's failures are.

## Problem Statement

> As a maintainer running Foreman coders, I want the gate to catch real-cluster integration regressions before a PR is opened, so coder solutions don't pass the gate and then fail CI on envtest/e2e.

The current coder gate runs entirely in the coder's workspace: gofmt, go vet, go build, golangci-lint, and (since #762/#763) a fast unit-test tier. It has **no Kubernetes cluster**, so it structurally cannot run envtest (`KUBEBUILDER_ASSETS`) or e2e (kind). Any coder change whose behavior only manifests against a real cluster — CRD reconciliation, operator logic, CLI-against-cluster — can pass the full in-workspace gate (including unit tests, which mock the cluster and give false confidence) and only fail in CI after the PR is open.

This is **systematic, not a one-off**. It has now bitten the #731 cache-list work twice and the earlier envtest trap once:

- #731: the coder produced a correct `cache list` per-InferenceService feature (verified working against a live 31-model cluster), but it regressed a pre-existing e2e: `cache list` inspects every labeled cache PVC, and a Pending (unbound) per-service PVC — created by an InferenceService whose fake e2e model never produces a running serving pod — sent the inspector into a `waitForPodRunning` block until the suite's global timeout. The unit tests passed (mocked client); only the kind e2e caught it.
- The earlier envtest trap: the coder ran envtest in its workspace where `KUBEBUILDER_ASSETS` is unset and hung.

No amount of better unit testing closes this: mocked tests cannot model real PVC layout, real pod/volume lifecycle, or real reconciliation timing. It is integration territory by nature.

## Proposed Solution

A post-fast-gate tier that runs the integration suite on a cluster-backed Job:

- After the in-workspace fast gate (gofmt/vet/build/lint/unit) passes a coder's GO, dispatch a clean-room Job on a node that has a cluster (Shadowstack; ties to #659 spot GPU NodePool for gate Jobs).
- The Job checks out the coder's branch and runs `make test` (envtest) and/or `make test-e2e` (kind), bounded by a timeout.
- On failure, feed the output back to the coder loop as gate feedback (same mechanism as the fast gate), so the coder fixes and resubmits; on pass, the GO stands.
- Keep the fast in-workspace gate as the first, cheap line of defense; the cluster Job is the heavyweight second tier (only run when the fast gate is green).

## Alternatives Considered

- **Run e2e/envtest in the coder workspace** — rejected: the workspace has no cluster, and provisioning one per coder is heavy and slow; the envtest trap shows the coder should not run these directly.
- **Rely on better unit tests** — insufficient: mocked tests gave false confidence on exactly the #731 regression.
- **Accept CI as the only integration gate** — status quo; it pushes the failure past the GO and into review, which is what this issue exists to fix.

## Additional Context

- Related: #659 (spot GPU NodePool for Foreman gate Jobs), #762/#763 (fast unit-test tier), #731 (the motivating regression).
- The fast gate's design (deterministic checks + feedback loop, #749) is the pattern to extend: same feedback mechanism, heavier tier, cluster-side.

## Priority

## Willingness to Contribute

- [x] Yes, I can submit a PR
- [x] Yes, I can help test

Defilan:
## What

Adds an in-executor **envtest feedback loop** to the Foreman coder gate. When a coder's post-push envtest gate fails, the executor now re-runs the coder against the same workspace with the gate output injected, then re-commits, force-with-lease pushes, and re-gates, bounded by a new `Agent.spec.maxEnvtestIterations` (default 1). Exhausting the bound falls back to the existing behavior (downgrade to INCOMPLETE, outcome `ENVTEST-GATE-FAILED`).

## Why

Fixes #768

The post-push envtest gate (#859/#864) was terminal: it captured the failing `make test` output but fed it nowhere, so a coder got exactly one blind shot at any envtest-backed reconcile change (it cannot run envtest in its workspace). Landing such a change required a human to re-dispatch the coder by hand with the gate output pasted in. This is the envtest half of #768's "auto-wire + feedback" item, which the later triage had marked delivered but was not: the gate ran, but its failure never fed back.

## How

An outer retry loop in `NativeAgentLoopExecutor.runLLMPath` wraps the existing commit -> push -> envtest-gate sequence. Attempt 0 is byte-identical to prior behavior; only a gate failure with budget remaining triggers a re-run. Coverage is universal (every coder issue-fix task, any Workload shape), which is why the loop lives in the executor rather than the Workload controller. Design decisions:

- **New knob** `Agent.spec.maxEnvtestIterations *int32`, three-valued (nil -> 1; explicit 0 -> pre-#768 fail-on-first; N -> up to N retries), mirroring `Workload.spec.maxReviewIterations`.
- **Feedback injection** via `envtestFeedbackPrompt` + `retryCfg`, a sibling of the reviewer-NO-GO `reviewFeedbackPrompt`: the retry rebuilds the user prompt as the original issue context plus the gate output, so it iterates against its own workspace rather than blind.
- **Retries force-with-lease** on the task's own branch (`ReplaceOnReject: attempt > 0 || AllowOverwrite`); a coder NO-GO / MaxTurns on a retry surfaces that terminal and pushes nothing new.
- Advisories and the work-class policy run once, after the loop settles on a GO.
- Extracted `mapLoopError` (behavior-preserving) and split `commitPushAttempt` / `retryCoderTerminalResult` to keep `runLLMPath` under the gocyclo ceiling.

Full design in `docs/proposals/768-envtest-feedback-loop.md`. Follow-up (e2e/kind tier) tracked separately per the issue.

Behavior note for reviewers: the default of 1 means every envtest-touched coder GO that fails the gate now gets one automatic retry before INCOMPLETE, where it previously downgraded immediately. Bounded by the cap and each loop's own MaxTurns; `explicit 0` restores the old terminal behavior.

## Checklist

- [x] Tests added/updated (three end-to-end retry-loop tests + resolver/prompt unit tests)
- [x] Affected package suite passes locally (`go test ./pkg/foreman/agent/` green); full `make test` (envtest) runs in CI
- [x] `golangci-lint` passes locally (incl. `GOOS=linux`)
- [x] Documentation updated (design proposal added; CRD field documented)

## AI assistance disclosure

Assisted-by: Claude Code (Claude Opus 4.8), running locally under my direction. I drove a design-then-plan-then-execute flow: the assistant explored the existing gate code, we settled the architecture together (an in-executor retry loop reusing the proven #946 re-dispatch semantics, chosen over two alternatives), and it wrote the committed design proposal and an implementation plan. Execution ran task-by-task with a fresh implementer per task and an independent code-review pass after each, plus a whole-branch review at the end; I reviewed the diffs, confirmed attempt-0 behavior preservation, and adjudicated the review findings. A human (me) is accountable for this change per CONTRIBUTING.md.

### DIFF
--- a/api/foreman/v1alpha1/agent_types.go
+++ b/api/foreman/v1alpha1/agent_types.go
@@ -220,6 +220,18 @@ type AgentSpec struct {
 	// +optional
 	MaxTurns int32 `json:"maxTurns,omitempty"`
 
+	// MaxEnvtestIterations bounds how many times the executor re-runs a
+	// coder after the post-push envtest gate fails, feeding the gate
+	// output back so the coder fixes the failure and the branch is
+	// re-gated (#768). Three-valued: nil defaults to 1; an explicit 0
+	// restores the pre-#768 behavior (a failing gate downgrades to
+	// INCOMPLETE on the first failure); N allows up to N retries before
+	// the downgrade. Applies only to coder issue-fix tasks whose change
+	// touches an envtest-backed package and whose gate runner is wired.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	MaxEnvtestIterations *int32 `json:"maxEnvtestIterations,omitempty"`
+
 	// MaxRetries bounds how many times the loop retries a single turn on
 	// recoverable errors (notably llama.cpp #22072 truncated tool_call
 	// argument JSON). Bounded exponential backoff with jitter.
--- a/api/foreman/v1alpha1/zz_generated.deepcopy.go
+++ b/api/foreman/v1alpha1/zz_generated.deepcopy.go
@@ -276,6 +276,11 @@ func (in *AgentSpec) DeepCopyInto(out *AgentSpec) {
 		*out = new(string)
 		**out = **in
 	}
+	if in.MaxEnvtestIterations != nil {
+		in, out := &in.MaxEnvtestIterations, &out.MaxEnvtestIterations
+		*out = new(int32)
+		**out = **in
+	}
 	if in.StuckLoopDetection != nil {
 		in, out := &in.StuckLoopDetection, &out.StuckLoopDetection
 		*out = new(StuckLoopDetectionSpec)
--- a/charts/foreman/templates/crds/agents.yaml
+++ b/charts/foreman/templates/crds/agents.yaml
@@ -220,6 +220,19 @@ spec:
                     type: string
                 type: object
                 x-kubernetes-map-type: atomic
+              maxEnvtestIterations:
+                description: |-
+                  MaxEnvtestIterations bounds how many times the executor re-runs a
+                  coder after the post-push envtest gate fails, feeding the gate
+                  output back so the coder fixes the failure and the branch is
+                  re-gated (#768). Three-valued: nil defaults to 1; an explicit 0
+                  restores the pre-#768 behavior (a failing gate downgrades to
+                  INCOMPLETE on the first failure); N allows up to N retries before
+                  the downgrade. Applies only to coder issue-fix tasks whose change
+                  touches an envtest-backed package and whose gate runner is wired.
+                format: int32
+                minimum: 0
+                type: integer
               maxRetries:
                 default: 3
                 description: |-
--- a/config/crd/bases/foreman.llmkube.dev_agents.yaml
+++ b/config/crd/bases/foreman.llmkube.dev_agents.yaml
@@ -219,6 +219,19 @@ spec:
                     type: string
                 type: object
                 x-kubernetes-map-type: atomic
+              maxEnvtestIterations:
+                description: |-
+                  MaxEnvtestIterations bounds how many times the executor re-runs a
+                  coder after the post-push envtest gate fails, feeding the gate
+                  output back so the coder fixes the failure and the branch is
+                  re-gated (#768). Three-valued: nil defaults to 1; an explicit 0
+                  restores the pre-#768 behavior (a failing gate downgrades to
+                  INCOMPLETE on the first failure); N allows up to N retries before
+                  the downgrade. Applies only to coder issue-fix tasks whose change
+                  touches an envtest-backed package and whose gate runner is wired.
+                format: int32
+                minimum: 0
+                type: integer
               maxRetries:
                 default: 3
                 description: |-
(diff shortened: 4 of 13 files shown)


joryirving (approved):
Approve. Clean design, well-executed. The in-executor retry loop is the right architectural choice — universal coverage without controller-level plumbing.

**Verified:**

- Build, go vet, gofmt clean. Agent package tests pass (6.4s). The two `tools/run_reconcile_test.go` failures are pre-existing on main (environmental git pathspec issue, not from this PR).
- Attempt-0 behavior preservation confirmed: `ReplaceOnReject` evaluates to `allowOverwrite` on attempt 0, gate classification is identical. The only behavior change is the default gate-failure response (retry once vs immediate INCOMPLETE), which is the intended change. Explicit `maxEnvtestIterations: 0` restores old behavior.
- Tri-state gate verdict (`envtestGateOK` / `envtestGateFailed` / `envtestGateUnverified`) correctly classifies outcomes. Caller correctly distinguishes unverified-by-attempt: attempt 0 settles (pre-#768), retry downgrades (prevents false-GO).
- `gateJobName` fix verified: truncates task portion only, preserves `-<unix-ms>` suffix. Test confirms two submissions of the same long task name with different timestamps produce distinct ≤63-char names.
- `mapLoopError` extraction is byte-identical to the original inline switch.
- Retry loop is bounded by `maxEnvtestIters`; cannot spin infinitely. Each attempt persists transcript. Non-GO retry surfaces via `retryCoderTerminalResult` without pushing.
- Force-push safety: retries use force-with-lease on the task-scoped branch (same as #946 reviewer re-dispatch).
- CRD types, deepcopy, config/crd/bases, and chart CRDs all consistent.
- DCO, conventional commits, AI disclosure, PR template all present.

**Non-blocking:**

1. Grounding rail + no-functional-change advisory moved from pre-push to post-loop. On a push-failure terminal they no longer run — intentional and arguably better, but a minor delta from the "byte-identical attempt 0" framing.
2. Cap exhaustion tested with N=0 only. A multi-retry test (N≥2, gate fails then converges, or fails N+1 times) would round out coverage. Low risk — the bound is a simple `>=` check.

PR #1166, opened 2026-07-20
feat(foreman): make the verify gate optional in the issue-batch decomposition

### LINKED ISSUE #1151: Make the verify gate optional in the issue-batch decomposition (code → review)
## Summary

Requesting the ability to run the issue-batch decomposition **without** the deterministic `verify` gate — i.e. `code → review` instead of `code → verify → review` — by making `verifierAgentRef` optional in the issue-batch shortcut rather than required.

## Motivation

Operators running polyglot fleets pay a real tax for the deterministic gate: every repo needs a `GateProfile` (language, image, exact build/test commands), and repos with no profile fall through to `generic`, which has **no image** and errors the gate outright. Maintaining per-repo gate profiles (or a fat polyglot gate image) is significant ongoing work.

But the deterministic gate is arguably redundant with two checks that already exist:

1. **The coder already self-verifies.** The `issue-fix` task's own contract is "edit the repo, *run the verification*, commit, push" — an agent that reads the repo (AGENTS.md, Makefile, package.json) and runs its tests, polyglot by nature.
2. **The repo's real CI runs on the opened PR** — the repo's actual tests, defined once in the repo, maintained for the repo's own sake, polyglot for free. (In our setup dispatch already ingests those `check_run` results and drives a fix loop.)

For teams that want the deterministic signal to be the repo's real CI rather than a re-declared command subset, the pre-PR gate is pure overhead. Letting it be optional means: coder self-verifies → PR opens → repo CI is the deterministic verdict → review judges. No per-repo gate config anywhere.

## Current behavior

Verify is mandatory in the issue-batch path:

- `chooseSteps` (`internal/foreman/controller/workload_controller.go`): the issue-batch case requires `CoderAgentRef != nil && VerifierAgentRef != nil`; without the verifier it errors (`"coderAgentRef and verifierAgentRef are required for the issue-batch shortcut"`).
- `synthesizeIssueBatch` unconditionally emits `verify-<N>` (deref `*VerifierAgentRef`), and `review-<N>-<i>` `dependsOn` it.
- The `#548` gate-cascade, `#946` review-iteration, and coder-escalation logic all key off the verify step.

An explicit hand-authored `spec.pipeline` can omit verify today, but that opts out of the whole issue-batch orchestration (`#946`/`#548`/escalation), so it isn't an equivalent path.

## Proposed behavior

When `Issues` is set with `CoderAgentRef` but **no** `VerifierAgentRef`:

- `chooseSteps` accepts it (no error).
- `synthesizeIssueBatch` skips the `verify-<N>` step; each `review-<N>-<i>` `dependsOn` `code-<N>` directly.
- The cascade / review-iteration / escalation logic treats "no verify step" gracefully (a review round begins once the coder step is terminal, rather than gating on GATE-PASS).

Verify stays the default when `verifierAgentRef` is set — this is purely additive and opt-in.

## Appetite check

Is this a direction you're open to? Happy to implement it (I've been through this code for #1042/#1044/#1120) — the touch points look like `chooseSteps`, `synthesizeIssueBatch`, and the verify-assuming branches in the cascade/iteration/escalation paths. Wanted to gauge whether an optional-gate model fits foreman's philosophy before opening a PR.

joryirving:
## What

Make `verifierAgentRef` optional in the issue-batch decomposition: when unset, each issue's pipeline is `code → review` (no deterministic verify gate), across the base round, the #946 review-fix iterations, and the coder-escalation tier.

## Why

Fixes #1151

The shortcut required `VerifierAgentRef` and always emitted a verify step, so every repo needed a `GateProfile` with a real image + command set — the sole remaining per-repo configuration cost in our loop. For operators whose deterministic signal is the repo's own CI on the opened PR (with the coder already self-verifying inside the issue-fix loop, per its task contract), the pre-PR gate is redundant: unmapped repos fall to the imageless `generic` profile and GATE-ERROR, and maintaining per-repo gate commands duplicates what the repo's CI already declares.

Opt-in and fully backward compatible: with `verifierAgentRef` set, the shape is byte-identical to today (`code → verify → review` everywhere), pinned by regression tests.

## How

- `chooseSteps`: the issue-batch case requires only `CoderAgentRef`; the no-coder error message updated accordingly.
- `synthesizeIssueBatch`: emits the verify step only when a verifier is set; reviews `dependsOn` the verify step, or the code step when gateless.
- `reviewIterationSteps` (#946) and `coderEscalationSteps`: previously **disabled themselves entirely** on a nil verifier; now they emit gateless iterations/escalations (`code-rK → review-rK`, `code-esc → review-esc`) so a gateless Workload keeps fix iterations and coder escalation.
- `#548` cascade: no change needed — it keys off declared dependencies, so reviews still cascade-fail when their code dependency fails; there's simply no gate verdict in between. Rollup likewise classifies by phase+verdict generically.
- API doc + CRD description regenerated (`make manifests && make chart-crds`).

## Checklist

- [x] Tests added/updated — `workload_gateless_test.go`: gateless batch shape + review wiring, verifier-set regression pin, `chooseSteps` acceptance, gateless coder-escalation, gateless review-iteration
- [x] `make test` passes locally (controller suite 82.7% coverage)
- [x] Conventional commits
- [x] Signed off (DCO)
- [x] AI assistance disclosed — authored with Claude Code
- [x] Documentation updated — API field doc + generated CRD description

### DIFF
--- a/api/foreman/v1alpha1/workload_types.go
+++ b/api/foreman/v1alpha1/workload_types.go
@@ -43,11 +43,12 @@ const (
 //     one AgenticTask owner-ref'd to this Workload. The reconciler
 //     rewrites step-local DependsOn names to absolute task names. Used
 //     for full control over a hand-authored pipeline.
-//  2. Issue-batch shortcut (Issues non-empty + CoderAgentRef +
-//     VerifierAgentRef set): emit one code+verify pair per issue. The
-//     most common shape for v0.1. When ReviewerAgentRefs is also set
-//     (v0.2), each issue additionally fans out one parallel review
-//     task per listed reviewer Agent, depending on the verify task.
+//  2. Issue-batch shortcut (Issues non-empty + CoderAgentRef set): emit
+//     one code step per issue, plus a verify step when VerifierAgentRef
+//     is set. The most common shape for v0.1. When ReviewerAgentRefs is
+//     also set (v0.2), each issue additionally fans out one parallel
+//     review task per listed reviewer Agent, depending on the verify
+//     task (or directly on the code task when no verifier is set).
 //     When EscalationReviewerAgentRefs is also set, a second reviewer
 //     tier is emitted per issue only after a base reviewer returns
 //     NO-GO.
@@ -105,9 +106,13 @@ type WorkloadSpec struct {
 	// +optional
 	CoderAgentRef *corev1.LocalObjectReference `json:"coderAgentRef,omitempty"`
 
-	// VerifierAgentRef is required when Issues is set: names the same-
-	// namespace Agent the verify-<N> steps reference. Ignored in explicit
-	// Pipeline mode.
+	// VerifierAgentRef names the same-namespace Agent the verify-<N>
+	// steps reference. Optional in the issue-batch shortcut: when unset,
+	// no deterministic verify gate is emitted and each issue's pipeline
+	// is code -> review (reviews depend directly on the code step);
+	// verification then rests on the coder's own test run and the
+	// repository's CI on the opened PR. Ignored in explicit Pipeline
+	// mode.
 	// +optional
 	VerifierAgentRef *corev1.LocalObjectReference `json:"verifierAgentRef,omitempty"`
 
--- a/config/crd/bases/foreman.llmkube.dev_workloads.yaml
+++ b/config/crd/bases/foreman.llmkube.dev_workloads.yaml
@@ -771,9 +771,13 @@ spec:
                 type: object
               verifierAgentRef:
                 description: |-
-                  VerifierAgentRef is required when Issues is set: names the same-
-                  namespace Agent the verify-<N> steps reference. Ignored in explicit
-                  Pipeline mode.
+                  VerifierAgentRef names the same-namespace Agent the verify-<N>
+                  steps reference. Optional in the issue-batch shortcut: when unset,
+                  no deterministic verify gate is emitted and each issue's pipeline
+                  is code -> review (reviews depend directly on the code step);
+                  verification then rests on the coder's own test run and the
+                  repository's CI on the opened PR. Ignored in explicit Pipeline
+                  mode.
                 properties:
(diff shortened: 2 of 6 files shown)


Defilan (approved):
Thanks Jory, this is a really clean one. I went through the diff against the claims rather than taking the description on faith, checked out the branch, and ran build, vet, and the unit suite (all green; only the envtest suite skips without KUBEBUILDER_ASSETS, which is unrelated).

The backward-compat story holds up across all three decomposition paths (base `synthesizeIssueBatch`, the #946 review iterations, and coder escalation): with a verifier set, the emitted step graph and `DependsOn` edges come out byte-identical to today, and the regression test pins the base shape. The gateless wiring is correct everywhere too. Reviews depend on the code step across the base round, the `code-rK -> review-rK` iterations, and the `code-esc -> review-esc` escalation, and the new tests assert the actual step names and edges rather than just that it runs.

I also confirmed the #548 cascade claim: `cascadeFailIfDepFailed` and `allDepsSucceeded` key off `DependsOn` plus `SucceededOnTarget()`, so a coder NO-GO cascade-fails the gateless review exactly like a GATE-FAIL verify would, with no rollup misclassification. All three `VerifierAgentRef` dereferences are nil-guarded now, and the other issue-batch paths (reviewer escalation, review advisories) never assume a verify step, so they compose fine with gateless.

Two non-blocking nits before you squash-merge:

1. A few doc comments still describe the old required-verifier behavior and now read as stale, despite the "docs updated" checkbox: the `Issues` and `ReviewerAgentRefs` field docs in `workload_types.go`, and the `WorkloadReconciler` mode-2 doc in `workload_controller.go`. One-line touch-ups keep the package docs coherent.

2. Consider a single combined test exercising gateless plus escalation plus iteration together. Each is covered on its own, but nothing locks in that they compose now that dropping the verifier guard activates both tiers for gateless workloads.

Neither blocks. Approving, nice contribution.

PR #1182, opened 2026-07-21
feat(controller): bound InferenceService pod lifetime via maxPodLifetimeSeconds

### LINKED ISSUE #1173: feat: expose a maximum inference pod lifetime
## Feature description

Expose an optional maximum lifetime for pods managed by an `InferenceService`, ideally by passing Kubernetes `activeDeadlineSeconds` through to the generated Pod template (or an equivalent controller-managed restart-after field).

## Problem

Some long-running inference backends need periodic process recycling even while their health endpoint remains healthy. One concrete case is llama.cpp on AMD Vulkan: pinned GTT memory can accumulate or remain resident outside normal cgroup accounting and is released only when the process exits. The closest upstream report is ggml-org/llama.cpp#5380.

LLMKube currently owns the Deployment and correctly preserves externally added `kubectl.kubernetes.io/restartedAt` annotations, but an operator wanting a bounded pod lifetime must still deploy separate CronJob/RBAC machinery to patch the generated Deployment. A liveness probe is not appropriate because inference remains functional until node memory pressure becomes severe.

## Proposed behavior

- Add an optional positive duration/seconds field to `InferenceService` for maximum pod lifetime.
- Copy it to the generated Pod template as `activeDeadlineSeconds`, or provide equivalent controller-owned semantics.
- Keep the current behavior when omitted.
- Ensure expiry is compatible with the existing Deployment strategy and rollout policy.

## Alternatives considered

- Scheduled `kubectl rollout restart`: requires external automation and Deployment patch RBAC.
- CronJob patching `spec.template.metadata.annotations`: works, but adds four Kubernetes resources per workload group.
- Liveness probe: models process health, not process age or driver memory lifetime.

Would exposing `activeDeadlineSeconds` on `InferenceService` fit the project API, or is there another preferred lifecycle mechanism?

Defilan:
## What

Add `InferenceService.spec.maxPodLifetimeSeconds` (`*int64`, minimum 1). When set, the reconciler copies it to the pod template's `PodSpec.ActiveDeadlineSeconds`, so Kubernetes terminates the pod after the given duration and the Deployment replaces it.

## Why

Some workloads need periodic process recycling to release driver memory that the runtime does not free on its own, for example llama.cpp on AMD Vulkan with pinned GTT memory. Today pods run indefinitely until manually restarted or the Deployment is updated.

Fixes #1173

## How

- New optional field `maxPodLifetimeSeconds` on `InferenceServiceSpec` with a `Minimum=1` validation marker.
- `constructDeployment` sets `ActiveDeadlineSeconds: isvc.Spec.MaxPodLifetimeSeconds` on the pod template (nil when unset, so default behavior is unchanged).
- Regenerated deepcopy + CRDs (`make generate`, `make manifests`, `make chart-crds`).

## Checklist

- [x] Tests added/updated (`internal/controller/inferenceservice_deployment_test.go`: covers unset -> nil and set -> ActiveDeadlineSeconds equals the value)
- [x] `make test` passes (verified by the Foreman gate; build/vet/lint also run locally)
- [x] `make lint` passes locally (golangci-lint GOOS=linux, 0 issues)
- [ ] Documentation updated (field is self-documenting via its CRD description; no separate docs page)

## AI assistance

AI-assisted (band 3 per CONTRIBUTING.md): the implementation was produced by an agentic coding tool (Foreman, dense Qwopus3.6-27B with Perplexity + Context7 MCP grounding) and passed the automated verify gate; a human reviewed the diff and tests, re-ran build/vet/lint, and owns this submission.


Defilan:
Hey @Tanguille here is my fix for your request. If this looks good I can get it merged. My testing showed this worked but would love another set of eyes.

Issue #1225, opened 2026-07-22
[BUG] maxPodLifetimeSeconds makes InferenceService reconciliation fail


Tanguille:
## Bug Description

`maxPodLifetimeSeconds` is accepted by the CRD, but using it makes the LLMKube controller fail when it updates the generated Deployment. I found this while trying to recycle llama.cpp Vulkan embedders on AMD iGPUs, where pinned GTT memory is not visible to cgroups.

## Steps to Reproduce

1. Run LLMKube 0.9.10.
2. Apply the InferenceService below.
3. Let the controller reconcile it.
4. Check the controller logs and generated Deployment.

## Expected Behavior

The service should reconcile successfully and the pod should be recycled after the configured lifetime.

## Actual Behavior

The Deployment update is rejected by the API server:

```text
Deployment.apps "qwen3-embedding" is invalid:
spec.template.spec.activeDeadlineSeconds:
Forbidden: activeDeadlineSeconds in ReplicaSet is not Supported
```

The existing pod keeps running, but the controller repeatedly fails to apply the desired Deployment.

## Environment

**LLMKube Version:**
```bash
llmkube version
# Output: 0.9.10
```

**Kubernetes Version:**
```bash
kubectl version
# Output: not recorded; the API server returns the validation error above
```

**Cluster Type:**
- [ ] GKE
- [ ] EKS
- [ ] AKS
- [ ] minikube
- [ ] kind
- [ ] K3s
- [x] Other: Talos Linux

**GPU (if applicable):**
- [ ] NVIDIA T4
- [ ] NVIDIA L4
- [ ] NVIDIA V100
- [ ] NVIDIA A100
- [ ] Metal (Apple Silicon)
- [ ] None (CPU only)
- [x] Other: AMD Radeon 660M iGPU (Vulkan)

## Logs

**Controller Logs:**
```bash
kubectl logs -n llmkube-system deployment/llmkube-controller-manager --tail=50
# Relevant output:
# Deployment.apps "qwen3-embedding" is invalid: spec.template.spec.activeDeadlineSeconds:
# Forbidden: activeDeadlineSeconds in ReplicaSet is not Supported
```

**Inference Pod Logs:**

Not applicable. The existing inference pod remains healthy; the failure is on the Deployment update.

## YAML Manifests

**Model:**
```yaml
apiVersion: inference.llmkube.dev/v1alpha1
kind: Model
metadata:
  name: qwen3-embedding
spec:
[...]

Issue #1224, opened 2026-07-22
[BUG] llmkube_gpu_queue_depth reports a queue position, not a depth, and never resets


Tanguille:
## Bug Description

`llmkube_gpu_queue_depth` is documented as "Number of InferenceServices waiting for GPU resources" but it's set to one specific InferenceService's 1-based FIFO position. It's a package-level singleton Gauge, so with N services queued it's last-writer-wins across 1..N and moves around as each object reconciles. It never reliably equals N, and it never goes back to 0 when the queue drains.

Filing this as an issue rather than a PR because the fix changes an exported metric's signature, so it needs a call on deprecate-vs-change first.

## Expected Behavior

The metric matches its name and help text: the number of InferenceServices currently waiting for GPU resources, going back to 0 when none are.

## Actual Behavior

An arbitrary queue position from whichever object reconciled last, which then sticks forever once the queue empties.

## Where it goes wrong

`internal/metrics/metrics.go:71` declares a plain `Gauge`, not a `GaugeVec`. One writer, at `internal/controller/status_builder.go:205`:

```go
if phase == PhaseWaitingForGPU {
	queuePos, err := r.calculateQueuePosition(ctx, isvc)
	...
	isvc.Status.QueuePosition = queuePos
	llmkubemetrics.GPUQueueDepth.Set(float64(queuePos))
} else {
	isvc.Status.QueuePosition = 0   // status gets reset, the metric doesn't
}
```

`calculateQueuePosition` (`internal/controller/scheduling.go:151`) lists all InferenceServices, filters to `PhaseWaitingForGPU`, sorts FIFO on creation timestamp and returns `pos + 1` for this object at L187-191. That's a position.

So three queued services write 1, 2 and 3 in whatever order they reconcile and the gauge keeps whichever was last. And because the else branch doesn't touch the metric, once everything is scheduled it sits at its last non-zero value forever, which would latch a `> 0` alert permanently.

Reads 0 here only because nothing has ever actually queued for a GPU on this cluster, so this is a latent one rather than something that's bitten us.

## Suggested Fix

Your call, since it's exported:

- Change it in place to a GaugeVec keyed by namespace, set to `len(waitingServices)` (already available inside `calculateQueuePosition`) on every reconcile including the zero case. Correct, but breaks existing dashboards and alerts.
- Or leave the singleton alone, add a correctly-named one alongside it, mark the old one deprecated in the help text.

[...]

Issue #1226, opened 2026-07-22
[BUG] Shipped Grafana dashboards query metric names that nothing emits


Tanguille:
## Bug Description

Went to adopt the shipped dashboards instead of maintaining our own and found that most of the panels can't return data — they query metric names that nothing in the tree emits. Same failure mode as #786 and #1223: it installs clean and renders empty, which looks identical to "no traffic yet".

Three separate dashboards affected, and the pinned metric contract in the #700 slices repeats two of the bad names.

### `docs/grafana/llmkube-inference.json` — blank on every install

All 4 targets query `llmkube_gpu_queue_wait_duration_seconds_bucket`. That metric is declared at `internal/metrics/metrics.go:78` and registered at `:232`, but there is no `Observe()` call anywhere in the tree:

```
$ grep -rn "GPUQueueWaitDuration" --include="*.go" . | grep -v _test
internal/metrics/metrics.go:78:  GPUQueueWaitDuration = prometheus.NewHistogramVec(
internal/metrics/metrics.go:232:         GPUQueueWaitDuration,
```

Declaration and registration, no producer. So this dashboard is empty 100% of the time, on any install, regardless of hardware or runtime.

### `config/grafana/llmkube-inference-dashboard.json` — llama.cpp half is dead

It queries `llamacpp_requests_processing`, `llamacpp_requests_deferred`, `llamacpp_prompt_tokens_total`, `llamacpp_tokens_second`, `llamacpp_kv_cache_usage_ratio`.

llama.cpp uses **colons**, not underscores. Raw exposition from `ghcr.io/ggml-org/llama.cpp:server-vulkan`, which is what LLMKube runs — this is the complete list, 11 metrics:

```
llamacpp:n_busy_slots_per_decode
llamacpp:n_decode_total
llamacpp:n_tokens_max
llamacpp:predicted_tokens_seconds
llamacpp:prompt_seconds_total
llamacpp:prompt_tokens_seconds
llamacpp:prompt_tokens_total
llamacpp:requests_deferred
llamacpp:requests_processing
llamacpp:tokens_predicted_seconds_total
llamacpp:tokens_predicted_total
```

So three of the five are just misspelled (`_` for `:`). The other two don't exist under any spelling on current builds:

- `llamacpp_tokens_second` — closest real ones are `llamacpp:predicted_tokens_seconds` and `llamacpp:prompt_tokens_seconds`
[...]

PR #1229, opened 2026-07-22
feat(metrics): publish operator state metrics from observed state

### LINKED ISSUE #1219: [BUG] llmkube_model_status is never emitted for Models that are already Ready
## Bug Description

`llmkube_model_status` never appears on the controller metrics endpoint for a Model in its normal resting state. The gauge is only set on a phase transition, and every steady-state reconcile path returns before it reaches a `Set()`. So a Model that has settled on Ready exports nothing, and after a controller restart the series never comes back.

Same shape as #786, except here the emitting code exists, it's just unreachable on the common path.

## Steps to Reproduce

1. Create a Model with a HuggingFace source (`source: hf://org/repo@sha`)
2. Wait for `kubectl get model` to report Ready
3. `kubectl rollout restart deploy/llmkube-controller-manager`
4. Scrape the controller metrics endpoint, grep `llmkube_model_status`

## Expected Behavior

`llmkube_model_status{model=...,namespace=...,phase="Ready"} 1` for every Model the controller manages, for as long as it exists.

## Actual Behavior

Nothing. Four Models here, all Ready, controller up 24 days. Every `llmkube_*` name on the controller scrape job:

```
llmkube_gpu_queue_depth
llmkube_inferenceservice_info
llmkube_inferenceservice_phase
llmkube_inferenceservice_ready_duration_seconds_{bucket,count,sum}
llmkube_reconcile_duration_seconds_{bucket,count,sum}
llmkube_reconcile_total
```

No `llmkube_model_status`. Meanwhile `llmkube_reconcile_total{controller="model"}` reads 22, so the model controller is clearly running. It just never gets to a `Set()`.

(`llmkube_model_download_duration_seconds` is missing too, but that one's expected — nothing here uses a controller-downloaded source.)

## Where it goes wrong

`internal/controller/model_controller.go` sets `ModelStatus` in nine places, all on transitions. The three resting paths each return before any of them:

`reconcileRuntimeResolvedSource` (L590) handles `hf://` and remote HTTP. It's reached from `reconcileBySourceType` (L551), which runs near the top of `Reconcile` at L178. Both branches of the `if model.Status.Phase == PhaseReady` guard at L600-611 return with only `ReconcileTotal.Inc()`, so the `ModelStatus.Set(1)` further down at L689 is dead for anything already Ready.

`handleReadyCachedModel` (L422) returns `skip=true` at L456-458, again with only `ReconcileTotal.Inc()`. `reconcilePVCSource` (L463) is the same.

Since the `hf://` dispatch happens before any of the cache-key logic, a HuggingFace-sourced Model can't set the gauge again after its first transition to Ready. Which is what we're seeing.

## Suggested Fix

Drop the nine transition writes and set it once from observed state, so it doesn't matter which return path a reconcile takes:

```go
func (r *ModelReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
	reconcileStart := time.Now()
	model := &inferencev1alpha1.Model{}
	defer func() {
		llmkubemetrics.ReconcileDuration.WithLabelValues("model").Observe(time.Since(reconcileStart).Seconds())
		if model.Name != "" {
			llmkubemetrics.ModelStatus.DeletePartialMatch(prometheus.Labels{
				"model": model.Name, "namespace": model.Namespace,
			})
			llmkubemetrics.ModelStatus.
				WithLabelValues(model.Name, model.Namespace, model.Status.Phase).Set(1)
		}
	}()
```

Same pattern as the `ReconcileDuration` defer that's already there. Fewer lines than what it replaces. The `DeletePartialMatch` is so old phases don't pile up.

One thing to decide while you're in there: the gauge's Help says the value encodes the phase (`0=Unknown, 1=Downloading, ... 5=Failed`) but no call site does that. They all set 1 and vary the `phase` label, same as `llmkube_inferenceservice_phase`. The help text should probably just match what the code does.

Happy to send the PR if you want it.

## Environment

```
Kubernetes: v1.36.2
```

---

Found in the same pass as #1221, #1222, #1223 and #1224. #1221 is the closest relative — happy to do both as a single PR if that's fewer review surfaces for you.

### LINKED ISSUE #1221: [BUG] InferenceService and Model metric series are never removed when the CR is deleted
## Bug Description

Nothing cleans up metrics when a CR goes away. Series for a deleted InferenceService stay in the registry until the controller process restarts, so counting ready services over-reports, and an alert keyed on the phase gauge keeps firing for something that doesn't exist any more.

## Steps to Reproduce

1. Create an InferenceService, wait for Ready
2. `kubectl delete inferenceservice <name>`
3. Scrape the controller metrics endpoint

## Expected Behavior

`llmkube_inferenceservice_phase` and `llmkube_inferenceservice_info` for that name/namespace stop being exported.

## Actual Behavior

Both are still there with `phase="Ready"` at 1, until the controller restarts.

## Where it goes wrong

The only metric cleanup anywhere in `internal/` is `status_builder.go:177`, a `DeletePartialMatch` on `InferenceServiceInfo` that covers an in-place accelerator/runtime change. Deletion isn't handled at all. Both reconcilers just fall out of the not-found branch:

- `internal/controller/inferenceservice_controller.go:168` → `return ctrl.Result{}, nil`
- `internal/controller/model_controller.go:153` → same

Related, same metric: `updateStatus` sets the new phase to 1 and the previous phase to 0 (`status_builder.go:159-162`) but never removes older ones, so they accumulate per service. Here:

```
count by (inferenceservice) (llmkube_inferenceservice_phase)
  qwen3-embedding   1
  qwen35-2b         1
  qwen36-27b        2     <- phase="Creating" 0, phase="Ready" 1
  vmcp-embedding    1
```

Doesn't break one-hot queries, but it grows and never shrinks.

## Suggested Fix

Delete the object's series in the not-found branch:

```go
if apierrors.IsNotFound(err) {
	labels := prometheus.Labels{"inferenceservice": req.Name, "namespace": req.Namespace}
	llmkubemetrics.InferenceServicePhase.DeletePartialMatch(labels)
	llmkubemetrics.InferenceServiceInfo.DeletePartialMatch(labels)
	return ctrl.Result{}, nil
}
```

plus the `model`/`namespace` version for `ModelStatus` in the model controller. Together with the delete-before-set suggested in #1219 that also sorts out the stale phases.

This is best-effort obviously — if the controller is down when the delete happens it'll miss it — but it fixes the steady-state leak and a restart clears anything left over.

## Environment

```
Kubernetes: v1.36.2
```

---

Found in the same pass as #1219. Happy to do both as a single PR if that's fewer review surfaces for you.

### LINKED ISSUE #1222: [FEATURE] Export InferenceService replica counts as metrics
## Feature Description

`InferenceServiceStatus` already tracks `ReadyReplicas`, `DesiredReplicas` and `Replicas` (`api/v1alpha1/inferenceservice_types.go:1084-1090`) but none of it is exported. A gauge would make replica health queryable.

## Problem Statement

I want to alert on "InferenceService says Ready but has 0 ready replicas", and right now I can't. `llmkube_inferenceservice_phase{phase="Ready"}` tells me the phase, nothing tells me how many pods are actually serving.

Rebuilding it from `kube_deployment_status_replicas_available` means guessing at the operator's Deployment naming, which is internal detail I'd rather not depend on.

## Proposed Solution

One GaugeVec in `internal/metrics/metrics.go`:

```go
InferenceServiceReplicas = prometheus.NewGaugeVec(
	prometheus.GaugeOpts{
		Name: "llmkube_inferenceservice_replicas",
		Help: "InferenceService replica counts. Use the state label: ready, desired.",
	},
	[]string{"inferenceservice", "namespace", "state"},
)
```

Set in `status_builder.go` where `Status.ReadyReplicas` and `Status.DesiredReplicas` are already being assigned around L149-152. No new plumbing or API calls.

## Alternatives Considered

Joining on `kube_deployment_status_replicas_available` (depends on internal naming), or kube-state-metrics `CustomResourceState` against the CRD. The latter works and it's what we're deploying locally as a stopgap, but it seems wrong as the only route to a basic operator metric.

## Additional Context

- Related issues: #1219, #1221 — same files, found in the same pass over the metrics emit paths.

## Priority

- [x] Medium - Nice to have

## Willingness to Contribute

- [x] Yes, I can submit a PR

Would fold into the fix for #1219 and #1221, it's the same files.

Tanguille:
Closes #1219, closes #1221, closes #1222.

All three are the same defect shape — the operator's state gauges are written at transitions rather than derived from observed state — so they're one change rather than three PRs.

## Observed on a live install

Scraped from a 25-day-old cluster running the released `0.9.10`, with four Models and four InferenceServices, all `Ready`:

```
llmkube_model_status                                                  <- absent entirely
llmkube_inferenceservice_replicas                                     <- does not exist
llmkube_inferenceservice_phase{inferenceservice="qwen36-27b",phase="Creating"} 0
llmkube_inferenceservice_phase{inferenceservice="qwen36-27b",phase="Ready"}    1
```

Every Model is `Ready` and `llmkube_model_status` has zero series, which is #1219. The stranded `phase="Creating"` at `0` next to `phase="Ready"` at `1` is the zeroing behaviour — and it is also the honest version of the claim: the count was never wrong, the zero-valued series just never shed.

I then built this branch, swapped the controller image on that same cluster, and rescraped:

```
llmkube_model_status{model="qwen35-2b",namespace="ai",phase="Ready"}       1
llmkube_model_status{model="qwen36-27b-awq",namespace="ai",phase="Ready"}  1
llmkube_model_status{model="qwen3-embedding",namespace="ai",phase="Ready"} 1
llmkube_model_status{model="vmcp-embedding",namespace="ai",phase="Ready"}  1
llmkube_inferenceservice_replicas{inferenceservice="qwen35-2b",state="desired"} 1
llmkube_inferenceservice_replicas{inferenceservice="qwen35-2b",state="ready"}   1
llmkube_inferenceservice_replicas{inferenceservice="qwen36-27b",state="desired"} 1
llmkube_inferenceservice_replicas{inferenceservice="qwen36-27b",state="ready"}   1
```

All four Models report, against zero before. Replicas exist for both InferenceServices that complete a reconcile (the other two are the #1225 case described under *Not in scope*).

For the deletion half I applied a throwaway Model, confirmed `llmkube_model_status{model="metrics-probe-model"} 1`, deleted the CR, and rescraped: zero series for it, the other four untouched. Before this change that series would have persisted until the process restarted.

The four inference Deployments were never restarted during any of this — the operator is not in the data path — and the cluster was reverted to `0.9.10` afterwards.

## What changed

**`llmkube_model_status` is republished on every reconcile** (#1219), from `model.Status.Phase` in a defer. It was set at nine call sites, all on transitions, and every steady-state path returns before reaching one: `hf://` and remote-HTTP take the already-Ready exit in `reconcileRuntimeResolvedSource`, cached files take `handleReadyCachedModel`, PVC sources take `reconcilePVCSource`.

**Series are dropped when the CR is gone** (#1221), in the `IsNotFound` branch of both reconcilers. Best-effort: if the controller is down when the delete lands, client-go's relist on restart won't replay the missed event, so that one series survives until the process restarts. Bounded and self-healing, rather than the current permanent leak.

**Exactly one phase series per object.** Zeroing only the immediately-previous phase left every earlier phase in place at 0, so the vec grew with phase churn and never shed. Both gauges now delete before setting.

**New `llmkube_inferenceservice_replicas{state="ready"|"desired"}`** (#1222), set where `Status.ReadyReplicas`/`DesiredReplicas` are already assigned.

The publish/delete pairs live in `internal/metrics` rather than the controllers, so the label sets sit with the declarations they belong to and a controller never spells out a label map. That's a structural call — happy to move them back if you'd rather `internal/metrics` stay declarations-only.

I also replaced the hand-copied collector list in `init()` with an exported `AllCollectors` that `init()` and the tests both range over. The new gauge was silently missing from `TestMetricsRegistered`'s parallel list, which is the bug that list invites.

## Two behaviour changes worth your attention

**`llmkube_model_status` no longer emits `phase="Cached"`.** `PhaseCached` is declared but never assigned to `Model.Status.Phase`, so that label never matched what `kubectl get model` reports; the gauge now mirrors the CR. The metric was absent in steady state anyway, so nothing could have depended on it — I didn't mark this breaking, but say the word and I'll carry the cache-hit distinction as its own label.

**A phase change now deletes the old series instead of zeroing it.** `TestInferenceServicePhase` asserted the zeroing convention, so it's updated here. Worth being precise about the old behaviour, because the issue text overstated it: zeroing kept exactly one series at `1`, so `count(... == 1)` was never wrong. What accumulated was zero-valued series that never shed — a cardinality leak, not a wrong count.

I also corrected the `ModelStatus` help text, which documented a numeric phase encoding (`0=Unknown, 1=Downloading, …`) that no call site implemented.

## On cost

The delete is unconditional rather than gated on an observed transition. I tried the gated version — cheaper, since `DeletePartialMatch` takes an exclusive lock and scans the vec — and backed it out: it isn't self-healing. Seven paths set `Status.Phase = PhaseFailed` and then call `updateStatus`, which can fail; the defer has already published `Failed`, the next reconcile reads the older stored phase back, the gate doesn't fire, and the orphan is permanent. That's the bug this PR exists to fix, so the scan stays. It's a map walk over one entry per object.

## Verification

- `make lint` — 0 issues
- `make test` — green (`internal/controller` 86.0%, `internal/metrics` 100%)
- `internal/controller` no longer imports prometheus outside tests: the label and `state=` vocabulary lives only next to the declarations
- **All six new specs confirmed to fail without the production change.** The five Model/deletion specs: the three controller files checked out from `upstream/main`, tests left in place — 5 failures, re-confirmed after the rebase. The sixth (InferenceService early return) can't use that method, since it calls a helper `main` doesn't have; removing only the deferred publish from `Reconcile` fails it and nothing else — 601 of 602 pass.

The transition specs drive a real two-phase publish rather than patching `Status.Phase` directly — the Model goes Failed (host-path allowlist rejection, no network needed) then Ready, because the gauge only accumulates when the controller itself writes the second phase. They also deliberately don't clear the vec between reconciles, since clearing is exactly what would mask an accumulating gauge.

## InferenceService, in full

I first wrote this up as narrow, on the reasoning that the happy path always publishes. Running the branch on a real cluster showed that is wrong. Two of four InferenceServices there are `Ready` and serving, but fail `reconcileDeployment` on every pass with `activeDeadlineSeconds in ReplicaSet is not Supported` (#1225) — so they return early every time and export no phase, replicas or info series at all. Not a transient: for as long as an earlier reconcile step errors, a healthy service is invisible.

You said either way was fine, so it's in rather than deferred. `Reconcile` hoists the object above the defer and publishes phase, replicas and info from stored status, the same shape as the Model half. `updateStatusWithSchedulingInfo` no longer publishes at all — it is unreachable on exactly the passes that need reporting — so each object has one writer. `model` fell out unused there and in `reconcileService`; both params are removed.

The spec drives the live failure rather than approximating it: a seeded Deployment puts `reconcileDeployment` on its in-place `Update` path, an interceptor rejects that `Update` the way the API server does, and the assertions are that a `Ready` service with 2 replicas still reports phase, replica counts and the Model's `cuda` accelerator afterwards. #1222 is now fully closed rather than half — replicas no longer depend on the reconcile completing.

## Not in scope

`llmkube_gpu_queue_depth` (#1224) is untouched — signature change, and I'd rather have your call on deprecate-vs-change first.

GPUQuota has the same deletion bug; filed as #1230 rather than widened into this PR, since #1221 is scoped to InferenceService and Model. Happy to follow up once this lands — it reuses `DeleteInferenceServiceSeries`'s shape.

### DIFF
--- a/internal/controller/inferenceservice_controller.go
+++ b/internal/controller/inferenceservice_controller.go
@@ -157,15 +157,24 @@ func initContainerSecurityContext(isvc *inferencev1alpha1.InferenceService) *cor
 
 func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	reconcileStart := time.Now()
+	inferenceService := &inferencev1alpha1.InferenceService{}
+	// Assigned by getModelForInferenceService below; declared here so the
+	// deferred publish can label the info series with the Model's accelerator.
+	var model *inferencev1alpha1.Model
 	defer func() {
 		llmkubemetrics.ReconcileDuration.WithLabelValues("inferenceservice").Observe(time.Since(reconcileStart).Seconds())
+		// Republish from observed state: a pass that returns before
+		// updateStatusWithSchedulingInfo publishes nothing at the call site.
+		// A live cluster hit this via #1225 — two Ready, serving services
+		// failed reconcileDeployment every pass and exported no series at all.
+		publishInferenceServiceState(inferenceService, model)
 	}()
 
 	log := logf.FromContext(ctx)
 
-	inferenceService := &inferencev1alpha1.InferenceService{}
 	if err := r.Get(ctx, req.NamespacedName, inferenceService); err != nil {
 		if apierrors.IsNotFound(err) {
+			llmkubemetrics.DeleteInferenceServiceSeries(req.Name, req.Namespace)
 			return ctrl.Result{}, nil
 		}
 		log.Error(err, "Failed to get InferenceService")
@@ -188,7 +197,7 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	// from before the allowlist was introduced (or tightened).
 	if valErr := validateLocalSourceAllowed(model.Spec.Source, r.AllowedHostPathRoots); valErr != nil {
 		log.Error(valErr, "rejected local model source by host-path allowlist", "model", model.Name, "source", model.Spec.Source)
-		blockResult, updateErr := r.updateStatusWithSchedulingInfo(ctx, inferenceService, model, PhaseFailed, modelReady, 0, 0, "",
+		blockResult, updateErr := r.updateStatusWithSchedulingInfo(ctx, inferenceService, PhaseFailed, modelReady, 0, 0, "",
 			valErr.Error(), nil)
 		return blockResult, updateErr
 	}
@@ -201,7 +210,7 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if effectiveModelCacheKey(model) != "" && r.ModelCachePath != "" {
 		if err := r.ensureModelCachePVC(ctx, inferenceService); err != nil {
 			log.Error(err, "Failed to ensure model cache PVC exists", "namespace", inferenceService.Namespace)
-			return r.updateStatusWithSchedulingInfo(ctx, inferenceService, model, PhaseFailed, modelReady, 0, desiredReplicas, "",
+			return r.updateStatusWithSchedulingInfo(ctx, inferenceService, PhaseFailed, modelReady, 0, desiredReplicas, "",
 				fmt.Sprintf("Failed to ensure model cache PVC: %v", err), nil)
 		}
 	}
@@ -228,7 +237,7 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return ctrl.Result{}, err
 	}
 
-	service, result, err := r.reconcileService(ctx, inferenceService, model, modelReady, desiredReplicas, isMetal)
+	service, result, err := r.reconcileService(ctx, inferenceService, modelReady, desiredReplicas, isMetal)
 	if err != nil || result != nil {
 		if result != nil {
 			return *result, err
@@ -243,7 +252,7 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	endpoint := r.constructEndpoint(inferenceService, service)
 	phase, schedulingInfo := r.determinePhase(ctx, inferenceService, readyReplicas, desiredReplicas, isMetal, deployment, metalSnap)
 
-	finalResult, statusErr := r.updateStatusWithSchedulingInfo(ctx, inferenceService, model, phase, modelReady, readyReplicas, desiredReplicas, endpoint, "", schedulingInfo)
+	finalResult, statusErr := r.updateStatusWithSchedulingInfo(ctx, inferenceService, phase, modelReady, readyReplicas, desiredReplicas, endpoint, "", schedulingInfo)
 	if statusErr != nil {
 		return finalResult, statusErr
 	}
@@ -279,7 +288,7 @@ func (r *InferenceServiceReconciler) getModelForInferenceService(ctx context.Con
 	if err := r.Get(ctx, modelName, model); err != nil {
 		if apierrors.IsNotFound(err) {
 			log.Info("Referenced Model not found", "model", isvc.Spec.ModelRef)
-			result, updateErr := r.updateStatusWithSchedulingInfo(ctx, isvc, nil, PhaseFailed, false, 0, 0, "", "Model not found", nil)
+			result, updateErr := r.updateStatusWithSchedulingInfo(ctx, isvc, PhaseFailed, false, 0, 0, "", "Model not found", nil)
 			return nil, false, &result, updateErr
 		}
 		log.Error(err, "Failed to get Model")
@@ -289,7 +298,7 @@ func (r *InferenceServiceReconciler) getModelForInferenceService(ctx context.Con
 	modelReady := model.Status.Phase == PhaseReady
 	if !modelReady {
 		log.Info("Model not ready yet", "model", model.Name, "phase", model.Status.Phase)
-		result, updateErr := r.updateStatusWithSchedulingInfo(ctx, isvc, model, "Pending", false, 0, 0, "", "Waiting for Model to be Ready", nil)
+		result, updateErr := r.updateStatusWithSchedulingInfo(ctx, isvc, "Pending", false, 0, 0, "", "Waiting for Model to be Ready", nil)
 		return nil, false, &result, updateErr
 	}
 
@@ -325,7 +334,7 @@ func (r *InferenceServiceReconciler) reconcileDeployment(ctx context.Context, is
 	// message; admission-time rejection is the follow-up (#1196 story 5).
 	if _, err := resolveGPUSharing(isvc, model, r.GPUSharingSharedPool); err != nil {
 		log.Info("Rejecting InferenceService with invalid gpuSharing spec", "reason", err.Error())
-		result, updateErr := r.updateStatusWithSchedulingInfo(ctx, isvc, model, PhaseFailed, modelReady, 0, desiredReplicas, "", fmt.Sprintf("Invalid gpuSharing: %v", err), nil)
+		result, updateErr := r.updateStatusWithSchedulingInfo(ctx, isvc, PhaseFailed, modelReady, 0, desiredReplicas, "", fmt.Sprintf("Invalid gpuSharing: %v", err), nil)
 		return nil, 0, nil, &result, updateErr
 	}
 
@@ -347,7 +356,7 @@ func (r *InferenceServiceReconciler) reconcileDeployment(ctx context.Context, is
 		deployment.Annotations[AnnotationDesiredTemplateHash] = tmplHash
 		if err := r.Create(ctx, deployment); err != nil {
 			log.Error(err, "Failed to create Deployment")
-			result, updateErr := r.updateStatusWithSchedulingInfo(ctx, isvc, model, PhaseFailed, modelReady, 0, desiredReplicas, "", "Failed to create Deployment", nil)
+			result, updateErr := r.updateStatusWithSchedulingInfo(ctx, isvc, PhaseFailed, modelReady, 0, desiredReplicas, "", "Failed to create Deployment", nil)
 			return nil, 0, nil, &result, updateErr
 		}
 		return deployment, 0, nil, nil, nil
@@ -486,7 +495,7 @@ func (r *InferenceServiceReconciler) reconcileDeployment(ctx context.Context, is
 	return deployment, existingDeployment.Status.ReadyReplicas, nil, nil, nil
 }
 
-func (r *InferenceServiceReconciler) reconcileService(ctx context.Context, isvc *inferencev1alpha1.InferenceService, model *inferencev1alpha1.Model, modelReady bool, desiredReplicas int32, isMetal bool) (*corev1.Service, *ctrl.Result, error) {
+func (r *InferenceServiceReconciler) reconcileService(ctx context.Context, isvc *inferencev1alpha1.InferenceService, modelReady bool, desiredReplicas int32, isMetal bool) (*corev1.Service, *ctrl.Result, error) {
 	log := logf.FromContext(ctx)
 
 	if isMetal {
@@ -513,7 +522,7 @@ func (r *InferenceServiceReconciler) reconcileService(ctx context.Context, isvc 
 		log.Info("Creating new Service", "name", service.Name)
 		if err := r.Create(ctx, service); err != nil {
 			log.Error(err, "Failed to create Service")
-			result, updateErr := r.updateStatusWithSchedulingInfo(ctx, isvc, model, PhaseFailed, modelReady, 0, desiredReplicas, "", "Failed to create Service", nil)
+			result, updateErr := r.updateStatusWithSchedulingInfo(ctx, isvc, PhaseFailed, modelReady, 0, desiredReplicas, "", "Failed to create Service", nil)
 			return nil, &result, updateErr
 		}
 	} else if err != nil {
--- a/internal/controller/inferenceservice_service_test.go
+++ b/internal/controller/inferenceservice_service_test.go
@@ -213,7 +213,7 @@ var _ = Describe("reconcileService Metal path", func() {
 			Spec:       inferencev1alpha1.InferenceServiceSpec{ModelRef: "some-model"},
 		}
 
-		svc, result, err := reconciler.reconcileService(context.Background(), isvc, nil, true, 1, true)
+		svc, result, err := reconciler.reconcileService(context.Background(), isvc, true, 1, true)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(result).To(BeNil())
 		Expect(svc).NotTo(BeNil())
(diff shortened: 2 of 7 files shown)


Defilan (changes requested):
Overall: strong work, and I would approve the logic. There is one blocker (merge conflict) and one scope question before merge. This is an unusually well-reasoned PR. The defect model is correct, the live-cluster before/after is convincing, and the "all five specs fail against unmodified main" check is exactly the rigor that makes a metrics change trustworthy.

### What I verified independently

- **The `PhaseCached` removal is safe.** `PhaseCached` is never assigned to `Model.Status.Phase` anywhere in `internal/`, so the old `phase="Cached"` label never matched the CR. Mirroring `Status.Phase` is a correctness fix, not a loss.
- **No dashboard or alert regression.** The `Model Status` stat panel in `llmkube-gpu-dashboard.json` queries `llmkube_model_status` by value, but since the "0-5 encoding" in the old help text was never implemented (the gauge always emitted `1`), the panel value is unchanged. It just stops showing the stranded zero-valued series, which is an improvement. The alert on `llmkube_inferenceservice_phase{phase="Creating"} == 1` is unaffected: delete-on-transition and zero-on-transition both leave it non-firing after a transition.
- **The delete primitive is right.** `DeletePartialMatch` with the `{inferenceservice, namespace}` (or `{model, namespace}`) subset drops all phase series for an object regardless of the open-ended phase label.
- **The model defer is sound.** Hoisting `model` above the defer so it reads observed `Status.Phase`, with `PublishModelPhase` no-op'ing on the empty phase, cleanly handles the `IsNotFound` and early-Get-error paths without emitting a bogus empty-phase series.

### Blocker

- **Merge conflict, needs a rebase on `main`.** GitHub shows the branch conflicting. `model_controller.go` is a hot file and has moved under this branch. Please rebase, then re-confirm the fail-on-main specs still fail after the rebase. The resolution should not change behavior, but the specs are the proof.

### Scope question (not blocking, your call)

- **InferenceService is only half-converted.** Model now derives its phase from observed state in a `defer`, but InferenceService still publishes phase from the call site inside `updateStatusWithSchedulingInfo`. So an ISvc that returns early (the real `activeDeadlineSeconds` / #1225 case you found on the live cluster) still exports no phase, replicas, or info series at all, which is the same bug this PR fixes for Model. You are explicit about this and propose a follow-up with its own fail-on-main spec. Deferring is reasonable given #1225 needs its own repro, but it does mean this PR half-fixes #1222 (replicas exist only when the reconcile completes). I would land it as-is provided the follow-up is filed and linked. Fine either way on whether the ISvc defer comes into this PR for symmetry.

### Minor

- `PublishModelPhase` does delete-then-set on every reconcile, which is two non-atomic ops under the vec lock. A scrape landing in the sub-microsecond gap would momentarily see no series for that model. Negligible at scrape intervals and inherent to the self-healing design; a one-line comment noting the window is accepted would be enough.
- Pre-existing, not yours: the `Model Status` panel's `legendFormat` is `{{name}}`, but the label is `model`, so it renders empty. Since you are already in metrics territory, `{{model}}` would be a nice drive-by, but equally fine to leave for a dashboard PR.

### Worth stating

The `AllCollectors` refactor (init and the tests ranging the same slice) closes the class of bug that let the new gauge go missing from `TestMetricsRegistered`, and moving the label and `state=` vocabulary into `internal/metrics` so the controllers stop importing prometheus is the right structural call. Keep both.

Thanks for the thorough write-up and the live-cluster verification. Once the rebase is in and the follow-up issue for the InferenceService half is linked, this is good to go.

Tanguille:
Rebased onto 7c1161b. I also pulled the InferenceService half in, since you said either way was fine.

The conflict was only in `model_controller.go`, against #1218 moving the cache-hit block into `reconcileCachedModelFile`. Kept that extraction and deleted the `phase="Cached"` `Set(1)` that came with it, same as the other call sites this PR drops. That path sets `Status.Phase = PhaseReady`, so it reports Ready now. `range-diff` against what you reviewed is that hunk plus the two things below.

Side effect worth mentioning: prefetch writes Downloading/Ready/Failed to `Status.Phase` and returns early in a few branches, so the defer picks all of it up without a new call site. On main that path exports nothing.

Re-ran the fail-on-main check after the rebase (three controller files from `upstream/main`, tests left alone) and got the same 5 failures.

For the ISvc side, `Reconcile` hoists the object above the defer and publishes phase, replicas and info from stored status. `updateStatusWithSchedulingInfo` doesn't publish at all anymore — it can't help on the passes where publishing actually matters. `model` ended up unused there and in `reconcileService`, so I dropped the param from both.

The new spec goes at the real failure rather than faking it: seed the Deployment so `reconcileDeployment` takes its in-place `Update`, have an interceptor reject that `Update`, then assert a Ready service with 2 replicas still reports phase, replica counts and the Model's `cuda` accelerator. The fail-on-main proof runs backwards for this one, since it calls a helper main doesn't have — pull just the deferred publish out and it's the only spec that fails, 601 of 602 pass.

That closes #1222 properly rather than halfway, and there's no follow-up issue left to file. Remaining metrics debt I know of is #1230 and #1224.

Added the scrape-window note you asked for:

```go
// Delete-then-set is not atomic: a scrape landing between the two ops sees no
// series for this Model. Accepted — the window is sub-microsecond.
```

Left `legendFormat` alone. The dashboard PR for #1226 rewrites those panels anyway, so fixing it here would just be a stray hunk.

`make test` and `make lint` are green on the rebase (`internal/controller` 86.0%, `internal/metrics` 100%).

Defilan (approved):
Re-reviewed after the rebase and the InferenceService half. Both blockers from the first pass are resolved, and I verified the new code rather than just the writeup.

### Verified

- **Rebase is correct.** The only conflict was in `model_controller.go` against #1218 moving the cache-hit block into `reconcileCachedModelFile`. You kept that extraction and dropped the `phase="Cached"` `Set(1)` that came with it, consistent with the other call sites this PR removes; that path sets `Status.Phase = PhaseReady`, so it reports Ready. Nice catch that #1218's prefetch writes Downloading/Ready/Failed and returns early, so the Model defer now picks that up for free.
- **The InferenceService half is right.** `Reconcile` hoists both `inferenceService` and `var model` above the defer, and `model, modelReady, result, err := r.getModelForInferenceService(...)` sits at function-body scope, so it reassigns the hoisted `model` rather than shadowing it. The defer sees the real object, which is the subtle part that would silently break if it were a `:=` inside a block. `publishInferenceServiceState` no-ops on an empty phase and falls back to cpu/llamacpp on a nil model, matching what the old call site did. `updateStatusWithSchedulingInfo` no longer publishes, and dropping the now-unused `model` param from it and `reconcileService` is the right cleanup. No controller touches prometheus directly anymore.
- **The new spec goes at the real failure.** Seeding the Deployment so `reconcileDeployment` takes its in-place `Update`, then rejecting that `Update` with an interceptor and asserting a Ready service still reports phase, replicas, and the Model's `cuda` accelerator, is exactly the #1225 shape rather than a faked `Status.Phase`. That closes #1222 properly.
- Scrape-window comment added. Leaving `legendFormat` for the #1226 dashboard PR is the right call. CI is green and coverage holds.

No remaining issues. Thanks for the thorough turnaround and the clear rebase notes. Approving.

Issue #1240, opened 2026-07-23
[FEATURE] Declarative bindAddress on InferenceService instead of overriding --host via extraArgs


Tanguille:
## Feature Description

Right now the only way to change the address a runtime listens on is to re-declare `--host` in `spec.extraArgs` and count on the parser keeping the last occurrence. I'd like a real field for it.

To be clear up front, the `::` default is correct and I'm not asking to change it. #972/#973 got that call right, and I checked it rather than assuming: bound a socket to `::` inside a pod on an IPv4-only cluster and connected to it over IPv4 from another pod, works fine (`bindv6only=0`, and none of these servers set `IPV6_V6ONLY`). I'm running it as shipped. This is only about how you'd change it if you had to.

## Problem Statement

The flag is hardcoded in five places now: `runtime_llamacpp.go:77`, `runtime_llamacpp_router.go:98`, `runtime_vllm.go:96`, `runtime_sglang.go:92` (all `--host ::`), and `runtime_tgi.go:44` (`--hostname ::`). #973 added the first batch, then sglang (#1059) and llamacpp-router (#1152) each copied the literal and its comment forward.

Three things bother me about the extraArgs workaround:

**TGI can't use it.** `runtime_tgi.go` never reads `Spec.ExtraArgs` at all, so there's no override path there short of switching to `runtime: generic` and hand-writing the whole arg list. #973's own description flagged this ("TGI does not currently consume `extraArgs`; a small parity follow-up could add that if needed") but I couldn't find a follow-up issue for it.

**The CRD doc is wrong.** `inferenceservice_types.go:380` still says extraArgs is "Supported by the `llamacpp` and `vllm` runtimes. Ignored by others." That's wrong in both directions today, since sglang and llamacpp-router do consume it and tgi genuinely doesn't, so `kubectl explain` misleads either way. There's nothing about the `--host` override in `docs/` or the README either; grepping for ipv6/dual-stack/bindv6only comes back empty. The workaround effectively only exists in a merged PR description.

**Nothing actually guarantees extraArgs is last.** `runtime_llamacpp.go:155` appends `--metrics` after it. Harmless for `--host` specifically, but no test guards the ordering and the whole workaround rests on it. The last-wins parser behaviour is the same story — asserted in comments, not tests, across four third-party parsers. llama.cpp does honour it but logs a warning on every duplicated flag, so you'd be trading a config problem for permanent startup noise.

## Proposed Solution

```yaml
spec:
  runtime: sglang
[...]

Issue #1230, opened 2026-07-22
[BUG] GPUQuota metric series are never removed when the CR is deleted


Tanguille:
## Bug Description

`GPUQuota` has the same "no cleanup on delete" bug as #1221, in the same shape. Filing separately because #1221 is scoped to InferenceService and Model, and CONTRIBUTING asks that a change match the scope of the issue it claims to fix — so I left it out of #1229 rather than widening that PR.

## Steps to Reproduce

1. Create a GPUQuota, let it reconcile
2. `kubectl delete gpuquota <name>`
3. Scrape the controller metrics endpoint

## Expected Behavior

The four per-quota gauges stop being exported for that quota.

## Actual Behavior

`llmkube_gpuquota_used_gpu_count`, `llmkube_gpuquota_gpu_count_limit`, `llmkube_gpuquota_used_vram_bytes` and `llmkube_gpuquota_vram_bytes_limit` keep reporting the deleted quota's last known values until the controller process restarts.

## Where it goes wrong

`internal/controller/gpuquota_controller.go:74-80` — the `IsNotFound` branch returns with no cleanup:

```go
gq := &inferencev1alpha1.GPUQuota{}
if err := r.Get(ctx, req.NamespacedName, gq); err != nil {
    if errors.IsNotFound(err) {
        return ctrl.Result{}, nil
    }
    ...
```

and `:155-158` sets all four gauges unconditionally on every successful reconcile — the current-state-gauge shape that needs a matching delete on the way out.

The practical effect is on the multi-tenancy dashboard the metrics were added for (the comment cites #416): a "used GPUs by quota" sum over-counts by every deleted quota's last value.

## Suggested Fix

Same pattern as #1229. That PR adds `DeleteInferenceServiceSeries` to `internal/metrics` so the label sets live next to the declarations; a `DeleteGPUQuotaSeries(name, namespace)` alongside it, called from the `IsNotFound` branch, would cover this.

Worth noting *why* this was easy to miss, since it's the same reason it'll be missed again: nothing ties a new per-object gauge to a cleanup path. #1229 adds a third InferenceService-scoped gauge and only stays correct because the cleanup helper had to be edited by hand at the same time. If you'd rather solve that structurally than one controller at a time, the kube-state-metrics approach — deriving gauges from the controller-runtime cache in a custom Collector instead of holding mutable state — makes the whole bug class unrepresentable. That's clearly a bigger piece of work and a separate discussion; flagging it rather than proposing it.

[...]

PR #1243, opened 2026-07-23
fix(metrics): report GPU queue depth per namespace

### LINKED ISSUE #1224: [BUG] llmkube_gpu_queue_depth reports a queue position, not a depth, and never resets

Tanguille:
Closes #1224

## The defect

`llmkube_gpu_queue_depth` was a single global gauge (`prometheus.NewGauge`, no labels) set from `calculateQueuePosition`'s return value. That return value is the reconciling service's **1-based FIFO position**, not a depth, so the gauge reported whichever position reconciled last and had no namespace dimension. With three services queued it reads 1, 2 or 3 depending purely on reconcile ordering.

## The change

`evaluateGPUQueue` returns the per-namespace depth map alongside the position, and `PublishGPUQueueDepth` makes that map the whole of the metric: a namespace absent from it stops being reported, so a namespace draining to 0 reports 0 instead of holding a stale value.

Position and depth are derived in a single pass over the listing. The previous hand-rolled selection sort over the waiting set is gone: counting services created before the subject gives the same FIFO position without ordering the slice.

## Two things that go beyond the metric, declared

**1. The publish is unconditional, not gated on a queue transition.**

An earlier revision of this branch gated the publish on entering or leaving `WaitingForGPU`. That is the shape `Reconcile`'s own defer warns about:

```go
// A live cluster hit this via #1225 — two Ready, serving services
// failed reconcileDeployment every pass and exported no series at all.
```

Gating also left the `GaugeVec` exporting no series at all on an idle cluster after an operator restart, so the stat panel would read "No data" rather than 0. The publish is now unconditional, matching `publishInferenceServiceState`. Cost is one cache-backed `List` per status write; `gpuquota_controller.go:120` already does one `List` per in-scope namespace per reconcile.

**2. `Status.QueuePosition` changes, not only the metric.**

The subject's phase is taken from `isvc` rather than from the cache-backed listing, which still reports the previous phase on the reconcile that changes it, and may not list a freshly created object at all. Without this a service entering the queue reports position 0, which reaches the user through the Progressing condition as "Queued at position 0 waiting for GPU". It self-corrects on the next reconcile, so this fixes a transient user-visible status string as well as the metric.

The dashboard changes are required by the new label rather than cosmetic: `legendFormat` becomes `{{namespace}}` (a fixed string would render N identically-labelled series) and the stat panel is wrapped in `sum()` to stay a single cluster-wide number. The alert expression needs no change — `llmkube_gpu_queue_depth > 0` now fires per namespace — and its annotation gains `{{ $labels.namespace }}` to match the neighbouring `InferenceServiceStuckCreating` rule.

## Verification

`make lint` → `0 issues`. `make test` → all packages `ok`, `internal/controller` 86.2%, `internal/metrics` 100.0%.

Behavioural fail-on-main proof: main's semantics were restored **behind the new signatures**, so the build stays intact and the specs fail on values rather than on compilation. Under that build, `TestGPUQueueDepth` fails with `expected team-a to stop being reported, got 1 series`, and all 11 new controller specs fail, including:

- `head of the queue` — `{"default": 0}` vs `{"default": 2, "team-a": 1, "team-b": 0}`
- `reports how many services are queued, not the position of one` — `2` vs `3`
- `counts a service on the reconcile that puts it in the queue` — `0` vs `3`

The lifecycle spec that enters the queue was rewritten so it cannot pass by coincidence: `delta` is now created *first*, so its position (1) and the depth (3) are different numbers. Previously it was created last, making position numerically equal to depth, which main's last-writer-wins gauge satisfied by accident. A `BeforeEach` resets the gauge so no spec inherits another's value.

## Known limitation, not addressed here

Deleting an InferenceService never routes through `updateStatusWithSchedulingInfo` — `Reconcile`'s `IsNotFound` branch calls `DeleteInferenceServiceSeries` and returns. Deleting the last queued service in a namespace can therefore strand a depth series until some other transition republishes the map. This is strictly better than main, where the single gauge stuck at the last position forever with no reset at all, but it is not fully fixed. Happy to add a republish on the deletion path here if you would rather it land in one go.

## Update: tie-break fix

Fixed the requested-changes item: `evaluateGPUQueue` now breaks ties on equal `CreationTimestamp` (1s resolution) by name, so two services created in the same wall-clock second get a strict `1..N` ordering instead of colliding on the same reported position. Added `evaluateGPUQueue tie-breaking`, a dedicated `DescribeTable` with two same-second entries verifying the ordering both ways.

Also rebased onto current `main` now that #1244, #1230 and #1240 have merged. #1230 touches the same `internal/metrics/metrics.go`/`metrics_test.go` files this PR does (different metric, `GPUQuota` deletion vs `GPUQueueDepth`); the rebase applied clean with no manual conflict resolution needed.

### DIFF
--- a/internal/controller/scheduling.go
+++ b/internal/controller/scheduling.go
@@ -176,7 +176,12 @@ func (r *InferenceServiceReconciler) evaluateGPUQueue(
 		depth := depths[svc.Namespace]
 		if svc.Status.Phase == PhaseWaitingForGPU {
 			depth++
-			if svc.CreationTimestamp.Before(&isvc.CreationTimestamp) {
+			// CreationTimestamp is 1s resolution, so a Helm release or a
+			// scripted loop can create several services within the same
+			// second; break the tie by name so positions stay a strict
+			// 1..N permutation instead of colliding.
+			if svc.CreationTimestamp.Before(&isvc.CreationTimestamp) ||
+				(svc.CreationTimestamp.Equal(&isvc.CreationTimestamp) && svc.Name < isvc.Name) {
 				ahead++
 			}
 		}
--- a/internal/controller/scheduling.go
+++ b/internal/controller/scheduling.go
@@ -181,7 +181,9 @@ func (r *InferenceServiceReconciler) evaluateGPUQueue(
 			// second; break the tie by name so positions stay a strict
 			// 1..N permutation instead of colliding.
 			if svc.CreationTimestamp.Before(&isvc.CreationTimestamp) ||
-				(svc.CreationTimestamp.Equal(&isvc.CreationTimestamp) && svc.Name < isvc.Name) {
+				(svc.CreationTimestamp.Equal(&isvc.CreationTimestamp) &&
+					(svc.Namespace < isvc.Namespace ||
+						(svc.Namespace == isvc.Namespace && svc.Name < isvc.Name))) {
 				ahead++
 			}
 		}
--- a/internal/controller/scheduling.go
+++ b/internal/controller/scheduling.go
@@ -23,7 +23,6 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
@@ -148,49 +147,48 @@ func (r *InferenceServiceReconciler) getPodSchedulingInfo(ctx context.Context, i
 	return nil, nil
 }
 
-func (r *InferenceServiceReconciler) calculateQueuePosition(ctx context.Context, isvc *inferencev1alpha1.InferenceService) (int32, error) {
-	if isvc.Status.Phase != PhaseWaitingForGPU {
-		return 0, nil
-	}
-
+// evaluateGPUQueue returns isvc's 1-based position in the cluster-wide FIFO GPU
+// queue, and the number of services waiting for GPU in every namespace that
+// holds an InferenceService. Position is 0 when isvc is not waiting.
+//
+// isvc carries the authoritative phase and creation time: the cache-backed
+// listing still reports the previous phase on the reconcile that changes it,
+// and may not list a freshly created object at all.
+func (r *InferenceServiceReconciler) evaluateGPUQueue(
+	ctx context.Context,
+	isvc *inferencev1alpha1.InferenceService,
+) (int32, map[string]int32, error) {
 	allServices := &inferencev1alpha1.InferenceServiceList{}
 	if err := r.List(ctx, allServices); err != nil {
-		return 0, err
+		return 0, nil, err
 	}
 
-	type queueEntry struct {
-		name      string
-		namespace string
-		created   metav1.Time
-	}
+	// Every namespace holding an InferenceService is keyed, queued or not, so a
+	// namespace that drains to 0 reports 0 instead of losing its series.
+	depths := map[string]int32{isvc.Namespace: 0}
 
-	var waitingServices []queueEntry
-	for _, svc := range allServices.Items {
-		if svc.Status.Phase == PhaseWaitingForGPU {
-			waitingServices = append(waitingServices, queueEntry{
-				name:      svc.Name,
-				namespace: svc.Namespace,
-				created:   svc.CreationTimestamp,
-			})
+	var ahead int32
+	for i := range allServices.Items {
+		svc := &allServices.Items[i]
+		if svc.Name == isvc.Name && svc.Namespace == isvc.Namespace {
+			continue // counted once below, from isvc's own phase
 		}
-	}
-
-	// Sort by creation timestamp (FIFO)
-	for i := 0; i < len(waitingServices)-1; i++ {
-		for j := i + 1; j < len(waitingServices); j++ {
-			if waitingServices[j].created.Before(&waitingServices[i].created) {
-				waitingServices[i], waitingServices[j] = waitingServices[j], waitingServices[i]
+		depth := depths[svc.Namespace]
+		if svc.Status.Phase == PhaseWaitingForGPU {
+			depth++
+			if svc.CreationTimestamp.Before(&isvc.CreationTimestamp) {
+				ahead++
 			}
 		}
+		depths[svc.Namespace] = depth
 	}
 
-	for pos, entry := range waitingServices {
-		if entry.name == isvc.Name && entry.namespace == isvc.Namespace {
-			return int32(pos + 1), nil //nolint:gosec // G115: queue index bounded by waitingServices size
-		}
+	if isvc.Status.Phase != PhaseWaitingForGPU {
+		return 0, depths, nil
 	}
+	depths[isvc.Namespace]++
 
-	return 0, nil
+	return ahead + 1, depths, nil
 }
 
 func (r *InferenceServiceReconciler) resolvePriorityClassName(isvc *inferencev1alpha1.InferenceService) string {
--- a/charts/llmkube/templates/servicemonitor.yaml
+++ b/charts/llmkube/templates/servicemonitor.yaml
@@ -15,6 +15,9 @@ spec:
     port: {{ if .Values.metrics.secure }}https{{ else }}http{{ end }}
     scheme: {{ if .Values.metrics.secure }}https{{ else }}http{{ end }}
     interval: {{ .Values.prometheus.serviceMonitor.interval }}
+    # llmkube_* metrics label the workload's namespace themselves. Without this
+    # the scrape target's namespace wins and theirs becomes exported_namespace.
+    honorLabels: true
     {{- if .Values.metrics.secure }}
     tlsConfig:
       insecureSkipVerify: true
--- a/charts/llmkube/tests/metrics_test.yaml
+++ b/charts/llmkube/tests/metrics_test.yaml
@@ -89,6 +89,18 @@ tests:
           path: spec.endpoints[0].scheme
           value: http
 
+  # Without honorLabels the scrape target's namespace wins and the workload
+  # namespace llmkube_* metrics carry is renamed exported_namespace, which the
+  # dashboards and the GPUQueueBacklog annotation do not select on.
+  - it: should honor the metric's own namespace label
+    template: servicemonitor.yaml
+    set:
+      prometheus.serviceMonitor.enabled: true
+    asserts:
+      - equal:
+          path: spec.endpoints[0].honorLabels
+          value: true
+
   # ServiceMonitor: tlsConfig only when secure
   - it: should include tlsConfig when secure
     template: servicemonitor.yaml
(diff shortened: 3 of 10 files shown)


Tanguille:
Verified against a live cluster, which turned up a defect in the first revision of this branch and a second commit to fix it.

**The metric change alone was not enough.** The chart's ServiceMonitor sets no `honorLabels`, so Prometheus renames any label on a scraped metric that collides with a target label. `llmkube_*` metrics label the workload's namespace themselves, and the target label is the operator's own namespace, so the workload namespace was being stored as `exported_namespace`.

This is observable on the existing metric, before any change:

```
llmkube_inferenceservice_phase{
  namespace="ai",              # the operator pod's namespace, from the scrape
  exported_namespace="ai",     # the label the operator actually set
  inferenceservice="qwen36-27b", phase="Stopped"
}
```

Applied to this branch that would have made every presentation change wrong: `{{namespace}}` would render the operator's namespace for every series, `{namespace=~"$namespace"}` would filter on it, and the `GPUQueueBacklog` annotation would name it. Exactly the defect #1224 reports, in a different coat.

Proven with a throwaway InferenceService in a separate namespace, `honorLabels` off then on:

| | `namespace` | `exported_namespace` |
|---|---|---|
| before | `ai` | `llmkube-qtest-a` |
| after | `llmkube-qtest-a` | absent |

Hence the second commit. `charts/llmkube/tests/metrics_test.yaml` covers it, and the test fails when `honorLabels` is removed from the template rather than passing regardless.

**Separately, live confirmation of the metric this branch touches.** Querying 30 days of production data:

| metric | series in 30d |
|---|---|
| `llmkube_gpu_queue_depth` | 44 |
| `llmkube_gpu_queue_wait_duration_seconds` | 0 |

The queue-depth gauge currently resolves to a single series whose only namespace label is the operator's, which is the reporting gap this PR closes.

Note the chart change affects every `llmkube_*` metric that carries its own `namespace` label, not only queue depth. I read that as a fix rather than a regression, since the operator's namespace is not useful on a per-workload series, but it is a visible label change for anyone with existing queries and I would rather call it out than slip it in.

The operator image was not swapped for this: the collision and its fix are properties of the scrape configuration, so they are provable against the released 0.9.10 build. The cluster was returned to its prior state and diffed to confirm it (CRD and webhook inventory identical, image matching `helm get manifest`, no Flux object left suspended).

Defilan (changes requested):
Reviewed the code against a real apply of the diff onto current main, building and running all the new specs. This is a strong fix and most of it is exactly right; there's one real correctness regression in the position math to fix before merge, plus two non-blocking notes.

### Request changes

**FIFO positions collide on tied CreationTimestamps.** `evaluateGPUQueue` (`internal/controller/scheduling.go`) counts `ahead` only on strict `svc.CreationTimestamp.Before(&isvc.CreationTimestamp)`. Kubernetes `CreationTimestamp` is 1-second resolution, so two InferenceServices created in the same wall-clock second (a Helm release that deploys several, a batch `kubectl apply -f`, a script loop) get identical timestamps and therefore the same `ahead` count and the same reported position. The old `calculateQueuePosition` selection-sorted the waiting set and took `position = index+1`, which is always a strict `1..N` permutation even under ties, so positions never collided. So this contradicts the PR's stated invariant ("so positions don't shift"): they do not shift, they collide, and two different services' Progressing condition can simultaneously read "Queued at position 1 waiting for GPU."

It is not destructive: the `GPUQueueDepth` gauge is tie-independent and correct, and scheduling is unaffected; it's only the user-facing position number/message. Fix is small: add a secondary tie-break by name, counting a service as ahead when it is `Before` OR (equal timestamp AND `Name <` the subject). Please also add a `DescribeTable` case with two equal timestamps, since the current entries (10, 20, 30, 5, 40, 15, 50) never tie and pass either way.

### Non-blocking notes

- **`GPUQueueDepth` doesn't get #1229's defer-based republish.** It's published inline in `updateStatusWithSchedulingInfo`, not from `Reconcile`'s defer the way `publishInferenceServiceState` republishes phase/replicas/info to survive a reconcile that returns early (#1225). Because `PublishGPUQueueDepth` rebuilds the whole map from a cluster-wide List, any other InferenceService reconciling successfully refreshes it, so the only real stranding case is deleting the last queued service in the whole cluster. That matches your "DELETE path can strand one" caveat and is no worse than main. A defer republish would fully close it, but I would not block on it.
- **Unconditional publish adds a cluster-wide List on every status write**, where the old List was gated to `PhaseWaitingForGPU`. It's informer-cache-backed so each call is cheap, but it now runs for every phase (including the early `Failed`/`Pending` call sites), O(N) per reconcile in total InferenceServices. Fine at current scale, worth watching in the thousands. This is the deliberate trade-off matching the #1225 lesson, so I am fine with it as chosen.

### Verified correct

- **Reset semantics.** `PublishGPUQueueDepth` does `Reset()` then rebuilds from the map, so a namespace absent from the map is genuinely gone and depth-0 is `Set(0)`, not silently dropped. And on a List error `evaluateGPUQueue` returns `nil` depths and the publish is skipped, so a transient cache hiccup does not blank the gauge. Good.
- **The `Status.QueuePosition` side-fix is correct and not gratuitous.** Taking phase/timestamp from the in-memory `isvc` rather than the cache-listed copy fixes a real transient "position 0" bug (a service entering the queue this reconcile has a stale cached phase and is excluded from the old waiting set), and the same self-listing staleness would under-count the per-namespace depth if not handled, so it is structurally required for a correct single-pass function.
- **Dashboard and alert** are right for a GaugeVec: `sum(...)` for the cluster-wide stat, `{{namespace}}` legend, and the untouched `> 0` alert fires per namespace with the added `{{ $labels.namespace }}` annotation.

### Nice catch worth calling out

The `honorLabels: true` ServiceMonitor change is a genuine fix beyond this metric: without it, Prometheus's target-namespace relabeling overwrites any `llmkube_*` metric's own `namespace` label (renaming it to `exported_namespace`), which means #1229's namespace-labeled metrics (`llmkube_inferenceservice_phase`, etc.) were already silently broken the same way. This repairs all of them, not just the new one.

### Coordination

This overlaps #1244 (dashboard phantom-panel cleanup): both edit `internal/metrics/metrics.go` and the GPU dashboard JSON. Different metrics, no logical conflict, but whichever merges second will likely need a quick rebase.

Fix the tie-break (and its test) and I'm happy to re-approve. Everything else is solid.

Tanguille:
Fixed. `evaluateGPUQueue` now breaks ties on equal `CreationTimestamp` by name, so services created in the same second get a strict 1..N ordering instead of colliding on the same position. Added a dedicated `DescribeTable` covering both orderings.

Also rebased onto current main now that #1244, #1230 and #1240 have merged. #1230 touches the same metrics.go/metrics_test.go, applied clean.

internal/controller/scheduling.go
```diff
@@ -148,49 +147,53 @@ func (r *InferenceServiceReconciler) getPodSchedulingInfo(ctx context.Context, i
...
-				waitingServices[i], waitingServices[j] = waitingServices[j], waitingServices[i]
+		depth := depths[svc.Namespace]
+		if svc.Status.Phase == PhaseWaitingForGPU {
+			depth++
+			// CreationTimestamp is 1s resolution, so a Helm release or a
+			// scripted loop can create several services within the same
+			// second; break the tie by name so positions stay a strict
+			// 1..N permutation instead of colliding.
+			if svc.CreationTimestamp.Before(&isvc.CreationTimestamp) ||
+				(svc.CreationTimestamp.Equal(&isvc.CreationTimestamp) && svc.Name < isvc.Name) {
```
Defilan: Name is only unique per namespace, so same-name services in different namespaces created in the same second still collide. Including namespace makes the tuple a total order:

```suggestion
			if svc.CreationTimestamp.Before(&isvc.CreationTimestamp) ||
				(svc.CreationTimestamp.Equal(&isvc.CreationTimestamp) &&
					(svc.Namespace < isvc.Namespace ||
						(svc.Namespace == isvc.Namespace && svc.Name < isvc.Name))) {
```

Defilan (approved):
Re-reviewed after the tie-break commit and the rebase. The fix is right, the new DescribeTable proves both orderings, verified the honorLabels ServiceMonitor change and the chart test, and the rebase is clean against current main (no interaction with #1254: suspended services short-circuit before WaitingForGPU, so they never enter this queue). CI is green across the full matrix.

Approving with one non-blocking suggestion attached inline: the tie-break tuple is (CreationTimestamp, Name), but name is only unique within a namespace. Two same-named services in different namespaces (the same Helm-loop scenario your comment describes, deployed into team-a and team-b) created in the same second still tie, and both report the same position. Including namespace in the tuple makes the ordering total. One click to apply, and the existing table could gain a cross-namespace entry whenever convenient.

Good to merge once applied. Thanks for the fast turnaround on both rounds, this landed well.

Issue #786, opened 2026-06-21
[BUG] #409 ships TTFT histogram + request-error counter that no production code emits


Defilan:
## Bug Description

PR #769 (#409) registers two Prometheus metrics that no production code ever emits, and ships recording rules + a Grafana panel that read them. The result is observability that silently shows zero forever.

- `llmkube_inference_ttft_seconds` (histogram, `internal/metrics/metrics.go:81`)
- `llmkube_inference_request_errors_total` (counter, `internal/metrics/metrics.go:90`)

Both are only ever referenced in `metrics.go` (definition + registration). No controller or runtime path calls `.Observe()` / `.Inc()`:

```
$ grep -rn 'InferenceTTFT\|InferenceRequestErrors' --include='*.go' | grep -v _test.go
internal/metrics/metrics.go:81:  InferenceTTFT = prometheus.NewHistogramVec(...)
internal/metrics/metrics.go:126: InferenceTTFT,
internal/metrics/metrics.go:90:  InferenceRequestErrors = prometheus.NewCounterVec(...)
internal/metrics/metrics.go:127: InferenceRequestErrors,
```

The unit tests pass because they call `.Observe()` / `.Inc()` *inside the test itself* (`internal/metrics/metrics_test.go`), so they validate the Prometheus client library, not any production emission path.

## Impact

- `charts/llmkube/templates/prometheusrule.yaml` recording rules `llmkube:inference:ttft:p95_5m` and `llmkube:inference:request_errors:rate5m` evaluate over empty series.
- The Grafana panel "Request error rate by status class" (`docs/grafana/llmkube-inference.json`) uses `noValue: "0"`, so it actively displays a healthy-looking flat zero for a metric that is never populated. An operator would read "0 errors" when in reality nothing is measured.

## Expected Behavior

Either:
1. Wire the emission: increment `InferenceRequestErrors` on real 4xx/5xx responses and observe `InferenceTTFT` on first-token latency in the request path the metrics claim to measure; or
2. Remove the inert metrics + their recording rules + dashboard panel until the emission path exists.

## Root cause / class

Same test-weakness class as #784: a test that passes in isolation without exercising the real production path (here, a self-incrementing metric test). Surfaced by the weekend test-quality audit.

## Notes

The `--enable-metrics` vLLM-flag portion of #409 is correct and well-tested; only the two Go metrics above are affected.

## Environment

**LLMKube Version:** 0.8.12

PR #1244, opened 2026-07-23
fix(grafana): delete dashboard panels for metrics nothing emits

### LINKED ISSUE #1226: [BUG] Shipped Grafana dashboards query metric names that nothing emits

Tanguille:
Closes #1226

## The defect

Shipped dashboards query metric names nothing in the repo produces. A panel on a name that is never emitted renders empty, which looks exactly like an idle cluster, so the dashboard silently misleads rather than failing.

Confirmed against a live Prometheus over a 30-day window: neither `llamacpp:tokens_per_second` nor `llamacpp:kv_cache_usage_ratio` has ever been scraped. llama.cpp exposes `llamacpp:predicted_tokens_seconds` and has no KV-cache-ratio series at all.

## The change

Phantom panels are deleted, and `internal/metrics/dashboard_sync_test.go` walks every dashboard's PromQL and fails when a queried name is not one of:

- an fqName declared by a collector in `AllCollectors` or `agent.AgentRegistry`
- a name recorded by the chart's PrometheusRule
- owned by an exporter `config/grafana/SETUP.md` tells operators to install

That last rule is the one worth stating explicitly, because it is what stops the allowlist becoming a hole. `otel_span_metrics` was **deleted rather than allowlisted**: nothing in this repo emits it, nothing documents installing a spanmetrics connector, and listing it would have meant shipping a guard test that suppresses the exact defect it was written to catch.

The known-names set is built from the registries the repo already maintains rather than a hand-written list, so it cannot drift from what the operator actually registers.

## What this guard does not do, stated plainly

It checks a name is **declared**, not that anything ever writes it. It would **not** have caught #786, where `llmkube_inference_ttft_seconds` was declared, registered, dashboarded, and never observed — `declaredNames()` would have returned it and the test would have passed. The doc comment says so rather than implying broader coverage.

That limit is why this PR also deletes `GPUQueueWaitDuration`. It is declared at `metrics.go:86` and registered at `:243`, and the only `Observe()` call in the entire tree is in `metrics_test.go`. Its panels are among those being deleted here, so leaving the declaration behind would let them be re-added tomorrow with CI green. Removing the declaration is what makes the guard's answer correct rather than accidentally correct, and it is the same class of removal `9e42217` did for #786 in the same file.

## Verification

`make lint` → `0 issues`. `make test` → all packages `ok`, `internal/metrics` 100.0%.

The guard discriminates rather than merely passing. Re-injecting the phantom name this PR removes:

```
--- FAIL: TestDashboardsQueryEmittedMetrics (0.01s)
    dashboard_sync_test.go:251: ../../docs/grafana/llmkube-inference.json queries
    "llamacpp:tokens_per_second", which no registered collector, chart recording
    rule or allowlisted exporter emits
        query: rate(llamacpp:tokens_per_second[5m])
```

Panel deletion left four grid rows half empty; those are repacked so the survivors fill the width. Verified programmatically that every row in every shipped dashboard now sums to 24 columns, except one pre-existing gap in `llmkube-inference.json` that predates this change and is left alone.

## Adjacent problems found and deliberately not fixed here

1. **`model-router-dashboard.json` cannot populate at all.** `internal/router/proxy.go` registers seven `llmkube_router_*` collectors, but `proxy.Mount` registers only `/v1/chat/completions`, `/v1/models`, `/health` and `/healthz` — there is no `/metrics` handler in `cmd/router-proxy`. This is a missing scrape path, not a naming bug, and wiring it needs a port, Service, PodMonitor and chart values. The guard cannot see it, because a `Desc` exists. Documented in the README table so the dashboard is not mistaken for broken-by-naming; happy to file it separately.

2. **`docs/site/guides/metrics-driven-autoscaling.md`** ships a prometheus-adapter rule matching `llamacpp_.+` and renaming to `llamacpp:$1`. Per the metric names confirmed here, llama.cpp emits `llamacpp:` natively, so that rule matches nothing and the custom metric never exists — while `runtime_llamacpp.go:67` and `hpa_reconciler.go:114` make `llamacpp:requests_processing` the default HPA metric. Same bug class as this issue with an HPA that never scales instead of a panel that never fills. Left alone because it is a larger blast radius than dashboards; say the word and I will widen the guard to cover shipped YAML `expr:` and adapter `metricName:` too.

### DIFF
--- a/config/grafana/SETUP.md
+++ b/config/grafana/SETUP.md
@@ -256,12 +256,12 @@ curl http://prometheus:9090/api/v1/targets
 
 | Metric | Description |
 |--------|-------------|
-| `llamacpp_tokens_second` | Current generation speed (tokens/sec) |
-| `llamacpp_tokens_predicted_total` | Total generated tokens |
-| `llamacpp_prompt_tokens_total` | Total prompt tokens processed |
-| `llamacpp_kv_cache_usage_ratio` | KV cache utilization (0-1) |
-| `llamacpp_requests_processing` | Currently processing requests |
-| `llamacpp_requests_deferred` | Deferred/queued requests |
+| `llamacpp:predicted_tokens_seconds` | Current decode speed (tokens/sec) |
+| `llamacpp:prompt_tokens_seconds` | Current prefill speed (tokens/sec) |
+| `llamacpp:tokens_predicted_total` | Total generated tokens |
+| `llamacpp:prompt_tokens_total` | Total prompt tokens processed |
+| `llamacpp:requests_processing` | Currently processing requests |
+| `llamacpp:requests_deferred` | Deferred/queued requests |
 
 ### LLMKube Controller Metrics
 
@@ -273,8 +273,6 @@ curl http://prometheus:9090/api/v1/targets
 | `llmkube_inferenceservice_ready_duration_seconds` | Time to Ready phase |
 | `llmkube_reconcile_total` | Total reconciliation cycles |
 | `llmkube_reconcile_duration_seconds` | Reconciliation cycle duration |
-| `llmkube_active_models_total` | Models in Ready/Cached phase |
-| `llmkube_active_inferenceservices_total` | Inference services in Ready phase |
 
 ### LLMKube Metal Agent Metrics
 
--- a/config/grafana/llmkube-gpu-dashboard.json
+++ b/config/grafana/llmkube-gpu-dashboard.json
@@ -1545,17 +1545,7 @@
           "color": {
             "mode": "thresholds"
           },
-          "mappings": [
-            {
-              "options": {
-                "Ready": { "color": "green", "index": 0 },
-                "Downloading": { "color": "yellow", "index": 1 },
-                "Pending": { "color": "blue", "index": 2 },
-                "Failed": { "color": "red", "index": 3 }
-              },
-              "type": "value"
-            }
-          ],
+          "mappings": [],
           "thresholds": {
             "mode": "absolute",
             "steps": [
@@ -1587,8 +1577,8 @@
             "type": "prometheus",
             "uid": "${datasource}"
           },
-          "expr": "llmkube_model_status{instance=~\"$instance\"}",
-          "legendFormat": "{{name}} ({{namespace}})",
+          "expr": "llmkube_model_status",
+          "legendFormat": "{{model}} ({{namespace}}) {{phase}}",
           "refId": "A"
         }
       ],
@@ -1605,16 +1595,7 @@
           "color": {
             "mode": "thresholds"
           },
-          "mappings": [
-            {
-              "options": {
-                "Running": { "color": "green", "index": 0 },
-                "Pending": { "color": "yellow", "index": 1 },
-                "Failed": { "color": "red", "index": 2 }
-              },
-              "type": "value"
-            }
-          ],
+          "mappings": [],
           "thresholds": {
             "mode": "absolute",
             "steps": [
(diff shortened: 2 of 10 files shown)


Tanguille:
Confirmed against 30 days of production data on a live cluster. Every name this PR deletes has never been scraped once; every name it keeps has data.

| metric | series in last 30d | this PR |
|---|---|---|
| `llamacpp:tokens_per_second` | **0** | deleted |
| `llamacpp:kv_cache_usage_ratio` | **0** | deleted |
| `otel_span_metrics` | **0** | deleted |
| `llmkube_gpu_queue_wait_duration_seconds` | **0** | declaration removed |
| `llamacpp:predicted_tokens_seconds` | 187 | kept |
| `llmkube_inferenceservice_phase` | 250 | kept |

The `llmkube_gpu_queue_wait_duration_seconds` row is the one I want to draw attention to, because it was the least comfortable call in this PR. The argument for removing the declaration was static: registered, but no `Observe()` anywhere outside a test. Zero series across 30 days of a cluster that actively queues on GPU is the runtime confirmation, and it is the same evidence that would have justified 9e42217 for #786.

No operator image was deployed to obtain this. These are scrape-path facts, so they hold against the released 0.9.10 build, and the cluster was left exactly as found.

Defilan (approved):
Reviewed the code against a real checkout of the PR head (not just the diff), building and running the tests. Approving. This is a genuinely well-built guard and the deletions are all justified.

### Verified

- **The guard discriminates, proven both directions.** Renaming a real metric to a bogus name in a dashboard makes `TestDashboardsQueryEmittedMetrics` fail with the exact "no registered collector, chart recording rule or allowlisted exporter emits" error; the real content passes. The allowlist's collector half is genuinely dynamic, not hand-maintained: it walks `AllCollectors` (the same slice `init()` registers) and `agent.AgentRegistry.Describe(...)`, and parses the chart PrometheusRule's `- record:` lines directly. PromQL name extraction handles `rate(x[5m])`, `sum by(ns)(x)`, `label_values(metric,label)`, and colon-bearing recording-rule names. It runs over all 7 shipped dashboards across `config/grafana` and `docs/grafana`, not a subset.
- **The deleted names are genuinely unemitted.** Cross-checked against the current llama.cpp server metrics list: it exposes `llamacpp:predicted_tokens_seconds` and has no KV-cache-ratio series of any kind, so the deleted "KV Cache Occupancy" panels queried a series that never existed in shipped llama.cpp, not a renamed one. `otel_span_metrics` appears nowhere in the repo and nothing documents a spanmetrics connector, so deleting it (rather than allowlisting it) is the right call and keeps the guard honest.
- **`GPUQueueWaitDuration` removal is safe.** Its only pre-PR references were the declaration/registration and one `TestHistogramBuckets` case; no production path observed it. Post-PR there are zero references anywhere (Go/JSON/YAML/MD), build and tests green.
- **Grid integrity holds.** Every row touched by the diff sums to 24 columns with no x-overlaps. The one non-24 row in `docs/grafana/llmkube-inference.json` predates this PR and isn't touched.

### Minor (non-blocking)

- `internal/metrics/dashboard_sync_test.go` (the `externalPrefixes` comment): it says these are "exporters SETUP.md tells operators to install," but only `DCGM_FI_DEV_` and `node_` are actually in SETUP.md. `vllm:`/`up:` are Pyrra-derived recording rules (documented in the SLO doc) and `amdgpu_` is documented in the grafana README, not SETUP.md. The prefixes are all legitimate, so this masks nothing today, but `externalPrefixes` is the one hand-maintained piece of an otherwise-dynamic allowlist, so it's worth a one-line comment fix to point at the real sources so it doesn't drift into the exact hole this test exists to prevent.

### Coordination note

This overlaps #1243 (per-namespace `gpu_queue_depth`): both edit `internal/metrics/metrics.go` and the GPU dashboard JSON. They touch different metrics (wait-duration vs depth), so there's no logical conflict, but whichever merges second will likely need a quick rebase.

Thanks for the fail-injection proof and the honest "what this guard does NOT catch" section (the declared-but-never-observed case, #786). Approving.

PR #1245, opened 2026-07-23
feat(grafana): chart the InferenceService metrics with no panel

Tanguille:
Follow-up to #1244, now rebased onto `main` with #1244's own commit merged (`9c586d9`).
#1226 and #1244 handle one direction: panels querying names nothing emits. The opposite gap was never measured, so I measured it.

## The gap

Of the 43 names this repo declares or records, **16 appear on a dashboard and 27 do not**. Cross-referencing against 30 days of production data narrows what is worth acting on:

| category | count | action |
|---|---|---|
| emitted per InferenceService, no panel | 2 | **charted here** |
| Metal agent (`llmkube_metal_agent_*`) | 19 | left alone, see below |
| router (`llmkube_router_first_token_seconds`) | 1 | blocked on the missing `/metrics` endpoint noted in #1244 |
| vLLM TTFT recording rule | 1 | not a gap; empty only where `prometheus.prometheusRule.enabled=false` |
| histogram base names whose `_bucket` is charted | 4 | not a gap; artefact of my first count |

## What this adds

**Replicas ready vs desired** — `llmkube_inferenceservice_replicas`, plotted per service and state (the metric is already unique on those two labels, so no aggregation is needed). The phase timeline reports `Ready` as soon as one replica is, so a service sitting at 1/3 looks healthy on every existing panel. Ready trailing desired is a partial rollout; a flat gap is a stuck one.

**Service inventory** — a table over `llmkube_inferenceservice_info`, aggregated server-side with `max by (inferenceservice, namespace, accelerator, runtime)` so ServiceMonitor scrape-metadata labels never reach the table, even if a future label gets added. The value is always 1 and the labels are the payload, so a timeseries is the wrong shape; an `organize` transformation renames the four real labels for display. Answers "what is actually running, on what" without reading CRs.

`InferenceService Phase` goes back to half width. #1244 widened it to 24 only because deleting a phantom neighbour left the row half empty — that was filling a hole, not a layout decision, and the row now has a real second panel.

## Verification

Both names pass #1244's guard test (`TestDashboardsQueryEmittedMetrics`), which now runs against the merged collector set on `main`.

Live data, same cluster and window as the #1244 evidence:

- `llmkube_inferenceservice_info` — 4 live series, e.g. `{inferenceservice="qwen36-27b", runtime="sglang", accelerator="rocm"}`
- `llmkube_inferenceservice_replicas` — no current series, and that is expected rather than a phantom: `v0.9.10` (the running release) contains **no reference to this metric at all**. It was added in #1229 and is published from `status_builder.go:150` via `Reconcile`'s defer, so it starts emitting with the next release.

Grid verified programmatically: JSON valid, no duplicate panel ids, every row sums to 24 columns.

## Why the Metal agent set is not charted

19 metrics with zero coverage is the largest remaining gap, and I deliberately did not fill it. Nothing in this repo's test or staging surface emits them — they need a macOS host running the Metal agent — so the panels, their units, and their thresholds would all be guesswork. Writing dashboards for metrics nobody has observed is exactly how the panels #1244 removes came to exist. Happy to do it against a real agent, or to pair with someone who has one.

### DIFF
--- a/config/grafana/llmkube-inference-dashboard.json
+++ b/config/grafana/llmkube-inference-dashboard.json
@@ -524,7 +524,7 @@
         },
         "overrides": []
       },
-      "gridPos": { "h": 8, "w": 24, "x": 0, "y": 33 },
+      "gridPos": { "h": 8, "w": 12, "x": 0, "y": 33 },
       "id": 42,
       "options": {
         "legend": { "calcs": ["lastNotNull"], "displayMode": "table", "placement": "bottom" },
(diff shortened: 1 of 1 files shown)


Defilan (approved):
Reviewed against the actual metric definitions and the dashboard JSON at the PR head.

Verified:
- Both queries match the emitters exactly: `llmkube_inferenceservice_replicas` carries `inferenceservice/namespace/state` (legend renders correctly), and `llmkube_inferenceservice_info` carries exactly the four labels your `max by` groups on (internal/metrics/metrics.go:65-80).
- JSON parses, 23 panels with no duplicate ids (new 43/60/61 clean), `${datasource}` var consistent with the rest of the file, gridPos does not collide.
- The server-side `max by` on the inventory table is a smart choice given the `exported_namespace` renaming you diagnosed in #1243; this table stays correct whichever way that lands.
- The measured 16-charted/27-uncharted gap analysis in the body is exactly the kind of legwork that makes a dashboard PR easy to trust. Leaving the 19 metal-agent metrics alone was the right call.

One clarification, non-blocking: the stated motivation says the phase timeline reports Ready as soon as one replica is ready. determinePhase actually returns Ready only when readyReplicas equals desiredReplicas (internal/controller/scheduling.go:69); a service at 1/3 shows Progressing, not Ready. The panel is still the right addition, and arguably for a stronger reason: Progressing tells you neither the magnitude of the gap (1/3 vs 2/3) nor how long it has been stuck, and this chart shows both. Worth a one-line tweak to the PR description if you feel like it; the code needs no change.

Approving. Two CI jobs were still running at review time; merge on green as usual.

Issue #1227, opened 2026-07-22
[FEATURE] No SGLang dashboard, and the inference dashboard is llama.cpp-only


Tanguille:
## Feature Description

SGLang is a first-class runtime — `RuntimeSGLANG` in `internal/controller/runtime_sglang.go:42`, its own backend, in the valid-runtime map at `deployment_builder.go:582` — but there's no dashboard for it anywhere in `docs/grafana/` or `config/grafana/`.

And the inference dashboard that does exist is llama.cpp-only by construction, so on a mixed-runtime fleet there's no single view. Two asks in one issue because the fix is the same edit; happy to split if you'd rather.

## Problem Statement

We run four InferenceServices across two runtimes: one SGLang (27B dense on a Radeon R9700) and three llama.cpp (2B generation plus two embedding models on a 660M iGPU). There's no shipped dashboard that shows both, so we maintain our own fleet view and our own SGLang view.

The gap isn't just cosmetic. `config/grafana/llmkube-inference-dashboard.json` puts a service selector on the dashboard, but the panels underneath are all `llamacpp*`. Pick an SGLang service and you get empty panels rather than an error, so it reads as "service is idle" instead of "this dashboard doesn't cover this runtime". Its restart panel has the same problem — it filters on a hardcoded container name, and SGLang's container is `sglang`. (The chart's recording rule has the same bug, filed separately as #1223.)

## Proposed Solution

**Make the fleet dashboard runtime-agnostic.** No recording-rule layer needed — a two-target `or` union on the `service` label the PodMonitor already relabels does it:

```promql
sum by (service) (rate(llamacpp:tokens_predicted_total[$__rate_interval]))
  or sum by (service) (rate(sglang:generation_tokens_total[$__rate_interval]))
```

The clean pairings across the two runtimes:

| signal | llama.cpp | SGLang |
|---|---|---|
| generated tokens | `llamacpp:tokens_predicted_total` | `sglang:generation_tokens_total` |
| prompt tokens | `llamacpp:prompt_tokens_total` | `sglang:prompt_tokens_total` |
| in flight | `llamacpp:requests_processing` | `sglang:num_running_reqs` |
| queued | `llamacpp:requests_deferred` | `sglang:num_queue_reqs` |

For restarts, drop the container filter and join on the `inference.llmkube.dev/service` pod label instead — that's runtime-independent and won't need editing when a new runtime lands.

[...]

PR #1248, opened 2026-07-23
feat(grafana): union llama.cpp, sglang and vLLM metrics on the inference dashboard

Tanguille:
First PR against #1227 — this half unions llama.cpp, sglang and vLLM metrics on the shared fleet dashboard; #1266 adds the standalone SGLang and vLLM dashboards the issue also asked for.

Follow-up to #1244 and #1245 (both merged into `main`). Rebased onto current `main`, which already contains both squashed. This is now a single, self-contained commit: `feat(grafana): union llama.cpp, sglang and vLLM metrics on the inference dashboard`.

## The problem

The shared Inference Monitor panels (tokens/sec, requests processing, requests deferred, generation and prompt tokens/sec) queried llama.cpp metric names only. Every one of those panels renders blank in a namespace running SGLang or vLLM.

## What this does

Panel queries now union all three runtimes with PromQL `or`, e.g.:

```
sum(rate(llamacpp:tokens_predicted_total{namespace=~"$namespace"}[5m]) or rate(sglang:generation_tokens_total{namespace=~"$namespace"}[5m]) or rate(vllm:generation_tokens_total{namespace=~"$namespace"}[5m]))
```

### Behaviour change on the speed panels

The decode- and prefill-speed panels changed from `sum(llamacpp:predicted_tokens_seconds{...})` to `sum(rate(llamacpp:tokens_predicted_total{...}[5m]) or ...)`. These measure different things: the old one is llama.cpp's own gauge of speed *while generating*, the new one is a 5 minute average that includes idle time, so it reads substantially lower on bursty traffic. SGLang and vLLM expose no instantaneous-speed gauge, so a union has to agree on the counter shape. SETUP.md notes this and no longer lists the two abandoned gauges as if a panel used them.

`config/grafana/SETUP.md` documents all three columns side by side instead of a llama.cpp-only table.

`internal/metrics/dashboard_sync_test.go` gains `TestInferenceDashboardCoversEveryRuntime`, which fails if a future panel on this dashboard reads one runtime's metrics without the others'. It also replaces the hardcoded llama.cpp name list with `internal/metrics/testdata/{llamacpp,sglang,vllm}-metrics.txt`, one metric name per line, so a new runtime widens the coverage check by dropping in a fixture rather than editing the test.

Because vLLM is now a fixture runtime rather than an allowlisted external namespace, the `externalPrefixes` entry for it is narrowed from the whole `vllm:` namespace to `vllm:e2e_request_latency_seconds:`, the prefix Pyrra actually generates. The runtime's own metrics are verified against the fixture like every other runtime's.

## Verification

`go build ./...` and `go test ./internal/metrics/...` pass, rebased against current `main`.

The coverage guard was checked for teeth, not just for green: removing vLLM from a single panel expression makes `TestInferenceDashboardCoversEveryRuntime` fail with `reads [llamacpp sglang] only, so the panel is blank on every other runtime`.

Every vLLM metric name used here was taken from a live vLLM `/metrics` scrape on a running cluster, not from documentation.

## Note on scope

This branch was originally built against plain `main`, independently of #1244. Rebasing it collapsed most of the original diff: two files (`llmkube-gpu-dashboard.json`, `model-router-dashboard.json`) turned out byte-identical to what #1244 shipped, and a third (`llmkube-inference.json`) collapsed after picking #1244's gridPos fix over this branch's, which left a dangling half-empty row. What's left is only the cross-runtime union work specific to this PR.


Defilan (commented):
Reviewed the diff and traced the claims rather than taking the description on faith. The core work is right, and one part of it is right in a way that is easy to get wrong, so I want to call that out before the nits.

**The union is actually sound, and that is not obvious.**

My first concern was that this would silently drop runtimes. PromQL's `or` only adds right-hand series whose label set does not already exist on the left, and `rate()` strips `__name__`. So `rate(llamacpp:tokens_predicted_total[5m]) or rate(sglang:generation_tokens_total[5m])` looks like it should swallow the SGLang series in any namespace running both.

It does not, because `inference-podmonitor.yaml` (lines 34-35) relabels a `runtime` label onto every scraped inference series. The label sets always differ by `runtime`, so nothing is dropped and the `sum` is a true total in a mixed namespace. Worth a comment somewhere, because the correctness of every panel in this PR rests on a relabeling rule three files away.

I also like that the guard was checked for teeth rather than for green. "Removing vLLM from a single panel makes the test fail" is the right thing to verify and the right thing to put in a PR description.

**Blocking: DCO.** The sign-off check is failing. Everything else is green.

**One thing I would fix before merge: SETUP.md now documents two metrics no panel uses.**

The decode-speed panel changed from `sum(llamacpp:predicted_tokens_seconds{...})` to `sum(rate(llamacpp:tokens_predicted_total{...}[5m]) or ...)`. Those measure different things. The old one is llama.cpp's own gauge of decode speed *while generating*; the new one is a 5 minute average that includes idle time, so the number will read substantially lower on bursty traffic.

I think that is the correct call, since SGLang and vLLM expose no instantaneous-speed gauge and a union has to agree on a shape. But it is a silent behaviour change to an existing panel and it deserves a line in the PR body.

The doc problem follows from it. The new table still lists:

| Description | llama.cpp | SGLang | vLLM |
|---|---|---|---|
| Current decode speed (tokens/sec) | `llamacpp:predicted_tokens_seconds` | | |
| Current prefill speed (tokens/sec) | `llamacpp:prompt_tokens_seconds` | | |

Neither of those metrics is used by any panel after this change; both panels now derive from the `_total` counters one and two rows down. So two of the six rows point an operator at metrics the dashboard abandoned, which is precisely the wrong signpost for someone debugging a blank panel. Either drop those rows, or note that they are exposed but unused.

**One question, not a blocker: dashboard discovery got narrower in a PR about making coverage wider.**

```go
-// Every grafana directory in the repo, so a new one is covered without an edit.
-const dashboardGlob = "../../*/grafana/*.json"
+var dashboardDirs = []string{"../../config/grafana", "../../docs/grafana"}
```

Only two grafana directories exist today, so this is equivalent right now. But the comment you removed states the glob's purpose was to cover a new directory without an edit, and the hardcoded list gives that up: a third directory added later is silently unchecked.

It sits oddly next to the part of this PR I like most, where a new runtime widens the check by dropping in a fixture instead of editing the test. Fixtures became data-driven and dashboard discovery became manual in the same commit. Was there a reason the glob had to go? If it was just tidiness, I would keep the glob.

**Verified and fine:** narrowing `externalPrefixes` from the whole `vllm:` namespace to `vllm:e2e_request_latency_seconds:` is a real tightening, and the fixture covers every `vllm:` name the dashboards currently reference. Good change, and it is the kind that only pays off later when someone adds a vLLM metric that does not exist.

So: sign off the commit, fix the two stale SETUP.md rows, and tell me whether the glob change was deliberate. The rest is good work, and pulling the metric names from a live scrape rather than documentation is the detail that makes me trust the fixtures.

Tanguille:
All four addressed, force-pushed as 5ac0688.

**DCO** — commit amended with `Signed-off-by`.

**Glob** — not deliberate, just tidiness while I was in that file, and you are right that it traded away the property the comment named. Restored `dashboardGlob = "../../*/grafana/*.json"` and its comment; `dashboardDirs` and the per-directory loop are gone. A third grafana directory is covered again on creation, same as the fixtures.

**Stale SETUP.md rows** — dropped both. The gauges are still exposed by llama.cpp, so rather than delete the fact I put it in a note above the table: the speed panels derive from the `_total` counters via `rate()` over 5m, and `llamacpp:predicted_tokens_seconds` / `llamacpp:prompt_tokens_seconds` exist but no panel uses them because the other two runtimes expose no counterpart.

**Behaviour change** — added to the PR body under "Behaviour change on the speed panels", including that the 5m average includes idle time and so reads lower than the gauge on bursty traffic.

**The `runtime` relabeling** — documented where it actually protects something, on `TestInferenceDashboardCoversEveryRuntime`: `or` drops a right-hand series whose label set already exists on the left and `rate()` strips `__name__`, so the union is only lossless because the PodMonitor relabels a distinct `runtime` label onto every scraped series. Drop that rule and every panel here starts undercounting.

Defilan (approved):
All three addressed, and one of them better than I asked for. Approving.

**DCO** signed off. **The glob is back**, comment and all, so a future `grafana/` directory stays covered without an edit.

**The stale SETUP.md rows are gone.** `predicted_tokens_seconds` and `prompt_tokens_seconds` no longer sit in the table pointing operators at metrics no panel reads, and the four rows that remain are exactly the ones the panels use.

The part I did not ask for and appreciate:

> The speed panels derive from the `_total` counters below via `rate()` over 5 minutes, so they read lower than an instantaneous gauge on bursty traffic.

That was the real hazard in the change and it is now documented rather than silent. Someone comparing a dashboard against last week's screenshot will find the explanation instead of filing a bug.

Everything I verified on the first pass still holds: the `or` union is sound because the PodMonitor relabels a `runtime` label onto every scraped series, so nothing gets swallowed in a mixed namespace; the coverage guard is intact; the `externalPrefixes` narrowing to `vllm:e2e_request_latency_seconds:` is a real tightening with the fixture covering current usage. 23 checks green.

Nice work, and thanks for pulling the metric names off a live scrape rather than documentation. That is the detail that makes the fixtures trustworthy.

PR #1266, opened 2026-07-24
feat(grafana): add SGLang and vLLM runtime dashboards

Tanguille:
Second PR against #1227. The shared fleet dashboard's cross-runtime union shipped in #1248; this adds the standalone per-runtime dashboards the issue also asked for, one for SGLang and one for vLLM.

## What this does

- `docs/grafana/sglang-dashboard.json` (uid `llmkube-sglang`): throughput, TTFT/E2E/inter-token latency, queue and concurrency, KV pool, prefix cache hit rate, and scheduler backpressure. Mamba pool and HiCache host pool are in collapsed rows since those series only exist on hybrid-attention models and with `--hicache` respectively. Sanitized from our own production dashboard: no hardware-specific panels, no hardcoded service or node names, templated on `$service`.
- `docs/grafana/vllm-dashboard.json` (uid `llmkube-vllm`): the same shape over vLLM's own metrics. Throughput, TTFT/E2E/inter-token latency, running vs waiting, GPU KV cache utilization, prefix cache hit rate, preemptions and aborts. Prefill/decode/queue request-phase timing is in a collapsed row.
- `charts/llmkube/templates/prometheusrule.yaml`: extends the two shared latency recording rules from vLLM-only to vLLM|SGLang (both runtimes export the same histograms), and adds p50 siblings of the existing p95 rules so the dashboards' p50 series read a shared rule instead of every viewer re-scanning the raw histogram on refresh.
- `docs/grafana/README.md`: new table rows.

## Stacked on #1248

This branch is rebased onto #1248's tip rather than `main`, because the exact-fixture guard test that recognizes each dashboard's metric names only exists on #1248, not yet on `main`. Base is `main` since a fork branch cannot be selected as a GitHub PR base; the diff will shrink to just this PR's own content once #1248 merges and this rebases onto it.

## Verification

Every vLLM metric name was validated against a live vLLM `/metrics` scrape, and every panel query confirmed to return data from a real Prometheus-compatible datasource. Each dashboard's `description` field was then checked back against the panels actually shipped, which corrected two of them: the SGLang latency panels plot p50 alongside p95, and the vLLM backpressure row counts aborts rather than every finish reason. `go test ./internal/metrics/...`, `helm lint` / `helm template` in all three PrometheusRule states, and `helm unittest charts/llmkube` all pass.


Defilan (approved):
Approving. I checked the claims rather than taking the description on trust,
and they all hold.

## What I verified

**Both dashboards parse and are properly templated.** 20 panels on SGLang, 18
on vLLM, both templated on `$service` and `DS_PROMETHEUS`, no hardcoded service
names, node names or addresses anywhere in either file. The sanitization you
mention actually happened.

**Every metric name is real.** All 17 `sglang:*` and all 13 `vllm:*` series the
dashboards query are present in the `internal/metrics/testdata` fixtures. I also
confirmed the guard is not vacuous: renaming one series to
`sglang:totally_invented_metric` fails `TestDashboardsQueryEmittedMetrics` with
a precise error naming the file and the query. So the fixture check is doing
real work on these two files, which is the thing I most wanted to know before
merging 1000 lines of JSON.

Worth noting for anyone reading later: that guard globs `*/grafana/*.json`
rather than listing files, so these dashboards were covered the moment they
landed in the directory. Nice property, and it is why this PR needed no test
edit.

**The recording-rule claim is true.** Both dashboards reference
`llmkube:inference:request_latency:p50_5m` and
`llmkube:inference:ttft_seconds:p50_5m`, so the p50 panels really do read the
shared rules rather than re-scanning the histogram per viewer. That was the
justification for adding the rules and it checks out.

**Collapsed rows are actually collapsed**, and exactly the conditional ones:
Mamba pool and HiCache on SGLang, request-phase timing on vLLM. Someone
importing this against a plain non-hybrid model without `--hicache` sees a clean
dashboard, which is the right default.

**The chart change is consistent.** Rule count assertion moved 3 to 5 to match
the two new p50 rules, and you caught the `llmkube-inference.json` README row
that was still saying "vLLM only" now that the shared rules cover both runtimes.

```
go test ./internal/metrics/...   ok
helm lint charts/llmkube         0 failed
helm unittest charts/llmkube     40 passed, 5 suites
```

## Two minor notes, neither blocking

**The "stacked on #1248" section is stale.** #1248 merged, and this branch now
diffs cleanly against `main` at 5 files. Worth trimming from the description so
the next reader does not go looking for a dependency that no longer exists.

**p50 and p95 share a `latency_kind` value.** Both e2e rules carry
`latency_kind: end_to_end` and both TTFT rules carry `latency_kind: ttft`. That
is harmless as written since the record names differ and every query names the
record, but a future query that selects on `latency_kind` alone would match both
quantiles. Not worth changing now, just something to know is there.

On the `or` between the two runtime histograms: I convinced myself it is safe
here. `or` only drops right-hand series whose label set already exists on the
left, and a vLLM series and an SGLang series can never share one, since they
differ by `service` at minimum. Same pattern as #1248, so it stays consistent.

## Merging

This is the standalone half of #1227 and it does what the issue asked. Good
work, particularly the decision to keep the conditional pools collapsed and to
check the dashboard descriptions back against the panels actually shipped.
That last one catches a class of drift most people never look for.

PR #1267, opened 2026-07-24
feat(chart): template Grafana dashboards, publishing runtime ones only while their runtime serves

### LINKED ISSUE #1228: [FEATURE] Chart doesn't template Grafana dashboards, and has no grafana-operator GrafanaDashboard support
## Feature Description

The chart templates every other observability resource — `prometheusrule.yaml`, `servicemonitor.yaml`, `inference-podmonitor.yaml` — but not dashboards. Those live in `docs/grafana/` and `config/grafana/` as loose JSON you're expected to apply yourself.

Asking for two things: template them from the chart, and support the grafana-operator `GrafanaDashboard` CR alongside the sidecar-ConfigMap convention.

## Problem Statement

> As an operator installing the chart, I want the dashboards to arrive with the release like the ServiceMonitor and PrometheusRule do, instead of copying JSON out of the repo by hand and keeping it in sync myself.

Two gaps, and the second is the one that actually blocks us.

**No dashboard templating in the llmkube chart at all.** `charts/llmkube/values.yaml` has no `grafana` or `dashboard` key anywhere, and there's no dashboard template. So the dashboards are effectively docs, not deliverables — they drift from the metrics with nothing tying them to a release, which is roughly how #1226 happened.

**No grafana-operator support anywhere in the repo.** The one dashboard-templating precedent is `charts/foreman/templates/grafana-dashboard.yaml`, which renders a ConfigMap labelled `grafana_dashboard: "true"`. That's the kube-prometheus-stack sidecar convention and it works fine if you run that stack. It does nothing if you run grafana-operator, which wants a `GrafanaDashboard` CR (`grafana.integreatly.org/v1beta1`) with an `instanceSelector`. There's no values path to get one, so we hand-write them.

## Proposed Solution

The two mechanisms don't need separate copies of the JSON. `GrafanaDashboard` supports `configMapRef`, so one ConfigMap can serve both: label it for the sidecar, and optionally emit a CR pointing at it.

Rough shape, adjust to taste:

```yaml
dashboards:
  enabled: false
  # sidecar  - labelled ConfigMap only (kube-prometheus-stack)
  # operator - ConfigMap plus GrafanaDashboard CRs referencing it
  # both     - as it says
  mode: sidecar
  sidecarLabel: grafana_dashboard
  operator:
    instanceSelector:
      matchLabels:
        dashboards: grafana
    allowCrossNamespaceImport: true
```

What we currently hand-write, for reference:

```yaml
apiVersion: grafana.integreatly.org/v1beta1
kind: GrafanaDashboard
metadata:
  name: llmkube-inference
spec:
  allowCrossNamespaceImport: true
  instanceSelector:
    matchLabels:
      grafana.internal/instance: grafana
  datasources:
    - datasourceName: VictoriaMetrics
      inputName: DS_PROMETHEUS
  configMapRef:
    name: llmkube-dashboards
    key: llmkube-inference.json
```

plus a kustomize `configMapGenerator` to build `llmkube-dashboards` from the JSON files.

The `datasources` block there is worth stealing regardless of mode: it remaps the dashboard's `DS_PROMETHEUS` input to whatever the cluster's datasource is actually called. Chart-shipped dashboards otherwise have to guess, and ours is a VictoriaMetrics datasource, not one named "Prometheus".

## Sequencing

Worth landing after #1226. Right now most panels in the shipped dashboards query names nothing emits — templating them from the chart before that's fixed just distributes the dead panels more efficiently. #1227 is also relevant since it changes which dashboards there are to ship.

## Alternatives Considered

- Keep applying them by hand, which is what we do. Fine for one cluster, but it means the dashboards silently drift from the metrics, and there's no version relationship between the dashboard and the operator that emits the metrics.
- Sidecar ConfigMap only, matching the foreman chart. Simpler, and it'd be a real improvement over nothing — but grafana-operator installs still get nothing, and the CR is about 10 extra lines given the ConfigMap already exists.
- A separate `llmkube-dashboards` chart. More release surface than this needs.

## Additional Context

- Related: #1226 (dashboards query names nothing emits), #1227 (no SGLang dashboard, inference dashboard is llama.cpp-only)
- Precedent in-repo: `charts/foreman/templates/grafana-dashboard.yaml`
- We run grafana-operator with VictoriaMetrics rather than kube-prometheus-stack, so we're squarely in the case that currently has no path

## Priority

- [x] Medium - Would significantly improve my workflow

## Willingness to Contribute

- [x] Yes, I can submit a PR

Happy to do this one, though I'd rather #1226 lands first so I'm not templating dashboards that need rewriting anyway.

---

Found in the same pass as #1219, #1221, #1222, #1223, #1224, #1226 and #1227.

Tanguille:
Closes #1228. Rebased onto `main` now that #1266 has merged, so the diff is just this PR's own content.

## Why

The chart templates every other observability resource (PrometheusRule, ServiceMonitor, PodMonitor) but no dashboards. They sit as loose JSON under `docs/grafana/` and get copied in by hand. `charts/foreman` has a dashboard ConfigMap, but it only covers the kube-prometheus-stack sidecar convention, so grafana-operator users get nothing.

## What

The seven dashboards move to `charts/llmkube/dashboards/` (Helm's `.Files.Glob` can't read outside the chart root) and `grafana.dashboards.enabled` ships them as one ConfigMap, labelled for the sidecar. `grafana.dashboards.operator.enabled` adds a `GrafanaDashboard` CR per dashboard whose `configMapRef` points at the same key, so the JSON exists once regardless of which delivery mechanism you use. `config/grafana/*.json` is an unrelated set and stays where it is.

Two of the dashboards, SGLang and vLLM, read a single runtime's metrics. On a cluster serving neither they render blank, and blank looks exactly like an idle cluster, which is the confusion #1227 set out to fix. The chart can't decide this: which runtimes exist is an InferenceService fact that doesn't exist at `helm install` time, and Helm leaves nothing running to notice later.

So with `operator.mode: auto` (the default) the chart writes those two manifests into a candidates ConfigMap instead of applying them, and a new `GrafanaDashboardReconciler` publishes each one while an InferenceService on its runtime exists, retiring it when the last one goes. Deploy a vLLM service and the vLLM dashboard shows up; delete it and the dashboard goes away. `operator.mode: all` restores install-time publishing.

The controller gets the whole manifest, not a set of knobs, so `instanceSelector`, `folder`, `datasources` and the rest stay owned by `values.yaml`. It decides whether a dashboard exists, never what's in it. Retirement is scoped by a `managed-by: llmkube` label, so a hand-authored dashboard sharing a name is never touched. Both gates are self-disabling: no `--grafana-dashboard-source` flag or no grafana-operator CRD and the reconciler no-ops.

Also here:

- `grafana.dashboards.only` narrows the shipped set by filename, for the sidecar path or for dropping `amd-gpu-observability` and `llmkube-slo`, which have no runtime signal. A name matching no file fails the render.
- `operator.folder` and `operator.resyncPeriod`, matching what `home-operations/tuppr`'s chart exposes for the same integration. Without `folder` the operator path had no way to place dashboards at all.

## Verification

`go build ./...`, `go test ./internal/...`, `helm lint`, `helm unittest charts/llmkube` (59 tests).

I ran this end to end on my own cluster, which runs llama.cpp, SGLang and vLLM services, in a scratch namespace with an `instanceSelector` matching no Grafana:

- Helm applied the five unconditional CRs and deferred SGLang and vLLM into the candidates ConfigMap.
- First reconcile published the SGLang dashboard (a service on that runtime is Ready) and not the vLLM one, whose only service is `Stopped`.
- `folder`, `configMapRef` and the runtime annotation survived the ConfigMap round trip.
- Pointing the candidate at a runtime nobody serves retired it, leaving Helm's five alone.
- A dashboard labelled `managed-by: SomeoneElse` survived retirement.

That run caught two bugs the unit tests missed, both fixed in 672aaa3: a `Stopped` service still counted as serving, which would have published a guaranteed-blank dashboard, and the published CR dropped its `requires-runtime` annotation, so nothing on the object said why it was there.

The reconcile loop ran from a one-shot driver, not the operator image, so the manager wiring in `cmd/main.go` builds but hasn't run in-cluster.

## Note

The auto mode is an opinionated call, particularly deferring manifests through a ConfigMap and making it the default rather than an opt-in. Happy to change any of it: flip the default to `all`, drop the deferral and keep only `grafana.dashboards.only`, or split the reconciler into its own PR if that reviews better.


internal/controller/grafana_dashboard_controller.go
```diff
@@ -0,0 +1,281 @@
...
+	live.SetNamespace(candidate.GetNamespace())
+
+	_, err = controllerutil.CreateOrUpdate(ctx, r.Client, live, func() error {
+		live.Object["spec"] = desiredSpec
+		live.SetLabels(candidate.GetLabels())
+		// The requires-runtime annotation rides along so `kubectl get -o yaml`
+		// answers why a dashboard is here, not just that it is.
+		live.SetAnnotations(candidate.GetAnnotations())
+		return nil
+	})
```
Defilan: `retire` is guarded by `managed-by`, but `publish` is not, which makes the guard launderable.

`CreateOrUpdate` adopts whatever object exists at that name, and `SetLabels(candidate.GetLabels())` replaces the map wholesale. So after one publish, a hand-authored dashboard carries `managed-by: llmkube`, and the guard in `retire` then matches it.

The sequence: someone hand-authors `GrafanaDashboard/llmkube-vllm-dashboard` with their own folder and team labels. vLLM is serving, so publish overwrites the spec and drops their labels. Their last vLLM service goes away, retire sees our label, and the dashboard is deleted. Config gone in two steps, no warning. Not a contrived case either, given #1228 notes these are currently hand-written.

This also means we replace annotations wholesale on every reconcile, which will fight Argo CD by stripping `argocd.argoproj.io/tracking-id`.

Suggest refusing to adopt an object we did not publish: `Get` first, and if it exists with a different `managed-by`, return without touching it. Worth a regression test that mirrors `TestUnmanagedDashboardSurvivesRetirement` but with a serving InferenceService, since that is the gap the current test leaves open.

charts/llmkube/templates/grafana-dashboards.yaml
```diff
@@ -0,0 +1,112 @@
...
+{{- $operator := .root.Values.grafana.dashboards.operator -}}
+apiVersion: grafana.integreatly.org/v1beta1
+kind: GrafanaDashboard
+metadata:
+  name: {{ printf "%s-%s" .fullname (trimSuffix ".json" .base) }}
+  namespace: {{ .namespace }}
+  labels:
+    {{- .labels | nindent 4 }}
+    {{- if .deferred }}
+    app.kubernetes.io/managed-by: llmkube
```
Defilan: This renders a duplicate `app.kubernetes.io/managed-by` key.

`llmkube.labels` already emits `app.kubernetes.io/managed-by: {{ .Release.Service }}` (`_helpers.tpl:40`), so every deferred manifest comes out with the key twice:

```yaml
        app.kubernetes.io/managed-by: Helm
        app.kubernetes.io/managed-by: llmkube
```

It works today only because `sigs.k8s.io/yaml.Unmarshal` is non-strict and takes the last value. `UnmarshalStrict` rejects it outright.

The risk is asymmetric and silent: if the label order in the helper ever changes, published dashboards carry `managed-by: Helm`, the retirement guard never matches, and dashboards leak forever with no error surfaced anywhere. Nothing in CI catches it, because this manifest is a string value inside a ConfigMap so kubeconform never parses it, `helm lint --strict` passes, and the helm-unittest assertion is a contains-check that the duplicate satisfies.

Either replace the inherited value rather than appending, or use a distinct key such as `dashboards.llmkube.dev/managed-by` and update `managedByLabel` in the controller to match. I mildly prefer the distinct key, since it stops overloading a well-known label, but your call.

charts/llmkube/templates/grafana-dashboards.yaml
```diff
@@ -0,0 +1,112 @@
+{{- if .Values.grafana.dashboards.enabled }}
+{{- /* Globbed once: the ConfigMap keys and the CRs' configMapRef keys are the
+       same file list by construction. */}}
+{{- $dashboards := .Files.Glob "dashboards/*.json" }}
+{{- with .Values.grafana.dashboards.only }}
+{{- $dashboards = $.Files.Glob (printf "dashboards/{%s}.json" (join "," .)) }}
+{{- if not $dashboards }}
+{{- fail (printf "grafana.dashboards.only matches no dashboard in dashboards/: %v" .) }}
```
Defilan: `only` fails the render only when every name misses, not when any name misses. I checked both:

- `only={does-not-exist}` correctly fails.
- `only={llmkube-inference,does-not-exist}` renders one dashboard and no error.

So a single typo among several names silently ships less than asked, which is the outcome the chart README says this prevents. A per-name `Files.Glob` check with `fail` would close it, and then the README sentence becomes true as written.

internal/controller/grafana_dashboard_controller.go
```diff
@@ -0,0 +1,281 @@
...
+	serving := map[string]bool{}
+	for i := range list.Items {
+		// Stopped and Suspended tear the workload down, so the series stop
+		// too and the dashboard would go blank while the object lingers.
+		// Every other phase either has pods now (Ready) or is on its way to
+		// them, and a dashboard that arrives a reconcile early is harmless.
+		switch list.Items[i].Status.Phase {
+		case inferencev1alpha1.PhaseStopped, inferencev1alpha1.PhaseSuspended:
+			continue
+		}
```
Defilan, line 188: Should `Failed` count as serving?

The phase set is `Ready`, `Failed`, `Cached`, `Downloading`, `Creating`, `Stopped`, `Suspended`, `WaitingForGPU`, and only `Stopped` and `Suspended` are skipped. So a `Failed` or crashlooping service publishes a dashboard with no pods and no series behind it, which is the same blank-panel symptom #1227 set out to remove. The comment says every other phase "either has pods now or is on its way to them", which is not quite true of `Failed`.

Genuinely arguable, since a `Failed` service may recover and flapping has its own cost, so I am asking rather than asserting. Either way, extending the table in `TestTornDownServiceDoesNotCountAsServing` to `Failed`, `Cached` and `Downloading` would pin whichever behaviour you pick. Right now the omission is what makes the question invisible.

internal/controller/grafana_dashboard_controller.go
```diff
@@ -0,0 +1,281 @@
...
+// diffing the one InferenceService that changed. The published set is a
+// function of which runtimes exist cluster-wide, so a single service's
+// deletion can be the event that retires a dashboard still wanted by its
+// siblings — there is no per-service answer to cache, and a full sweep is one
+// List against a handful of candidates.
+func (r *GrafanaDashboardReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result, error) {
+	log := logf.FromContext(ctx).WithName(dashboardControllerName)
+
+	if r.Source.Name == "" {
+		return ctrl.Result{}, nil
```
Defilan, line 98: Both disable paths orphan already-published dashboards.

`grafana.dashboards.enabled=false` drops the flag so this early return fires and nothing is retired. `operator.enabled=false` deletes the candidates ConfigMap, so the `len(candidates) == 0` return below does the same. `helm uninstall` has it too, since operator-published CRs are not in the release manifest.

Not blocking, but it is the one lifecycle case where the published set cannot converge. Even a line in the chart README pointing at `kubectl delete grafanadashboard -l app.kubernetes.io/managed-by=llmkube` would give an operator a way to clean up.

internal/controller/grafana_dashboard_controller.go
```diff
@@ -0,0 +1,281 @@
...
+	// from --grafana-dashboard-source=<namespace>/<name>. Zero value disables
+	// the reconciler: charts that publish every dashboard directly, and any
+	// non-Helm install, never create it.
+	Source types.NamespacedName
+
+	detectorOnce sync.Once
+	detector     *crdDetector
+}
+
+// +kubebuilder:rbac:groups=grafana.integreatly.org,resources=grafanadashboards,verbs=get;list;watch;create;update;patch;delete
```
Defilan: Minor: these verbs are broader than the controller uses. It does `Get` in `retire`, `Create`/`Update` via `CreateOrUpdate`, and `Delete`. It never lists or watches GrafanaDashboards (no `Owns()`, no `List`) and never patches, so `get;create;update;delete` would cover it.

For what it is worth I verified the marker, `config/rbac/role.yaml` and the chart ClusterRole are all perfectly in sync, so the plumbing is right and this is just tightening the verb set. Remember `make manifests` and `make check-helm-rbac` after changing it.

Defilan (changes requested):
Thanks for this, it is careful work and the write-up made it quick to review. Being upfront about the `cmd/main.go` wiring never having run in-cluster, and about the two bugs your own live testing caught that the unit tests missed, is exactly the kind of thing that makes a large PR reviewable.

Things I verified and want to call out as good, since they are the parts that usually go wrong here. RBAC is exactly in sync: the kubebuilder marker, `config/rbac/role.yaml` and the chart ClusterRole all agree, `make manifests` produces zero drift and `make check-helm-rbac` passes. No stale `docs/grafana/*.json` references survive anywhere in the tree, and you caught the `curl -d @` example in the README that most people would have missed. `values.schema.json` matches `values.yaml` exactly. And the List-error path in `servingRuntimes` returns before the candidate loop, so an API error requeues rather than deleting, which was the first thing I went looking for.

I also mutation-tested three of the safety paths: removing the `managed-by` guard from `retire`, removing the `Stopped`/`Suspended` skip from `servingRuntimes`, and forcing the publish predicate true. Your tests caught all three with clear failures. That is a real safety net rather than shape-checking, and it is why the two findings below stood out as the exceptions rather than the pattern.

Two things I would like fixed before this lands.

The first is that `publish` is not guarded by `managed-by` the way `retire` is, so the guard can be laundered. `CreateOrUpdate` adopts whatever object is at that name and `SetLabels` replaces the label map wholesale, so after one publish a hand-authored dashboard carries our own label and becomes eligible for deletion on the next retirement. `TestUnmanagedDashboardSurvivesRetirement` passes only because that scenario has no serving InferenceService, so `publish` never runs in it. This is the one I would call blocking, because it destroys user config silently and in two steps. It also means we stamp over GitOps tracking annotations on every reconcile.

The second is smaller but nastier to debug later: the deferred manifests render `app.kubernetes.io/managed-by` twice, because `llmkube.labels` already emits it. It resolves to `llmkube` today only because `sigs.k8s.io/yaml` is non-strict. If the label order in the helper ever changes, published dashboards silently carry `managed-by: Helm`, retirement stops matching, and they leak forever with no error. Nothing in CI can see it, since the manifest is a string inside a ConfigMap.

Smaller items inline: `only` fails the render only when every name misses, so a single typo quietly ships less than the README promises; published CRs orphan when the feature is disabled or on `helm uninstall`; and `Failed` currently counts as serving, which I think is genuinely arguable rather than wrong.

On your question about the opinionated auto-mode default: keep it. The reasoning in your description is sound, the sidecar path is unaffected, and deferring the manifest rather than a set of knobs keeps every CR field owned by `values.yaml`, which is the right layering. I would not split the reconciler out either. The thing that actually needs deciding is whether `mode: all` still works as the documented escape hatch once auto mode has published CRs that Helm does not own, since Helm should refuse to adopt objects without its ownership metadata. I could not test that without a live grafana-operator, and you already have that cluster, so could you check?

internal/controller/grafana_dashboard_controller.go
Tanguille: Fixed in d30f912. `publish` now `Get`s first and returns without touching anything whose `managed-by` is not ours, so a hand-authored dashboard is never adopted and so never becomes retirement-eligible. It also merges labels and annotations instead of replacing the maps, so `argocd.argoproj.io/tracking-id` and anything else foreign survives each reconcile. Two regression tests: `TestUnmanagedDashboardIsNotAdoptedByPublish` (serving InferenceService, asserts the label is not stamped, the hand-authored spec is intact, and the tracking annotation survives) and `TestPublishPreservesForeignMetadata`.

charts/llmkube/templates/grafana-dashboards.yaml
Tanguille: Fixed in d30f912, taking your preferred option: the key is now `dashboards.llmkube.dev/managed-by`, with `managedByLabel` in the controller updated to match. No more duplicate key, and it stops overloading a well-known label with a meaning it does not have.

charts/llmkube/templates/grafana-dashboards.yaml
Tanguille: Fixed in d30f912. Each name in `only` is now globbed on its own and fails on the first that matches nothing, before the combined glob runs. Added a helm-unittest case for exactly the `{llmkube-inference,llmkube-inferance}` shape you found, which passed before. README sentence updated to say per-name.

internal/controller/grafana_dashboard_controller.go
Tanguille: Kept `Failed` as serving, and pinned it: `TestTornDownServiceDoesNotCountAsServing` now covers `Failed`, `Cached` and `Downloading` explicitly. Reasoning is that `Failed` still intends to run pods and usually has crashlooping ones, so there are series to look at, and it is precisely when someone opens the dashboard. Retiring on `Failed` would also flap the CR out and back on every recovery. Only a deliberate teardown retires now, which is a rule that is easy to state. Fixed the comment, which did overclaim.

internal/controller/grafana_dashboard_controller.go
Tanguille: Documented in d30f912 rather than fixed — converging on disable would need a finalizer or a pre-delete hook, and neither covers `operator.enabled=false` deleting the ConfigMap out from under a running reconciler. The chart README now names all three cases and gives `kubectl delete grafanadashboard -A -l dashboards.llmkube.dev/managed-by=llmkube`, which is safe precisely because the label is only on what we published.

internal/controller/grafana_dashboard_controller.go
Tanguille: Fixed in d30f912: marker is now `get;create;update;delete`. Ran `make manifests` and hand-synced the chart ClusterRole; `make check-helm-rbac` passes (228 rules covered).

Defilan:
This is a very good round of fixes. I checked each one against the head rather than the commit messages, and they all hold.

The adoption guard is right, and you went further than I asked: `Get` first, return untouched on a foreign `managed-by`, and merge rather than replace the metadata maps, which also closes the Argo CD annotation-stripping I mentioned in passing. `TestUnmanagedDashboardIsNotAdoptedByPublish` is the test the old one should have been, since it runs with a serving InferenceService so `publish` actually executes, and it asserts all three things that matter: our label is not stamped, the hand-authored spec survives, and the tracking annotation survives.

Taking the distinct `dashboards.llmkube.dev/managed-by` key was the better of the two options I offered. It removes the duplicate render and stops overloading a well-known label with a meaning it does not have. Controller constant matches. RBAC is tightened to `get;create;update;delete` and still in three-way sync across the marker, `config/rbac/role.yaml` and the chart ClusterRole.

The `mode: all` work is the part I want to single out. You ran the live test, confirmed the failure with the exact ownership error, and then fixed it rather than settling for a documented manual step. Stamping the Helm ownership annotations onto the deferred manifests is the right call and your reasoning for why it is honest rather than a trick holds up: the operator is applying those on the release's behalf, and they are inert in `auto` because Helm never looks outside its own manifest. Re-running the procedure in both directions afterwards is what makes it convincing.

Two things I especially appreciated. You flagged the leftover `dashboards.llmkube.dev/managed-by` label on an adopted object yourself rather than leaving me to find it, and correctly scoped the README cleanup command to the case it is written for. And you checked that the new helm-unittest assertion actually bites by flipping the expected namespace and watching it fail, which is the difference between a test and a decoration.

On `Failed` counting as serving: your reasoning convinces me. A `Failed` service still intends to run pods and usually has crashlooping ones, so there are series to look at, and that is exactly when someone opens the dashboard. Retiring on `Failed` would flap the CR on every recovery. "Only a deliberate teardown retires" is a rule that is easy to state and easy to reason about, which counts for a lot in a controller. Pinning `Failed`, `Cached` and `Downloading` in the table is the right way to lock it in.

Documenting the orphan-on-disable case rather than fixing it is also the right trade. A finalizer would not cover `operator.enabled=false` deleting the ConfigMap out from under a running reconciler, and the label-scoped cleanup command is safe precisely because that label only lands on what we published.

One thing before I approve: the workflow runs on `a5e3cdc3e` are sitting at `action_required`, so nothing has actually run on this head. I will approve the runs and let the full matrix go green, then merge. Nothing further from me on the code.

Tanguille:
Addressed everything, in `2bbeaec` and `a5e3cdc`.

The publish adoption bug was the real one. `publish` now `Get`s first and returns untouched if the object doesn't carry our label, so a hand-authored dashboard is never adopted and therefore never becomes retirement-eligible. It also merges labels and annotations instead of replacing the maps, so Argo's tracking-id survives each reconcile. Two tests cover it: one with a serving InferenceService, which is the case `TestUnmanagedDashboardSurvivesRetirement` couldn't reach, and one for the metadata merge.

The duplicate label is gone via your preferred option, `dashboards.llmkube.dev/managed-by`, with `managedByLabel` updated to match. `only` is now validated per name, with a test for the `{good,typo}` shape that passed before. RBAC is down to `get;create;update;delete`, `make manifests` and `check-helm-rbac` clean.

I kept `Failed` as serving, and pinned it along with `Cached` and `Downloading`. A failed service usually has crashlooping pods and is exactly when someone opens the dashboard; retiring on `Failed` would also flap the CR out and back on every recovery. The rule is now just "a deliberate teardown retires". The comment that overclaimed is fixed.

On `mode: all`: you were right to suspect it. Tested against a live grafana-operator and it failed with `invalid ownership metadata`, since Helm won't adopt an object it didn't create. Rather than document the workaround I made it a non-issue: the ownership check wants the `managed-by: Helm` label, which the deferred manifests already inherit, plus `meta.helm.sh/release-name` and `-namespace`. Stamping those two makes `auto` -> `all` adopt in place. `all` -> `auto` already worked. Both directions verified live, no manual step in either.

One residue worth knowing: an adopted dashboard keeps the `dashboards.llmkube.dev/managed-by` label, because Helm doesn't strip fields it didn't set. It's inert, since `mode: all` renders no candidate ConfigMap and the reconciler returns before it looks at anything, but it does mean the README's cleanup selector would match a live Helm-owned dashboard, so the README now scopes when to run it. Disabling the feature still orphans published CRs; that's documented rather than fixed, since converging would need a finalizer and still wouldn't cover `operator.enabled=false` deleting the ConfigMap out from under a running reconciler.

> I rewrote this comment since my last set of comments were a bit convoluted and hard to read explaining the unexpected chronologic timeline

Defilan (approved):
Approving. CI has now run on `a5e3cdc` and is green across all 23 checks, including `CRD Sync Check` and `RBAC Sync Check`, which independently confirm the regeneration and the three-way RBAC sync you reported.

Everything I raised is addressed, and two of the fixes went further than what I asked for: merging metadata instead of replacing it, which closes the GitOps annotation-stripping as a side effect, and turning the `mode: all` ownership problem into a non-issue rather than a documented manual step.

What made this quick to review, and worth saying because it is rarer than it should be: you tested the thing I flagged as unverifiable from a diff, reported that my prediction held with the exact error text, then fixed it and re-ran both directions. You flagged the leftover label on adopted objects yourself instead of leaving it for me. And you checked the new helm-unittest assertion actually bites by flipping the expected namespace and watching it fail. That last one is the difference between a test and a decoration, and most contributors do not do it.

Your reasoning on keeping `Failed` as serving convinced me. "Only a deliberate teardown retires" is a rule that is easy to state and easy to reason about in a controller, and pinning `Failed`, `Cached` and `Downloading` in the table means the next person changing it has to do so deliberately.

Merging.

PR #1341, opened 2026-07-29
fix(foreman): accept multi-segment repo slugs (GitLab subgroups, nested Forgejo orgs) (#1299)

### LINKED ISSUE #1299: [BUG] Repo identity assumes a single-slash GitHub slug, excluding GitLab subgroups and nested Forgejo orgs
## Bug Description

Repo identity is modeled as a GitHub `owner/name` slug with exactly one slash. GitLab subgroups (`group/subgroup/project`) and nested Forgejo organizations cannot be expressed, so they are rejected before any provider code runs.

## Current Behavior

The same pattern is duplicated in two places:

```go
// pkg/foreman/agent/codehost/codehost.go:34
// pkg/foreman/agent/executor_native.go:2240
var repoSlugPattern = regexp.MustCompile(`^[A-Za-z0-9._-]+/[A-Za-z0-9._-]+$`)
```

Exactly one slash, so `group/subgroup/project` fails to match and `ResolveCloneURL` returns empty.

Splitting compounds it. `codehost.go:90` and `:105` use `strings.Cut(repoSlug, "/")`, which splits on the **first** slash. Relaxing the regex without fixing the split yields `owner=group`, `name=subgroup/project` and builds wrong API paths, which is worse than the current clean rejection because it fails at request time rather than at parse time.

## Expected Behavior

A repo identity that carries an arbitrary-depth namespace path. Provider adapters decide how to render it: GitHub joins two segments, GitLab URL-encodes the full path, Gitea and Forgejo take the last two segments as owner and repo.

Concretely: split on the last slash rather than the first, deduplicate the pattern to one definition, and keep rejecting `..`, whitespace, and absolute paths.

## Additional Context

Part of #1297, and a prerequisite for the GitLab and Gitea-family adapters. Verified against `v0.9.12`.

Defilan:
## What

Repo identity was modeled as a GitHub `owner/name` slug with exactly one slash, so GitLab subgroups and nested Forgejo orgs (`group/subgroup/project`) were rejected before any provider code ran. This routes every repo-slug parse/split/validate through a shared last-slash seam (`SplitRepoSlug`/`IsValidRepoSlug` in `pkg/foreman/agent/codehost`), so a multi-segment slug is accepted end to end with the final segment as the repository name and the prefix as the namespace.

## Why

Fixes #1299

## Contract

- `group/subgroup/project` accepted and split (final = name, prefix = namespace); `owner/name` unchanged.
- Path traversal (`..`), whitespace, empty segments, and absolute paths stay rejected (the relaxed character class is closed by an explicit `..`/empty-segment check).
- Every slug-parsing call site now uses the shared seam, closing the "accepted here, wrongly split on the first slash there" failure mode that sank a prior attempt.

## Provenance

Produced by a Foreman coder run (Laguna, wide enumeration intent), which converged to GO; the clean-room verify gate returned GATE-PASS. Independently reviewed: all repo-slug split sites were confirmed consistent on the shared seam (no remaining first-slash split). Rebased onto current `main` before opening.

## Recommended fast-follow (not this PR)

An independent review flagged that `githubissue.ParseRepo` reimplements the split correctly but does not enforce the character-class allowlist that `codehost.IsValidRepoSlug` does, so `ParseRepo("owner/name with space")` is accepted where `codehost` rejects it. This is **pre-existing** (the old `ParseRepo` never checked whitespace either) and not the multi-slash bug class this PR fixes, but `ParseRepo` should delegate to `codehost.IsValidRepoSlug`/`SplitRepoSlug` (no import cycle) as a follow-up.

## AI-assisted (band 3)

Agent-authored, gate-verified (GATE-PASS), independently reviewed, rebased and opened by me.

### DIFF
--- a/pkg/foreman/agent/codehost/codehost.go
+++ b/pkg/foreman/agent/codehost/codehost.go
@@ -28,17 +28,57 @@ import (
 	"github.com/defilantech/llmkube/pkg/foreman/agent/githubpr"
 )
 
-// repoSlugPattern matches a single "owner/name" GitHub slug. Each segment
-// is limited to git/GitHub-safe characters and exactly one slash is allowed,
-// so "..", multiple path segments, and whitespace are rejected.
-var repoSlugPattern = regexp.MustCompile(`^[A-Za-z0-9._-]+/[A-Za-z0-9._-]+$`)
+// RepoSlugPattern matches a repo slug with one or more path segments
+// separated by slashes, e.g. "owner/name" (GitHub) or
+// "group/subgroup/project" (GitLab / nested Forgejo). Each segment is
+// limited to git/GitHub-safe characters. The pattern requires at least
+// two segments (one slash) and rejects empty segments, whitespace, and
+// absolute paths. Path-traversal segments ("..") are rejected by
+// IsValidRepoSlug, which wraps this pattern.
+var RepoSlugPattern = regexp.MustCompile(`^[A-Za-z0-9._-]+(/[A-Za-z0-9._-]+)+$`)
+
+// IsValidRepoSlug reports whether slug is a valid repo slug: it matches
+// RepoSlugPattern and contains no ".." path-traversal segment. The
+// character class in RepoSlugPattern includes ".", so ".." would
+// otherwise match; this check closes that gap so multi-segment slugs
+// cannot smuggle traversal segments.
+func IsValidRepoSlug(slug string) bool {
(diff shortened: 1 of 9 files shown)


Defilan:
The `githubissue.ParseRepo` char-class fast-follow noted above is now tracked as #1343 (not blocking this PR: pre-existing, and the split itself is correct here).

PR #1363, opened 2026-07-30
fix(controller): skip model-cache PVC for pvc:// sources

### LINKED ISSUE #1353: [BUG] Unused per-InferenceService model-cache PVC provisioned for pvc:// (pre-staged) sources
### What happened

When a `Model` uses a pre-staged `pvc://` source, the operator still provisions a per-InferenceService model-cache PVC (`<isvc>-model-cache`) — even though a `pvc://` source is mounted read-only and never uses the cache. The result is an empty, unused, ISVC-owned PVC (and its PV) for every such InferenceService.

### Expected

For a `pvc://` source there is no download, so no model-cache PVC should be provisioned — the weights are already on a volume and are mounted read-only.

### Observed (perService mode)

`<isvc>-model-cache` is created and bound (empty). The serving pod correctly mounts the pre-staged `pvc://` volume (via `buildPVCStorageConfig`) and has **no** `model-downloader` init container — so the cache PVC is never touched. It just sits there.

### Root cause — the mount path is source-aware, the provisioning path is not

- **Mount** (`internal/controller/model_storage.go`): `buildModelStorageConfig` returns `buildPVCStorageConfig(model)` when `isPVCSource(model.Spec.Source)`, bypassing the cache entirely.
- **Provision** (`internal/controller/inferenceservice_controller.go`): `ensureModelCachePVC` is called whenever `effectiveModelCacheKey(model) != "" && r.ModelCachePath != ""` — with **no `isPVCSource` check**.

So a `pvc://` source that never uses the cache still gets one provisioned. The codebase already "knows" `pvc://` needs no cache (`warnIgnoredModelCacheClaim` special-cases it); the provisioning call site just doesn't apply the same knowledge.

Note: in the default **shared** mode the cache is a single namespace-wide PVC (largely harmless). The per-ISVC waste shows up in **perService** mode.

### Reproduce

1. Operator with model caching enabled (`--model-cache-path` set / chart `modelCache.enabled: true`), `perService` mode.
2. Pre-stage a model on a PVC; create a `Model` with `source: pvc://<pvc>/model.gguf` and an `InferenceService`.
3. Observe: `<isvc>-model-cache` PVC is created (empty); the pod has no `model-downloader` init container and serves from the pre-staged volume.

Verified live on llmkube **0.9.10** (AKS); the logic is unchanged on `main` (**0.9.13**).

### Why it matters

Pre-staged `pvc://` sources are the natural pattern for **air-gapped / on-prem** deployments (no HuggingFace pull) and for **shared read-only** model volumes. In those setups every InferenceService leaves behind an unused per-ISVC disk.

For context on why teams pre-stage onto a shared volume, we benchmarked cold read throughput of the storage classes you'd pre-stage onto (`dd iflag=direct`, ~2 GB, AKS westus2) — pre-staged read speed is the real cost of shared serving:

| storage class | access | read | write |
|---|---|---|---|
| managed disk (`default`) | RWO | **398 MB/s** | 82 MB/s |
| `azurefile-csi-premium` (SMB) | RWX | 154 MB/s | 377 MB/s |
| `azurefile-csi` (SMB) | RWX | 107 MB/s | 320 MB/s |
| `azureblob-default` (NFS) | RWX | 44 MB/s | 209 MB/s |

These motivate pre-staging on a shared RWX volume + `pvc://` — exactly the path that hits the unused cache PVC.

### Suggested fix

Guard the provisioning call with the same `isPVCSource` check the mount side already uses. A PR with a unit test (mutation-checked: it fails if the guard is removed) is prepared and can be opened against this issue.

### Environment

- llmkube: `0.9.10` deployed; logic confirmed on `main` / `0.9.13`
- Platform: AKS (Azure); `managed-csi` / `azurefile-csi` / `azureblob` storage classes

brandonliuli:
## What

A `pvc://` model source is pre-staged and mounted read-only (`buildModelStorageConfig` dispatches to `buildPVCStorageConfig`, never the cache), so provisioning a per-InferenceService model-cache PVC for it only leaves an empty, unused, ISVC-owned claim.

The **mount** path already special-cases `pvc://`; the **provisioning** call site did not. This adds a small `modelNeedsCachePVC` predicate that applies the same `isPVCSource` check, and uses it to gate the `ensureModelCachePVC` call.

Fixes #1353.

## Change

- `internal/controller/model_storage.go`: new `modelNeedsCachePVC(model, modelCachePath)` predicate (caching enabled + has cache key + **not** `pvc://`).
- `internal/controller/inferenceservice_controller.go`: gate the `ensureModelCachePVC` call with it (was an inline condition missing the `pvc://` check).

No behavior change for downloadable sources (http/https/s3) or shared-mode caching — only `pvc://` sources stop getting an unused per-ISVC cache PVC.

## Test

`TestModelNeedsCachePVC` (table-driven) covers http/https/s3 (cache provisioned), `pvc://` (no cache — the fix), caching-disabled, and no-cache-key. Mutation-checked: removing the `isPVCSource` guard flips the `pvc://` case and fails the test.

```
go test ./internal/controller/ -run '^TestModelNeedsCachePVC$'
```

## Checklist
- [x] Commit signed off (DCO)
- [x] `gofmt` / `go vet` clean on the changed package
- [x] Test exercises real behavior (fails when the guard is removed)
- [x] Scope matches the issue (no unrelated changes)

Assisted-by: Claude Code (drafted the fix, the test, and this description; a human reviewed the diff and ran the focused test — it passes and fails when the guard is removed).

### DIFF
--- a/internal/controller/inferenceservice_controller.go
+++ b/internal/controller/inferenceservice_controller.go
@@ -216,7 +216,7 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		desiredReplicas = 0
 	}
 
-	if effectiveModelCacheKey(model) != "" && r.ModelCachePath != "" {
+	if modelNeedsCachePVC(model, r.ModelCachePath) {
 		if err := r.ensureModelCachePVC(ctx, inferenceService); err != nil {
 			log.Error(err, "Failed to ensure model cache PVC exists", "namespace", inferenceService.Namespace)
 			return r.updateStatusWithSchedulingInfo(ctx, inferenceService, PhaseFailed, modelReady, 0, desiredReplicas, "",
--- a/internal/controller/model_cache_pvc_test.go
+++ b/internal/controller/model_cache_pvc_test.go
@@ -1,0 +1,45 @@
+package controller
+
+import (
+	"testing"
+
+	inferencev1alpha1 "github.com/defilantech/llmkube/api/v1alpha1"
+)
+
+// TestModelNeedsCachePVC guards the fix for the unused cache PVC on pvc://
+// sources: a pre-staged pvc:// model is mounted read-only (no download), so the
+// operator must not provision a per-ISVC cache PVC for it. Each model below
+// sets Status.CacheKey so effectiveModelCacheKey is non-empty -- the pvc:// case
+// is therefore false ONLY because of the isPVCSource guard. Remove that guard
+// and the pvc:// case flips to true, failing this test.
+func TestModelNeedsCachePVC(t *testing.T) {
+	withKey := func(source string) *inferencev1alpha1.Model {
+		return &inferencev1alpha1.Model{
+			Spec:   inferencev1alpha1.ModelSpec{Source: source},
+			Status: inferencev1alpha1.ModelStatus{CacheKey: "deadbeef"},
+		}
+	}
+	const cachePath = "/models"
+
+	tests := []struct {
+		name      string
+		model     *inferencev1alpha1.Model
+		cachePath string
+		want      bool
+	}{
+		{"https source is downloaded -> needs a cache", withKey("https://example.com/model.gguf"), cachePath, true},
+		{"http source is downloaded -> needs a cache", withKey("http://example.com/model.gguf"), cachePath, true},
+		{"s3 source is downloaded -> needs a cache", withKey("s3://bucket/model.gguf"), cachePath, true},
+		{"pvc:// source is pre-staged/read-only -> no cache PVC", withKey("pvc://my-models/model.gguf"), cachePath, false},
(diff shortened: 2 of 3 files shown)


Defilan (approved):
brandonliuli, welcome to LLMKube, and thank you! Great first contribution. I verified this end to end: the guard is byte-for-byte equivalent for non-`pvc://` sources (zero behavior change for http/s3/hf), the regression test genuinely bites (drop the guard and it flips to `true` and fails), and it reuses the same `isPVCSource` the mount path uses, so provision and mount can't disagree on `pvc://`. No #928/#1139 cache-PVC oscillation risk reintroduced. Scope matches #1353 exactly, DCO is signed, and the AI disclosure is in the right place (the PR body). Approving. I'll get CI authorized and squash-merge. Looking forward to more.

PR #1365, opened 2026-07-30
fix(foreman): fail loudly when a revision's prior attempt cannot be restored

### LINKED ISSUE #1364: [BUG] revision restore silently falls back to base and force-pushes away the prior attempt (reviewed work destroyed)
## Bug Description

The revision/pr-fix restore path can silently destroy a prior attempt: when `CreateBranchFromRemoteRef` fails to find (or fetch) the prior attempt's ref, it **falls back to cutting from the base branch without error** (`found=false, err=nil` — the documented #951 fallback), and `allowOverwrite` then force-pushes that base-cut over the existing branch. The reviewed work is erased, the PR body still describes it, and every rail downstream operates on the wrong belief.

## Observed blast radius (one day, one cluster)

`misospace/pr-reviewer-action#444`: a reviewed, GO'd 554-line shell→Python migration (`a8917a73`: `pr_reviewer/precheck.py` + 37 tests) was force-pushed away by a pr-fix task chasing a CI failure. Three force-pushes in one day (06:35, 09:19, 18:26 on 2026-07-30), each a fresh base-cut containing **only** that attempt's CI tweak (`+3/−1` on `ci.yaml` in the final state). Each force-push auto-dismissed the reviewer's approval, and the fix loop happily re-reviewed the empty husk.

Worse, two sibling PRs **merged in this state**: `misospace/miso-chat#726` (body: adds `resetMobileUpdateCache()` + tests; merged diff: `package-lock.json` only) and `#727` (merged diff: lockfile + `package.json`, zero fix) — both auto-closed their issues (#718, #712) as "fixed" with the work absent.

## Root cause

Two compounding decisions:

1. `CreateBranchFromRemoteRef` treats "ref not found" as a *soft* miss (fall back to base) — reasonable for a first attempt, catastrophic for a revision: the caller *knows* a prior attempt exists (that's why `reviseFromBranch` is set), so a miss is an error, not a fallback.
2. `allowOverwrite` on the same payload converts the silent fallback into silent destruction: the base-cut force-pushes over the surviving ref via force-with-lease (the lease is against the fetched remote state, so it passes).

## Proposed fix

When `payload.reviseFromBranch` is set, a restore miss must **fail the task loudly** (e.g. `FailureRestoreFailed` / verdict `INCOMPLETE` with an explicit message) instead of falling back to base — at minimum whenever `allowOverwrite` is also set. The base-branch fallback remains correct only for the non-revision path (no prior attempt claimed to exist). Also worth logging the ls-remote output on the miss: the interesting question is *why* the probe misses a ref that demonstrably exists (transient transport/auth failures currently return an error, so the miss path implies the remote genuinely lacked the ref at probe time — a race with a concurrent force-push from a parallel fix attempt is the prime suspect, and loud failure would surface that too).

Happy to submit the PR — I've been in this code for #1042/#1120/#1166.

joryirving:
## What

A `reviseFromBranch` restore miss now fails the task with an explicit error instead of silently falling back to a fresh base-cut.

## Why

Fixes #1364

The fallback was a data-loss primitive: revisions and pr-fixes always carry `allowOverwrite`, so the base-cut force-pushed over the surviving remote ref — reviewed work destroyed with no error anywhere. Observed blast radius in one day: a reviewed 554-line migration replaced by a +3/−1 CI tweak (pr-reviewer-action#444, three force-pushes, reviewer approval auto-dismissed each time), plus two sibling PRs merged as lockfile-only husks that falsely closed their issues (miso-chat#726/#727 → #718/#712 reopened).

A caller that names a prior attempt is asserting it exists. A miss means the ref was pruned or — the prime suspect given three same-day occurrences — a **concurrent fix attempt raced this one**; both are conditions to surface. The failed task lands in the existing retry/escalation ladder instead of eating the branch.

## How

- `setupTaskBranch`: the `found == false` branch returns an error (task fails as `FailureCloneFailed` with the full message: ref, suspected causes, what was refused). The base-branch path is unchanged for the `reset` strategy and non-revision tasks, where no prior attempt is claimed. Happy to add a dedicated `FailureRestoreFailed` reason instead if you'd prefer the enum churn.
- `TestSetupTaskBranch_MissingRefFallsBackToBase` (which pinned the old fallback) inverted into `TestSetupTaskBranch_MissingRefFailsLoud`; the restore/reset/default-strategy tests are untouched and green.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (agent 88.0%, controller 82.8%)
- [x] Conventional commits, DCO signed
- [x] AI assistance disclosed — authored with Claude Code
- [ ] Documentation updated — n/a (internal executor behavior; doc comments updated in-code)

### DIFF
--- a/pkg/foreman/agent/executor_native.go
+++ b/pkg/foreman/agent/executor_native.go
@@ -456,8 +456,9 @@ func (e *NativeAgentLoopExecutor) Execute(ctx context.Context, task *foremanv1al
 //     earlier commits forward on top of work merged since instead of reverting
 //     it. The executor owns this restore+rebase — prompt-driven git proved
 //     fragile under stuck-loop forcing windows. When the ref is gone from the
-//     remote (pruned, or the prior attempt never pushed) it logs and falls
-//     through to (2). Under the default "reset" strategy the restore is skipped
+//     remote the task FAILS (#1364) — falling through to (2) would rebuild
+//     from base and, with allowOverwrite, force-push over the prior attempt,
+//     silently destroying it. Under the default "reset" strategy the restore is skipped
 //     entirely: a retry or repair re-dispatch redoes the work fresh from base,
 //     so a stale prior branch can never drift from base and revert merged work.
 //  2. Upstream base fetch (#813): when the task carries a repo slug, fetch the
@@ -489,7 +490,8 @@ func setupTaskBranch(
 
 	// rebase (in-review revision): restore the prior attempt AND rebase it onto
 	// the current base, so its commits replay on top of work merged since rather
-	// than reverting it. reset (the default) skips the restore entirely and cuts
+	// than reverting it. A restore miss fails the task (#1364) — see below.
+	// reset (the default) skips the restore entirely and cuts
 	// fresh from the current base below, so a retry or repair re-dispatch can
 	// never carry a stale branch that drifts from base.
 	if strategy == foremanv1alpha1.BranchStrategyRebase &&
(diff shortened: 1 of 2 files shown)


Defilan (approved):
Jory, clean safety fix, thank you. I traced it: the `found == false` path now returns instead of silently cutting from base, it's correctly gated to rebase revisions (first attempts carry no `reviseFromBranch`, so they're untouched), and it strengthens the honest-verdict guarantee rather than threatening it, a restore miss now surfaces as a failure instead of a GO on a wrong base. The regression test bites, and CI is green. One tiny thing: the AI-assistance line is in the commit message, but our policy keeps commits attribution-free and puts the disclosure in the PR body (it's already there), so please drop that line from the commit on the squash. Approving.

PR #1368, opened 2026-07-30
refactor(chart): apply three dashboard conventions across every board

Tanguille:
First of four. The three layers stacked on this branch (operator-health row, SGLang panels, llama.cpp board) follow one at a time as this merges, so each arrives as its own reviewable diff rather than one large change.

Adds no panels and no features: it exists so those layers inherit one convention instead of each inventing their own.

## What

| Convention | Applied to |
|---|---|
| `min: 0` / `max: 1` on percentunit ratios against a hard limit | llmkube-quota ids 1, 5 |
| `noValue: "0"` where the outer PromQL is `sum()`/`count()` | model-router 1/2/4, sglang 1/6/8/14/15, vllm 1/6 |
| `palette-classic-by-name` where series vary with a template variable | 30 timeseries panels across sglang, vllm, llmkube-slo |

Plus `TestDashboardConventions` in `internal/metrics/dashboard_sync_test.go`, so a future dashboard cannot silently violate any of the three. It walks `fieldConfig.defaults` per panel, pairing each panel with its own target exprs, and scopes the rules structurally rather than with an exemption list: error budgets (`1 - (...)`) and burn-rate thresholds (`14 * (...)`) fall out of rule 1 by shape, and `sum(a)/sum(b)` ratio-of-sums falls out of rule 2 because the `sum()` match does not close at the end of the string.

## Why

`palette-classic` assigns colour by series order, so filtering `$service` repaints the survivors and a reader who learned "service X is green" is misled. Ratio panels without a `max` autoscale, so a 2% blip renders identically to saturation. `count()` over an empty vector reads "No data" on a fresh cluster instead of 0.

## Not in scope

Legend calcs (`["mean","max"]` is the existing convention and stays), fillOpacity (a `drawStyle: bars` panel legitimately needs a high fill), colours on fixed series sets, and `config/grafana`s two boards, which predate this pass.

## Evidence

- Audited all 7 boards, not only the 5 changed: `amd-gpu-observability` and `llmkube-inference` have no qualifying panels.
- Diffing both trees with min/max/noValue/color stripped is byte-identical on all 7 files, so no legend, fill, panel-id or formatting churn.
- `helm unittest charts/llmkube` 60/60, `go test ./internal/metrics/...` ok, all dashboard JSON valid.
- The test was falsified rather than assumed: removing one `noValue` makes it fail naming the exact panel, restoring makes it pass. It has already caught six real violations in the layers stacked on top, two of them in a commit that claimed to have fixed them.

Happy to split or reorder if you would rather see the test separately from the JSON changes.

## Live validation

Chart rendered from all four branches merged, applied to a production cluster (AMD R9700, SGLang v0.5.16 + three llama.cpp services) behind a suspended HelmRelease, then reverted and proved clean.

**33 of 37 panels return live data.** The four empties: two are `histogram_quantile` over an idle operator, one was the queue-wait panel awaiting its recording rules (which materialised on apply at p95 18.0s / p50 1.0s), and one was a real defect in the llama.cpp branch, now fixed there.

Revert proof: recording rules back to the 5 the release ships, dashboards back to the 2 Flux owns, zero suspended Flux objects.

### DIFF
--- a/charts/llmkube/dashboards/llmkube-quota.json
+++ b/charts/llmkube/dashboards/llmkube-quota.json
@@ -49,7 +49,9 @@
               {"color": "red", "value": 0.9}
             ]
           },
-          "unit": "percentunit"
+          "unit": "percentunit",
+          "min": 0,
+          "max": 1
         },
         "overrides": []
       },
@@ -221,7 +223,9 @@
               {"color": "red", "value": 0.9}
             ]
           },
-          "unit": "percentunit"
+          "unit": "percentunit",
+          "min": 0,
+          "max": 1
         },
         "overrides": []
       },
--- a/charts/llmkube/dashboards/llmkube-slo.json
+++ b/charts/llmkube/dashboards/llmkube-slo.json
@@ -111,7 +111,7 @@
       "datasource": {"type": "prometheus", "uid": "${DS_PROMETHEUS}"},
       "fieldConfig": {
         "defaults": {
-          "color": {"mode": "palette-classic"},
+          "color": {"mode": "palette-classic-by-name"},
           "custom": {"axisLabel": "budget remaining", "drawStyle": "line", "fillOpacity": 10, "lineWidth": 1, "pointSize": 5, "showPoints": "never", "spanNulls": true},
           "unit": "percentunit"
         },
@@ -138,7 +138,7 @@
       "datasource": {"type": "prometheus", "uid": "${DS_PROMETHEUS}"},
       "fieldConfig": {
         "defaults": {
-          "color": {"mode": "palette-classic"},
+          "color": {"mode": "palette-classic-by-name"},
           "custom": {"axisLabel": "budget remaining", "drawStyle": "line", "fillOpacity": 10, "lineWidth": 1, "pointSize": 5, "showPoints": "never", "spanNulls": true},
           "unit": "percentunit"
         },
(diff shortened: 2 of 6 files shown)


Defilan (commented):
Tanguille, this is one of the cleaner refactors I've seen, thank you. Zero PromQL touched (I diffed it: 0 changed `expr` lines), the key-change tally matches your description exactly, and the guard test is genuinely falsifiable, it already caught 41 real inconsistencies in the pre-PR dashboards. One thing before I approve: please add the AI-assistance disclosure line to the PR body per CONTRIBUTING (`Assisted-by: <tool> (what it did; what you verified)`). Add that and I'll approve and merge this one first, since #1369/#1370/#1371 all stack on it and will rebase onto it.

Tanguille:
Thanks for the compliment and done with pleasure. I added the AI-assistance disclosure to the PR body.

Defilan (approved):
The value here is that the conventions are executable rather than written down
somewhere and forgotten. All three encode a real failure I recognise:

- `percentunit` against a hard limit pinning `min:0 max:1`, so the panel reads
  against the limit instead of autoscaling to whatever the current range happens
  to be;
- `sum()`/`count()` as the outer operator setting `noValue:"0"`, so a fresh
  cluster shows a real zero instead of "No data";
- a template variable inside a label matcher forcing
  `palette-classic-by-name`, so filtering the variable doesn't repaint the
  series that remain.

Walking panel by panel rather than flattening to expr strings is the right call:
a convention about unit and colour needs to see them alongside the targets.
Scoping the glob to `charts/*/dashboards` and saying plainly in the comment that
`config/grafana`'s standalone pair predates the pass is better than quietly
globbing everything and leaving someone to discover the exclusion later.

Verified locally: `go test ./internal/metrics/ -run TestDashboard` passes on this
branch.

Approving.

Worth noting given my review on #1369: I criticised `noValue:"0"` there for
rendering a plausible-but-fake zero. That criticism only holds when the metric
is not emitted at all, which was my error on that PR, not a property of the
convention. Where the series is genuinely emitted, your rationale is the correct
one and this convention is right.

PR #1369, opened 2026-07-30
feat(dashboards): add an operator-health row to llmkube-inference

### LINKED ISSUE #1362: [FEATURE] Operator-health row in llmkube-inference (reconcile rate, reconcile p95, ready duration)
Chart 0.9.13, `llmkube-inference.json` has four panels and none of them cover the operator itself, even though the chart already scrapes it (`prometheus.serviceMonitor`) and `ControllerDown` alerts on it. When the operator is up but reconciling badly there is no board that shows it.

Series already exported and present in this cluster:

| Metric | Panel |
|---|---|
| `llmkube_reconcile_total` by `result` | reconcile rate, errors vs success |
| `llmkube_reconcile_duration_seconds_bucket` | reconcile p95 |
| `llmkube_inferenceservice_ready_duration_seconds_bucket` | time to Ready p95, the number you watch during a rollout |
| `llmkube_inferenceservice_phase` | services Ready vs total, single-stat |
| `llmkube_inferenceservice_replicas`, `llmkube_gpu_queue_depth` | fleet size and pending GPU work |

Ask: an Operator row on `llmkube-inference` with the above. It is the one board that is safe to publish unconditionally, since these series exist whenever the chart is installed at all.

Tanguille:
Fixes #1362. Stacked on #1368 — this branch contains that PR commits, so review the top commits only; the layer own change is 2 files, +191/-2.

## What

An `Operator` row on `llmkube-inference.json`: reconcile rate by result, reconcile p95, time-to-Ready p95, Services Ready (stat), fleet replicas ready vs desired, GPU queue depth.

## Why

The chart already scrapes the controller and alerts on it (`ControllerDown`), but no board showed it, so an operator that is up and reconciling badly surfaced nowhere. All six series exist whenever the chart is installed, so the row is unconditional and the board correctly stays out of `$runtimeDashboards`: these are controller-scoped metrics, not an InferenceService fact.

## Overlap with config/grafana

Six panels overlap `config/grafana/llmkube-inference-dashboard.json` (ids 52, 51, 32, 3, 43, 21), verified id-for-id. The chart board supersedes them, noted in `docs/grafana/README.md` rather than left for a reviewer to find. Happy to instead narrow the row if you would rather keep that board canonical.

## Live validation

Applied to a production cluster behind a suspended HelmRelease, then reverted and proved clean.

```
GrafanaDashboard llmkube-llmkube-inference: DashboardSynchronized=True
  "Dashboard was successfully applied to 1 instances"
```

Panel data: reconcile rate 2 series (by result), Services Ready 1+1, fleet replicas 2, GPU queue depth 1. `Reconcile duration p95` and `Time to Ready p95` return empty on an idle operator, which is correct: they are `histogram_quantile` over a 5m rate and populate during a rollout.

### DIFF
--- a/charts/llmkube/dashboards/llmkube-inference.json
+++ b/charts/llmkube/dashboards/llmkube-inference.json
@@ -104,6 +104,7 @@
         "defaults": {
           "color": {"mode": "palette-classic"},
           "custom": {"axisLabel": "reconciles/s", "drawStyle": "line", "fillOpacity": 8, "lineWidth": 1, "pointSize": 5, "showPoints": "never", "spanNulls": true},
+          "noValue": "0",
           "unit": "ops"
         },
         "overrides": [
@@ -226,6 +227,7 @@
         "defaults": {
           "color": {"mode": "palette-classic"},
           "custom": {"axisLabel": "replicas", "drawStyle": "line", "fillOpacity": 8, "lineWidth": 1, "pointSize": 5, "showPoints": "never", "spanNulls": true},
+          "noValue": "0",
           "unit": "short"
         },
         "overrides": [
--- a/charts/llmkube/dashboards/llmkube-inference.json
+++ b/charts/llmkube/dashboards/llmkube-inference.json
@@ -89,6 +89,192 @@
       ],
       "title": "Inference container restart rate (5m)",
       "type": "timeseries"
+    },
+    {
+      "collapsed": false,
+      "gridPos": {"h": 1, "w": 24, "x": 0, "y": 18},
+      "id": 102,
+      "panels": [],
+      "title": "Operator health",
+      "type": "row"
+    },
+    {
+      "datasource": {"type": "prometheus", "uid": "${DS_PROMETHEUS}"},
+      "fieldConfig": {
+        "defaults": {
+          "color": {"mode": "palette-classic"},
+          "custom": {"axisLabel": "reconciles/s", "drawStyle": "line", "fillOpacity": 8, "lineWidth": 1, "pointSize": 5, "showPoints": "never", "spanNulls": true},
+          "unit": "ops"
+        },
+        "overrides": [
+          {"matcher": {"id": "byRegexp", "options": "/success/"}, "properties": [{"id": "color", "value": {"mode": "fixed", "fixedColor": "green"}}]},
+          {"matcher": {"id": "byRegexp", "options": "/error/"}, "properties": [{"id": "color", "value": {"mode": "fixed", "fixedColor": "red"}}]}
+        ]
+      },
+      "gridPos": {"h": 8, "w": 12, "x": 0, "y": 19},
+      "id": 5,
+      "options": {
+        "legend": {"calcs": ["lastNotNull", "max"], "displayMode": "table", "placement": "bottom"},
+        "tooltip": {"mode": "multi", "sort": "desc"}
+      },
+      "targets": [
+        {
+          "datasource": {"type": "prometheus", "uid": "${DS_PROMETHEUS}"},
+          "expr": "sum by (controller, result) (rate(llmkube_reconcile_total[5m]))",
+          "legendFormat": "{{controller}} ({{result}})",
+          "refId": "A"
+        }
+      ],
+      "title": "Reconcile rate by result (5m)",
+      "type": "timeseries"
+    },
+    {
+      "datasource": {"type": "prometheus", "uid": "${DS_PROMETHEUS}"},
+      "fieldConfig": {
+        "defaults": {
+          "color": {"mode": "palette-classic-by-name"},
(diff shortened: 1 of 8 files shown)


Defilan (changes requested):
Tanguille, the operator-health row is a genuinely useful gap to close, and the four reconcile/ready/phase panels are wired correctly to real controller metrics, thank you. Two things block merge:

1. **Fleet replicas: ready vs desired** (`llmkube-inference.json:247`) queries `llmkube_inferenceservice_replicas`, which no collector, recording rule, or exporter in this repo emits (and there's no `state` label anywhere). With `noValue: "0"` it renders a plausible-but-fake `0` forever rather than "No data", so it reads as "0 replicas ready" instead of "wired to nothing". Our own `TestDashboardMetricsAreEmitted` guard should flag this. Please point it at a real series (e.g. a kube-state-metrics `kube_deployment_status_replicas` / `_available` join) or drop the panel until the operator emits replica counts.

2. **GPU queue depth by namespace** (`llmkube-inference.json:275`) reads `llmkube_gpu_queue_depth`, which is a label-less global gauge (Set to a single queue position in `status_builder.go:184`). There's no `namespace` label, so `{{namespace}}` renders empty and the "by namespace" title promises a breakdown that can't exist. Retitle to "GPU queue depth" and drop the `{{namespace}}` legend, or add a `namespace` label to the collector first.

Also: could you reconcile the "Live validation: fleet replicas 2" note? That panel's metric doesn't exist, so that number can't have come from this expr. And if any of this was AI-assisted, please add the `Assisted-by:` line to the PR body per CONTRIBUTING (DCO is clean). Happy to help wire the replica panel to a real series if that was the intent.

Tanguille:
Thanks for the careful review @Defilan. Both series are emitted by the operator on current main, so the panels are already wired to real metrics as far as I can tell.

1. Fleet replicas: 
llmkube_inferenceservice_replicas is declared in internal/metrics/metrics.go (InferenceServiceReplicas, GaugeVec with state label) and published every reconcile via PublishInferenceServiceReplicas (internal/controller/status_builder.go), setting state="ready"/state="desired" from Status.ReadyReplicas/Status.DesiredReplicas. Added in #1229, before this branch's base. TestDashboardsQueryEmittedMetrics passes against it locally, so no kube-state-metrics dependency is needed.

2. GPU queue depth: 
#1243 changed llmkube_gpu_queue_depth from a label-less gauge to a per-namespace GaugeVec (namespace label, PublishGPUQueueDepth(map[string]int32)). The status_builder.go:184 single-Set reference matches the pre-#1243 code; with the namespace label present, the {{namespace}} legend and "by namespace" title are accurate.


I adjusted the PR body: "fleet replicas 2" meant 2 series (ready + desired states), matching the current metric.

Happy to switch the replica panel to a kube-state-metrics join regardless if you'd prefer.

I also added the Assisted-by disclosure. Sorry that I missed that before.

Defilan (approved):
Tanguille, I need to retract both blockers. I was wrong on both, and you were
right on both. Apologies for the churn.

**1. `llmkube_inferenceservice_replicas` is real and does have a `state` label.**
I claimed nothing emits it and that no `state` label exists anywhere. Both false:

```go
// internal/metrics/metrics.go:74
InferenceServiceReplicas = prometheus.NewGaugeVec(
    prometheus.GaugeOpts{
        Name: "llmkube_inferenceservice_replicas",
        Help: "InferenceService replica counts. Use the state label: ready, desired.",
    },
    []string{"inferenceservice", "namespace", "state"},
)
```

It is published on every status write, `PublishInferenceServiceReplicas`
(metrics.go:288) from `status_builder.go:172`, setting `state="ready"` and
`state="desired"`. So `sum by (state) (llmkube_inferenceservice_replicas)` is
exactly right, and your "fleet replicas 2" validation note was accurate. Sorry
for questioning it.

**2. `llmkube_gpu_queue_depth` is per-namespace, not a global gauge.**

```go
// internal/metrics/metrics.go:82
GPUQueueDepth = prometheus.NewGaugeVec(
    prometheus.GaugeOpts{
        Name: "llmkube_gpu_queue_depth",
        Help: "Number of InferenceServices waiting for GPU resources, per namespace.",
    },
    []string{labelNamespace},
)
```

`PublishGPUQueueDepth` resets and repopulates it per namespace. The
"by namespace" title and the `{{namespace}}` legend are correct as written. The
`status_builder.go:184` line I cited is a function signature, not a `Set` call.
I misread it.

The part that should have stopped me is that this PR adds
`TestDashboardsQueryEmittedMetrics`, which fails when a dashboard queries
anything no collector, recording rule or allowlisted exporter emits. I appealed
to that guard in my review to support the opposite conclusion without running
it. It is green on your branch, which was the answer the whole time.

Approving. The operator-health row is wired correctly throughout, and
`TestDashboardsQueryEmittedMetrics` is a genuinely valuable addition that makes
this class of mistake, mine included, impossible to land quietly. Thanks for
building the guard along with the feature.

Nothing blocking left. Optional, entirely your call: `Services Ready vs total`
carries two targets with the same panel title, so the legend may read ambiguously
in the rendered board.

PR #1370, opened 2026-07-30
feat(dashboards): SGLang queue-wait, stage latency, token mix and HiCache L3 panels

### LINKED ISSUE #1360: [FEATURE] SGLang dashboard: queue-wait, per-stage latency, new_token_ratio and an L3 HiCache row
Chart 0.9.13, `sglang-dashboard.json`. The board covers throughput, TTFT/e2e/ITL, running vs queued, KV pool, prefix hit rate, retracted/paused/aborted, mamba pool, and the HiCache host (L2) pool. Panels that earned their place in a hand-written board this cluster replaced with the chart one, and that the chart does not have:

| Panel | Series | Why |
|---|---|---|
| Queue wait p50/p90/p99 | `sglang:queue_time_seconds_bucket` | separates "slow model" from "queued behind other requests"; running-vs-queued counts do not tell you how long anyone waited |
| Per-stage latency | `sglang:per_stage_req_latency_seconds_bucket` by `stage` | points at prefill vs decode vs grammar directly |
| New token ratio | `sglang:new_token_ratio` | the prefill/decode mix, the thing that moves when chunked prefill or a batch-size change lands |
| Prompt length distribution | `sglang:prompt_tokens_histogram_bucket` | tells you whether a latency change is workload or config |
| Request rate | `rate(sglang:num_requests_total[5m])` | throughput board has tokens/sec but no req/sec |
| HiCache L3 row | `sglang:load_back_tokens_total`, `sglang:backuped_tokens_total`, `sglang:evicted_tokens_total`, `sglang:cached_tokens_total` | with `--hicache-storage-backend`, L2-only panels hide whether the disk tier is doing anything |

The L3 row is the biggest gap. Running HiCache with the file backend on NVMe, the host pool panel shows the same picture whether L3 is serving hits or dead, so the backup and load-back token rates are what tells you the tier works. All six series are present on SGLang v0.5.16 in this cluster.

Tanguille:
Fixes #1360. Stacked on #1368 — this branch contains that PR commits; the layer own change is 3 files, +281/-9.

## What

| Row | Panels |
|---|---|
| Throughput | request rate, prompt-length distribution (heatmap), token throughput by mode |
| Latency | queue wait p95/p50, per-stage latency by stage |
| Scheduler | new token ratio |
| HiCache L3 (collapsed) | backup vs load-back rate, cached vs evicted tokens |

Queue-wait quantiles are backed by new `llmkube:inference:queue_time_seconds:{p50,p95}_5m` recording rules that mirror the TTFT and e2e siblings exactly, rather than three inline `histogram_quantile` calls recomputing one aggregation per refresh.

## Live validation

Production cluster, SGLang v0.5.16 on an AMD R9700 under real traffic. The recording rules materialise and carry signal no existing panel showed:

| Series | Before apply | After apply |
|---|---|---|
| `llmkube:inference:queue_time_seconds:p95_5m` | absent | **18.0 s** |
| `llmkube:inference:queue_time_seconds:p50_5m` | absent | **1.0 s** |

A 1 s median with an 18 s tail is queueing, not slow inference. Every other new panel returns live data: request rate 1 series, per-stage latency 3 (one per stage), new token ratio 1, prompt-length heatmap 36 buckets, L3 backup/load-back 2, L3 cached/evicted 2.

Token throughput by mode returns 3 series (`decode`, `prefill_cache`, `prefill_compute`) and is not a duplicate of the generation-vs-prompt panel: it is the same prefill total split by cache hit versus recompute. Verified on the same cluster, `prefill_cache + prefill_compute` equals `prompt_tokens_total` (45,898,768 + 5,503,461 vs 51,385,859 sampled moments apart).

The L3 row is collapsed by default, matching the mamba and HiCache host rows, so a non-HiCache install does not open onto empty panels.

## Merge-order note

This branch and the GPU-alerts work both edit `charts/llmkube/tests/prometheusrule_test.yaml` and conflict; whichever lands second needs the rule-count assertion reconciled.

Happy to be told otherwise on any of this: if you would rather these panels lived elsewhere, or not at all, say so and I will revert it.

### DIFF
--- a/charts/llmkube/dashboards/llmkube-quota.json
+++ b/charts/llmkube/dashboards/llmkube-quota.json
@@ -49,7 +49,9 @@
               {"color": "red", "value": 0.9}
             ]
           },
-          "unit": "percentunit"
+          "unit": "percentunit",
+          "min": 0,
+          "max": 1
         },
         "overrides": []
       },
@@ -221,7 +223,9 @@
               {"color": "red", "value": 0.9}
             ]
           },
-          "unit": "percentunit"
+          "unit": "percentunit",
+          "min": 0,
+          "max": 1
         },
         "overrides": []
       },
--- a/charts/llmkube/dashboards/llmkube-slo.json
+++ b/charts/llmkube/dashboards/llmkube-slo.json
@@ -111,7 +111,7 @@
       "datasource": {"type": "prometheus", "uid": "${DS_PROMETHEUS}"},
       "fieldConfig": {
         "defaults": {
-          "color": {"mode": "palette-classic"},
+          "color": {"mode": "palette-classic-by-name"},
           "custom": {"axisLabel": "budget remaining", "drawStyle": "line", "fillOpacity": 10, "lineWidth": 1, "pointSize": 5, "showPoints": "never", "spanNulls": true},
           "unit": "percentunit"
         },
@@ -138,7 +138,7 @@
       "datasource": {"type": "prometheus", "uid": "${DS_PROMETHEUS}"},
       "fieldConfig": {
         "defaults": {
-          "color": {"mode": "palette-classic"},
+          "color": {"mode": "palette-classic-by-name"},
           "custom": {"axisLabel": "budget remaining", "drawStyle": "line", "fillOpacity": 10, "lineWidth": 1, "pointSize": 5, "showPoints": "never", "spanNulls": true},
           "unit": "percentunit"
         },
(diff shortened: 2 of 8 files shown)


Defilan (approved):
Tanguille, the metric hygiene here is excellent, thank you. I pulled the current SGLang collector source and confirmed all ten metrics your panels and recording rules reference exist with the right types and labels: `queue_time_seconds` / `per_stage_req_latency_seconds` / `prompt_tokens_histogram` are histograms (correctly using `_bucket`), `new_token_ratio` is a gauge (correctly `avg`), `realtime_tokens_total` carries `mode` and `per_stage` carries `stage`. Nothing renders empty. Approving on the code.

Two small asks before I merge: please add the `Assisted-by:` disclosure line to the PR body per CONTRIBUTING, and heads-up your rule-count assertion (now 7) will collide with #1372 on `prometheusrule_test.yaml`, whoever lands second reconciles it. Optional nit: `new_token_ratio`'s `max:1` may clip, since it's a scheduler estimate rather than a strict 0-1 ratio.

PR #1371, opened 2026-07-30
feat(dashboards): add a llama.cpp runtime dashboard

### LINKED ISSUE #1359: [FEATURE] llama.cpp runtime dashboard (the runtime with no board of its own)
Chart 0.9.13 ships runtime dashboards for SGLang and vLLM. llama.cpp, the default runtime for small models and embedders, has none, and `llmkube-inference.json` is four panels (e2e latency p95/p50, restart rate) whose latency panels read `vllm:*`/`sglang:*` recording rules only, so a llamacpp-only cluster sees an almost empty board.

Three of the four InferenceServices in this cluster are llamacpp (two embedders plus a 2B chat model on iGPUs), and their `/metrics` already exposes enough for a real board today:

| Metric | Panel |
|---|---|
| `llamacpp:requests_processing`, `llamacpp:requests_deferred` | in-flight vs queued (deferred is the saturation signal for router work) |
| `llamacpp:n_busy_slots_per_decode` | slot utilization per decode |
| `llamacpp:n_decode_total` | decode rate |
| `llamacpp:prompt_tokens_total`, `llamacpp:tokens_predicted_total` | prefill and decode throughput |
| `llamacpp:kv_cache_usage_ratio`, `llamacpp:kv_cache_tokens` | KV pool |

Ask: a `llamacpp-dashboard.json` covering the above, added to `$runtimeDashboards` so auto mode publishes it only while a llamacpp InferenceService serves. Related: #1186 (llama.cpp /metrics to Prometheus) and #1146 (latency histogram, which would add a latency row later).

Tanguille:
Fixes #1359. Stacked on #1368 — this branch contains that PR commits; the layer own change is 5 files.

## What

`llamacpp-dashboard.json`, registered in `$runtimeDashboards` as `"llamacpp-dashboard.json" "llamacpp"` so auto mode publishes it only while a llamacpp InferenceService serves. That name is what `runtimeNameLabel` returns for an empty `spec.runtime` (`internal/controller/runtime.go:115`), so the existing gate handles it with no controller change.

Rows: load (in-flight vs deferred, busy slots per decode), throughput (recent 5m rate and lifetime average), decode activity, context.

## Why

SGLang and vLLM each have a runtime board. llama.cpp, had none, and `llmkube-inference.json` reads `vllm:*`/`sglang:*` recording rules only, so a llamacpp-only cluster saw an almost empty board. `requests_deferred` is the saturation signal, so it sits on the first row.

## A doc bug this surfaced, worth its own issue

An earlier revision of this branch carried a KV-cache panel, added because `docs/site/guides/metrics-driven-autoscaling.md` documents `llamacpp:kv_cache_usage_ratio` and builds an HPA on it. Live testing against `ghcr.io/ggml-org/llama.cpp:server-vulkan@31bca24` shows the server exports 11 `llamacpp:` series and that is not one of them:

```
n_busy_slots_per_decode  n_decode_total  n_tokens_max
predicted_tokens_seconds  prompt_seconds_total  prompt_tokens_seconds
prompt_tokens_total  requests_deferred  requests_processing
tokens_predicted_seconds_total  tokens_predicted_total
```

The panel and its metrics-fixture entry are removed. **The autoscaling guide still needs fixing**: as written, anyone following it gets an HPA on a series that never appears. I can open that separately.

## Live validation

All 7 panels return live data across three llama.cpp InferenceServices (two embedders, one 2B chat model), 3 series each. Applied behind a suspended HelmRelease, reverted afterwards, cluster proved clean.

Happy to be told otherwise on any of this: if you disagree with the board's shape or the fixture change, say so and I will revert it.


Defilan (commented):
Tanguille, the dashboard itself is in great shape, thank you. I verified all 11 llama.cpp metrics against the fixture and they match exactly, so no empty-panel risk, and the `kv_cache_usage_ratio` omission (with your live-verified 11-series enumeration and the self-correcting revert) is exactly right. Nice catch, too, on `metrics-driven-autoscaling.md` building an HPA on `kv_cache_usage_ratio` that this runtime doesn't expose, please do open that as its own issue so it isn't lost. One thing before I approve: add the `Assisted-by:` disclosure line to the PR body per CONTRIBUTING. I'll merge #1368 first, then this rebases down to the ~5-file change.

Tanguille:
I filed the issue #1390 and add the AI-disclosure. Thanks for the review.

Defilan (approved):
llama.cpp being the default runtime for small models and embedders while having
no board of its own was a real gap. This closes it properly.

Every series here is one llama.cpp actually exports. I checked this set against a
live server's `/metrics` earlier today while working #1373, and
`prompt_tokens_total`, `tokens_predicted_total`, `requests_processing`,
`requests_deferred`, `n_busy_slots_per_decode`, `n_decode_total`,
`prompt_seconds_total`, `tokens_predicted_seconds_total` and `n_tokens_max` are
all present. No invented series, which is the failure mode this class of PR
usually carries.

The row structure earns its keep: throughput, queue and concurrency, decode
activity, context. Pairing rate-based tokens/sec with the lifetime
`prompt_tokens_seconds` gauges in the same panel is a nice touch, since the two
disagree in exactly the situations worth looking at.

Registering it in `$runtimeDashboards` rather than adding another bespoke
template keeps the chart's existing seam intact.

Approving.

PR #1372, opened 2026-07-30
fix(chart): make GPU alerts exporter-agnostic instead of DCGM-only

### LINKED ISSUE #1356: [BUG] gpu alert group is DCGM-only and default-on, so AMD clusters get silent no-op alerting
Chart 0.9.13, `prometheus.prometheusRule.rules.gpu.enabled: true` by default.

All four rules key on NVIDIA DCGM series:

| Alert | expr |
|---|---|
| GPUHighUtilization | `DCGM_FI_DEV_GPU_UTIL > 90` |
| GPUHighTemperature | `DCGM_FI_DEV_GPU_TEMP > 85` |
| GPUMemoryPressure | `(DCGM_FI_DEV_FB_USED / DCGM_FI_DEV_FB_TOTAL) * 100 > 90` |
| GPUPowerLimit | `DCGM_FI_DEV_POWER_USAGE > 250` |

On an AMD cluster (R9700 + Radeon 660M iGPUs) `count(DCGM_FI_DEV_GPU_UTIL)` returns zero series, so all four rules are permanently no-data. They never fire and never error, which reads as healthy GPUs. The chart advertises GPU alerting that an AMD install does not get, and nothing in values.yaml says the group is NVIDIA-only.

Asks, any of:

1. Make the group vendor-aware (`rules.gpu.vendor: nvidia|amd`) with an AMD expression set, or
2. default `rules.gpu.enabled: false` and document the DCGM prerequisite, or
3. add an absent-metric guard so a missing exporter is visible instead of silent.

AMD series available in this cluster today, from a drm-exporter DaemonSet: `drm_temperature_celsius`, `drm_power_watts`, `drm_memory_used_bytes{pool="vram"}`, `drm_memory_total_bytes`, `drm_engine_utilization_ratio{engine="gpu"}`, `drm_fan_speed_rpm`, `drm_frequency_hertz`. Related to the metric contract in #1185.

Workaround in use: `rules.gpu.enabled: false`.

Tanguille:
Fixes #1356. Off `main`, independent of #1368, touches no dashboard.

All four GPU rules key on `DCGM_FI_DEV_*` while `rules.gpu.enabled` defaults to true, so a cluster without DCGM gets four rules that never fire and never error.

## Change

Each rule ORs the normalized expression for every exporter the chart knows, so the one a cluster runs is the one that evaluates. `rules.gpu.vendor` is gone: nothing needs configuring.

| Role | Expression |
|---|---|
| util | `DCGM_FI_DEV_GPU_UTIL or amdgpu_gpu_busy_percent or drm_engine_utilization_ratio{engine="gpu"} * 100` |
| temp | `DCGM_FI_DEV_GPU_TEMP or node_hwmon_temp_celsius{sensor="temp1"} * on(chip, instance) group_left() node_hwmon_chip_names{chip_name="amdgpu"} or drm_temperature_celsius{sensor="edge"}` |
| power | `DCGM_FI_DEV_POWER_USAGE or node_hwmon_power_watt * on(chip, instance) group_left() node_hwmon_chip_names{chip_name="amdgpu"} or drm_power_watts` |
| mem | `(DCGM_FI_DEV_FB_USED / DCGM_FI_DEV_FB_TOTAL) * 100 or (amdgpu_vram_used_bytes / amdgpu_vram_total_bytes) * 100 or (drm_memory_used_bytes{pool="vram"} / drm_memory_total_bytes{pool="vram"}) * 100` |

Each union is parenthesised before the comparison: `(a or b or c) > 90`, not `a or b or (c > 90)`. Rule bodies are written once from a role-keyed helper. `GPUMetricsMissing` is now absent() over all three families, ANDed.

`GPUMemoryPressure` ships off by default (`rules.gpu.memoryPressure.enabled`): on an inference host VRAM utilization reports the fraction the runtime was configured to reserve (`mem-fraction-static`, `gpu-memory-utilization`), not pressure. It is held for the process lifetime and does not fall under load, so no static threshold on it distinguishes healthy from saturated.

## Normalization, measured on a live AMD cluster

| Finding | Measurement | Fix |
|---|---|---|
| Sensor split | edge 69 C, junction 82 C, mem 79 C on one GPU; DCGM reports one edge-equivalent series | pin `sensor="edge"` / `temp1` |
| node_hwmon scope | unscoped, fired GPUHighTemperature at 89.25 C off a chipset sensor while the GPU sat at 69 C | join `node_hwmon_chip_names{chip_name="amdgpu"}`, as amd-gpu-observability.json does |
| Memory pool | drm carries `pool={vram,smem}`; unfiltered denominator worked only by implicit label matching | pin `pool="vram"` both sides |
| VRAM threshold | 99% on this host, but the number is the configured reservation, not demand | GPUMemoryPressure opt-in |

## Evidence

Rendered from this branch, evaluated against live VictoriaMetrics:

| Rule | Result |
|---|---|
| GPUHighUtilization | fires, 99% busy under load |
| GPUHighTemperature | does not fire; union resolves to GPU sensors only (39 / 69 / 44 C) |
| GPUPowerLimit | does not fire, 246 W vs 250 |
| GPUMetricsMissing | does not fire |
| GPUMemoryPressure | absent from rendered output by default |

Both exporters run on that cluster, so the temp union returns the node_hwmon and drm branches for the same GPU: they agree at 69 C.

`helm unittest` 67/67, `go test ./internal/metrics/...` ok, `helm lint` clean. The nvidia path renders byte-identical to the current release apart from `GPUMetricsMissing`.

If you would rather keep an explicit vendor knob, or disagree with any of the normalization choices, say so and I will revert it.


Defilan (changes requested):
Tanguille, excellent, well-researched change, thank you. I verified every AMD/DCGM/drm metric name against `amd-gpu-observability.json` and the `dashboard_sync_test.go` allowlist and they all check out, including the correct call to source temp/power from `node_hwmon_*` rather than a nonexistent `amdgpu_temperature_celsius`.

One blocker: `GPUMetricsMissing` (`prometheusrule.yaml:75`) never fires. PromQL `and` matches only on exactly-equal label sets, and `absent(drm_engine_utilization_ratio{engine="gpu"})` produces `{engine="gpu"}` while the other two `absent()` terms produce `{}`, so the intersection is empty even when all three exporter families are absent, the exact silent no-op #1356 exists to kill. `promtool test rules` against an empty series set confirms it. Drop the `{engine="gpu"}` matcher inside that `absent()` call (keep it in the util alert, where it's a value comparison and unaffected), and update the expected string in `prometheusrule_test.yaml:196` to match.

Two smaller things: please add the `Assisted-by:` disclosure line to the PR body per CONTRIBUTING, and consider keeping `{{ $labels.instance }}` in the descriptions so alerts still name a GPU (`instance` is present on all three families). I'll authorize the CI runs so we get a green signal.

Defilan:
CI heads-up, and it's a real consequence of the change rather than a flake. The `Test Prometheus Integration` job fails at `.github/workflows/helm-chart.yml:276`:

```
grep -q "DCGM_FI_DEV_GPU_UTIL > 95"   →   ❌ Custom GPU threshold not applied
```

That assertion pins the old DCGM-only alert shape. Your generalized `GPUHighUtilization` now renders as:

```
(DCGM_FI_DEV_GPU_UTIL or amdgpu_gpu_busy_percent or drm_engine_utilization_ratio{engine="gpu"} * 100) > 95
```

so the custom threshold **is** applied correctly, but the contiguous string `DCGM_FI_DEV_GPU_UTIL > 95` no longer exists (the `or` terms split it). `promtool` itself passes ("SUCCESS: 12 rules found"), so the rules are valid, it's purely the test string that's stale.

Since this PR intentionally changes the alert shape, please update that assertion in the same pass, e.g.:

```bash
if ! grep -qE 'DCGM_FI_DEV_GPU_UTIL.*> 95' /tmp/custom-thresholds.yaml; then
```

which confirms the DCGM term is still present and the custom threshold applied, without requiring the old contiguous form. That, plus the `GPUMetricsMissing` `absent()` fix, should get you green.

Tanguille:
Thanks for the verification of the metric names against amd-gpu-observability.json and the allowlist. All three points are addressed in 4ea9325:

GPUMetricsMissing: 
you're right, absent() outputs the vector for its own label set, so absent(drm_engine_utilization_ratio{engine="gpu"}) yields {engine="gpu"} while the other two absent() terms yield {}, and and on empty-intersection label sets is always false. Dropped the {engine="gpu"} matcher inside the absent() call; it remains in GPUHighUtilization, where it scopes a value comparison. helm unittest 67/67 and promtool check rules pass on the rendered output.

Descriptions: 
re-added {{ $labels.instance }} to the util/temp/memory/power alerts since instance is present on all three exporter families; left it out of GPUMetricsMissing as an absent() alert carries no labels.

CI: 
the Test Prometheus Integration assertion now uses grep -qE 'DCGM_FI_DEV_GPU_UTIL.*> 95', matching the OR-split expression shape while still pinning the DCGM term and the custom threshold.

I added the Assisted-by line to the PR body per CONTRIBUTING.

Defilan (approved):
The `absent()` fix is right, and the test expectation moved with it
(`prometheusrule_test.yaml:177`), so all three families now produce `{}` and the
`and` actually intersects. Thanks for keeping `$labels.instance` too, the alerts
name a GPU now.

The `llmkube.gpuMetricNames` contract is the part I like most here. One
OR-fallback per role means adding an exporter is one branch instead of another
rule body, and the comment carries the reasoning rather than just the result.
Three details worth calling out because they are the ones people get wrong:

- scoping `node_hwmon` through `node_hwmon_chip_names{chip_name="amdgpu"}`
  instead of matching every sensor on the node, with the 89 C chipset
  false-positive recorded as the reason;
- pinning `sensor="temp1"` / `sensor="edge"` rather than letting edge, junction
  and mem alias into one series;
- pinning `pool="vram"` on **both** sides of the memory ratio instead of relying
  on implicit label matching.

Defaulting `memoryPressure` off is the right call, and the rationale is correct:
on an inference host the VRAM figure reflects what the runtime reserved at
startup, not pressure, so it would alert on a healthy steady state.

Approving. `values.schema.json` moved with `values.yaml`, and the chart tests
cover the new expressions.

One optional note, no change requested: the `* on(chip, instance) group_left()
node_hwmon_chip_names{...}` join multiplies by that series' value to filter. That
is the conventional idiom and correct while the series is always 1, but it is
worth a comment if anyone later meets an exporter that emits something else
there.

Tanguille:
Optional note is covered in cda1533f: _helpers.tpl:144-146 now records that the group_left() join filters by multiplying by that series' value, correct only while it is 1, with a bool match as the fallback if an exporter emits something else.

The * on(chip, instance) group_left() join multiplies by that series' value
to filter: the conventional idiom, correct only while the value is always 1.
If an exporter ever emits something else there, switch to a bool match.

Defilan (approved):
Re-reviewed at `8f8a9649`, since my approval on 08-01 predates the last push and an approval does not carry forward across commits.

**What actually changed since then:** only a 3-line comment in `_helpers.tpl` — the `group_left()` note you mentioned. The other movement in the range (`prometheusrule.yaml` +11, `values.yaml` +36/-6) is `main` arriving through the merge; those 11 lines are the SGLang queue-time recording rules, nothing to do with this PR.

Re-verified everything on the current head rather than trusting the earlier run:

| check | result |
|---|---|
| `helm unittest charts/llmkube` | 67/67, 6 suites |
| `promtool check rules` (rendered, default values) | SUCCESS, 13 rules |
| `promtool check rules` with `memoryPressure.enabled=true` | SUCCESS, 14 rules |
| `GPUMemoryPressure` gated off by default / on when enabled | confirmed both ways |

I had to extract `.spec` before promtool would read it — pointed at the raw `PrometheusRule` it reports `SUCCESS: 0 rules found`, which is a false pass worth knowing about if anyone scripts this later.

Two things I checked specifically because they are the kind that pass tests and still be wrong:

**Operator precedence in the `or` chains.** `*` binds tighter than `or`, so `a or b or c{engine="gpu"} * 100` scales only the drm term. That is what you want — DCGM and amdgpu already report percent, drm reports a ratio — but it is load-bearing and invisible. Confirmed by parse.

**The `absent()` fix.** Dropping `{engine="gpu"}` inside `absent()` is right: with no equality matchers all three calls return the empty label set, so the `and` intersects instead of being permanently false. That closes the finding from my first pass.

## One optional note, not blocking

`GPUMetricsMissing` uses unscoped `drm_engine_utilization_ratio` while `GPUHighUtilization` scopes `{engine="gpu"}`. An exporter emitting engine labels but no `engine="gpu"` would make the metric "present" — suppressing GPUMetricsMissing — while GPUHighUtilization never fires, which is the silent-blind-spot shape this PR exists to remove.

`and on()` would let both stay scoped, since it makes the conjunction label-agnostic instead of relying on empty label sets:

```
absent(DCGM_FI_DEV_GPU_UTIL) and on() absent(amdgpu_gpu_busy_percent) and on() absent(drm_engine_utilization_ratio{engine="gpu"})
```

It parses. I am not asking for it in this PR — the current form is correct for every exporter the chart knows about, and you reasoned your way to it deliberately. Worth a follow-up issue at most.

Approving. Thanks for the thoroughness on this one, particularly chasing the `absent()` label-set semantics rather than working around them.

Issue #516, opened 2026-05-21
[FEATURE] llamacpp-router runtime for multi-model InferenceService


Defilan:
## Feature Description

Add a new `runtime: llamacpp-router` option on `InferenceService.spec`
that lets a single `llama-server` Pod host **multiple** models via
llama.cpp's built-in router mode, swapping which model is resident in
VRAM on demand. Today an InferenceService is 1:1 with a Model and 1:1
with a Pod; this would let one InferenceService advertise N models
behind one Service endpoint, with the upstream `llama-server` handling
the load / unload / dispatch cycle.

This is purely additive. The existing single-model runtime path stays
unchanged. The ModelRouter CRD also stays unchanged: it would now front
a mix of single-model and multi-model InferenceServices transparently.

## Problem Statement

As a cluster operator with a multi-site / edge fleet (NVIDIA L4 nodes
at manufacturing sites, Mac mini, similar low-VRAM hardware), I want
to host several distinct models on the same node without paying the
"one Pod per model" cost in GPU memory, so that I can offer my users
choice of model on hardware that physically can't hold them all in
VRAM simultaneously.

Concrete scenarios:

- An L4 site needs a planner model during the day, a code-completion
  model in dev hours, a summarization model at end-of-day. Three Pods
  fighting for one 24 GB L4 is wasteful. One Pod with router mode is
  the right shape.
- A 24 GB Mac mini cannot hold two ~13B models in unified memory at
  the same time but can swap between them in ~5-15s.
- Dev clusters where someone wants to A/B between two quantizations
  of the same base model without doubling the Pod footprint.

## Proposed Solution

Add `spec.runtime: llamacpp-router` as an enum value on InferenceService,
with `spec.models[]` accepting a list of Model refs (in addition to
the existing `spec.modelRef` for the single-model case). Mutual
exclusion: setting `runtime: llamacpp-router` requires `models[]`
and forbids `modelRef`; the default `runtime` value (or any other
value) requires `modelRef` and forbids `models[]`.

```yaml
apiVersion: inference.llmkube.dev/v1alpha1
kind: InferenceService
metadata:
  name: edge-multi-model
spec:
  runtime: llamacpp-router
  models:
    - name: phi4-mini
    - name: qwen3-7b-coder
    - name: gemma3-4b-summarizer
  resources:
    requests:
      nvidia.com/gpu: 1
  # one Service, one endpoint; clients pick the model via the OAI
  # `model` field on /v1/chat/completions just like today.
```

[...]

PR #1393, opened 2026-08-02
feat(modelpool): cross-pod sticky model swapping on a shared GPU slot

### LINKED ISSUE #1111: [FEATURE] ModelPool: cross-pod sticky model swapping on a shared GPU slot
## Feature Description

Add a `ModelPool` CR that lets **N independent single-model `InferenceService`s share one exclusive GPU slot** on a node, swapping which member is resident on demand, fronted by `ModelRouter`. At most one member is `Ready` at a time; the incumbent is fully drained and unloaded (VRAM freed) before the next member loads.

This is the **cross-pod** counterpart to #516's in-process `runtime: llamacpp-router`. Per the discussion on #516, the two are complementary and were split into separate issues to avoid conflation: #516 stays the single-Pod in-process swap; this issue is the cross-pod, VRAM-gated swap where **LLMKube owns the queue and swap policy in Go**. A `ModelRouter` fronts both kinds transparently.

## Problem Statement

On a constrained single node (a 64 GB unified-memory mini-PC, or a single 24 GB GPU), two large models can't co-reside: a ~32B judge/eval model and a similarly-sized coder need the *same* memory. Operators want to offer both behind one endpoint, swapping which is resident on demand, **across independent Pods** so each keeps its own image / context size / runtime (llama.cpp vs vLLM).

## Why a separate shape (not just the in-process runtime)

- **Per-member runtime divergence.** Different images, context sizes, flags, or engines per model, impossible in one shared `llama-server` process.
- **Policy lives in LLMKube, in Go.** The swap scheduler is the `ModelRouter` proxy, which already owns the per-backend queue, so sticky/anti-thrash policy needs no upstream llama.cpp change (in-process mode is bounded by ggml-org/llama.cpp#24849 / #22560, and no anti-thrash work is filed upstream).
- **Reuses existing machinery.** `Replicas: 0` -> `PhaseStopped`, the `/slots` idle contract (#1088), the `priorityClassMap` preemption path, and queue-depth metrics already exist; this composes them.

## Proposed Shape

```yaml
kind: ModelPool
metadata:
  name: heavy-slot
  namespace: lab
spec:
  nodeSelector:
    lab.example/inference: heavy
  gpu: 1
  swapPolicy: sticky        # sticky | priority
  members:
    - inferenceServiceRef: { name: judge }
    - inferenceServiceRef: { name: coder }
  default: judge            # optional: warm member on a cold pool
```

Controller invariant: **at most one member `Ready` per pool**; the rest held at `PhaseStopped`. Members are otherwise ordinary single-model `InferenceService`s.

## Swap Policy

- **Sticky (default).** The resident member stays until a *different* member is requested. No auto-restore of a default.
- **Swap only when the incumbent is idle (anti-thrash).** A cross-model request for B is *held open* (never 503'd on this path) and the slot flips **only once every slot on the incumbent A is idle**. While A is busy, A is never preempted; a same-model request arriving during the hold is served on the warm A and B keeps waiting. This is the drain-before-unload / `/slots` contract below, and it is the *entire* anti-thrash mechanism in v1: no request counting, no timers. It prevents GPU ping-pong because the flip is gated on real idleness, not on demand order.
- **`priority` mode (opt-in).** Reuse `EffectivePriority` so a designated member reclaims the slot when the preemptor goes idle.

## Drain-before-unload (reusing the `/slots` contract)

Per @joryirving's point on #516, drain-before-swap reuses the existing `/slots` idleness contract (merged in #1088) rather than a new definition of "drained":

1. Hold the request for member B at `ModelRouter`.
2. Poll incumbent A's `/slots`.
3. Keep A resident while any slot is busy **or** idleness can't be established (fail-closed).
4. Once A is fully idle, scale it to zero and wait until its pod terminates / GPU is released.
5. Only then scale B to one and forward the held request after B is `Ready`.

**Critical difference from rollouts:** a rollout force-proceeds after `idleTimeoutSeconds`; a swap must **never force-unload the incumbent on timeout**. On swap-timeout, leave A resident and return `503` + `Retry-After` for B. Force behavior must require explicit admin action, never be reachable through request traffic.

## Scope for v1

- **Kubernetes-GPU-only.** On a one-GPU node the device plugin gates co-scheduling, so the displaced member sits `Pending` until the incumbent releases the device (the "never two resident" invariant lives on the scheduler). Metal hosts have no such gating, so metal-backed members are refused in v1 (`MetalSupported=False`); metal-agent unload-before-load enforcement is a follow-up.
- Fast-interleaved-swap must be tested explicitly so a device-plugin reclaim lag can't briefly co-schedule two residents and OOM.

## Cost to Document

- Cold-swap latency on the first request after a model change (~15-45s for a ~32B Q4, ~30-90s for a ~70B Q4). Default is the activator hold; the request just takes longer inside the router budget.
- `503` + `Retry-After` is the **only** fallback (budget exceeded); `429` is wrong (implies rate limiting).
- Client timeout must outlive the swap; `httpx`'s 5s default will fail mid-swap. Document for callers.
- Per-member HPA is constrained: a pooled member scales 0->1 within its slot.

## Acceptance Criteria

- [ ] An in-flight request to A completes when a request for B arrives.
- [ ] B remains held until every A slot is idle.
- [ ] Failure to read `/slots` fails closed and leaves A resident.
- [ ] Swap timeout leaves A resident and returns a bounded error for B; it does not unload A.
- [ ] B is not started until A has terminated and relinquished the shared GPU.
- [ ] A same-model request arriving while B is held is served on the warm A and does **not** trigger a swap; B stays held until A is idle.
- [ ] Fast-interleaved swaps never produce two `Ready` members.

## PR Split

1. **PR1** - `ModelPool` CR + exclusive-slot sticky reconciler (idle-gated, fail-closed, drain-before-unload), Ginkgo coverage under `internal/controller/`.
2. **PR2** - activation + coalescing + `priority` mode on `ModelRouter`.

## Out of Scope (v2 follow-ups)

- **Guaranteed min-dwell / batched residency** (`minResidency: N`): guarantee A serves N requests before yielding the slot to B, holding A resident even across brief idle windows to amortize swap cost, and symmetrically for B. The v1 default never holds an idle slot, so under rapid A/B/A/B interleaving it can swap on every idle gap; a min-dwell floor would batch that. Deferred because it trades tail latency for throughput in a way not every workload wants, and no concrete demand exists yet (flagging it because someone will).
- **Prompt-cache save/restore across swaps** (@JJGadgets): save llama-server slot prompt cache to a PVC before unload, restore on reload via the slot API, keyed by a model hash. Attacks the swap's biggest cost (context reprocessing) for planner-with-subagents flows.
- **Metal path**: `metal-agent`-enforced unload-before-load on hosts without device-plugin GPU gating.

Refs #516

### LINKED ISSUE #1123: [BUG] `ModelRouter` activator serves a stale resident member and never self-heals after an out-of-band residency change (instant 502s until proxy restart)
## Bug Description

The router-proxy activator tracks a ModelPool's resident member in an
**in-memory** value (`poolRuntime.resident`) that is **seeded once** and
thereafter updated **only when the activator itself completes a swap**. Nothing
reconciles it against the pool's actual `status.residentMember`.

When residency changes for any reason the activator did not drive (controller
fallback to `spec.default`, OOM-kill of the resident, node drain, manual scale,
or a pod roll caused by editing the resident member's spec), the activator's
belief goes **permanently stale**. The `Acquire` fast path trusts that stale
value, so requests for the believed-resident member are dispatched to a backend
that has **no pod** — returning an instant `502 all_backends_failed` — and
**never trigger a corrective swap**. There is no self-healing: the only recovery
is restarting the router-proxy, which forces a re-seed.

Notably, a routine `kubectl apply` that edits the spec of the *currently
resident* member is enough to trigger this.

## Steps to Reproduce

1. Deploy a single-slot `ModelPool` (`swapPolicy: sticky`, `default: coder`)
   with two members, `coder` and `gemma-longctx`, fronted by a `ModelRouter`
   with `--enable-activation`. Drive a request to `gemma-longctx` so it becomes
   resident and serves 200s. The activator now holds `resident = gemma-longctx`.
2. Edit the spec of the **resident** member (`gemma-longctx`) — any field that
   rolls the pod, e.g. `contextSize` / `parallelSlots` / `extraArgs` — and
   `kubectl apply` it.
3. The controller rolls `gemma-longctx`'s Deployment; its pod is deleted. With
   the slot now empty the controller activates `spec.default` (`coder`). The pool
   converges to `residentMember: coder`, `phase: Ready`.
4. Send an OpenAI request with `"model": "gemma-longctx"` to the router.
5. Observe an immediate `502` on every such request, with no swap ever attempted,
   until the router-proxy pod is restarted.

## Expected Behavior

The `gemma-longctx` request triggers a swap back to `gemma-longctx` (drain
`coder`, load `gemma-longctx` with its new spec) and is held open per the
activator pattern — or, at minimum, fails in a way that lets the pool recover on
a subsequent request without a manual proxy restart.

## Actual Behavior

Every `gemma-longctx` request returns immediately (`latencyMs` 0–23 ms) with a
`502`, dispatched to the resident member's ClusterIP which has no backing pod
(`gemma-longctx` Deployment `replicas=0`, `ENDPOINTS: <none>`). No swap is ever
attempted. This persists across many requests indefinitely. Restarting the
router-proxy pod (which re-runs the one-shot seed) is the only fix; after
restart the activator seeds `resident = coder` and routing works again.

Root cause is in `internal/router/activator.go`: `pr.resident` is authored only
in `seed()` (runs once, gated by `if pr.seeded { return nil }`) and in the
`startSwap()` goroutine (`pr.resident = target` after a swap the activator itself
performs). No path reconciles it with `ModelPool.status.residentMember`, and the
`Acquire` fast path trusts it without verification:

```go
// Fast path: our member already owns the slot and no swap is unwinding it.
if !pr.swapping && pr.resident == member {
    pr.inflight[member]++
    ...
    return release, nil   // dispatch, no swap
}
```

A dispatch failure does **not** invalidate the belief, so there is no
self-healing.

**Suggested fixes** (any one breaks the wedge; ideally 1 + 3):
1. Reconcile `pr.resident` against `ModelPool.status.residentMember` (watch or
   periodic re-query) — the controller already publishes the truth.
2. Drop the one-shot `pr.seeded` latch / allow a forced re-seed so residency can
   be re-derived from member phase after any disruption, not just at startup.
3. Self-heal on dispatch failure: on a connection-level error to the
   believed-resident member, invalidate `pr.resident` (or re-check its phase) so
   the next request re-evaluates and can trigger a swap.
4. Verify the believed resident actually has a ready endpoint before taking the
   no-swap fast path.

## Environment

**LLMKube Version:**
```bash
llmkube version
# Fork feat/modelpool-cross-pod-swapping @ commit 55783f2
# (defilantech/LLMKube ModelPool + ModelRouter cross-pod GPU-swapping feature)
```

**Kubernetes Version:**
```bash
kubectl version
# Server Version: v1.36.2+k3s1
```

**Cluster Type:**
- [ ] GKE
- [ ] EKS
- [ ] AKS
- [ ] minikube
- [ ] kind
- [x] K3s
- [ ] Other:

**GPU (if applicable):**
- [ ] NVIDIA T4
- [ ] NVIDIA L4
- [ ] NVIDIA V100
- [ ] NVIDIA A100
- [ ] Metal (Apple Silicon)
- [ ] None (CPU only)
- [x] Other: AMD Radeon 8050S iGPU (gfx1151 / Strix Halo), Vulkan runtime — but
  the bug is GPU-agnostic; it is in the router activator, not the serving path.

## Logs

**Controller Logs:**
```bash
```
```
# Contributing rough edge seen in the same repro: the controller probes the
# member's /slots to enforce drain-before-swap, and when the member has no pod
# the probe times out and the swap is deferred (rather than treating "no pod" as
# "not busy, safe to proceed"):
INFO  member idle check failed; deferring swap  {"controller":"modelpool",
  "controllerKind":"ModelPool","ModelPool":{"name":"heavy-slot","namespace":"lab"},
  "member":"gemma-longctx","error":"failed to query /slots: Get
  \"http://gemma-longctx.lab.svc.cluster.local:8080/slots\": context deadline
  exceeded (Client.Timeout exceeded while awaiting headers)"}
```

**Router-Proxy Logs:**
```bash
kubectl logs <router-proxy-pod> --tail=50
```
```
# Every gemma-longctx request: instant 502, backend named, no swap.
{"msg":"router.dispatch","model":"gemma-longctx","status":502,
  "outcome":"all_backends_failed","latencyMs":9,"timeoutMs":900000}
{"msg":"router.dispatch","model":"gemma-longctx","status":502,
  "outcome":"all_backends_failed","latencyMs":0,"timeoutMs":900000}
```
```
# HTTP body returned to the client:
502 {"error":{"code":502,"message":"all backends failed: upstream request:
Post \"http://gemma-longctx.lab.svc.cluster.local:8080/v1/chat/completions\":
dial tcp 10.43.107.107:8080: connect: connection refused"}}
```

**Inference Pod Logs:** n/a — the resident member (`gemma-longctx`) has no pod
(`replicas=0`, `ENDPOINTS: <none>`); that is precisely the state the activator
fails to observe.

## YAML Manifests

**ModelPool:**
```yaml
kind: ModelPool
metadata:
  name: heavy-slot
  namespace: lab
spec:
  gpu: 1
  swapPolicy: sticky
  default: coder
  members:
    - inferenceServiceRef:
        name: coder
    - inferenceServiceRef:
        name: gemma-longctx
```

**ModelRouter:**
```yaml
kind: ModelRouter
metadata:
  name: heavy-router
  namespace: lab
spec:
  defaultRouteStrategy: BackendNameMatch
  proxy:
    responseHeaderTimeout: 900s
  backends:
    - name: coder
      tier: local
      inferenceServiceRef:
        name: coder
      timeout: 900s
    - name: gemma-longctx
      tier: local
      inferenceServiceRef:
        name: gemma-longctx
      timeout: 900s
```

**InferenceService (the resident member whose spec edit triggers the repro):**
```yaml
kind: InferenceService
metadata:
  name: gemma-longctx
  namespace: lab
spec:
  modelRef: gemma-longctx
  replicas: 1
  contextSize: 65536      # editing this (or parallelSlots/extraArgs) rolls the
  parallelSlots: 1        # resident pod and triggers the stale-resident wedge
  flashAttention: true
  jinja: true
  endpoint:
    port: 8080
    type: ClusterIP
```

## Additional Context

- Impact: for a single-slot ModelPool (the headline cross-pod GPU-swapping use
  case), any disruption of the resident member — including an operator-initiated
  spec change, which users do routinely — can leave the router returning hard
  502s for the other member until the proxy is manually restarted. It defeats the
  pool's self-managing promise and is not observable without correlating router
  and controller logs.
- Two lower-priority contributing behaviors observed in the same repro (could be
  split into their own issues):
  1. The controller falls back to `spec.default` when the *resident* member's pod
     rolls, instead of reloading the same member with its new spec.
  2. The `/slots` idle-probe times out when a member has no pod, deferring the
     swap rather than treating "no pod" as "not busy."

sylvainsf:
## What

Adds `ModelPool`, a CRD plus controller that lets several single-model `InferenceService`s share one exclusive GPU slot. At most one member is `Ready` at a time; the controller drains and unloads the incumbent (freeing VRAM) before the next member loads. The `ModelRouter` proxy fronts the members, activates a stopped member on demand, and enforces a sticky, anti-thrash swap policy across the slot.

Four commits:

1. `feat(modelpool): add ModelPool CRD and exclusive-slot reconciler`
2. `feat(router): activate ModelPool members with sticky idle-gated swapping`
3. `feat(modelpool): decouple swap hold budget and self-heal residency`
4. `feat(router): pin the router-proxy with spec.proxy.nodeSelector`

## Why

Fixes #1111.
Fixes #1123.
Refs #516.

`ModelPool` is tracked as its own feature issue (#1111), distinct from the in-process `runtime: llamacpp-router` shape in #516. `ModelPool` is the cross-pod shape: N independent single-model services sharing one exclusive GPU slot, so LLMKube owns the queue and swap policy in Go and the anti-thrash behavior needs no llama.cpp change. Members can also diverge in runtime (different images, context sizes, vLLM vs llama.cpp). The two shapes compose: a `ModelRouter` can front both transparently.

On a single-GPU node only one model can be resident, and there was no safe primitive to switch models on demand. `ModelPool` provides scale-from-zero on request, a VRAM-gated exclusive slot, and a sticky policy that will not thrash under mixed demand. The residency self-heal in this PR also closes the stale-resident activator bug (#1123): a member whose pod is gone no longer keeps serving stale 502s until a proxy restart.

Priority-reclaim swap mode and metal-backed members are deliberately deferred and tracked in #1394.

## How

Design notes (full write-up in `docs/proposals/516-modelpool-cross-pod-swapping.md`):

- Exclusive slot via Kubernetes device-plugin gating: a busy incumbent is never scaled down, and the displaced member stays `Pending` until the incumbent releases the device. v1 is Kubernetes-GPU-only; metal-backed members are refused with `MetalSupported=False`.
- Drain-before-unload reuses the runtime idle-probe abstraction (`IdleDetector.IdleProbe`, `/slots` for llama.cpp): an unreachable idle check fails closed (the incumbent stays resident). There is no controller-side force-unload; the only bound on a waiting request is the router hold budget, surfaced as 503 plus `Retry-After` (never 429).
- The activator runs swaps under a base context, not the caller context, so a client disconnecting mid-swap never aborts an in-progress load. Cross-model requests are held open as real connection state; same-model demand coalesces onto the warm member to avoid GPU ping-pong.
- `spec.swapBudget` (default 300s) bounds the swap hold independently of the response-header (generation) timeout, so a large cold load does not force operators to inflate the generation cap for every request. The dispatch clock starts only after the member is resident.
- Residency self-heal: the activator re-verifies its cached resident at a bounded interval and on a connection-level dispatch failure, so a member whose pod is gone does not keep serving stale 502s until a proxy restart.
- Least-privilege activation RBAC: the operator provisions a per-router ServiceAccount, Role, and RoleBinding (get/list/watch/update/patch on InferenceServices) owner-referenced to the `ModelRouter`; the proxy runs with `--enable-activation` only when the router actually fronts pools.
- `spec.proxy.nodeSelector` pins the router-proxy pod, which fixes a node-local-registry setup where the proxy scheduled onto a node that lacked its image.

## Metrics

New ModelPool observability, exported by the router-proxy at `:8081/metrics` (registered via the shared `AllCollectors` list so nothing is silently dropped):

- `llmkube_modelpool_resident{namespace,pool,member}`: which member owns the slot (1 resident, 0 not).
- `llmkube_modelpool_swaps_total{router,pool,from,to}`: completed slot swaps by direction.
- `llmkube_modelpool_swap_duration_seconds{router,pool}`: incumbent unload plus target load.
- `llmkube_modelpool_hold_duration_seconds{router,pool,member}`: how long a request was held open waiting for activation.
- `llmkube_modelpool_coalesced_total{router,pool,member}`: requests served on the warm member without a swap.
- `llmkube_modelpool_held_requests{router,pool,member}`: requests currently held open per member.

### Cluster validation (author, on the ModelPool functionality)

Environment: k3s v1.36.2+k3s1, two nodes (helpy control-plane; corsair agent, Ubuntu 24.04.4, kernel 6.17-oem). Single AMD Radeon 8050S iGPU on corsair (gfx1151 / Strix Halo), served via the operator's Vulkan image `ghcr.io/defilantech/llmkube-llama-vulkan` scheduled against `devices.es/dri-render` (generic-device-plugin, count 1), not `amd.com/gpu` or NVIDIA. Node label `lab.sylvain.dev/inference=heavy` on corsair. Models pre-staged on a local NVMe hostPath PVC (`pvc://`), no download.

Live-tested commit: `ad4d8e0` (the pre-rebase "decouple swap hold budget and self-heal residency" tip). Built controller and router-proxy with podman, imported into containerd via `k3s ctr images import`, rolled the controller-manager onto the new image, and applied the branch's ModelPool CRD (the live CRD predated `swapBudget`).

ModelPool `heavy-slot` (namespace `lab`, `gpu: 1`, `swapPolicy: sticky`, `default: coder`, `nodeSelector: {lab.sylvain.dev/inference: heavy}`, `swapBudget: 300s`) with two Vulkan members on the single GPU:
- `coder` = Qwen3.6-27B-Q6_K (about 21 GB)
- `gemma-longctx` = gemma-4-31B-it-Q4_K_M (about 18 GB)

Fronted by a `ModelRouter` (`defaultRouteStrategy: BackendNameMatch`); proxy and per-backend `responseHeaderTimeout` initially 60s, later dropped to 15s for the decoupling proof.

Results:
- swapBudget compiles into the proxy: the generated router-proxy ConfigMap showed `"swapBudget": 300000000000` (300s) per backend pool alongside the backend `"timeout"`, confirming `ModelPool.spec.swapBudget` reaches `BackendPool.SwapBudget`, decoupled from the dispatch timeout.
- Cold-pool convergence and on-demand activation: `coder` warmed to Ready, `gemma-longctx` held Stopped, pool Ready with `residentMember=coder`.
- Cross-model swap and drain-before-unload: requesting `gemma-longctx` drained `coder` to Stopped, then loaded `gemma-longctx` to Ready. First run (60s timeout): HTTP 200 in 37s, server swap_duration about 34s.
- swapBudget decoupling (the definitive test): dropping the dispatch timeout to 15s (below the about 34s swap) and triggering `coder -> gemma-longctx` returned HTTP 200 in 35.2s with swap_duration about 32s. A 32s hold succeeding under a 15s generation cap is only possible because `swapBudget` (300s) governs the hold; under the old coupling that request would 503 at 15s.
- Never-two-resident invariant: member phases never showed both Ready during or after swaps; the residency gauge showed exactly one resident.
- Metrics scraped from the proxy pod IP `:8081/metrics`: `llmkube_modelpool_swaps_total{from="coder",to="gemma-longctx"} 1`, `swap_duration_seconds` sum about 32-34s (count 1), `resident{member="coder"} 0` and `resident{member="gemma-longctx"} 1`; the full family (swaps_total, swap_duration, resident, coalesced_total, hold_duration, held_requests) was present.
- Activation RBAC: the operator provisioned the proxy's ServiceAccount/Role/RoleBinding and the proxy Deployment came up with `--enable-activation` (after the RBAC note below).
- Sticky: the pool held the resident member until a different member was requested; repeats did not swap.

Issues hit and worked around during the live run:
- Stale operator ClusterRole: the deployed `manager-role` predated the activation rules, so the proxy Deployment was never created (operator logged `serviceaccounts is forbidden ... cannot list`). Applying the branch's updated `config/rbac/role.yaml` and restarting the operator fixed it. A fresh install from this branch's chart provisions it correctly.
- Proxy placement: `ad4d8e0` had no proxy placement field, so the proxy scheduled onto helpy, which lacked the node-local image (ImagePullBackOff). Cordoning helpy rescheduled it to corsair. This is exactly the gap `spec.proxy.nodeSelector` (commit 4) closes.

Not live-exercised (verified at the code/test level only, stated to avoid overclaiming):
- `spec.proxy.nodeSelector` pinning: implemented in commit 4, verified via build/vet/tests and generated CRD/deepcopy match, but not run on the cluster (the live run used the helpy-cordon workaround instead).
- The #1123 stale-resident self-heal: present and covered by unit/controller tests, but no deliberate out-of-band residency change was staged live; the live swaps completed without stale-502 symptoms.

### Local validation (on the rebased branch)

This branch was rebased onto current `upstream/main` (0 behind), resolving the conflicts from upstream's idle-check refactor (the ModelPool drain now routes through `IdleDetector.IdleProbe` instead of a private `/slots` helper) and the deployment-update path (merged upstream's suspension-aware HPA handling with the retry-on-conflict update that absorbs concurrent ModelPool replica scaling). Manifests were regenerated with the pinned controller-gen v0.19.0 (this also dropped stale `minResidency` / `priority` drift from the ModelPool CRD) and the CRDs and ClusterRole were synced into the Helm chart.

On the rebased tip:
- `go build ./...` (linux): clean
- `go vet ./...` (linux): clean
- `golangci-lint run` v2.12.2 on the changed packages (`internal/router/...`, `internal/metrics/...`, `internal/controller/...`, `api/v1alpha1/...`): 0 issues
- `go test ./internal/router/...`: pass (about 122s), covering the activator sticky/coalesce/cancel-on-timeout paths, the swapBudget decoupling test, and the #1123 self-heal test
- The envtest controller suite (`go test ./internal/controller/...`) was not re-run on the rebase host (Windows, no envtest binaries); it passed in the author's lab on the pre-rebase code. Reviewers on Linux CI will exercise it.

## Checklist

- [x] Tests added/updated
- [x] `make lint` passes (golangci-lint v2.12.2, 0 issues on the changed packages)
- [x] Build and vet pass; router unit tests pass (envtest controller suite runs in CI, see Testing)
- [x] All commits are signed off per DCO (author identity matches sign-off)
- [x] Documentation updated (design proposal under `docs/proposals/`)


Defilan (commented):
Thanks for this, and apologies for the slow first pass. The shape is good: the RBAC is a namespaced Role rather than a ClusterRole, scoped to `inferenceservices`, bound to a per-router ServiceAccount and owner-referenced so it garbage-collects with the ModelRouter. CRD, chart CRD, and RBAC are all in sync, and the test coverage is substantial.

One thing I want to resolve before merging: **activation does not appear to be coordinated across proxy replicas.**

`Activator.mu` is a single in-process mutex, and `spec.proxy.replicas` is user-settable with nothing constraining it when a pool is attached (`replicas := int32(1)` in `newRouterDeployment` is only a default). The member writes in `activator_k8s.go` use `client.Patch(ctx, patched, client.MergeFrom(cur))` without `WithOptimisticLock`, so two proxies racing to activate different members do not conflict; the later write silently wins.

The controller-side exclusive-slot invariant still holds, so this is not corruption and two members never go resident at once. The cost is thrash: at two or more replicas, proxies receiving requests for different members each drive a swap, and every flip is a full drain plus a multi-gigabyte reload. That is the GPU ping-pong the coalescing logic exists to prevent, reappearing at replicas >= 2, with nothing in status or events to indicate it is happening.

The `Activator` doc comment says all swaps for a pool are serialized so at most one member is ever made resident at a time, which a reader would reasonably take as a cluster-wide guarantee rather than a per-process one.

Two resolutions I would be happy with:

1. **Validation.** Reject or pin `spec.proxy.replicas` to 1 while a ModelPool references the router, and state the constraint in the CRD field docs. Smallest change, and honest about what the current design guarantees.
2. **Real single-writer.** `client.MergeFromWithOptions(cur, client.MergeFromWithOptimisticLock{})` so competing activations conflict instead of silently last-write-winning, plus a lease so one replica owns swap decisions for a pool.

Which did you have in mind? If it is 1, that is fine for this PR and 2 can be a follow-up issue.

sylvainsf:
Thanks for the thorough pass. I went with option 1 (pin) for this PR and filed #1477 to track option 2 (the real cross-replica single-writer).

What changed:
- The operator now pins `spec.proxy.replicas` to 1 whenever a router has pooled backends, in `newRouterDeployment` where `hasPools` is already known. I put it there rather than in the ModelRouter validating webhook because that webhook is pure-spec-only by design (pool resolution needs a cluster lookup, which its doc deliberately keeps reconciler-side).
- It logs the override at reconcile time when a user asked for more than 1, and the constraint is documented on the CRD field (`RouterProxySpec.Replicas`), regenerated into the base CRD and the chart.
- Test: `TestNewRouterDeploymentPinsReplicasWhenPooled` (pooled pins to 1 even with `replicas: 3`; unpooled honors it).

You are right that the `Activator` doc comment overreached by implying a cluster-wide guarantee. With the pin it is now an accurate per-process guarantee for the only topology the operator allows.

#1477 covers the real fix: `MergeFromWithOptimisticLock` member writes so competing activations conflict instead of last-write-winning, plus a Lease so one replica owns swap decisions for a pool, after which the replicas=1 pin can be lifted.

Also rebased onto latest main to clear the merge conflict. Worth flagging one resolution: the router-proxy metrics server collided with the new `--metrics-bind-address` work that landed on main. I resolved to main's flag name and port (9090), but kept the handler serving `ctrlmetrics.Registry` rather than the default registry, since that is where the router and ModelPool collectors actually register (internal/metrics init) and the default-registry handler would expose none of them.

Defilan (approved):
Option 1 is the right call for this PR, and #1477 is the right place for the
rest. Approving.

I verified each of the four claims rather than taking them:

- The pin is real and in the right place: `if hasPools { replicas = 1 }` in
  `newRouterDeployment`, with the override logged at reconcile time when a user
  asked for more. Putting it there rather than in the validating webhook is
  correct; pool resolution needs a cluster lookup and that webhook is
  deliberately pure-spec.
- `TestNewRouterDeploymentPinsReplicasWhenPooled` asserts both directions,
  pooled pinning to 1 with `replicas: 3` set and unpooled still honouring 3.
  The negative half is what makes it a real test.
- The CRD field doc carries the constraint and it is regenerated into both the
  base CRD and the chart.
- The drain-path conflict was resolved by adapting `memberIdle` to main's newer
  `IdleDetector` interface rather than taking either side wholesale. That is the
  resolution I was most concerned about and it preserves the fail-closed
  behaviour.

**On the metrics registry: you found a live bug on main, and you are right.**

I checked this empirically because it is exactly the kind of thing that silently
serves an empty endpoint. Built both binaries and scraped them:

```
main: 38 metric families
PR:  149 metric families, including certwatcher_* and controller_runtime_*
```

Those are different registries. `internal/metrics` registers every collector
into `ctrlmetrics.Registry` in its `init()`, while main's handler is
`promhttp.Handler()`, which serves the Prometheus default gatherer. So on main
the `llmkube_router_*` collectors exist in a registry nothing exposes, and
`/metrics` returns Go runtime and process metrics only. #1427 is effectively
still open, and the PodMonitor in #1473 is currently scraping an endpoint with
none of the series it was added for.

Your `promhttp.HandlerFor(ctrlmetrics.Registry, ...)` is the correct fix.

One request, and it is not a change to this PR: that fix is small, independent,
and repairs something broken on main right now, while this PR is large and will
take longer to land. Would you mind pulling it out into its own PR against main?
Happy to do it myself and credit you if you would rather not carry another
branch. This one can keep the change too; the duplicate resolves trivially
whichever lands second.

One non-blocking observation for later, arguably pre-existing rather than yours.
`GenericBackend.IdleProbe` returns `errIdleUnsupported` when the
`AnnotationIdleEndpoint` annotation is absent, and `memberIdle` surfaces that as
an error, which the caller treats like any probe failure: defer with
`ReasonIdleCheckFailed`. For a generic-backend member without that annotation
that is a permanent wedge reported as a transient condition, since the probe can
never start succeeding. The six runtimes that implement `IdleDetector` are all
fine, so this only bites the annotation-less generic case. Worth distinguishing
"cannot ever be probed" from "probe failed this time" at some point, perhaps
alongside #1477.

PR #1413, opened 2026-08-05
fix(runtime): disable the llama.cpp prompt cache for embedding and rerank

### LINKED ISSUE #1406: [BUG] embedding and rerank InferenceServices inherit llama.cpp's 8 GiB prompt cache, which they can never read
## Summary

`mode: embedding` and `mode: rerank` both pass `--embedding` to llama-server. llama-server then fills its host prompt cache up to `--cache-ram` (default **8192 MiB**) and never releases it, while the cache read path is gated to completion tasks. The result is up to 8 GiB of host RAM per embedding InferenceService that can never produce a hit.

## Evidence

Measured on `qwen3-embedding` (Qwen3-Embedding-0.6B-Q8_0, `mode: embedding`, llama.cpp `b10257`, image `ghcr.io/ggml-org/llama.cpp:server-vulkan@sha256:a26d0488`), AMD Radeon 660M iGPU, values from cgroup `memory.stat` `anon`:

| distinct embedding requests (~200 tok each) | default | with `--cache-ram 0` |
|---|---|---|
| 0 (fresh pod, model loaded) | 0.08 GiB | 0.10 GiB |
| 100 | 7.95 GiB | 0.10 GiB |
| 200 | 8.06 GiB | 0.10 GiB |
| 300 | 8.06 GiB (saturated) | 0.10 GiB |

- Growth tracks the number of **distinct** prompts, not tokens per request. A single 4000-token input does not move RSS.
- Never released. Only pod restart resets it.
- Saturation matches the documented llama.cpp default exactly (`--cache-ram`, default 8192).
- Host memory, not device memory: GTT was 1.34 GiB of a 14.55 GiB aperture while `anon` was 8.13 GiB.

Impact on our cluster: a 22.4 GiB node ran at 106% of allocatable while the cache was held, versus 55-59% with it disabled. The node had been treated as chronically memory-pressured for weeks; it was this one allocation.

## Proposed change

In `internal/controller/runtime_llamacpp_args.go`, `appendModeArgs`, alongside the existing `--embedding` / `--pooling` injection:

```go
case servingModeEmbedding:
    if !hasMatchingExtraArg(extraArgs, "embedding") {
        args = append(args, "--embedding")
    }
    if !hasMatchingExtraArg(extraArgs, "pooling") {
        args = append(args, "--pooling", "last")
    }
    if !hasMatchingExtraArg(extraArgs, "cache-ram") {
        args = append(args, "--cache-ram", "0")
    }
```

The same applies to `servingModeRerank`, which also passes `--embedding`.

`hasMatchingExtraArg` preserves the existing override contract, so anyone who explicitly sets `--cache-ram` keeps control.

Alternative if a softer default is preferred: pass `--no-cache-idle-slots` instead. That disables only the ungated write path and leaves the cache usable, which matters if a future mode serves both embedding and completion tasks from one server. For pure embedding and rerank the two are equivalent in effect.

## Note on the override guard

`hasMatchingExtraArg` matches `--cache-ram` and `--cache-ram=N` but not llama.cpp's short form `-cram`. A user who sets `-cram 512` in `extraArgs` would get `--cache-ram 0` appended as well. That is pre-existing behaviour for every flag this helper guards, not new to this change, but it is worth deciding whether short forms should be recognised.

## Workaround today

```yaml
extraArgs:
  - "--cache-ram"
  - "0"
```

Confirmed by the startup log:

```
W srv init: --cache-idle-slots requires --cache-ram, disabling
```

## Environment

- LLMKube chart 0.9.13
- llama.cpp `b10257 (22dc605c4)`, Vulkan build
- Talos v1.13.8, kernel 6.18.42, Kubernetes v1.36.3
- AMD Radeon 660M (gfx1035), UMA

Reproduced only on the Vulkan build. The llama.cpp code path is backend-independent, so we expect the same on other backends, but we have measured one.

Defilan:
## What

Pass `--cache-ram 0` to llama-server for `mode: embedding` and `mode: rerank`, so those InferenceServices stop holding a prompt cache they can never read from.

## Why

Fixes #1406

llama-server fills a host prompt cache up to `--cache-ram` (default 8192 MiB) and never releases it, while the cache read path is gated to completion tasks. An embedding or rerank service therefore accumulates up to 8 GiB of host RAM per replica for no possible benefit.

The measurements in #1406 (cgroup `memory.stat` `anon`, `qwen3-embedding` on llama.cpp `b10257`):

| distinct embedding requests | default | with `--cache-ram 0` |
|---|---|---|
| 100 | 7.95 GiB | 0.10 GiB |
| 300 | 8.06 GiB (saturated) | 0.10 GiB |

Saturation matches the documented 8192 MiB default exactly.

## How

`appendModeArgs` appends `--cache-ram 0` for the embedding and rerank cases, guarded by the existing `hasMatchingExtraArg` helper so an explicit user `--cache-ram` in `extraArgs` still wins.

Applied at **both** sites that build llama-server arguments, per the two-site parity rule in `AGENTS.md`:

- `internal/controller/runtime_llamacpp_args.go` (operator)
- `pkg/agent/executor.go` (metal agent)

`chat` and empty modes are untouched. #1406 notes `--no-cache-idle-slots` as a softer alternative that leaves the cache usable; for pure embedding and rerank the two are equivalent in effect, and `--cache-ram 0` is the option the issue proposed.

## Testing

- New tests at both sites: embedding sets `--cache-ram 0`, rerank sets it, and an explicit user value in `extraArgs` is not overridden.
- Confirmed these are real regression tests: reverting only the four production hunks makes them fail at both sites (`TestAppendModeArgs/embedding_sets_--cache-ram_0`, `TestAppendModeArgs/rerank_sets_--cache-ram_0`, `TestBuildLlamaServerArgs_EmbeddingModeDisablesCacheRam`, `TestBuildLlamaServerArgs_RerankModeDisablesCacheRam`), and pass with the change in place.
- One existing subtest was updated: `does not duplicate flags already in extraArgs` now includes `--cache-ram` in the `extraArgs` it passes. Its premise is "when extraArgs already sets every flag the mode would add, append nothing", so it still asserts the same property. The new override subtest covers `--cache-ram` precedence independently.
- `go test ./pkg/agent/` passes (full package). `make lint` clean, 0 issues, CI-pinned golangci-lint v2.12.2. `gofmt` clean.
- The full suite including envtest ran in the Foreman clean-room gate, which returned GATE-PASS. I did not run the envtest suite in my own working tree.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (full suite via the clean-room gate; affected packages run directly)
- [ ] Documentation updated (no user-facing API change; behaviour is an internal default)

## AI assistance

Implemented by the Foreman agentic-coding harness (Qwopus3.6-27B-Fusion as the coder), which produced the change and the tests. I reviewed the diff, independently verified the regression tests fail before and pass after at both sites, checked that the one modified existing test preserves its original property, ran the affected packages and `make lint` locally, and rebased onto current `main`. Band-3 disclosure per the project's AI-assisted contribution policy.

### DIFF
--- a/internal/controller/runtime_llamacpp_args.go
+++ b/internal/controller/runtime_llamacpp_args.go
@@ -173,6 +173,11 @@ func appendNoWarmupArgs(args []string, noWarmup *bool) []string {
 // A reranker needs both --reranking and --embedding. Flags already in extraArgs
 // win and are not duplicated, so hand-wired manifests are left untouched; chat
 // (or empty) adds nothing.
+//
+// For embedding and rerank modes, --cache-ram 0 is appended because llama.cpp
+// fills its host prompt cache up to --cache-ram (default 8 GiB) and never
+// releases it, while the cache read path is gated to completion tasks.
+// (#1406)
 func appendModeArgs(args []string, mode string, extraArgs []string) []string {
 	switch mode {
 	case servingModeRerank:
@@ -185,12 +190,18 @@ func appendModeArgs(args []string, mode string, extraArgs []string) []string {
 		if !hasMatchingExtraArg(extraArgs, "pooling") {
 			args = append(args, "--pooling", "rank")
 		}
+		if !hasMatchingExtraArg(extraArgs, "cache-ram") {
+			args = append(args, "--cache-ram", "0")
+		}
 	case servingModeEmbedding:
 		if !hasMatchingExtraArg(extraArgs, "embedding") {
 			args = append(args, "--embedding")
 		}
 		if !hasMatchingExtraArg(extraArgs, "pooling") {
 			args = append(args, "--pooling", "last")
+		}
+		if !hasMatchingExtraArg(extraArgs, "cache-ram") {
+			args = append(args, "--cache-ram", "0")
 		}
 	}
 	return args
--- a/internal/controller/runtime_mode_test.go
+++ b/internal/controller/runtime_mode_test.go
@@ -62,7 +62,7 @@ func TestAppendModeArgs(t *testing.T) {
 		}
 	})
 	t.Run("does not duplicate flags already in extraArgs", func(t *testing.T) {
-		extra := []string{"--embedding", "--pooling", "cls"}
+		extra := []string{"--embedding", "--pooling", "cls", "--cache-ram", "0"}
 		args := appendModeArgs(nil, servingModeEmbedding, extra)
 		if len(args) != 0 {
 			t.Fatalf("expected no appended flags when extraArgs already set them, got %v", args)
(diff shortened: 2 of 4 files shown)


Tanguille (approved):
Thanks, that should resolve this.

Issue #1332, opened 2026-07-28
[BUG] Foreman coder gate cannot detect test-dilution: green gate earned by weakening checks


Defilan:
## Bug Description

In a harness-evaluation batch, two Foreman coder branches passed the deterministic gate and self-reported GO, but an independent review found each carried a critical defect the gate could not catch, because the coder made the checks pass by WEAKENING them rather than by being correct.

- #1309 (PR #1326): the revalidation script's size probe was always 0 (dead code, `curl -I -w '%{size_download}'`), but the tests only string-matched the generated command shape, so they were green on a script whose skip branch can never run.
- #1322 (PR #1325): the classifier reclassified file URLs as repos (a regression) and the serve path returned a URL vLLM rejects. The coder MOVED existing test fixtures off huggingface.co to example.com so the reclassification would not break them, deleting the only coverage of the regressed path, AND added assertions that encoded the buggy behavior as expected.

Common failure mode: the gate verifies that tests PASS, but cannot detect that the tests were diluted, fixtures relocated to dodge a change, assertions rewritten to expect wrong behavior, or shape-matching that can never fail on the real bug. A model optimizing for a green gate can satisfy it by eroding the checks.

## Steps to Reproduce

Give a coder a task whose correct fix would break some existing assertions. Observe that a submission which relocates those fixtures (or rewrites the assertions to the new, possibly-wrong behavior, or only adds string-match assertions for a logic change) passes the gate with no signal that coverage was weakened.

## Expected Behavior

The gate (or the reviewer packet) should surface when a submission weakens its own checks, so a green gate cannot be earned by diluting tests.

## Actual Behavior

GATE-PASS + coder-GO was not a sufficient signal: two of four GO branches in the batch shipped critical defects past the gate this way. Only independent review with real execution caught them.

## Proposed detection ideas (any subset)

- Coverage/mutation delta gate: run a mutation or coverage-of-changed-lines check so a diff that changes behavior without a test that FAILS on the old behavior is flagged.
- Test-diff heuristics: flag a submission that deletes or relocates existing test assertions/fixtures touching the changed code, or that only adds `ContainSubstring`/string-match assertions for a logic change, and surface it to the reviewer.
- Require at least one behavioral (executing) test for changes to generated shell or external-tool command strings, where string-matching cannot observe runtime behavior.
- Feed a test-diff summary into the reviewer prompt so the human sees what test coverage changed, not just that CI is green.

## Environment

Foreman coder gate + verification harness.

[...]

PR #1416, opened 2026-08-05
fix(foreman): detect assertion-value churn in the test-dilution gate

### LINKED ISSUE #1347: [BUG] Coder gate cannot detect assertion-value churn that encodes buggy behavior as expected (#1332 follow-up)
## Bug Description

Follow-up to #1332 (test-dilution advisory, slice 1), which deferred the "assertion-value churn" signal as the noisiest one.

The #1322 case: the coder rewrote existing assertions to encode the buggy behavior as expected (e.g. an `Expect(x).To(Equal(N))` whose expected `N` was changed to match a regression). A test that asserts the wrong behavior is still a real, biting test, so mutation/coverage execution cannot catch it; only comparing the test diff against the base can. #1332 shipped removed/relocated-coverage detection but not this.

## Expected Behavior

An advisory (consistent with the #1332 tier) that flags existing assertions whose EXPECTED right-hand-side values were rewritten in a changed test within a changed-prod package, so the reviewer can confirm the new expected value encodes correct behavior rather than a regression. This is inherently noisier than removed-coverage (a legitimate correct-behavior change also rewrites assertions), so it must be advisory-only and ideally scoped/ranked to keep volume manageable.

## Environment

Foreman coder gate. Extends `pkg/foreman/agent/test_dilution_gate.go`. Parent: #1332.

Defilan:
## What

Add an `assertionValueChurn` detector to the Foreman test-dilution gate, so an assertion whose expected value was rewritten in place is surfaced to the reviewer.

## Why

Fixes #1347

The gate had two detectors and neither covered this shape:

- `assertionErosion` only **counts** removed versus added assertion lines and fires when removed exceeds added. A one-for-one value rewrite nets to zero, so it is invisible.
- `fixtureLiteralChurn` compares URL hosts and `testdata/` paths only, never the expected value.

So a coder changing `Equal(42)` to `Equal(0)` to make a test agree with the behaviour it was meant to catch passed the gate silently. This is the signal #1332 (slice 1) deferred as the noisiest one.

## How

`assertionValueChurn` extracts matcher-plus-value tokens (`Equal(...)`, `ContainSubstring(...)`) from the removed and added sides of a changed test file in a changed production package, and reports a finding when a token appears on exactly one side. Pure additions or deletions are not churn and are left to `assertionErosion`. Output tokens are sorted so findings are deterministic.

One subtlety worth calling out for review: the gate computes its diff with `--unified=0`, so removed and added lines arrive **unpaired**. The detector therefore matches on the normalised matcher-plus-value token rather than assuming line-by-line correspondence between the two lists.

The detector is registered under `tierAdvisory` (`coder_gate.go:338`), which per its own comment "never blocks; the coder never sees it". A false positive is reviewer-visible noise, not a blocked branch.

## Testing

- New test `TestCheckTestDilution_FiresOnAssertionValueChurn_1347Shape`, mirroring the existing `..._FiresOnFixtureRelocation_1322Shape` case and its `dilutionRunner` fake.
- Confirmed it is a real regression test: unwiring only the `assertionValueChurn` call from `checkTestDilution` makes it fail; it passes with the detector wired.
- `go test ./pkg/foreman/agent/` passes (full package, 33s).
- The full suite ran in the Foreman clean-room gate: GATE-PASS.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (full suite via the clean-room gate; the affected package run directly)
- [ ] Documentation updated (internal gate behaviour, no user-facing surface)

## AI assistance

Implemented by the Foreman agentic-coding harness (Qwopus3.6-27B-Fusion as the coder). I reviewed the diff, independently verified the regression test fails when the detector is unwired and passes when it is not, ran the package suite and `make lint` locally, and rebased onto current `main`. Band-3 disclosure per the project's AI-assisted contribution policy.

Note for the record: the harness reviewer returned NO-GO on this branch, but its own structured result carried `verdictClaimed: "GO"` with `demotionReason: "issueAsk could not be verified as covering the fetched issue body"`. The reviewer approved it; an `issueAsk` verification rail demoted the verdict. I verified the branch by hand rather than relying on either signal.

### DIFF
--- a/pkg/foreman/agent/test_dilution_gate.go
+++ b/pkg/foreman/agent/test_dilution_gate.go
@@ -125,6 +125,9 @@ var (
 	urlHostRe = regexp.MustCompile(`https?://([^/\s"'` + "`" + `]+)`)
 	// testdataPathRe captures a testdata/ file path literal.
 	testdataPathRe = regexp.MustCompile(`[\w./-]*testdata/[\w./-]+`)
+	// assertionValueRe captures the matcher name and expected value in
+	// Gomega-style assertions like Equal(N) or ContainSubstring(N).
+	assertionValueRe = regexp.MustCompile(`(Equal|ContainSubstring)\(([^)]+)\)`)
 )
 
 // fixtureTokens extracts fixture-input identifiers from a set of diff lines:
@@ -168,6 +171,47 @@ func fixtureLiteralChurn(fh *fileHunks) []string {
 	sort.Strings(gone)
 	sort.Strings(appeared)
 	return []string{fmt.Sprintf("fixture input changed (removed %v, added %v)", gone, appeared)}
+}
+
+// assertionValueChurn flags an assertion whose expected value was rewritten
+// in place: a matcher+value pair (e.g. Equal(N)) present only on the removed
+// side while a different one appears only on the added side. This is the
+// #1347 shape: the coder changed the expected value to match buggy behavior.
+// Pure additions or deletions of assertions are not churn (those are caught
+// by assertionErosion). Deterministic: tokens are sorted.
+func assertionValueChurn(fh *fileHunks) []string {
+	rem := assertionValueTokens(fh.Removed)
+	add := assertionValueTokens(fh.Added)
+	var gone, appeared []string
+	for tkn := range rem {
+		if !add[tkn] {
+			gone = append(gone, tkn)
+		}
+	}
+	for tkn := range add {
+		if !rem[tkn] {
+			appeared = append(appeared, tkn)
+		}
+	}
+	if len(gone) == 0 || len(appeared) == 0 {
+		return nil
+	}
+	sort.Strings(gone)
+	sort.Strings(appeared)
+	return []string{fmt.Sprintf("assertion value changed (removed %v, added %v)", gone, appeared)}
+}
+
+// assertionValueTokens extracts matcher+value pairs from assertion lines:
+// e.g. "Equal(42)" or "ContainSubstring(boom)". The key is the full
+// matcher+value string so that Equal(42) vs Equal(0) is a churn signal.
+func assertionValueTokens(lines []string) map[string]bool {
+	set := map[string]bool{}
+	for _, l := range lines {
+		for _, m := range assertionValueRe.FindAllStringSubmatch(l, -1) {
+			set[m[0]] = true
+		}
+	}
+	return set
 }
 
 // nameStatusEntry is one file-level change from `git diff --name-status`.
@@ -300,6 +344,9 @@ func checkTestDilution(ctx context.Context, workspace string, run commandRunner)
 				file, removed-added, strings.Join(firstN(snippets, 3), "; ")))
 		}
 		for _, c := range fixtureLiteralChurn(fh) {
+			findings = append(findings, file+" "+c)
+		}
+		for _, c := range assertionValueChurn(fh) {
 			findings = append(findings, file+" "+c)
 		}
 	}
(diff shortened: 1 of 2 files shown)


joryirving (approved):
Approving. Verified the tier claim (`coder_gate.go:337-339` — advisory, never blocks), and the tests are solid: both negative shapes covered, plus the `_1347Shape` integration case, and unwiring the call to confirm it's a real regression test is the right check.

Three notes, none blocking.

**1. Assertion inversion is still invisible — probably worth its own issue.** `To(Equal(42))` → `NotTo(Equal(42))` slips past all three detectors: `assertionErosion` counts 1 removed / 1 added and nets zero (`test_dilution_gate.go:99-112`), and `assertionValueChurn` extracts the *identical* `Equal(42)` token from both sides, so `gone` and `appeared` are both empty. `BeTrue()` → `BeFalse()` isn't matched at all. Negating an assertion is at least as cheap as rewriting its expected value, so this is the same class of hole as #1347 rather than a separate concern. Not asking you to grow this PR — flagging it as the next slice.

**2. The description claims normalisation the code doesn't do.** It says the detector matches on "the normalised matcher-plus-value token", but `assertionValueTokens` stores the raw `m[0]`. Consequence: a gofmt-driven spacing change inside the parens reads as churn. Advisory-only so it's noise, not breakage — but worth reconciling the wording with the behaviour.

**3. Nested parens truncate the token — a false-negative direction.** `[^)]+` stops at the first `)`, so `Equal(f(x), 42)` tokenises to `Equal(f(x)`, and rewriting `42` to `0` yields the *same* token on both sides: silent. It holds up whenever the changed value precedes the first `)`, which covers the idiomatic shapes, but for a dilution gate the failure mode landing on false-negative rather than noise is the one that matters.

Minor, in the same vein: findings are scoped per file rather than per hunk, and `--unified=0` gives unpaired lines, so deleting an obsolete `Equal(1)` case while adding an unrelated `Equal(2)` one will fire. Fine for advisory; just means the noise floor isn't zero.

Defilan:
Thanks @joryirving , I made those changes. Appreciate the help, sir!

joryirving (approved):
Re-reviewed at `97230347`. Both findings are properly fixed — approving again.

**Finding 3 (nested parens):** balanced-paren matching with string-literal awareness is the right fix, and `matchParen` handles the cases that matter — depth tracking, quotes ignored inside literals, `skipString` returning `i` on an unterminated literal so a stray apostrophe in a comment degrades to a plain byte rather than swallowing the line. The `isIdentByte(line[start-1])` guard against `NotEqual(` is a good catch that I hadn't raised.

**Finding 2 (normalisation):** stripping whitespace outside literals while preserving it inside is exactly the right boundary, and `TestAssertionValueChurn_LiteralSpacingIsChurn` pins it so a later refactor can't quietly widen the normalisation into swallowing real signal. Description and behaviour now agree.

Verified locally at `97230347`: `go test ./pkg/foreman/agent/ -run TestAssertionValueChurn` passes, and `isIdentByte` resolves from `caller_impact_gate.go` in the same package. All 12 checks green.

**One remaining gap, same failure direction — not blocking.** The tokenizer is per-line, and the gate diffs with `--unified=0`, so an assertion split across lines never yields a token at all. Confirmed by running it against this commit:

```go
Removed: []string{`Expect(got).To(Equal(`, `	42,`, `))`}
Added:   []string{`Expect(got).To(Equal(`, `	0,`,  `))`}
// -> [] (no finding)
```

`matchParen` returns -1 on the unbalanced `Equal(` fragment, so the token is skipped. A 42 → 0 rewrite inside a wrapped assertion is invisible — the same silent-miss direction you just closed for nested parens, reached by a different route. Pre-existing rather than introduced here, and multi-line Gomega chains are common enough in table-driven tests to be worth a follow-up. Cheap mitigation if you want one: join a hunk's removed lines (and added lines) before tokenising, since the token extractor doesn't care about line boundaries.

For the record, the inversion gap from my first review is also still open at this commit — `To(Equal(42))` → `NotTo(Equal(42))` returns no finding, as expected since you're handling it separately.

PR #1420, opened 2026-08-05
fix(metrics): correct the AMD doc's metric contract and guard prose docs

### LINKED ISSUE #1419: [BUG] AMD observability contract doc names two llama.cpp metrics that are never emitted, and prose docs are unguarded
## Bug Description

`docs/observability/amd-gpu-metrics.md:51-53` names three llama.cpp metrics and asserts they are "Backend-agnostic and already emitted". Two of the three do not exist.

```
SLO metrics come from llama.cpp `/metrics` (`--metrics`), not this exporter:
`llamacpp:tokens_per_second`-family, `llamacpp:kv_cache_usage_ratio`,
`llamacpp:requests_processing`. Backend-agnostic and already emitted.
```

Checked against `internal/metrics/testdata/llamacpp-metrics.txt`, the scrape-captured fixture:

| named in the doc | actually emitted? |
|---|---|
| `llamacpp:tokens_per_second` | **no** — the real names are `llamacpp:predicted_tokens_seconds` and `llamacpp:prompt_tokens_seconds` |
| `llamacpp:kv_cache_usage_ratio` | **no** — llama.cpp emits no kv-cache series at all |
| `llamacpp:requests_processing` | yes |

This is the same defect class as #1373 (fixed by #1386), which corrected `llamacpp:kv_cache_usage_ratio` in the autoscaling guide. That fix corrected the guide it was filed against; this occurrence in the observability contract doc was not part of it and survived.

## Why it survived

`TestDashboardsQueryEmittedMetrics` (`internal/metrics/dashboard_sync_test.go:470`) is the guard that would normally catch this. It only globs dashboards:

```go
var dashboardGlobs = []string{"../../*/grafana/*.json", "../../charts/*/dashboards/*.json"}
```

Prose documentation is unguarded, so a doc can assert a metric exists indefinitely. A document that tells an operator to build an alert on `llamacpp:kv_cache_usage_ratio` produces a rule that silently never fires, which is precisely the failure #1386 was filed to stop.

## Expected Behavior

1. The contract section names only metrics that are actually emitted.
2. A guard fails if that ever drifts again.

## Proposed fix

**Part 1 — correct the text.** Use the fixture-verified names. `llamacpp:predicted_tokens_seconds` and `llamacpp:prompt_tokens_seconds` are the throughput pair; `llamacpp:requests_processing` stays. Drop `llamacpp:kv_cache_usage_ratio` entirely rather than substituting a replacement, since no kv-cache series exists. If a saturation signal is wanted, `llamacpp:requests_deferred` is the one #1386 settled on, and it is in the fixture.

**Part 2 — extend the guard, scoped.** Do **not** scan the whole file. `## Known gaps (deliberate)` deliberately names things that are not exported (`mclk`, `pp_dpm_mclk`), and a blanket scan would flag them as false positives.

Instead, delimit the guarded region with explicit markers:

```markdown
<!-- metrics-contract:begin -->
... only metric names inside here are guarded ...
<!-- metrics-contract:end -->
```

Then extend `dashboard_sync_test.go` with a sibling test that, for each doc containing the markers, extracts backticked metric names inside the marked region and asserts each is in the same `known` set the dashboard guard already builds (`declaredNames` + `chartRecordingRules` + `runtimeNames`).

## Acceptance

- `go test ./internal/metrics/` passes.
- The new test **fails** if `llamacpp:kv_cache_usage_ratio` is reinstated inside the marked region.
- The new test **stays silent** on the `## Known gaps (deliberate)` section, which sits outside the markers.
- No dashboard JSON changes; this is doc plus test only.

## Environment

Not applicable: found by inspection against `main` at `929f0a87`. No cluster or GPU is needed to reproduce; the fixture and the doc disagree in-tree.

## Additional Context

Refs #1373, #1386, #1186. The existing guard's doc comment notes it "checks that a name can be emitted, not that it ever is", which is the right bar here too.

Defilan:
## What

Correct two non-existent llama.cpp metric names in the AMD observability contract, and add a guard so prose documentation cannot assert a metric that nothing emits.

## Why

Fixes #1419

`docs/observability/amd-gpu-metrics.md` named three metrics and asserted they were "already emitted". Checked against `internal/metrics/testdata/llamacpp-metrics.txt`, the scrape-captured fixture, only one existed:

| named in the doc | emitted? |
|---|---|
| `llamacpp:tokens_per_second` | no — real names are `predicted_tokens_seconds` / `prompt_tokens_seconds` |
| `llamacpp:kv_cache_usage_ratio` | no — llama.cpp emits no kv-cache series at all |
| `llamacpp:requests_processing` | yes |

An operator following that contract builds an alert that silently never fires. #1386 fixed exactly this in the autoscaling guide; this second occurrence in the observability doc was not part of that change and survived.

## How

The kv-cache entry is dropped rather than given a replacement, because no such series exists.

`TestDocContractMetrics` asserts every backticked metric name inside a `<!-- metrics-contract:begin/end -->` region is one the repo can emit, reusing the same `known` set the dashboard guard builds (`declaredNames` + `chartRecordingRules` + `runtimeNames`).

The region is **delimited rather than whole-file** on purpose: the same document's `## Known gaps (deliberate)` section names metrics that intentionally do not exist (`mclk`, `pp_dpm_mclk`), and a whole-file scan would flag them as false positives.

## Reviewer note on the file walk

The original implementation used `filepath.Glob("../../docs/**/*.md")`. Go's `filepath.Glob` has **no `**` operator**, so that pattern silently means `docs/*/*.md`. Measured on this tree: it matches 26 files at one level and misses 35 nested deeper, including `docs/site/guides/metrics-driven-autoscaling.md` — the very guide where this defect class was fixed in #1386. A guard whose purpose is preventing recurrence would have skipped the file the recurrence happened in.

Replaced with `filepath.WalkDir`. This is my change on top of the generated branch, not the coder's, and is called out in the AI-assistance section below.

## Testing

- `go test ./internal/metrics/` passes.
- Verified the guard genuinely fires: reinstating `llamacpp:kv_cache_usage_ratio` inside the marked region fails with `names "llamacpp:kv_cache_usage_ratio" inside metrics-contract region, which no registered collector, chart recording rule or allowlisted exporter emits`.
- Verified it stays silent on `## Known gaps (deliberate)`, which sits outside the markers, and on all other docs now reachable by the walk.
- The full suite ran in the Foreman clean-room gate: GATE-PASS.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (full suite via the clean-room gate; `./internal/metrics/` run directly)
- [x] Documentation updated (this change corrects the doc)

## AI assistance

The doc correction, the marker design and `TestDocContractMetrics` were produced by the Foreman agentic-coding harness (Qwopus3.6-27B-Fusion as the coder), working from an issue that specified the fixture-verified names and the delimited-region requirement. I reviewed the diff, verified the guard fires and stays silent on the deliberate-gaps section, and **replaced the file walk after finding the `**` glob silently skipped 35 nested docs**. Band-3 disclosure per the project's AI-assisted contribution policy.

### DIFF
--- a/docs/observability/amd-gpu-metrics.md
+++ b/docs/observability/amd-gpu-metrics.md
@@ -48,9 +48,11 @@ versions of the same readings.
 
 ## Inference signals (slice #1186)
 
+<!-- metrics-contract:begin -->
 SLO metrics come from llama.cpp `/metrics` (`--metrics`), not this exporter:
-`llamacpp:tokens_per_second`-family, `llamacpp:kv_cache_usage_ratio`,
+`llamacpp:predicted_tokens_seconds`, `llamacpp:prompt_tokens_seconds`,
 `llamacpp:requests_processing`. Backend-agnostic and already emitted.
+<!-- metrics-contract:end -->
 
 ## Known gaps (deliberate)
 
--- a/internal/metrics/dashboard_sync_test.go
+++ b/internal/metrics/dashboard_sync_test.go
@@ -18,6 +18,7 @@ package metrics
 
 import (
 	"encoding/json"
+	"io/fs"
 	"maps"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


Tanguille (approved):
Nice spot, I didn't notice this before. Looks good to me.

Issue #1309, opened 2026-07-27
[BUG] Model cache never revalidates: every restart re-downloads full weights and truncates the cached copy first


Tanguille:
## Bug Description

The generated `model-downloader` init container re-downloads the entire weight set on every pod
restart even though the model cache PVC is intact. Three defects stack:

1. `revalidated` is logged on any curl exit 0, including a full `200` with the whole body, so the
   log gives no signal that a 17.4 GB transfer just happened.
2. The conditional request does not survive the HuggingFace redirect, so LFS-backed weights can
   never return `304`.
3. `-o "$dest"` truncates the valid cached file before curl knows whether the body is needed.

Defect 3 is also the mechanism behind cache corruption when a pod is replaced mid-download.

## Steps to Reproduce

1. Deploy a `Model` with an LFS-backed HF source and a persistent cache:
   `hf://mattbucci/Qwen3.6-27B-AWQ@f541031dfc1bf5da3f198b52b9bf44cbd5ceee49` (17.4 GB)
2. Let the `InferenceService` come up fully so the cache PVC holds a complete artifact set.
3. `kubectl delete pod <inference-pod>` to force a restart against the warm cache.
4. Watch `kubectl logs <new-pod> -c model-downloader` and time the init container.

## Expected Behavior

With a complete artifact already on the cache PVC, the downloader revalidates and skips the
transfer, so restart is bounded by weight loading rather than by network transfer.

## Actual Behavior

All 9 artifacts log `revalidated`, and the init container still takes 14m21s:

| init container | started | finished | duration |
|---|---|---|---|
| `model-downloader` | 18:08:38Z | 18:22:59Z | 14m21s |

14m21s at the observed 11 to 20 MB/s is the full weight set, not a set of conditional GETs.

The generated command is:

```sh
if curl -fsSL --etag-compare "$etag" --etag-save "$etag" -o "$dest" "$url"; then
    echo "Model artifact $rel revalidated"
elif [ -f "$dest" ]; then
    echo "Revalidation unreachable for $rel; kept cached copy"
else
    echo "ERROR: model artifact $rel missing and revalidation failed"; exit 1
fi
```

`echo ... revalidated` fires on curl exit 0, which a full `200` also satisfies. It does not
indicate a `304`.

**Why the conditional never applies to weights.** `huggingface.co/<repo>/resolve/<rev>/<file>`
returns `302` to a signed CDN URL on a different host, and `If-None-Match` is not reapplied
[...]

PR #1421, opened 2026-08-05
feat(foreman): advisory gate for command-string changes tested only by shape

### LINKED ISSUE #1346: [BUG] Coder gate cannot detect string-match-only tests for command-string/shell changes (#1332 follow-up)
## Bug Description

Follow-up to #1332 (test-dilution advisory, slice 1). The coder-gate test-dilution advisory ships the "removed/relocated coverage" signal but explicitly deferred detection of the #1309 mode: a change to command-string / shell-command generation (or branching logic) whose ONLY added tests are substring / shape matches (`ContainSubstring`, string equality on a generated command) with no executing/behavioral test.

The #1309 case: a revalidation shell script's size probe was dead code (`curl -I -w '%{size_download}'`, always 0), but the tests only string-matched the generated command shape, so they were green on a skip branch that can never run. String-matching cannot observe runtime behavior.

## Expected Behavior

The gate (advisory to the reviewer, consistent with the #1332 tier) should flag a submission where a change to a generated command string / shell script, or to non-trivial branching logic, adds only substring/shape-match assertions and no test that executes the produced command or exercises the branch at runtime. Surface it so the reviewer knows a behavioral test is missing where string-matching gives false confidence.

## Environment

Foreman coder gate. Extends the test-dilution work in `pkg/foreman/agent/test_dilution_gate.go`. Parent: #1332. Sibling deferred signals: assertion-value churn, coverage/mutation-delta.

Defilan:
## What

Add `command-string-test-dilution`, an advisory coder-gate check that surfaces a production change to a shell or exec command string whose only added test assertions are string-shape matches.

## Why

Fixes #1346

#1332 shipped the test-dilution advisory but deferred this signal. The #1309 case is the motivating one: a revalidation script's size probe (`curl -I -w '%{size_download}'`) was dead code that always returned 0, but the tests only string-matched the generated command shape, so they were green on a branch that can never run. String-matching cannot observe runtime behaviour.

## How

Fires when **both** hold:

1. the diff modifies a non-test Go file at a command-construction site: `exec.Command(`, an `"sh"`/`"-c"` pair, or a string literal carrying shell metacharacters used as a command; and
2. every assertion added to that package's changed `_test.go` files is a string-shape assertion: `ContainSubstring(`, `strings.Contains(`, or `Equal(` on a string literal.

Registered under `tierAdvisory` beside the existing test-dilution check, so it surfaces to the reviewer and can never block a branch. A false positive is noise, not breakage.

**Deliberately out of scope:** detecting "a test that actually executes the command". That is not mechanically decidable from a diff, and attempting it produces a fragile heuristic. The two-part predicate above is the accepted approximation, and this limitation is stated in the code comment rather than left implicit.

The gate diffs with `--unified=0`, so removed and added lines arrive unpaired; the detector matches on normalised tokens rather than assuming line correspondence.

## Testing

Verified as real behavioural coverage, not just presence:

- Neutering `checkCommandStringTestDilution` to always return no-finding fails `MustFire` and `ShellCommandChange`.
- Removing the registry entry fails `RegisteredAsAdvisory` — a registration-asserting test, so a future refactor cannot silently unregister the check and stay green.
- Five must-stay-silent cases: assertions on a parsed result, assertions on an exit code, no command change, no test-file change, no production change. Plus a mixed-assertions case.
- Two fail-open cases covering git errors, so a broken workspace cannot turn an advisory into a hard failure.
- `go test ./pkg/foreman/agent/` passes (full package, 28s).
- Foreman clean-room gate: GATE-PASS.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (full suite via the clean-room gate; the affected package run directly)
- [ ] Documentation updated (internal gate behaviour, no user-facing surface)

## AI assistance

Implemented by the Foreman agentic-coding harness (Qwopus3.6-27B-Fusion as the coder), from an issue plus a dispatching intent that supplied the predicate, since #1346 itself never defined a trigger. I reviewed the diff, verified the must-fire tests fail when the detector is neutered and the registration test fails when the entry is removed, checked the must-stay-silent cases, and ran the package suite and `make lint` locally. Band-3 disclosure per the project's AI-assisted contribution policy.

The reviewer stage did not produce a usable verdict on this branch: it returned INCOMPLETE with "loop wall-clock budget exhausted before the agent finished". That is a harness timeout, not a judgment on the code. Verification here is mine.

### DIFF
--- a/pkg/foreman/agent/coder_gate.go
+++ b/pkg/foreman/agent/coder_gate.go
@@ -338,6 +338,18 @@ func gateCheckRegistry(issueText, evidenceBaseSHA string, evidence []grounding.E
 			tier: tierAdvisory,
 			fn:   checkTestDilution,
 		},
+		{
+			// command-string-test-dilution (#1346): advisory. Surfaces to the
+			// reviewer when a submission that modifies a shell/exec command
+			// string in production code adds only string-shape test assertions
+			// (ContainSubstring, strings.Contains, Equal on a string literal)
+			// with no behavioral test that executes the produced command.
+			// No lang: command-string detection is language-agnostic.
+			// Never blocks; the coder never sees it.
+			name: "command-string-test-dilution",
+			tier: tierAdvisory,
+			fn:   checkCommandStringTestDilution,
+		},
 	}
 }
 
--- a/pkg/foreman/agent/command_string_gate.go
+++ b/pkg/foreman/agent/command_string_gate.go
@@ -1,0 +1,281 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package agent
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"strings"
+)
+
+// checkCommandStringTestDilution is a tierAdvisory gate check (#1346). It
+// surfaces to the reviewer when a submission that modifies a shell/exec
+// command string in production code adds only string-shape test assertions
+// (ContainSubstring, strings.Contains, Equal on a string literal) with no
+// behavioral test that executes the produced command or exercises the branch
+// at runtime.
+//
+// Fires when BOTH hold:
+//
+//	(a) the diff modifies a non-test Go file at a site that builds a shell
+//	    or exec command string (exec.Command(, "sh", "-c", or a string
+//	    literal with shell metacharacters used as a command), AND
+//	(b) the same package has a changed _test.go file whose ADDED assertion
+//	    lines are exclusively string-shape assertions.
+//
+// Never fails the gate; the coder never sees it.
+func checkCommandStringTestDilution(ctx context.Context, workspace string, run commandRunner) (bool, string) {
+	// Stage the working tree so a pre-commit diff includes new/untracked files.
+	if _, err := run(ctx, workspace, nil, "git", "add", "-A"); err != nil {
+		return false, ""
+	}
+	nsOut, err := run(ctx, workspace, nil, "git", "diff", "--name-status", "--cached", "HEAD")
+	if err != nil {
+		return false, ""
+	}
+	entries := parseNameStatus(nsOut)
+	prodPkgs := changedProdPackages(entries)
+	if len(prodPkgs) == 0 {
+		return false, ""
+	}
+
+	// Get the full diff for all changed files (not just tests).
+	diffOut, err := run(ctx, workspace, nil, "git", "diff", "--cached", "--unified=0",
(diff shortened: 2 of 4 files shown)


joryirving (changes requested):
**Request changes.** Two detection gaps, including the motivating #1309 case.

**Important — `command_string_gate.go:109-116` (`hasCommandStringChange`): multiline `exec.Command` changes are invisible.** The detector checks each diff line independently, but Go `exec.Command` calls commonly span multiple lines and only the changed argument appears in the diff. The #1309 motivating case — `"-w", "%{size_download}"` → `"-w", "%{size_total}"` in a multiline `exec.Command("curl", ...)` — is not detected: the changed line has no `exec.Command(`, no `"sh"/"-c"` pair, and `%` is not in the `shellMeta` set. The gate built to catch #1309 would miss it.

**Important — `command_string_gate.go:199-217` (`hasOnlyStringShapeAssertions`): returns true on the first shape-only test file, ignoring behavioral assertions in sibling test files.** If a package has `shape_test.go` (only ContainSubstring) and `behavior_test.go` (Equal(0) on exit code), the function returns true without checking the second file. The logic should verify that ALL changed test files with assertions in the package are shape-only.

**Minor — `command_string_gate.go:273-280` (`hasStringLiteral`):** The double-quote-count heuristic can misclassify a behavioral assertion as string-shape if the line happens to contain `Equal(` and any pair of double quotes (e.g. in a comment). Advisory-only so the false positive direction is noise, not breakage, but worth tightening to check for a string literal as the direct argument to Equal.

**What is good:**

- The two-part predicate is a sound design for an advisory gate. The "deliberately out of scope" note about not detecting actual command execution is the right call.
- Test coverage is thorough: must-fire, five must-stay-silent, two fail-open, registration, advisory-not-blocking, and mixed-assertions cases.

**Verified:**

- Reproduced both important findings with focused tests in the PR worktree; both fail.
- `go test ./pkg/foreman/agent/` passes for the submitted branch. `git diff --check` clean. All 25 CI checks green. DCO signed.
- #1309 confirmed via `gh issue view` — the `curl -I -w '%{size_download}'` case is exactly what the multiline gap misses.

joryirving (approved):
**Approve.** All three findings from the first review are fixed, each verified by failing tests that now pass.

The fix commit addresses each gap precisely:

1. **Multiline exec.Command (#1309 case):** Gate now diffs with `--unified=3` and adds an enclosing detector — a changed line that is argument-list material fires when surrounding context opens a command construction. `isStringLiteralArgument` requires the whole line to be string-literals-and-separators, so ordinary code near a command stays silent.
2. **`hasOnlyStringShapeAssertions` early-return:** Now checks ALL test files in the package and returns false as soon as one has a behavioral assertion. Doc comment explicitly documents the "every file" quantifier.
3. **`hasStringLiteral` quote-counting:** Replaced with `isEqualOnStringLiteral`, which checks the direct argument to `Equal(`. Table-driven coverage provided.

The `fileHunks.Context` field is populated only for non-zero `--unified`; the sibling test-dilution gate (`--unified=0`) is unaffected.

**Non-blocking nit:** `git diff --check` flags 9 lines in `command_string_gate_test.go` for "space before tab" — these are inside raw string literals simulating git diff output, not actual Go indentation. Cosmetic only.

**Verified:**

- Both original probe tests (multiline argument, sibling behavioral file) now pass.
- `go test ./pkg/foreman/agent/` passes (full package).
- All existing test-dilution tests pass.
- All 25 CI checks green. DCO signed on both commits.

PR #1425, opened 2026-08-06
feat(controller): diagnose CUDA driver/runtime mismatch from crashed pods

### LINKED ISSUE #1424: Cryptic crash when the image's CUDA userspace requires a newer driver than the node provides
## Problem

When an InferenceService image bundles a CUDA userspace newer than what the node's kernel driver supports, the engine dies at init — after a GPU node has been provisioned and a multi-GB image pulled — and the only explanation is buried in the crashed container's logs:

```
RuntimeError: The NVIDIA driver on your system is too old (found version 12040).
```

Hit in production-like usage on AKS: the current default vLLM image ships a CUDA 13 userspace, while Azure's GPU node image bakes in driver 550.144.03 (CUDA 12.4). The pod crashloops, `kubectl get isvc` shows nothing actionable, and diagnosing it means knowing to go read the terminated container's logs and recognize the message. The same wall applies to any pinned image whose CUDA major is newer than the fleet's driver — a class of failure that will recur as CUDA 13-built images become the ecosystem default while cloud node images lag on 12.x drivers.

The compatibility rule itself is simple and documented: the node driver's supported CUDA **major** must be >= the image's CUDA major. Minor versions within a major are compatible ([NVIDIA minor version compatibility](https://docs.nvidia.com/deploy/cuda-compatibility/minor-version-compatibility.html)) — verified live: a CUDA 12.9 vLLM image serves fine on a 12.4 driver, while a 13.0 image fails exactly this way.

## Proposal: diagnose the crash on the InferenceService (no prediction, no blocking)

The operator already turns opaque pod-level failures into named diagnoses (`getPodSchedulingInfo` → `InsufficientGPU` → `WaitingForGPU`). This is the same move for driver-mismatch crashes:

1. Set `terminationMessagePolicy: FallbackToLogsOnError` on the generated runtime container, so the fatal error lands in the pod's termination message (this alone improves every crash diagnosis, not just this one).
2. When the runtime container is in a terminated/crashloop state, match the known driver-too-old signatures in the termination message (PyTorch wording, the cudart canonical string, the two error-enum spellings).
3. On a match, do one targeted `Get` of the node the pod ran on (the manager already has nodes RBAC) and include its GFD-reported driver/CUDA-support labels in the message when present; when absent, diagnose from the error text alone.
4. Surface it as a status condition + transition-only Warning event:

```yaml
- type: DriverCompatible
  status: "False"
  reason: CUDADriverInsufficient
  message: >-
    Runtime container terminated with a CUDA driver/runtime incompatibility:
    "RuntimeError: The NVIDIA driver on your system is too old (found version 12040).".
    The image's CUDA userspace requires a newer GPU driver than the node provides.
    Node "aks-gpu-123" supports CUDA <= 12.4 (driver 550.144.03).
    Use an image built for CUDA 12.x, or schedule onto nodes with a newer driver.
```

The condition is purely advisory: absent until a mismatch is first diagnosed, never wired into phase/`Available`, flips back to `True` (reason `RuntimeStarted`) only when the runtime container is observed Ready.

**Non-goals, deliberately:** no preflight prediction (a fleet scan can't see autoscaled nodes that don't exist yet, and would end up reimplementing the scheduler), no registry client in the operator, no NFD/GFD requirement (labels enrich the message when present; their absence degrades gracefully), no blocking, and no image-version recommendations — the message states the constraint class only.

Happy to bikeshed the condition/reason naming here before the PR freezes it into API surface.

brandonliuli:
## What

Diagnose CUDA driver/runtime mismatch crashes on the InferenceService itself: an advisory `DriverCompatible` condition plus a transition-only Warning event, built from the crashed runtime container's termination message and the node it actually ran on. Purely post-hoc — no prediction, no blocking, never wired into phase or `Available`.

## Why

A runtime image whose CUDA userspace is newer than the node driver's supported major dies at engine init, after a GPU node has been provisioned and a multi-GB image pulled, with the only explanation buried in the crashed container's logs (`The NVIDIA driver on your system is too old`). Hit live on AKS (baked-in driver 550/CUDA 12.4 vs. a CUDA 13 image). Today diagnosing it takes log archaeology; with this change it's one `kubectl describe isvc`.

Fixes #1424

## How

Same design move as `getPodSchedulingInfo` (opaque pod failure → named diagnosis), observable-state-driven throughout:

- `terminationMessagePolicy: FallbackToLogsOnError` on the generated runtime container makes terminations self-describing (benefits every crash class, not just this one).
- `reconcileDriverCompatCondition` (called from `Reconcile` before the omnibus status update, non-metal only) lists the service's pods, and matches known driver-too-old signatures — PyTorch's wording, the cudart canonical string (also what llama.cpp's `ggml_cuda_init` surfaces), and both error-enum spellings — in the **runtime container's** terminated states only. Init containers and sidecars are never inspected. NVML's `Driver/library version mismatch` is deliberately excluded (node-internal skew, different fix — see the code comment).
- On a match: one targeted `Get` of `pod.spec.nodeName` (nodes RBAC already exists — no `make manifests` churn), enriching the message with GFD's driver/CUDA-support labels when present (current names, with the deprecated pre-0.15 fallbacks); graceful degradation when the node is gone or unlabeled. No node watch, no fleet listing.
- Condition semantics follow the house anti-churn convention (`VLLMSpecValid`): absent until first diagnosis; `False/CUDADriverInsufficient` while crash evidence stands; `True/RuntimeStarted` only when readiness proves recovery: an active (not-Ready) crash carrier always diagnoses even when other replicas are healthy (driver-skewed node pools must not be masked), readiness outranks stale `LastTerminationState` evidence only within the pod that produced it (a container that crashed once — e.g. driver daemonset race — and then started in place recovers correctly), and fleet readiness recovers a prior diagnosis only once no pod carries crash evidence anymore (rollout replaced the crashing pods). The Warning event (same reason string, for one `--field-selector` vocabulary) fires on transition only, computed against in-memory conditions before the reconcile's status write — effectively one event per episode, at-least-once if that status write fails and the next pass recomputes the transition.
- Signature matching folds case in ASCII only and sanitizes/caps the reported line: Unicode lowering can change byte length, and termination messages are arbitrary container output that must never be able to panic the controller or produce an over-long/invalid status write.

Design decisions worth noting:

- **Behavior change on upgrade:** the pod-template change (termination policy) rolls existing InferenceService Deployments once on operator upgrade (Recreate strategy for GPU workloads — one restart per service). Called out here so it can ride a release that's expected to roll templates anyway if preferred.
- **Coverage is bounded by the log tail:** `FallbackToLogsOnError` captures the last 2048 bytes/80 lines; a runtime that prints shutdown noise after the traceback can push the signature out of view. A miss fails safe (no condition, no noise). The docs phrase the feature accordingly.
- The condition constant lives controller-local (`driver_compat.go`), matching the `ConditionVLLMSpecValid`/`ConditionSGLangSpecValid` precedent for single-writer conditions.
- Naming (`DriverCompatible` / `CUDADriverInsufficient` / `RuntimeStarted`) follows the repo's uppercase-acronym convention (`GPUAvailable`, `InsufficientGPU`); happy to rename before merge — it freezes into user automation after.

Possible follow-up (separate PR if there's appetite): CUDA-major metadata on the operator's own default runtime images with a lockstep test (default-image tag bump fails CI unless the metadata moves with it), used to generate the compatibility doc so the hand-maintained matrix can retire.

Branch note: the head is based on my fork's last-synced main (0.9.13) for fork-permission reasons; the change was developed and fully validated against current `main` as well (same clean apply, `make lint`/`lint-all` 0 issues, suite green). Happy to rebase the head onto current main if you prefer.

AI assistance disclosure: implementation and tests were AI-assisted (Claude); design, review, and validation are owned by the submitter, per CONTRIBUTING.md.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (two pre-existing failures in `pkg/foreman/agent/tools` on my workstation come from Apple git 2.39 not supporting `--end-of-options` in `checkout` — unrelated to this change; everything else green)
- [x] `make lint` passes locally (`lint-all` too)

### DIFF
--- a/docs/gpu-setup-guide.md
+++ b/docs/gpu-setup-guide.md
@@ -319,6 +319,35 @@ kubectl logs $POD -c llama-server | grep -i cuda
 # If not: Image may not have CUDA support, check image tag
 ```
 
+### Crash at startup: "driver on your system is too old" / "CUDA driver version is insufficient"
+
+The image's CUDA userspace is newer than what the node's kernel driver supports.
+When the error appears in the crashed container's termination message (the log
+tail on error), the operator diagnoses it and reports it on the
+InferenceService, so start there instead of the pod logs:
+
+```bash
+kubectl get inferenceservice $NAME -o jsonpath='{.status.conditions[?(@.type=="DriverCompatible")]}'
+# status: "False", reason: CudaDriverInsufficient — the message names the node,
+# its supported CUDA version (from GPU feature labels, when present), and the
+# runtime's own error line.
+kubectl get events --field-selector reason=CUDADriverInsufficient
+```
+
+The compatibility rule: the node driver's supported CUDA **major** version must
+be >= the image's CUDA major (a CUDA 12.9 image runs on a 12.4 driver; a CUDA
+13.x image does not). Minor versions within a major are compatible per
+[NVIDIA's minor version compatibility](https://docs.nvidia.com/deploy/cuda-compatibility/minor-version-compatibility.html),
+with two exceptions: PTX JIT compilation, and APIs introduced in a newer minor
+that need a matching driver. Fix: use an image built for the node's CUDA
+major, or move the workload to nodes with a newer driver. Check what a node
+supports with (older gpu-feature-discovery releases publish the same value
+under the deprecated `nvidia.com/cuda.runtime.major`):
+
+```bash
+kubectl get node $NODE -o jsonpath='{.metadata.labels.nvidia\.com/cuda\.runtime-version\.major}'
+```
+
 ### Poor GPU performance
 ```bash
 # Check GPU utilization
--- a/docs/gpu-setup-guide.md
+++ b/docs/gpu-setup-guide.md
@@ -328,7 +328,7 @@ InferenceService, so start there instead of the pod logs:
 
 ```bash
 kubectl get inferenceservice $NAME -o jsonpath='{.status.conditions[?(@.type=="DriverCompatible")]}'
-# status: "False", reason: CudaDriverInsufficient — the message names the node,
+# status: "False", reason: CUDADriverInsufficient — the message names the node,
 # its supported CUDA version (from GPU feature labels, when present), and the
 # runtime's own error line.
 kubectl get events --field-selector reason=CUDADriverInsufficient
--- a/internal/controller/deployment_builder.go
+++ b/internal/controller/deployment_builder.go
@@ -363,6 +363,11 @@ func (r *InferenceServiceReconciler) constructDeployment(
 		Name:            backend.ContainerName(),
 		Image:           image,
 		SecurityContext: inferContainerSecurityContext(isvc),
+		// Engines print the fatal error (CUDA init, model load) to the log
+		// rather than /dev/termination-log, so fall back to the log tail on
+		// error to make terminations self-describing; the driver-compat
+		// diagnosis reads its signatures from this message.
+		TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
 		Ports: []corev1.ContainerPort{
 			{
 				Name:          "http",
--- a/internal/controller/driver_compat.go
+++ b/internal/controller/driver_compat.go
@@ -1,0 +1,336 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+	"unicode/utf8"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	inferencev1alpha1 "github.com/defilantech/llmkube/api/v1alpha1"
+)
+
+// CUDA driver/runtime mismatch diagnosis. A serving image bundles a CUDA
+// userspace; the node's kernel driver supports CUDA up to some version. When
+// the image's CUDA major is newer than the driver's, the engine dies at init
+// — after a GPU node has been provisioned and a multi-GB image pulled — with
+// the real error buried in the crashed container's logs. This file turns that
+// termination message into a named diagnosis on the InferenceService, the
+// same move getPodSchedulingInfo makes for Insufficient-GPU pod conditions.
+
+const (
+	// ConditionDriverCompatible reports whether the node's GPU driver
+	// satisfied the CUDA userspace bundled in the runtime image, as observed
+	// from actual container terminations. Diagnostic only: it never feeds
+	// phase or the Available condition. Absent until a mismatch is first
+	// diagnosed; True once the runtime container subsequently starts (proof
+	// CUDA init succeeded).
+	ConditionDriverCompatible = "DriverCompatible"
+
+	// ReasonCUDADriverInsufficient is set when DriverCompatible=False because
+	// the runtime container terminated with a CUDA driver/runtime version
+	// incompatibility. Also the reason of the Warning event emitted on that
+	// transition, so one string serves kubectl field-selectors and condition
+	// queries alike.
+	ReasonCUDADriverInsufficient = "CUDADriverInsufficient"
+
+	// ReasonRuntimeStarted is set when DriverCompatible flips back to True: a
+	// previously diagnosed service now has a ready runtime container, so CUDA
+	// initialization against the node driver succeeded.
+	ReasonRuntimeStarted = "RuntimeStarted"
+)
+
+// cudaDriverMismatchSignatures are substrings CUDA-based runtimes print when
+// the node's kernel driver is older than the CUDA userspace in the image:
+// PyTorch's wording ("The NVIDIA driver on your system is too old (found
+// version 12040)"), the CUDA runtime's canonical error string (also what
+// llama.cpp's ggml_cuda_init surfaces via cudaGetErrorString), and both
+// error-enum spellings (runtime API camelCase, driver API underscores).
+// Matching is case-insensitive; entries must be lowercase. Deliberately
+// absent: "Driver/library version mismatch" (NVML), which is node-internal
+// driver/userspace skew after an in-place driver change, not an
+// image-vs-driver incompatibility — diagnosing it with this message would
+// misdirect the fix.
+var cudaDriverMismatchSignatures = []string{
+	"driver on your system is too old",
+	"cuda driver version is insufficient",
+	"cudaerrorinsufficientdriver",
+	"cuda_error_insufficient_driver",
+}
+
+// asciiLower lowercases ASCII letters only, preserving byte length and
+// offsets. strings.ToLower is unusable here: Unicode simple case mapping can
+// change byte length (e.g. U+023A lowercases from 2 to 3 bytes), so an index
+// found in its output does not address the input — and the input is arbitrary
+// container output, which must never be able to panic the controller. The
+// signatures are pure ASCII, so ASCII folding is sufficient.
+func asciiLower(s string) string {
+	var b []byte
+	for i := 0; i < len(s); i++ {
+		c := s[i]
+		if c >= 'A' && c <= 'Z' {
+			if b == nil {
+				b = []byte(s)
+			}
+			b[i] = c + ('a' - 'A')
+		}
+	}
+	if b == nil {
+		return s
+	}
+	return string(b)
+}
+
+// sanitizeMessageLine trims the line, replaces control characters with
+// spaces, and caps its length on a rune boundary, so the condition message
+// and the (quoted, length-limited) event note stay valid and bounded no
+// matter what the container printed.
+func sanitizeMessageLine(line string) string {
+	line = strings.Map(func(r rune) rune {
+		if r < 0x20 || r == 0x7f {
+			return ' '
+		}
+		return r
+	}, strings.TrimSpace(line))
+	const maxLine = 200
+	if len(line) <= maxLine {
+		return line
+	}
+	cut := maxLine
+	for cut > 0 && !utf8.RuneStart(line[cut]) {
+		cut--
+	}
+	return line[:cut] + "..."
+}
+
+// matchCudaDriverMismatch reports whether msg contains a known CUDA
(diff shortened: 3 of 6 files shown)


brandonliuli:
Addressed the codecov report in aa26929: added an envtest spec that drives a full Reconcile over a crashlooping runtime pod and asserts the persisted condition (this was the meaningful gap — the call site in Reconcile was previously only covered indirectly; the new spec fails if the call is removed), plus unit cases for the pod-list-failure path, a crash on a pod with no node name, and the message-cap rune-boundary branch. driver_compat.go now sits at ~99% patch coverage locally; the remaining sliver is one branch of the ASCII-folding helper.

brandonliuli:
On the codecov flags — two of the three flagged spots are covered; the third now is:

- **`inferenceservice_controller.go` (the `if !isMetal { reconcileDriverCompatCondition(...) }` wiring):** exercised by `driver_compat_envtest_test.go`, which drives a full `Reconcile()` over a crashlooping pod and asserts the persisted condition — that spec exists precisely so this call site can't be deleted silently. A local run of the suite confirms the profile hits it (`inferenceservice_controller.go:256.2,256.14 1 1` / `256.14,258.3 1 1`), so the 0% patch attribution there looks like a report artifact rather than a real gap.
- **`driver_compat.go:101-103`:** that one was real — `asciiLower`'s zero-copy path for input with no uppercase bytes, which every existing signature case happened to skip. Covered in 2a94d9a with an already-lowercase matching message.

Defilan (changes requested):
Really strong PR. The design note is unusually good, the RBAC and condition-convention claims both hold up when checked, and the patch applies cleanly to current `main` despite being cut from 0.9.13. One functional gap to fix before merge, then a few small things.

## Blocking: a healthy replica masks a crash-looping one

`runtimeContainerReady` and `findCudaDriverCrash` each scan **every** pod of the service and return on the first hit, and their results are then OR'd:

https://github.com/defilantech/LLMKube/blob/2a94d9add763dc97ae82f8d1cc731d7f41bd59f7/internal/controller/driver_compat.go#L252-L258

So the two are evaluated fleet-wide rather than per pod. With `replicas: 2` split across a good node and a bad-driver node:

- **Fresh deploy:** replica A goes Ready on the good node, replica B crash-loops on the bad one. `ready` is true, so the function returns down the recovery path and the diagnosis is never surfaced at all.
- **Scale-up onto skew:** an existing `False/CUDADriverInsufficient` flips to `True/RuntimeStarted` the moment any one replica lands somewhere healthy, while the original pod keeps crashing.

Partial driver skew across a node pool is the case this feature is most useful for (mid-driver-upgrade pool, mixed node pools, a daemonset that only converged on some nodes), so this bites exactly where it should help. Pairing the readiness check to the pod, rather than to the service, should be enough: if the pod that carries the crash signature is itself now Ready, that is the stale-evidence case the comment describes and recovery is right; if a *different* pod is Ready, that says nothing about the crashing one.

The recovery reasoning in the comment above that block is sound, and I would keep it, this is just about the scope it is evaluated over.

## Docs: the reason string is misspelled in the one place a user would copy

https://github.com/defilantech/LLMKube/blob/2a94d9add763dc97ae82f8d1cc731d7f41bd59f7/docs/gpu-setup-guide.md#L330-L334

Line 331 says `CudaDriverInsufficient`, but the constant is `CUDADriverInsufficient` and the `kubectl get events` line three rows down uses the correct spelling. Since the whole point is greppability, worth fixing.

## Minor

**Node `Get` treats every error as deletion.** Any failure (timeout, throttling, transient API error) produces "no longer readable", which reads as "the node was deleted":

https://github.com/defilantech/LLMKube/blob/2a94d9add763dc97ae82f8d1cc731d7f41bd59f7/internal/controller/driver_compat.go#L313-L319

Worth distinguishing `apierrors.IsNotFound` from the rest, and logging the non-NotFound case the way the pod-list failure just above already does. Right now a persistent RBAC or API problem leaves no operator-visible trail.

**"One event per episode" can double-fire.** `transitioned` is computed from in-memory `isvc.Status.Conditions`, but the persist happens later in `Reconcile` via a separate status update that can fail on a 409. If that write fails, the next reconcile re-reads a status with no condition, recomputes `transitioned` as true, and emits a second Warning for the same unbroken episode. Low impact for an advisory event, and arguably not worth restructuring for, but the guarantee in the PR description is slightly stronger than what the code provides.

## Not yours, but adjacent

`normalizeContainers` in `drain_before_rollout.go` strips `TerminationMessagePolicy` before comparing container specs, on the assumption that the field is never operator-meaningful, which this PR changes. It does not bite today: that comparison is only the fallback path, and the primary path hashes the unnormalized template, so the rollout gate still engages. Flagging it so it is on the record rather than asking you to fix it here.

## Checked and fine

- The "nodes RBAC already exists, no `make manifests` churn" claim is correct. `config/rbac/role.yaml` and the chart clusterrole both already carry `nodes: get;list;watch`, granted for `model_controller.go`, and the tree is unmodified there.
- Condition shape matches the `ConditionVLLMSpecValid` precedent on constant placement, anti-churn, and reason casing.
- Init containers really are excluded structurally (`ContainerStatuses` vs `InitContainerStatuses`), and the byte-offset math in the sanitize and ASCII-lower path is safe against multibyte and invalid UTF-8.
- `pod.Spec.NodeName == ""` is handled and tested.
- No merge conflict with current `main`: `deployment_builder.go` has drifted since 0.9.13 (#1366), but in a different region. No need to rebase on my account.

On naming: `DriverCompatible` / `CUDADriverInsufficient` / `RuntimeStarted` are right, keep them.

brandonliuli:
Thanks for the review — the skew case was a real hole, and the fresh-deploy variant (diagnosis never surfaced at all) is the worse half of it. All addressed in 2afa8be + 6a8d8b7:

**Blocking — fixed, with one deliberate scope choice.** `findCudaDriverCrash` now pairs the evidence with the pod that carries it and prefers an active (not-Ready) carrier over a recovered one; the recovery gate is `!found || carrierReady` instead of fleet-wide `ready || !found`. Net semantics:

- Active carrier anywhere → diagnose `False`, regardless of other replicas' health (both of your bite cases: fresh-deploy skew and scale-up-onto-skew).
- Carrier itself Ready again → the stale-evidence recovery you agreed should stay, now scoped to the pod that produced it.
- **No carrier left + some runtime Ready → still recovers `True`.** This is the one place I stopped short of hard per-pod pairing: after a user fixes the image and the rollout replaces the crashing pods, the evidence is gone with them, and strict pairing would leave `False` stuck forever on a healthy service. Fleet readiness only gets a say once nothing contradicts it. If you'd rather have sticky-`False` there too, it's a two-line change — say the word.

Regression tests cover all three skew shapes (healthy-replica masking, scale-up flip, active-vs-stale-carrier precedence), and I verified they fail against the old gate before the fix (mutation check: reverting to `ready || !found` fails exactly those three).

**Docs casing** — fixed in 6a8d8b7 (`CudaDriverInsufficient` → `CUDADriverInsufficient` in the jsonpath comment).

**Node `Get` errors** — now split: `IsNotFound` keeps a deletion message ("no longer exists"), everything else logs like the pod-list failure path and says "not readable right now" instead of implying the node is gone. Test injects a Get failure via `interceptor.Funcs` (same idiom as the list-failure test).

**Event double-fire on a failed status write** — agreed it's real and agreed it's not worth restructuring for an advisory event; I've softened the PR description to "at-least-once per episode" so the claim matches the code rather than the other way around.

**`normalizeContainers`** — good catch on the record; happy to file a small follow-up issue so the assumption change is tracked, since the primary hash path is what keeps it from biting today.

Defilan (commented):
The skew fix is right and I verified it rather than taking the mutation check on trust. Two things left, one of them mine to own.

## Blocking finding: resolved, verified

`findCudaDriverCrash` now returns immediately on the first not-Ready carrier and only remembers a Ready one as fallback, with the gate at `!found || carrierReady`. I reverted that gate to the old `ready || !found` on your branch and ran the suite: exactly your three skew subtests fail, and nothing else does.

```
--- FAIL: TestReconcileDriverCompatConditionDriverSkew/healthy_replica_does_not_mask_a_crash-looping_one
--- FAIL: TestReconcileDriverCompatConditionDriverSkew/scale-up_onto_a_healthy_node_does_not_recover_a_still-crashing_carrier
--- FAIL: TestReconcileDriverCompatConditionDriverSkew/active_crash_outranks_another_pod's_stale_recovered_evidence
```

Precisely the claim you made. Also confirmed the node `Get` split reads correctly, with `IsNotFound` saying "no longer exists" and everything else logging and saying "not readable right now", and the docs casing is fixed.

## Your design question: keep what you have

> No carrier left + some runtime Ready → still recovers `True`. If you'd rather have sticky-`False` there too, it's a two-line change.

Keep it as is. Sticky-`False` is the worse failure: the normal path out of this condition is exactly "user fixes the image, rollout replaces the crashing pods, evidence leaves with them". Strict pairing would strand `False` on a healthy service with no way to clear it, which turns a diagnostic into a false alarm. Your framing is the right one, fleet readiness gets a say only once nothing contradicts it, and recovery still requires positive evidence (an actually-Ready runtime container) rather than mere absence of evidence.

The case that might look uncovered, a crashing pod evicted while a good replica serves, self-corrects: if the scheduler puts a new pod back on the skewed node it crashes and re-diagnoses. A condition that reports observed state is the right contract here.

Softening the PR description to "at-least-once per episode" is the right call over restructuring for an advisory event. Yes please on the `normalizeContainers` follow-up issue.

## One thing to fix: `make lint` is red on the current head

```
internal/controller/driver_compat_test.go:322:14: gfdNode - name always receives "node-a" (unparam)
```

All four call sites pass `"node-a"`. Run with the repo's pinned golangci-lint v2.12.2 (`./bin/golangci-lint run`, same as `make lint`), so this is not a version-skew artifact on my end. I suspect the checklist was ticked before 2afa8be and 6a8d8b7 landed. Dropping the parameter is enough.

Happy to approve as soon as that is green. Everything substantive is settled.

## Heads up, and this one is on my side

**No CI has run on this PR's head.** `6a8d8b7c` has zero check runs, and so does #1436, though an earlier commit here had 24. Something is gating workflow approval on fork PRs, so neither of us has had CI signal on your recent pushes, which is also how the lint break got through.

Since you had no build to check against, I ran it here instead: full `internal/controller` suite with envtest passes (248s), and lint is clean apart from the `unparam` above. I will sort out the workflow approval on my end. Sorry for the missing signal.

brandonliuli:
Fixed in 8579df5 — `gfdNode` no longer takes the parameter. `./bin/golangci-lint run` (pinned v2.12.2) is now **0 issues repo-wide** here, and the driver-compat tests pass unchanged.

You're right about the sequencing on my side too: the checklist tick predated 2afa8be/6a8d8b7 — I re-ran fmt/vet/tests after those but not lint, which is exactly the gap. Thanks for running the suite where CI couldn't and for re-verifying the mutation claim independently.

`normalizeContainers` follow-up filed as #1437, with both resolution options sketched — happy to send the PR for whichever direction you prefer.

Defilan (approved):
Approving. `8579df5b` clears it.

Verified on that head with the pinned golangci-lint v2.12.2 and a cleared cache: **0 issues**. Full `internal/controller` suite with envtest passes (248s).

Everything from the original review is settled: the skew fix is mutation-checked (reverting the gate fails exactly your three subtests and nothing else), the node `Get` split is right, and the docs casing is fixed. Keeping fleet readiness as the recovery signal when no carrier remains is the correct call, for the reason in my last comment.

Nice work on this one. The design note made it reviewable, and the mutation checks meant I could confirm your claims rather than re-derive them.

Note that CI still has not run here: the workflow runs on this branch are sitting at `action_required`, so the green above is from my machine, not from the build. I am sorting that out.

Issue #449, opened 2026-05-12
ci(release): build and push llmkube-router-proxy image alongside the controller


Defilan:
Part of epic #437. Follow-up to #448.

## Context

The Helm chart from #448 ships a `controllerManager.routerProxy` image block that the controller uses to spin up router-proxy Deployments for ModelRouter resources. The release pipeline currently builds and publishes `ghcr.io/defilantech/llmkube-controller:<version>` on every tagged release but does not build the corresponding `ghcr.io/defilantech/llmkube-router-proxy:<version>`.

As an interim mitigation, #448 pins the chart default for `controllerManager.routerProxy.tag` to `"dev"` and documents that users must build the image themselves (`make docker-build-router-proxy`) until this issue lands. That's awkward and limits Phase 1 adoption.

## Scope

Update the existing release workflow (most likely `.github/workflows/release-please.yml` or the goreleaser config) so that every tagged release also:

1. Builds `Dockerfile.router-proxy` for the same `linux/amd64,linux/arm64` matrix the controller uses.
2. Tags the image with the same version as the controller (chart appVersion).
3. Pushes to `ghcr.io/defilantech/llmkube-router-proxy:<version>`, `:latest`, and (ideally) a digest pin.
4. Signs the image with cosign if the controller is signed (verify which is true today).

## Once this ships

In a small follow-up PR:
- Flip the chart default `controllerManager.routerProxy.tag` from `"dev"` back to `""` (so it falls back to chart appVersion like the controller image).
- Remove the warning callout from `docs/site/concepts/model-router.md`.
- Update the warning comment in `charts/llmkube/values.yaml`.

## Success criteria

- A clean release produces both `llmkube-controller:<v>` and `llmkube-router-proxy:<v>` images at the same registry path conventions.
- `helm template charts/llmkube/` against a real release renders a controller `--router-proxy-image=ghcr.io/defilantech/llmkube-router-proxy:<v>` that exists in the registry.
- `kubectl apply -f config/samples/inference_v1alpha1_modelrouter.yaml` on a cluster with the chart-installed controller produces a ModelRouter whose proxy Deployment becomes Ready (not ImagePullBackOff).

## Out of scope

- Multi-platform / cross-arch optimization of the router-proxy image. Match the controller's existing matrix.
- ARM64-specific builds for DGX Spark / GB10. Already covered by the standard linux/arm64 target.

PR #1431, opened 2026-08-06
fix(chart): default the router-proxy image tag to the chart appVersion

### LINKED ISSUE #1426: [BUG] Chart defaults the router-proxy image to the dev tag although releases now publish versioned images
## Bug Description

A stock `ModelRouter` lands in `ImagePullBackOff` on a chart install: the chart defaults the router-proxy image to `ghcr.io/defilantech/llmkube-router-proxy:dev`, and that tag does not exist on ghcr.

The default predates #449. Now that #449 is closed, `.goreleaser.yaml` builds and pushes `ghcr.io/defilantech/llmkube-router-proxy:{{ .Version }}` on every release ([.goreleaser.yaml#L120-L122](https://github.com/defilantech/LLMKube/blob/36631a9/.goreleaser.yaml#L120-L122), [#L250-L257](https://github.com/defilantech/LLMKube/blob/36631a9/.goreleaser.yaml#L250-L257)) — I verified `ghcr.io/defilantech/llmkube-router-proxy:0.9.14` resolves (while `dev`, `latest`, and `v0.9.14` all 404). The chart helper already falls back to `.Chart.AppVersion` when the tag is empty ([_helpers.tpl#L105](https://github.com/defilantech/LLMKube/blob/36631a9/charts/llmkube/templates/_helpers.tpl#L105)), so the only thing standing between a default install and a working image is `tag: "dev"` in values.yaml plus the comment block above it that still says the release pipeline "does not yet build and push the router-proxy image" ([values.yaml#L96-L129](https://github.com/defilantech/LLMKube/blob/36631a9/charts/llmkube/values.yaml#L96-L129)).

## Steps to Reproduce

1. `helm install` chart 0.9.14 with default values.
2. Apply a minimal `ModelRouter` (dataPlane `Proxy`, one `inferenceServiceRef` backend) without `spec.proxy.image`.
3. The operator creates the router-proxy Deployment pointing at `ghcr.io/defilantech/llmkube-router-proxy:dev` → `ImagePullBackOff`.

## Expected Behavior

The default install pulls the router-proxy image that the release published alongside the controller image, the same way `controllerManager.container.image` defaults to the appVersion.

## Actual Behavior

`ImagePullBackOff` until the user discovers `spec.proxy.image: ghcr.io/defilantech/llmkube-router-proxy:0.9.14` (the bare version tag — not documented anywhere, since the values comment says the image doesn't exist).

## Environment

**LLMKube Version:** chart/operator 0.9.14 (also present on `main` @ 36631a9)
**Kubernetes:** AKS 1.33, ModelRouter verified serving end-to-end once the image override was set (120/120 chat completions proxied).

## Proposed Fix

`tag: ""` in values.yaml (letting the existing `| default .Chart.AppVersion` helper take over) and rewrite the stale comment. The in-code `defaultRouterProxyImage = ...:dev` constant ([router_deployment_builder.go#L28-L32](https://github.com/defilantech/LLMKube/blob/36631a9/internal/controller/router_deployment_builder.go#L28-L32)) can stay for flagless local dev, since chart installs always pass `--router-proxy-image`. PR incoming.

---

brandonliuli:
## What

Default the router-proxy image tag to the chart appVersion instead of the unpublished `dev` tag, drop the stale "the release pipeline does not yet build this image" comment block, and retire the same stale premise from the docs (model-router concept page, air-gapped guide).

## Why

Since #449 closed, `.goreleaser.yaml` builds and pushes `ghcr.io/defilantech/llmkube-router-proxy:{{ .Version }}` on every release — `0.9.14` resolves on ghcr today. But the chart still pins `tag: "dev"`, which was never published, so a stock `ModelRouter` lands in `ImagePullBackOff` until the user discovers the undocumented bare-version tag. (#609's own description names this chart default as the adoption blocker it was solving; the chart flip was the unfinished half.)

Fixes #1426

## How

`tag: ""` in values.yaml — the `llmkube.routerProxyImage` helper already falls back to `.Chart.AppVersion`, which matches the tag goreleaser publishes for that release, so no template change is needed. The `defaultRouterProxyImage = …:dev` constant in the operator is untouched: chart installs always pass `--router-proxy-image`, and the constant keeps flagless local dev working. It is also what the ModelRouter e2e cluster context keys on: `test/e2e/e2e_suite_test.go` side-loads exactly the constant's `:dev` tag into Kind so the operator dispatches without a `--router-proxy-image` rewrite — leaving the constant untouched keeps that flow unchanged.

Docs carrying the same premise updated in the same pass:
- `docs/site/concepts/model-router.md`: removed the "Router-proxy image not yet in the release pipeline" warning callout (the alpha Status callout stays).
- `docs/site/guides/air-gapped.md`: staged images now reference the published `0.9.14` tags. While there, fixed the guide's controller pins too — ghcr tags are bare versions, so the `v0.7.7`-style pulls the guide showed never resolved (`v0.7.7` 404s, `0.7.7` exists; verified against the registry).

New chart test suite (`router-proxy-image_test.yaml`) asserts the rendered `--router-proxy-image` arg: appVersion default, explicit tag override, digest precedence. Mutation-checked: reverting `tag` to `"dev"` fails the suite.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (`helm unittest`: 63/63 incl. the new suite; no Go change)
- [x] `make lint` / `make lint-all` pass locally aside from 6 pre-existing staticcheck findings in `pkg/foreman/agent` tests, identical on the base commit (nothing from this change)

---
Assisted-by: Claude (Anthropic). Investigated, verified, and reviewed by a human; the ImagePullBackOff and the working bare-version tag were confirmed on a live AKS cluster, and every image tag referenced in the doc changes was verified to resolve on ghcr.

### DIFF
--- a/charts/llmkube/tests/router-proxy-image_test.yaml
+++ b/charts/llmkube/tests/router-proxy-image_test.yaml
@@ -1,0 +1,31 @@
+suite: router-proxy default image
+templates:
+  - deployment.yaml
+
+tests:
+  # The published router-proxy image is tagged with the release version, so
+  # the default must track the chart appVersion — a stock install of a
+  # ModelRouter must never point at an unpublished tag (#1426).
+  - it: should default the router-proxy image tag to the chart appVersion
+    chart:
+      appVersion: 9.9.9
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: --router-proxy-image=ghcr.io/defilantech/llmkube-router-proxy:9.9.9
+
+  - it: should honor an explicit tag override
+    set:
+      controllerManager.routerProxy.tag: my-tag
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: --router-proxy-image=ghcr.io/defilantech/llmkube-router-proxy:my-tag
+
+  - it: should prefer a digest over the tag
+    set:
+      controllerManager.routerProxy.digest: sha256:0000000000000000000000000000000000000000000000000000000000000000
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: --router-proxy-image=ghcr.io/defilantech/llmkube-router-proxy@sha256:0000000000000000000000000000000000000000000000000000000000000000
--- a/charts/llmkube/values.yaml
+++ b/charts/llmkube/values.yaml
@@ -99,33 +99,19 @@ controllerManager:
   # InferenceServices and external provider backends, speaks
   # OpenAI chat completions, and enforces policy routing.
   #
-  # IMPORTANT (Phase 1 alpha): the release pipeline does not yet build
-  # and push the router-proxy image alongside the controller image.
-  # The default tag below is "dev" rather than the chart appVersion so
-  # users who never create a ModelRouter resource are unaffected, and
-  # users who do create one know they need to either:
-  #
-  #   1. Build the image locally (`make docker-build-router-proxy`)
-  #      and push it to a registry their cluster can reach, then
-  #      override repository / tag here.
-  #   2. Override `spec.proxy.image` per ModelRouter with their own
-  #      build.
-  #
-  # See https://github.com/defilantech/LLMKube/issues/449 for the
-  # tracking issue on integrating the router-proxy image into the
-  # release flow.
+  # The release pipeline publishes ghcr.io/defilantech/llmkube-router-proxy
+  # alongside the controller image with matching version tags (#449), so the
+  # empty tag below resolves to the chart appVersion.
   routerProxy:
     # Registry prefix for the router-proxy image (mirrors controller image).
     # Useful for air-gapped or enterprise proxy registries.
     registry: ""
     repository: ghcr.io/defilantech/llmkube-router-proxy
     pullPolicy: IfNotPresent
-    # Default tag. "dev" is intentional until the release pipeline
-    # ships versioned router-proxy images alongside the controller.
-    tag: "dev"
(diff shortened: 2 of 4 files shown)


Defilan (approved):
Approving. I checked all five load-bearing claims and every one holds.

- **The helper really does fall back to appVersion.** `_helpers.tpl` resolves `tag | default .Chart.AppVersion`, with a digest short-circuiting ahead of the tag, so `tag: ""` needs no template change. I applied your chart diff to a scratch copy and ran your new suite against it: 3/3 pass, including digest-over-tag precedence.
- **goreleaser publishes the image.** The `router-proxy` entry under `dockers_v2` pushes `ghcr.io/defilantech/llmkube-router-proxy` tagged `{{ .Version }}`, with no `skip_push`.
- **`0.9.14` resolves on ghcr.** 200 against the manifest endpoint.
- **The bare-vs-`v` tag claim is right.** `llmkube-controller:v0.7.7` returns 404, `0.7.7` returns 200. Good catch fixing the air-gapped guide's pins while you were in there; those pulls never would have worked.
- **Leaving `defaultRouterProxyImage` at `:dev` is correct, and load-bearing.** `test/e2e/e2e_suite_test.go` side-loads that exact literal into Kind, and the Kustomize `make deploy` path the e2e suite uses never passes `--router-proxy-image`, so the operator genuinely falls through to the Go constant. Changing it would have broken the ModelRouter e2e context. Worth noting these are two independent string literals with nothing enforcing the match, but that is pre-existing and exactly what you said, not more.

This is the unfinished half of #609 and it is good to have it closed.

PR #1432, opened 2026-08-06
fix(controller): publish model downloads atomically so an interrupted transfer is never cached

### LINKED ISSUE #1428: [BUG] Interrupted model download is cached as complete: the init script's existence check accepts a truncated file
## Bug Description

An interrupted model download is permanently cached as complete. The default (non-`OnChange`) init-container scripts download with `curl -f -L -o "$MODEL_PATH"` guarded only by `[ ! -f "$MODEL_PATH" ]` ([model_storage.go#L181-L205](https://github.com/defilantech/LLMKube/blob/36631a9/internal/controller/model_storage.go#L181-L205)): curl creates the destination file at transfer start, so if the init container is OOM-killed, evicted, or the node reboots mid-pull, a truncated file is left at `$MODEL_PATH`. Every subsequent restart sees the file exists, logs "Model already cached, skipping download", and hands the runtime a corrupt model.

Hit this in practice on a k3s lab: interrupted GGUF pull → llama.cpp crash-looping on "model is corrupted or incomplete" until the cache PVC was manually cleared. Because the cache is a PVC, the bad state survives pod restarts indefinitely — the failure needs manual intervention, and the error surfaces in the *runtime* container, far from the cause.

The repo has already fixed exactly this failure mode in the `RefreshPolicy=OnChange` paths (#1309): download to `"$dest.tmp"` then atomic `mv`, with a comment explaining that bare `curl -o` "truncates the cached file at request start, so a failed or full-200 transfer destroys the good cache" ([model_storage.go#L216-L233](https://github.com/defilantech/LLMKube/blob/36631a9/internal/controller/model_storage.go#L216-L233)). The default paths still use the pattern that comment condemns.

Affected sites (all in `model_storage.go`):

- single-file cached S3 and https branches, single-file non-cached S3 and https branches (`buildModelInitCommand`)
- the multi-file non-`OnChange` loop (`buildMultiFileInitCommand`, [#L443-L455](https://github.com/defilantech/LLMKube/blob/36631a9/internal/controller/model_storage.go#L443-L455))
- the local-source `cp` (same truncation window on an interrupted copy)

`Model.spec.sha256` mitigates when set, but it's optional and the failure shouldn't require it.

## Steps to Reproduce

1. Create a Model with an https source and `modelCache` enabled; kill the init container mid-download (or `kubectl delete pod` during the pull).
2. Let the pod restart.
3. Init logs "Model already cached, skipping download"; runtime container crashes on the truncated file, forever.

## Expected Behavior

A download either completes and is published atomically, or leaves nothing behind — the next attempt re-downloads.

## Actual Behavior

The partial file is treated as a valid cache entry until someone deletes the PVC contents.

## Environment

**LLMKube Version:** observed on 0.9.10; the affected scripts are unchanged on `main` @ 36631a9.

## Proposed Fix

Apply the existing `.tmp` + `mv` publish pattern from the `OnChange` scripts to the remaining download/copy branches. No behavior change on the happy path; regression coverage via the existing table tests. PR incoming.

---

brandonliuli:
## What

Make every non-`OnChange` model transfer publish atomically: download (or `cp`) to `"$MODEL_PATH.tmp"` / `"$dest.tmp"` and `mv` onto the final path on success. Covers the single-file https/S3/local branches (cached and uncached) in `buildModelInitCommand` and the multi-file `IfNotPresent` loop in `buildMultiFileInitCommand`.

## Why

The default init scripts wrote transfers straight to `$MODEL_PATH` behind a bare `[ ! -f ]` existence check. `curl -o` creates the destination at transfer start, so an init container interrupted mid-pull (OOM-kill, eviction, node reboot) leaves a truncated file that every subsequent restart reports as "Model already cached, skipping download" — and the runtime container then crash-loops on a corrupt model until someone clears the cache PVC by hand. Hit in practice: an interrupted GGUF pull left llama.cpp failing with "model is corrupted or incomplete" across restarts.

The `RefreshPolicy=OnChange` scripts already fixed exactly this in #1309 (`.tmp` + atomic `mv`, with a comment explaining that bare `curl -o` "destroys the good cache"); this applies the same publish pattern to the remaining branches.

Fixes #1428

## How

String-level change to the generated shell only; no controller logic touched. A failed transfer leaves at most a `.tmp` file, which the existence check ignores and the next attempt overwrites. Multi-file loop keeps its fail-fast `|| { …; exit 1; }` semantics (`a && b || c` runs the handler if either the download or the rename fails).

Tests: existing assertions updated to the atomic form, plus a new case asserting no non-`OnChange` variant writes the download target directly. Mutation-checked: reverting the cached-https branch to direct `-o "$MODEL_PATH"` fails the suite.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally for the affected packages (`internal/controller`, `api/...`; the only failures in a full run are 2 pre-existing `pkg/foreman/agent/tools` cases that fail identically on the untouched base commit — a local git incompatibility in that test harness, unrelated to this change)
- [x] `make lint` and `make lint-all` pass locally (6 pre-existing staticcheck findings in `pkg/foreman/agent` tests, identical on the base commit; nothing from this change)

---
Assisted-by: Claude (Anthropic). Investigated, verified, and reviewed by a human.

### DIFF
--- a/internal/controller/inferenceservice_storage_test.go
+++ b/internal/controller/inferenceservice_storage_test.go
@@ -391,7 +391,8 @@ var _ = Describe("buildMultiFileInitCommand", func() {
 		Expect(cmd).To(ContainSubstring(`mkdir -p "$CACHE_DIR"`))
 		Expect(cmd).To(ContainSubstring("printf '%s\\n' \"$MODEL_FILES\""))
 		Expect(cmd).To(ContainSubstring(`mkdir -p "$(dirname "$dest")"`))
-		Expect(cmd).To(ContainSubstring(`curl -f -L -o "$dest" "$url"`))
+		Expect(cmd).To(ContainSubstring(`curl -f -L -o "$dest.tmp" "$url" && mv "$dest.tmp" "$dest"`))
+		Expect(cmd).ToNot(ContainSubstring(`-o "$dest" `))
 		Expect(cmd).To(ContainSubstring("already cached, skipping download"))
 	})
 
@@ -1238,8 +1239,7 @@ var _ = Describe("buildModelInitCommand", func() {
 	It("should generate cached local copy command", func() {
 		cmd := buildModelInitCommand(true, false, true, RefreshPolicyIfNotPresent)
 		Expect(cmd).To(ContainSubstring(`mkdir -p "$CACHE_DIR"`))
-		Expect(cmd).To(ContainSubstring("cp /host-model/model.gguf"))
-		Expect(cmd).To(ContainSubstring(`"$MODEL_PATH"`))
+		Expect(cmd).To(ContainSubstring(`cp /host-model/model.gguf "$MODEL_PATH.tmp" && mv "$MODEL_PATH.tmp" "$MODEL_PATH"`))
 	})
 
 	It("should generate error exit for uncached local source", func() {
--- a/internal/controller/model_storage.go
+++ b/internal/controller/model_storage.go
@@ -178,30 +178,35 @@ func addCACertVolume(volumes *[]corev1.Volume, mounts *[]corev1.VolumeMount, cmd
 	*cmd = fmt.Sprintf("export CURL_CA_BUNDLE=/custom-certs/$(ls /custom-certs | grep -v '^\\.' | head -n 1) && %s", *cmd)
 }
 
+// All transfers write to "$MODEL_PATH.tmp" and mv onto "$MODEL_PATH" on
+// success: the guard is a bare existence check, so publishing the file
+// non-atomically would let an interrupted transfer (OOM-kill, eviction,
+// node reboot) leave a truncated artifact that every subsequent restart
+// treats as cached. See remoteRevalidateScript for the same pattern.
 func buildModelInitCommand(isLocal, isS3, useCache bool, refreshPolicy string) string {
 	if useCache {
 		if isLocal {
-			return `mkdir -p "$CACHE_DIR" && if [ ! -f "$MODEL_PATH" ]; then echo 'Copying model from local source...'; cp /host-model/model.gguf "$MODEL_PATH" && echo 'Model copied successfully'; else echo 'Model already cached, skipping copy'; fi`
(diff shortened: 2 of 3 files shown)


Defilan (approved):
Approving. I verified the shell rather than reading it, and it holds up on every point the description claims.

- **Operator precedence in the multi-file loop.** `curl && mv || handler` parses as `(curl && mv) || handler`. Ran all four outcomes through `sh -c`: curl fails, mv fails, both succeed, and the curl-ok/mv-fails case. The handler fires in exactly the three failure cases and no case silently succeeds. Fail-fast is preserved.
- **Exit status of the single-file branches.** No explicit `exit 1` is needed, since an `if` compound exits with the status of the last command in the taken branch. Confirmed the failing-curl path still exits non-zero and that this is byte-for-byte the same behavior as before. The new curl-ok/mv-fails path also correctly propagates.
- **Atomicity.** Traced `buildCachedStorageConfig` and `buildEmptyDirStorageConfig`: `MODEL_PATH`, `CACHE_DIR`, and `dest` are always `<mounted-dir>/<basename>` under a single mount, so `$X.tmp` and `$X` are always same-directory and therefore same-filesystem. `mv` is a real `rename(2)`, not a cross-device copy, in every branch including nested `rel` paths in the multi-file loop.
- **The trailing-space negative assertions** (``ToNot(ContainSubstring(`-o "$MODEL_PATH" `))``) are not vacuous today. I built both the fixed and the reverted command strings and grepped the literal needles: they are absent from the fixed form and present in the mutated form for all three branches. So the guard genuinely bites.
- Coverage matches #1428's three named sites, including the local `cp` branch that the prose summary does not call out by name but the diff does fix.
- Prefetch gets this for free, since #1218 reuses `buildModelStorageConfig` rather than keeping a second copy of the script.

One thing worth a follow-up rather than a change here.

**Nothing ever cleans up the `.tmp` files.** The fix is right, and a failed transfer leaving debris is much better than a truncated model being served forever. But there is no sweep anywhere: `llmkube cache clear` only prints `kubectl exec` instructions, the prefetch Job's owner-ref cleanup removes the Job and not files, and nothing in the reconcile path touches PVC contents. A transfer that is interrupted and then abandoned (Model deleted, or source re-pointed) leaves a `.tmp` on the shared PVC permanently. It also silently inflates `llmkube cache list`, since `cache_inspect.go` does `du -sb` over the whole cache-key subtree and cannot tell model bytes from tmp cruft.

This is not new in kind, the OnChange path has had it since #1326, but this PR extends it to every non-OnChange path, and `IfNotPresent` is the default refresh policy, so it will happen more often. Filing a follow-up; no need to grow this PR.

Good catch on the underlying bug, and the mutation check on the cached-https branch is the right way to have shown it.

PR #1433, opened 2026-08-06
docs(chart): dashboard datasources examples must use the datasource UID, not its display name

### LINKED ISSUE #1430: [BUG] grafana.dashboards.operator.datasources example passes a display name where grafana-operator substitutes a datasource UID
## Bug Description

The `grafana.dashboards.operator.datasources` example in values.yaml ([#L313-L317](https://github.com/defilantech/LLMKube/blob/36631a9/charts/llmkube/values.yaml#L313-L317)) and the chart README show `datasourceName: VictoriaMetrics` — a display name. But grafana-operator's `resolveDatasources` does a literal `${inputName}` → value `ReplaceAll` over the dashboard JSON, and these dashboards (schemaVersion 39, no `__inputs` block) carry the placeholder inside `"uid"` fields — so whatever string is configured lands verbatim as a datasource **UID**. Passing a display name writes `"uid": "VictoriaMetrics"`, which resolves to nothing, and every panel of every dashboard renders "datasource not found". No error is raised anywhere; the failure is silent.

The documented form only works in the coincidence where a datasource's UID equals its name. We hit this exactly: our datasource is name=`Mimir`, uid=`prom` — following the values.yaml example verbatim would have blanked all eight dashboards (caught pre-merge by inspecting the substituted JSON).

## Steps to Reproduce

1. Configure `grafana.dashboards.operator.datasources` per the values.yaml example, using your Prometheus datasource's display name.
2. Let grafana-operator reconcile the GrafanaDashboard CRs.
3. Every panel shows a datasource error, unless name == uid.

## Expected Behavior

The example demonstrates a value that works: the datasource **UID**, with a comment stating the constraint.

## Actual Behavior

The example demonstrates a display name, which silently produces dead dashboards on most real Grafana instances.

## Environment

**LLMKube Version:** `main` @ 36631a9 / chart 0.9.14; grafana-operator v5.

## Proposed Fix

Docs-only: change the example value to something uid-shaped and add one comment line ("must be the datasource *UID*, not its display name — the value is substituted into the dashboards' `uid` fields") in values.yaml and the README. PR incoming.

---

brandonliuli:
## What

Docs-only: the `grafana.dashboards.operator.datasources` examples in values.yaml and the chart README now use a UID-shaped value and state the constraint, instead of the display-name example (`datasourceName: VictoriaMetrics`).

## Why

grafana-operator's `resolveDatasources` does a literal `${inputName}` → value `ReplaceAll` over the dashboard JSON, and the shipped dashboards (schemaVersion 39, no `__inputs`) carry the placeholder inside `"uid"` fields — so the configured value lands verbatim as a datasource **UID**. Following the documented display-name example silently blanks every panel of every dashboard on any Grafana where a datasource's UID differs from its name (ours: name=`Mimir`, uid=`prom`; the documented form would have produced eight dead dashboards — caught pre-merge by inspecting the substituted JSON).

Fixes #1430

## How

Comment/README text only; no template or values-shape change. The test fixtures in `tests/dashboards_test.yaml` keep their existing placeholder value since they assert pass-through rendering, not datasource semantics.

## Checklist

- [ ] Tests added/updated (docs-only; `helm unittest` still 60/60, `helm template` clean)
- [x] `make test` passes locally (no Go change)
- [x] `make lint` passes locally (6 pre-existing staticcheck findings in `pkg/foreman/agent` tests, identical on the base commit; nothing from this change)

---
Assisted-by: Claude (Anthropic). Investigated, verified, and reviewed by a human; the blank-dashboard failure mode was reproduced against a live grafana-operator v5 install.

### DIFF
--- a/charts/llmkube/README.md
+++ b/charts/llmkube/README.md
@@ -268,8 +268,12 @@ desired state`), so an upgrade that flips it to `false` fails until the CRs are
 deleted.
 
 `operator.datasources` is a literal `${inputName}` -> `datasourceName` replace
-over the dashboard JSON before it reaches Grafana. Most of the shipped
-dashboards read their datasource from a `DS_PROMETHEUS` template variable;
+over the dashboard JSON before it reaches Grafana. `datasourceName` must be the
+datasource **UID**, not its display name: the placeholder sits inside the
+dashboards' `"uid"` fields, so a display name lands verbatim as a uid that
+resolves to nothing and every panel silently renders "datasource not found"
+(unless the UID happens to equal the name). Most of the shipped dashboards read
+their datasource from a `DS_PROMETHEUS` template variable;
 `amd-gpu-observability.json` uses one named `datasource`. None of them declares
 a dashboard `__inputs` block, so both names are plain template variables and
 covering every dashboard takes two entries:
@@ -285,9 +289,9 @@ grafana:
           dashboards: grafana
       datasources:
         - inputName: DS_PROMETHEUS
-          datasourceName: VictoriaMetrics
+          datasourceName: prometheus-uid # the datasource's UID, not its name
         - inputName: datasource
-          datasourceName: VictoriaMetrics
+          datasourceName: prometheus-uid
 ```
 
 The two dashboards under `config/grafana/` are not part of the chart and stay
--- a/charts/llmkube/values.yaml
+++ b/charts/llmkube/values.yaml
@@ -311,9 +311,14 @@ grafana:
       # CRD default (5m).
       resyncPeriod: ""
       # Literal ${inputName} -> datasourceName replace over the dashboard JSON.
+      # datasourceName must be the datasource UID, not its display name: the
+      # placeholder sits inside the dashboards' "uid" fields, so a display name
+      # lands verbatim as a uid that resolves to nothing and every panel
+      # silently renders "datasource not found" (unless the UID happens to
+      # equal the name).
       #   datasources:
       #     - inputName: DS_PROMETHEUS
-      #       datasourceName: VictoriaMetrics
+      #       datasourceName: prometheus-uid
       datasources: []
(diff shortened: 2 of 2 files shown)


Defilan (approved):
Approving. The diagnosis is right and I confirmed it against the shipped assets.

All eight dashboards under `charts/llmkube/dashboards/` carry the placeholder inside a `"uid"` field, e.g.

```json
"datasource": {"type": "prometheus", "uid": "${DS_PROMETHEUS}"}
```

and `grep -c '__inputs'` is 0 across all eight. So grafana-operator's literal `${inputName}` replace does land the configured value verbatim as a UID, and the old `datasourceName: VictoriaMetrics` example would blank every panel on any Grafana where the datasource's UID differs from its display name. Documenting the constraint is the right fix.

One spot the sweep missed, if you want to fold it in:

https://github.com/defilantech/LLMKube/blob/ab22a2898ec4648a1d8a5e0308452d71c36ade27/charts/llmkube/tests/dashboards_test.yaml#L154-L169

The fixture still uses `datasourceName: VictoriaMetrics`, the exact display-name shape this PR is correcting. It is functionally fine, since the assertion only checks pass-through into `GrafanaDashboard.spec.datasources` and the chart never interprets the value, so I am not going to hold the PR for it. But it is the same misleading example one directory over, and a fixture is a plausible place for someone to copy a pattern from. Your call whether to change it here or leave it.

Repo-wide grep for `datasourceName` finds nothing else outstanding: README and values.yaml hits are all covered by this diff, and there are no hits under `docs/site/` or `config/grafana/`.

brandonliuli:
The fixture spot you flagged is picked up in a follow-up (the PR merged before it could be folded in): #1436 — value-only, same reasoning as your note. Thanks for the thorough verification against the shipped assets.

PR #1442, opened 2026-08-07
fix(foreman): add chart validation to the coder gate, and install helm so it can run

### LINKED ISSUE #1441: [BUG] Coder gate runs no chart validation, so a chart-only change that breaks every helm test still returns GATE-PASS
## Bug Description

`DefaultGateChecks` contains no check that renders or validates the Helm chart, so a chart-only submission can break **every** chart test and still be reported GATE-PASS.

Observed on a real coder run (#1439, 2026-08-07): the branch added `controllerManager.caCertConfigMap` to `values.yaml` and a templated arg, but did not declare the key in `values.schema.json`. Since `controllerManager` sets `additionalProperties: false`, Helm rejects the value and the whole suite errors out:

| ref | `helm unittest charts/llmkube` |
|---|---|
| `main` | 70 passed |
| coder branch | **72 failed, 72 errored, 0 passed** |

The gate ran for 318s and returned **GATE-PASS**. It was not a false clone: the gate correctly cloned the fork where the branch lives (verified in the Job pod spec) and genuinely ran its checks. It simply has no check that would exercise a chart.

## To Reproduce

1. On a branch, add any key under `controllerManager` in `charts/llmkube/values.yaml` without adding it to `charts/llmkube/values.schema.json`.
2. Run the coder gate against that branch.
3. Gate reports GATE-PASS.
4. Run `helm lint charts/llmkube` or `helm unittest charts/llmkube` and see the chart is broken.

## Expected behavior

A submission that breaks the chart fails the gate.

## Actual behavior

GATE-PASS, because the check list is:

```go
// pkg/foreman/agent/tools/run_gate_job.go:66
var DefaultGateChecks = []string{
	"fmt", "vet", "lint", "test",
	"manifests", "chart-crds", "foreman-chart-crds",
}
```

`fmt`, `vet`, `lint`, `test` are Go-only. `chart-crds` and `foreman-chart-crds` regenerate CRDs *into* the chart and assert no drift; neither renders the chart, resolves values against the schema, nor runs the chart's own test suite. So the three chart-shaped checks all miss chart correctness.

This is a coverage gap rather than a wrong-ref problem, and it makes the gate structurally blind to an entire class of change. `charts/` is a first-class part of this repo: the chart carries 70+ of its own tests, a values schema, and a dedicated CI workflow.

## Proposed fix

Add a chart check to the gate. `helm lint` alone is sufficient to have caught this case (confirmed: it fails on the offending branch), and `helm unittest` is the stronger signal since the chart already ships 70+ tests.

Both already run in CI (`.github/workflows/helm-chart.yml` does `helm lint` and installs helm-unittest v1.1.1), so the gate is weaker than CI on exactly the surface where a coder is most likely to be wrong.

Two things to decide as part of the fix:

1. **Whether the chart check runs unconditionally or only when `charts/` is touched.** Unconditional is simpler and matches how the Go checks behave; conditional saves gate time on Go-only branches.
2. **Where the make target comes from.** There is currently no `helm lint`/`helm unittest` Makefile target, so either add one (e.g. `make test-chart`) and put it in `DefaultGateChecks`, or teach the gate job template to run helm directly. A Makefile target is more consistent, since every existing check is a make target.

The gate image also needs `helm` and the unittest plugin available, which CI already demonstrates how to install.

## Additional Context

Found while reviewing a Foreman submission for #1439. The coder's work was otherwise correct and well-scoped; the missing schema entry was a single line. The concern is not that the coder erred, it is that the gate reported the branch as verified when it was not, and only a human running `helm unittest` by hand caught it. The fix landed in #1440.

Related: #768 and #1135 on gate verdict trustworthiness.

## Priority

- [ ] Critical
- [x] High
- [ ] Medium
- [ ] Low

High: any chart-only coder submission currently gets an unearned GATE-PASS, and chart changes are a common Foreman task shape.

Defilan:
## What

Adds a `test-chart` make target (`helm lint` + `helm unittest` over both charts), puts it in `DefaultGateChecks`, and installs helm in the gate Job so the check can actually execute.

Fixes #1441

## Why

`DefaultGateChecks` was Go-only. Its three chart-*named* checks (`manifests`, `chart-crds`, `foreman-chart-crds`) regenerate CRDs into the chart and assert no drift; none renders the chart, resolves values against `values.schema.json`, or runs the chart's own tests.

Measured on a real coder branch (#1440): a chart-only change that broke **every** helm test still returned GATE-PASS after 318s of checks.

| ref | `helm unittest charts/llmkube` |
|---|---|
| `main` | 70 passed |
| that branch | 72 failed, 72 errored, 0 passed |

## How

**`make test-chart`** runs `helm lint` and `helm unittest` over `charts/llmkube` and `charts/foreman`, added to `DefaultGateChecks`.

**The gate Job installs helm.** This is the part that makes the check real rather than decorative: the gate image is `golang:1.26`, which has no helm and no unittest plugin. I verified this directly against the running gate image, and without the install the new check fails with `helm: command not found` on *every* gate run, Go-only branches included, which would be worse than the bug being fixed.

The install deliberately mirrors `.github/workflows/helm-chart.yml` instead of using `helm plugin install`, for the reason that workflow already documents: helm 3.x cannot skip signature verification, and the plugin's `plugin.yaml` carries `platformHooks`, a field added after 3.13 that older loaders reject. Confirmed empirically in the gate image:

```
$ helm plugin install https://github.com/helm-unittest/helm-unittest --version 1.1.1
Error: plugin is installed but unusable: failed to load plugin at
".../plugin.yaml": error unmarshaling JSON: unknown field "platformHooks"
```

So it fetches the pinned tarball, unpacks into `$(helm env HELM_PLUGINS)`, and strips the field, exactly as CI does.

Versions are pinned to CI's (`helm v3.13.0`, `unittest 1.1.1`) so the gate and CI validate charts with the same tooling. A gate that disagrees with CI is worse than no gate.

The binary and plugin land on the existing gate cache PVC (`/cache/bin`, `XDG_DATA_HOME=/cache/xdg`), so a warm cache makes this a no-op. The install renders **only when a chart check is requested**, so Go-only runs pay nothing.

## Testing

End-to-end in the actual gate image (`golang:1.26`), cloning this branch and running the real check:

```
install OK in 3s
=== make test-chart ===
Charts:      2 passed, 2 total
Test Suites: 8 passed, 8 total
Tests:       102 passed, 102 total
```

3 seconds against a gate run that already spends minutes fetching golangci-lint, controller-gen and the envtest binaries.

Unit tests: `TestGateJobInstallsHelmOnlyForChartChecks` asserts the install renders for chart checks, does **not** render for Go-only checks, pins the tarball version, and keeps the `platformHooks` strip. `TestNeedsHelm` covers the predicate and that `DefaultGateChecks` contains the chart check. Both bite-checked: deleting the install block from the template fails the first test and nothing else.

`go test ./pkg/foreman/agent/tools/` passes.

## Note on provenance

The `test-chart` target and the `DefaultGateChecks` entry came from the Foreman coder (Qwopus3.6-27B-Fusion on Strix), which returned GO and GATE-PASS. That GATE-PASS could not have meant anything: `DefaultGateChecks` is compiled into the agent binary, so the gate verifying the branch ran the *old* check list. It was structurally incapable of testing its own change. The missing helm install was found by probing the gate image by hand.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (the template and target carry the rationale inline)

## AI assistance

Assisted-by: Foreman harness (coder) for the make target and check-list entry; Claude Code for the gate-image investigation, the CI-mirroring install, the tests, and the end-to-end verification. I reviewed and ran everything before submitting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,6 +110,11 @@ vet: ## Run go vet against code.
 test: manifests generate fmt vet setup-envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
 
+.PHONY: test-chart
+test-chart: ## Lint and unit-test the Helm charts (requires helm + helm-unittest plugin).
+	helm lint charts/llmkube charts/foreman
+	helm unittest charts/llmkube charts/foreman
+
 # TODO(user): To use a different vendor for e2e tests, modify the setup under 'tests/e2e'.
 # The default setup assumes Kind is pre-installed and builds/loads the Manager Docker image locally.
 # CertManager is installed by default; skip with:
--- a/pkg/foreman/agent/tools/gate_job_template.yaml
+++ b/pkg/foreman/agent/tools/gate_job_template.yaml
@@ -88,6 +88,48 @@ spec:
               echo "=== {{ firstLine . }} ==="; ( {{ indentShell . }} ) || rc=1
               {{- end }}
               {{- else }}
+              {{- if .HelmVersion }}
+              # helm + helm-unittest for the chart checks (#1441). The gate
+              # image is a plain golang image with no helm, so a chart check
+              # would fail with "helm: command not found" without this.
+              #
+              # This deliberately mirrors .github/workflows/helm-chart.yml
+              # rather than using `helm plugin install`, for the reason that
+              # workflow documents: helm 3.x has no --verify skip and the
+              # runner has no GPG keyring, and the plugin's plugin.yaml
+              # carries `platformHooks`, a field added after 3.13 that older
+              # loaders reject outright ("unknown field platformHooks", which
+              # makes the next `helm plugin list` hard-fail). So: fetch the
+              # tarball, unpack into HELM_PLUGINS, strip the field.
+              #
+              # Versions are pinned to CI's so the gate and CI validate charts
+              # with the same tooling; a gate that disagrees with CI is worse
+              # than no gate.
+              #
+              # Installed under the gate cache PVC (/cache/bin, and the plugin
+              # under XDG_DATA_HOME=/cache/xdg), so a warm cache makes this a
+              # no-op. Cold it is ~17MB and a few seconds, against a run that
+              # already fetches golangci-lint, controller-gen and envtest.
+              export PATH="/cache/bin:$PATH"
+              mkdir -p /cache/bin
+              if ! command -v helm >/dev/null 2>&1; then
+                echo "=== install helm {{ .HelmVersion }} ==="
+                curl -sSfL "https://get.helm.sh/helm-{{ .HelmVersion }}-linux-amd64.tar.gz" -o /tmp/helm.tgz \
+                  && tar -xzf /tmp/helm.tgz -C /tmp \
+                  && install /tmp/linux-amd64/helm /cache/bin/helm \
+                  || { echo "GATE FAIL: helm install"; exit 1; }
+              fi
+              if ! helm plugin list 2>/dev/null | grep -qi unittest; then
+                echo "=== install helm-unittest {{ .HelmUnittestVersion }} ==="
+                PLUGIN_DIR="$(helm env HELM_PLUGINS)/unittest"
+                rm -rf "$PLUGIN_DIR"; mkdir -p "$PLUGIN_DIR"
+                curl -fsSL "https://github.com/helm-unittest/helm-unittest/releases/download/v{{ .HelmUnittestVersion }}/unittest-{{ .HelmUnittestVersion }}.tgz" \
+                  | tar -xz -C "$PLUGIN_DIR" --strip-components=1 \
+                  || { echo "GATE FAIL: helm-unittest download"; exit 1; }
+                sed -i '/platformHooks/,$d' "$PLUGIN_DIR/plugin.yaml"
+                helm plugin list >/dev/null 2>&1 || { echo "GATE FAIL: helm-unittest unusable"; exit 1; }
+              fi
+              {{- end }}
               {{- range .Checks }}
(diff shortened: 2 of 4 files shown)


joryirving (approved):
## Review: fix(foreman): add chart validation to the coder gate

**Overall: Looks good.** Problem is real, solution is correct, execution is clean.

### What's solid

- **Clear problem**: `DefaultGateChecks` was Go-only; the chart-*named* checks just regenerate CRDs. A chart-only change that broke every helm test still GATE-PASS'd.
- **Conditional install**: `helmVersionFor()` returns empty string for Go-only runs — they pay nothing. The `{{- if .HelmVersion }}` template guard is clean.
- **Version pinning**: Matches CI exactly (`v3.13.0` / `unittest 1.1.1`). A gate that disagrees with CI is worse than no gate.
- **Tests are bite-checked**: \`TestGateJobInstallsHelmOnlyForChartChecks\` asserts install renders for chart checks only, asserts pinned tarball URL, asserts platformHooks strip. \`TestDefaultGateChecks_ContainsTestChart\` is a proper regression guard. \`TestNeedsHelm\` covers the predicate.
- **Honest PR description**: The provenance note about the coder's GATE-PASS being structurally meaningless (DefaultGateChecks is compiled in) is exactly the kind of thing that should be called out.

### Follow-ups (not blocking)

1. **sed vs awk for platformHooks strip** — Gate uses `sed -i '/platformHooks/,$d'` (deletes to EOF). CI uses a scoped awk that only skips the indented block under the key. Fine today (platformHooks is last key), but fragile on version bump. Consider mirroring CI's awk for consistency.

2. **No SHA256 on helm binary download** — Low risk (pinned version, in-cluster, deterministic URL), but CI uses `azure/setup-helm` with verification. Worth noting if the gate ever runs in a less trusted context.

3. **Struct field re-alignment in \`rendererInput\`** — purely cosmetic (gofmt moved alignment). Noting so nobody wonders if those lines are meaningful.

### Verdict

**Approve.** The sed-vs-awk is the only thing worth a follow-up, and it's not blocking given the version pin.

PR #1446, opened 2026-08-08
fix(dashboard): AMD dashboard vendor hygiene — remove runtime-agnostic llama.cpp panels, standardize on DS_PROMETHEUS

### LINKED ISSUE #1429: [BUG] amd-gpu-observability: four runtime-agnostic llama.cpp panels populate on non-AMD fleets, and it alone uses a ${datasource} placeholder
## Bug Description

Two related consistency problems in `charts/llmkube/dashboards/amd-gpu-observability.json`, found while load-verifying all eight shipped dashboards:

**1. Four of its sixteen queries are runtime-agnostic llama.cpp panels, so the AMD dashboard populates on fleets with zero AMD GPUs.** The breakdown: 8 queries hit `amdgpu_*`, 4 hit `node_hwmon_*` (not AMD-specific), and 4 — the panels "Tokens / Second", "Tokens / Second History", "In-Flight Requests", "In-Flight Requests History" — hit `llamacpp:predicted_tokens_seconds` / `llamacpp:requests_processing`, which any llama.cpp workload emits regardless of GPU vendor. Verified live: on an all-NVIDIA cluster, "LLMKube AMD GPU Monitor" went from 0 to 4 populated panels purely from an NVIDIA T4 llama.cpp load test, and back to 0 when it stopped. A dashboard with that title showing live token throughput on a fleet with no AMD silicon reads as "AMD GPUs are active here" — actively misleading, the same class of problem as #1357.

Suggested fix: drop the four llama.cpp panels (the llamacpp-dashboard added in #1359 already covers them properly), or scope their queries to AMD-backed workloads. Given #1358 gated dashboards on their feature, panel-level vendor hygiene seems in the same spirit.

**2. It is the only dashboard using a `${datasource}` placeholder — the other seven use `${DS_PROMETHEUS}`.** Since `grafana.dashboards.operator.datasources` maps one `inputName` per entry, a consumer who configures the documented `DS_PROMETHEUS` mapping gets seven working dashboards and one silently unbound one (its panels point at a nonexistent datasource). The chart README works around it by showing both mappings, but the trap remains for anyone who maps only the conventional name. Standardizing on `${DS_PROMETHEUS}` removes the special case and simplifies the README.

## Steps to Reproduce

1. Deploy the eight chart dashboards to a Grafana on a cluster with NVIDIA GPUs only, mapping `inputName: DS_PROMETHEUS` per the docs.
2. Run any llama.cpp InferenceService under load.
3. amd-gpu-observability shows live token/request panels (false positive); it is also the one dashboard whose remaining panels can never bind if only `DS_PROMETHEUS` was mapped.

## Expected Behavior

The AMD dashboard is dark on a non-AMD fleet, and all eight dashboards bind with the single documented placeholder mapping.

## Environment

**LLMKube Version:** dashboards fetched from `main` @ 36631a9 (delivered via grafana-operator `spec.url`); load test on chart 0.9.14, AKS, NVIDIA T4.

Happy to send the JSON PR for either or both if the direction sounds right.

---

brandonliuli:
Fixes #1429. Two commits.

> **Updated after review.** Commit 2 now **removes** the four llama.cpp panels rather than gating them, per @Defilan's finding below. The gating revision is in this PR's history if it is ever wanted.

## 1. `datasource` variable renamed to `DS_PROMETHEUS`

`amd-gpu-observability.json` was the only shipped dashboard naming its datasource template variable `datasource`; the other seven use `DS_PROMETHEUS`. Since `grafana.dashboards.operator.datasources` maps one `inputName` per entry, a consumer who configured the documented `DS_PROMETHEUS` mapping got seven working dashboards and one whose panels pointed at a datasource that doesn't exist — silently, because an unbound placeholder renders as "datasource not found" per panel rather than failing the sync.

The variable definition was already byte-identical to the canonical one apart from its name, so this is purely the rename plus its 31 references. With every dashboard on the same name, the chart README's `operator.datasources` example drops from two mapping entries to the single one its surrounding prose already recommends.

## 2. The four llama.cpp panels are removed

`llamacpp:predicted_tokens_seconds` and `llamacpp:requests_processing` are emitted by any llama.cpp workload regardless of GPU vendor, so on an all-NVIDIA fleet the "LLMKube AMD GPU Monitor" went from zero to four populated panels purely from an NVIDIA load test, and back to zero when it stopped.

**Why removal rather than the gating this PR first proposed.** The gated form (`and on() (count(amdgpu_gpu_busy_percent) > 0)`) only darkens a fleet with *no* AMD hardware. @Defilan's fleet has both vendors in one cluster — `count(amdgpu_gpu_busy_percent)` returns 2 alongside 4 llama.cpp series — so the guard passes on the AMD metrics while NVIDIA-driven throughput still renders under the AMD title. Gating therefore fails on the exact topology the issue was reported against, which settles it: the panels come out.

Removed: "Tokens / Second", "In-Flight Requests", their two History counterparts, and the two now-empty Inference rows. The six-row gap that leaves in the grid is closed so the History section doesn't float (`y` is now contiguous: 0, 1, 6, 7, 12, 13, 21). Every remaining query is AMD-specific — `amdgpu_*`, or `node_hwmon_*` joined on `chip_name="amdgpu"`.

Nothing is lost: `llamacpp-dashboard.json` (#1359) already presents these signals as the richer per-service view, grouped by service and namespace. `docs/grafana/README.md` advertised the removed panels, so its catalog entry is updated and now points there for throughput.

For the record, per-workload scoping — showing only throughput from AMD-backed pods, which would have satisfied both topologies — was considered and is not portable: the `llamacpp` series carry `cluster`, `container`, `instance`, `job`, `namespace`, `pod`, `runtime` and `service` but no `node`, and their `instance` is a pod address while the `amdgpu` series are node-scoped. Attributing a pod to a GPU vendor needs a `kube_pod_info` join through kube-state-metrics, which this chart does not otherwise require.

## Verification

- `helm unittest charts/llmkube` — 72/72 across all 8 suites; `go test ./internal/metrics/...` passes; `make lint` — 0 issues. All on this branch, rebased onto current `main`.
- Post-change checks on the JSON: 14 panels, single `${DS_PROMETHEUS}` placeholder, zero remaining `llamacpp` references, contiguous grid.
- `TestDocContractMetrics` passes and `docs/observability/amd-gpu-metrics.md` is untouched — that test asserts contract-block metrics are *produced* by something, not that a dashboard queries them, so the `llamacpp:*` names there remain accurate.
- The false-positive behavior was observed live on an all-NVIDIA AKS cluster during the load-verification pass that produced #1429.

Assisted-by: Claude Code (made the JSON edits, the grid recompaction, and drafted this description; I verified the label sets that rule out a per-workload join, and ran the chart, metrics and lint suites myself).


Defilan (commented):
Commit 1 is unambiguous and I am happy with it. Confirmed `amd-gpu-observability.json` was genuinely the only one of the eight shipped dashboards naming its variable `datasource`.

Commit 2 works as designed, and I want to flag something you had no way to see from outside the cluster. Against my Prometheus:

```
count(amdgpu_gpu_busy_percent)           -> 2
count(llamacpp:predicted_tokens_seconds) -> 4
```

The guard metric is present and the gate passes, so the panels do render on AMD hardware. That resolves the obvious risk with gating on a metric that needs a textfile collector on gfx1151.

But it also means my fleet is exactly the mixed AMD and NVIDIA topology you flagged as not covered: the gate passes and NVIDIA-driven llama.cpp throughput still shows on a dashboard titled "LLMKube AMD GPU Monitor". On the fleet this was reported against, commit 2 does not remove the false positive it targets.

You offered to swap the commit for removal and noted removal is strictly stronger in the mixed case. I would like to take you up on that, since `llamacpp-dashboard.json` (#1359) already covers those signals as the richer per-service view. If you would rather split them, I am happy to merge commit 1 immediately and take the removal separately.

brandonliuli:
Swapped commit 2 to removal — thanks for running the counts, that settles it. Your fleet is the case gating cannot reach: the guard passes on your AMD metrics and the NVIDIA-driven throughput still renders under the AMD title, so on the exact topology this was reported against the gate buys nothing.

`65762c9` removes the four panels plus the two Inference rows they leave empty, and closes the resulting six-row gap so the History section does not float. The grid is contiguous again (`y` = 0, 1, 6, 7, 12, 13, 21), and every remaining query is `amdgpu_*` or `node_hwmon_*` joined on `chip_name="amdgpu"`. `docs/grafana/README.md` now points at `llamacpp-dashboard.json` for throughput.

I kept both commits in this PR since that is one merge for you, but I am happy to split if you would rather land commit 1 now and take the removal separately.

Worth recording for the future: per-workload scoping would have satisfied both topologies, but it is not portable here. The `llamacpp` series carry `cluster`, `container`, `instance`, `job`, `namespace`, `pod`, `runtime` and `service` — no `node` — and their `instance` is a pod address while the `amdgpu` series are node-scoped. Attributing a pod to a GPU vendor needs a `kube_pod_info` join through kube-state-metrics, which is more dependency than this dashboard warrants.

Defilan (approved):
Looks good! Approved.

Issue #644, opened 2026-06-10
[FEATURE] Reviewer integrity enforcement: demote verdicts when issueAsk cannot be verified


Defilan:
## Feature Description

PR #587 ground-truths the reviewer's `issueAsk` claim against the issue body the model itself fetched, recording `issueAskVerified: true|false` in the result extra. Today this is observe-only: the verdict stands even when the model demonstrably fabricated its understanding of the issue.

A 5-branch review battery with Mellum2-12B (2026-06-10) showed why observe-only is not enough: `issueAskVerified=false` on 5/5 runs, including a GO on a known scope-drift branch and a NO-GO justified by an entirely hallucinated ask ("add mlx model format support" for an issue that asks for a one-line docs nudge). Both confidently wrong verdicts stood.

## Proposed Behavior

In the native executor's reviewer path, after `reconcileReviewerIssueAsk`:

- `issueAskVerified == false` and verdict GO: demote to NO-GO. A reviewer that cannot prove it read the issue must not approve a branch.
- `issueAskVerified == false` and verdict NO-GO: keep NO-GO but mark it untrusted.
- In both cases annotate the result extra: `verdictDemoted: true`, `verdictClaimed: <original>`, `demotionReason: <why>` so operators and the escalation reviewer can see what happened.
- `issueAskVerified` absent (no fetch_issue body in the transcript, a harness-side gap, not model dishonesty): unchanged, observe-only.

Because #639 emits escalation reviewers on base NO-GO, demotion automatically routes integrity failures to a bigger model instead of letting them stand: no controller changes needed.

## Alternatives Considered

Demoting to INCOMPLETE is semantically cleaner ("could not produce a trustworthy review") but does not trigger escalation emission, which counts only NO-GO; extending the trigger is a larger #639 design change. Worth revisiting if NO-GO data hygiene becomes a problem.

## Additional Context

Direct follow-on to #587 (ground-truthing) and the reviewer-diversity findings. First of two "review integrity" levers; the second (computable scope-overlap pre-check) will be filed separately.

Issue #582, opened 2026-05-28
[BUG] Foreman reviewer (devstral) hallucinates submit_result.extra fields even when fetch_issue/read_file returned correct data


Defilan:
## Bug Description

The v0.4 reviewer Agent on Mac Studio (devstral-24b-instruct-2512 Q6_K)
calls `fetch_issue` and `read_file` correctly with the real diff's
files, but its terminal `submit_result.extra` payload includes
hallucinated fields:

- `filesTouched`: contains paths that are not in `git diff --stat
  main...HEAD`. Observed on rerun-10-rereview #449 (devstral listed
  `internal/controller/modelrouter_controller.go`,
  `router_builders_test.go`, `router_config_builder.go`; the actual
  diff is `.goreleaser.yaml`, `charts/llmkube/values.yaml`,
  `docs/site/concepts/model-router.md`).
- `issueAsk`: reverts to a training-prior summary that does not quote
  from the issue body, even though `fetch_issue` returned the real body
  to the model earlier in the same run. Observed twice on #526
  reviews: the same incorrect "Add a reconciler that cleans up
  orphaned Service+Endpoints..." text appears both pre-#581 and
  post-#581.

This means the terminal `submit_result` cannot be trusted on
multi-file diffs even when every upstream tool call returned correct
data. The reviewer's findings often cite the right code (devstral
read the right files), but its envelope fields (filesTouched,
issueAsk) are unreliable.

## Steps to Reproduce

1. Apply a review AgenticTask targeting a branch with a multi-file
   diff (3+ files) via the devstral-24b-reviewer Agent.
2. Inspect the transcript ConfigMap: `fetch_issue` and `read_file`
   tool results are correct; the model browses the right files.
3. Inspect `status.result.extra.modelExtra`: `filesTouched` does not
   match `git diff --stat main...HEAD`; `issueAsk` does not appear
   verbatim in the issue body the `fetch_issue` tool returned.

## Expected Behavior

`submit_result.extra.filesTouched` should always equal the diff's
file list. `extra.issueAsk` should be a verbatim quote from the issue
body the reviewer fetched.

## Proposed Fix (server-side, model-agnostic)

Have the executor compute `git diff --stat main...HEAD` after the
review branch is checked out and overwrite `result.extra.filesTouched`
with the ground-truth list at terminal time. Preserve the model's
claim under `extra.filesTouchedClaimed` for analysis. Tighten the
reviewer prompt to require `issueAsk` be a verbatim quote and to
[...]

PR #1448, opened 2026-08-08
fix: ground the PR body summary against the branch diff

### LINKED ISSUE #1411: [BUG] PR body is not grounded against the diff — coder summaries can claim work that does not exist
## Problem

The PR body foreman renders from the coder's `submit_result` summary is not grounded against the diff, so it can claim work that does not exist. Every other model-supplied claim in this pipeline already gets ground-truthed — `filesTouched` is rewritten from `git diff --name-only` (#582), `issueAsk` is verified against the fetched issue body (#644), findings must cite lines the diff changed — but the summary that becomes the **PR description a human reads at merge time** is passed through verbatim.

## Observed

`misospace/foreman-dispatch-bridge#90` (structured logging, issue #51). Body opened with:

> Introduces structured JSON logging across the bridge, replacing all bare `print()` calls with `logger.info()/warning()/error()`, **and adds a `/health` endpoint.**

The diff contains zero occurrences of "health" — no endpoint, no server, no route. The logging half was real and good (20 tests, CI green, reviewer APPROVED); the endpoint was invented. Nothing in the pipeline noticed, because:

- the deterministic scope-overlap rail only fires when the diff touches *none* of the files the issue names — it did touch them;
- `issueAsk` verification compares the reviewer's stated ask to the *issue*, not the coder's summary to the *diff*;
- the reviewer read the diff and approved it correctly — the false claim is in the PR body, which is not part of what the reviewer is asked to verify.

Secondary defect on the same PR: the rendered body was **truncated mid-sentence** (`... plus in \`claim.…`), suggesting a length cap applied to the summary without ellipsis-safe handling.

## Why it matters

The PR body is the artifact a human trusts when deciding to merge. A body that overstates the change is worse than no body: it invites a merge on false premises, and it silently "closes" gaps the issue raised. This is the same failure class as #582/#644 — model-authored metadata treated as fact — just pointed at the description instead of the fields.

It also interacts badly with `Fixes #N`: an overclaiming body plus auto-close means an issue can be closed as satisfied by work that was never done.

## Suggested fix

Ground the summary the same way the other claims are grounded. Cheapest useful version, in rough order of value:

1. **Cross-check the summary against the diff before rendering the body.** Extract concrete nouns/paths the summary claims (`/health`, file paths, symbol names) and verify each appears in the diff or the changed files; on mismatch either strip the sentence, or append a `⚠️ unverified claim` note, or fail the summary back to the coder for a rewrite.
2. **Archive the model's original summary** under `extra.summaryClaimed` (mirroring `issueAskClaimed`) whenever it is rewritten, so the drift is debuggable.
3. **Fix the truncation** — cap on a sentence/paragraph boundary, or state the cap explicitly rather than emitting a body that ends in `…`.

(1) is the load-bearing one. Even a narrow version — verify claimed file paths and quoted identifiers exist in the diff — would have caught this instance.

Happy to implement whichever shape you prefer; I've been through the adjacent rails in #1120/#1166/#1365.

joryirving:
## What

Cross-checks the summary that becomes the PR description against the branch diff
before rendering it, and caps `submit_result.summary` on a sentence boundary.

New `pkg/foreman/agent/summary_grounding.go` extracts the concrete, checkable
things a summary names — backticked file paths, route-shaped strings, code-shaped
identifiers — and requires each to appear in `git diff <base>...HEAD` or the
changed-file list. Claims the diff does not support get a `[!WARNING]` note
appended to the body; the model's original prose is archived under
`extra.summaryClaimed` with the failing claims in `extra.summaryUnverifiedClaims`.

Second commit: `truncateRuneSafe` backs off to the last sentence terminator when
one survives in the tail of the 280-byte budget.

**One correction to the issue's premise.** The issue says the body comes from the
*coder's* `submit_result` summary. It does not. `maybeOpenPullRequest` fires only
for `AgenticTaskKindReview` and passes `r.Summary` — the **reviewer's** summary.
The fix is unaffected (the reviewer read the diff to reach GO, but its prose is no
more ground-truthed than any other model-authored field, and it is still what a
human reads at merge time), but the mechanism in the issue text is wrong and the
code comments say so rather than repeating it.

## Why

Fixes #1411

Every other model-supplied claim in this pipeline is already ground-truthed —
`filesTouched` from `git diff --name-only` (#582), `issueAsk` against the fetched
issue body (#644), findings against the lines the diff changed (#1166) — but the
summary that becomes the PR description a human trusts at merge time was passed
through verbatim. Combined with `Fixes #N` auto-close, an overclaiming body can
close an issue as satisfied by work that was never done.

## How

**Mismatch handling: append, never rewrite.** The issue offered three options
(strip the sentence, append a note, fail back to the coder). Stripping requires
deciding which sentence a claim belongs to without any parse of the prose, and a
wrong guess silently destroys an accurate description — unrecoverable for the
human reading the PR. Appending is additive: the reviewer's words survive
verbatim and the disagreement between the words and the diff is stated where the
merge decision is made. The note renders as a GitHub `[!WARNING]` callout, so it
is impossible to miss and impossible to mistake for the reviewer's own prose. The
list is capped at 5 claims so a confabulated summary cannot bury the diff link.

**Deterministic, not semantic** — matching the rails it sits beside. No model
call, no embedding.

**Biased hard against false positives.** A correct summary that gets flagged puts
noise on every PR and trains humans to ignore the note, which is worse than the
bug it guards. Concretely:

- Only backticked spans are considered; unbackticked prose is never parsed, so
  ordinary English cannot trip the check.
- Grounding uses the **full** diff — added, removed, and context lines, plus file
  headers — not just added lines. "renames \`oldHandler\` to \`newHandler\`"
  therefore grounds `oldHandler` on the removal that renamed it.
- A claim with no literal hit falls back to its identifier atoms and passes if
  **any** atom appears. \`logger.info()\` grounds on `logger`.
- Identifiers must carry a code marker (underscore, dot, call suffix, or an
  uppercase letter) to be checkable at all. A bare lowercase word in backticks
  (\`logging\`, \`health\`) is indistinguishable from emphasis and is skipped.
- Path claims also ground on the file merely *existing* in the workspace, so
  "follows the pattern in \`pkg/foreman/agent/loop.go\`" is not reported as an
  invented path. Path claims skip the atom fallback in exchange — `pkg`,
  `internal`, and `cmd` appear in almost every diff and would ground a wholly
  fabricated path.
- Every missing input degrades open: no diff, a git error, an unresolved review
  base, or an empty summary all skip the check and preserve the model's prose.

**Plumbing.** `reviewBase` and `reviewDiff` are hoisted out of the reviewer-rails
block so the check reuses the base and name-only diff the grounded-finding and
scope-overlap rails already computed. One extra `git diff` is needed for the diff
*text* (name-only cannot ground `/health`), taken lazily after
`maybeOpenPullRequest`'s gate so non-PR paths do not pay for it.

**Truncation (issue item 3).** The cap is `MaxSubmitSummaryLen = 280` in
`pkg/foreman/agent/tools/submit_result.go`; it was a hard byte cut, which is what
produced `... plus in \`claim.…\`. It now backs off to the last sentence
terminator when the boundary keeps at least 60% of the budget, and keeps the hard
rune-safe cut otherwise so a summary whose only boundary sits near the start is
not gutted. A terminator counts only when followed by whitespace, so `v1.2` and
`logger.info` are not mistaken for sentence ends.

**Residual false-positive risk I could not eliminate.** An identifier claim about
pre-existing code the diff does not touch (\`Handler\`, when the change only calls
it and the call site is outside the diff context) can still be flagged. Grounding
identifiers against the whole repo instead of the diff would fix it but would also
have let this issue's `/health` through if "health" appeared anywhere in the tree,
so I kept the diff bound. The append-don't-rewrite choice is what makes this
acceptable: the cost is one visible line of noise, not a mangled description.

**Testing.** `pkg/foreman/agent/summary_grounding_test.go` covers the #1411
regression (the `/health` body against a logging-only diff) plus a false-positive
table — renames, call claims, bare words, backticked prose, fenced samples,
basename path matches, and missing ground truth. `executor_pr_test.go` covers the
wiring end to end: note in the body, `summaryClaimed` archived, `r.Summary` left
as the model's, and grounding skipped when there is no diff. Mutation-checked in
both directions: disabling the check fails 5 tests; making it over-eager (dropping
the English-word guard, the atom fallback, and the workspace-existence grounding)
fails 6, including the false-positive table.

## Checklist

- [x] Tests added/updated
- [x] `make lint` passes locally (and `make lint-all`, 0 issues on darwin + linux)
- [ ] Documentation updated (if user-facing change) — no user-facing docs surface; the behaviour is described in the package comment on `summary_grounding.go`

Assisted-by: Claude Code (wrote the grounding check, the truncation fix, and the tests from my issue and brief; I reviewed the design rationale and the change description, and ran `make test` and `make lint` locally before submitting)

### DIFF
--- a/pkg/foreman/agent/executor_native.go
+++ b/pkg/foreman/agent/executor_native.go
@@ -734,6 +734,12 @@ func (e *NativeAgentLoopExecutor) runLLMPath(
 		// (GO=APPROVE / NO-GO=REQUEST-CHANGES) is the authoritative
 		// signal regardless of findings validity (see pkg/foreman/agent/
 		// reviewer/findings.go for the schema).
+		//
+		// Declared out here so the PR-body summary grounding (#1411) can
+		// reuse the same base + name-only diff the reviewer rails ran on
+		// instead of resolving and shelling out for them a second time.
+		var reviewBase string
+		var reviewDiff []string
 		if agent.Spec.Role == foremanv1alpha1.AgentRoleReviewer &&
 			loopRes.Terminal != nil {
 			// Ground-truth filesTouched against the actual diff before
@@ -747,7 +753,7 @@ func (e *NativeAgentLoopExecutor) runLLMPath(
 			// reviewer clones the fork, whose local `main` lags upstream, so
 			// diffing against local `main` sweeps in the whole intervening
 			// upstream delta and neuters every rail (#1005).
-			reviewBase := e.reviewerDiffBase(ctx, log, task, workspace)
+			reviewBase = e.reviewerDiffBase(ctx, log, task, workspace)
 			reconcileReviewerFilesTouched(ctx, log, workspace, reviewBase, loopRes.Terminal.Extra)
 			// Ground-truth issueAsk against the fetch_issue tool result
 			// the model already had in its context. Devstral on the same
@@ -760,7 +766,8 @@ func (e *NativeAgentLoopExecutor) runLLMPath(
 			reconcileReviewerIssueAsk(log, loopRes.Transcript, loopRes.Terminal.Extra)
 			// Ground-truth the branch diff ONCE for the two computable rails
 			// below (grounded-finding + scope-overlap).
-			reviewDiff, reviewDiffErr := repo.DiffNameOnly(ctx, workspace, reviewBase)
+			var reviewDiffErr error
+			reviewDiff, reviewDiffErr = repo.DiffNameOnly(ctx, workspace, reviewBase)
 			// Grounded-finding rail: a NO-GO must be earned by >=1 blocking
 			// finding citing a line the diff changed; otherwise demote it to GO
 			// and archive the rejected findings. Mirror of scope-overlap, in the
@@ -810,7 +817,8 @@ func (e *NativeAgentLoopExecutor) runLLMPath(
 		// INCOMPLETE (this is NOT a GO); we only commit + push the branch and
 		// record it under r.Extra. Coder-role only (reviewers are read-only).
 		e.maybePreserveGateFailedBranch(ctx, log, agent, task, workspace, branch, auth, loopRes, r)
-		e.maybeOpenPullRequest(ctx, log, task, auth, verdict, r)
+		e.maybeOpenPullRequest(ctx, log, task, auth, verdict, r,
+			workspace, reviewBase, reviewDiff)
 		// Attach the normalized failure reason from the model-to-CRD
 		// mapping (e.g. ERROR→INCOMPLETE + ModelReportedError for #649). Only
 		// set when the normalizer produced a reason AND the result does
@@ -1065,7 +1073,10 @@ func (e *NativeAgentLoopExecutor) retryCoderTerminalResult(
 ) *Result {
 	r := e.modelDecidedResult(start, tref, lr, verdict)
 	e.maybePreserveGateFailedBranch(ctx, log, agent, task, workspace, branch, auth, lr, r)
-	e.maybeOpenPullRequest(ctx, log, task, auth, verdict, r)
+	// No reviewer rails ran on this path, so there is no resolved review base
+	// to ground the summary against; #1411's check degrades open on the empty
+	// base rather than grounding against a base it had to guess at.
+	e.maybeOpenPullRequest(ctx, log, task, auth, verdict, r, workspace, "", nil)
 	if normReason != "" && r.FailureReason == "" {
 		r.FailureReason = normReason
 	}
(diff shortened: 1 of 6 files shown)


Defilan (approved):
Approving. The false-positive discipline is systematic rather than incidental, which is the right instinct for a rail that annotates every PR body: only backticked spans are considered, bare lowercase words are skipped as indistinguishable from emphasis, grounding uses the full diff so a rename grounds on its removal line, and any missing input skips the check entirely. Path claims deliberately skipping the atom fallback, so `pkg` or `internal` cannot ground a fabricated path, is the detail that shows the care.

Verified independently:

**Truncation math holds on both branches.** The boundary cut returns `b + len(" …")`, bounded by maxLen since `b <= len(cut) <= maxLen - 4`; the fallback is `maxLen - 3 + 3`. The terminator rule correctly rejects `v1.2` and `logger.info` by requiring trailing whitespace or end of string.

**The executor change converts the inner `reviewDiff, reviewDiffErr :=` to `=`.** Without that the new outer `reviewDiff` would have been shadowed and always nil at the call site, which is exactly the bug this shape invites.

One cosmetic edge, not worth blocking: a terminator at the very end of the cut counts as a boundary, so truncating mid-decimal can manufacture one ("... version 1. …"). The 60 percent rule bounds how often that is reachable.

Your correction to the issue premise is right. #1411 describes the wrong mechanism; the code comments are now the better record, and I will fix the issue text rather than leave the two disagreeing.

PR #1452, opened 2026-08-08
feat(inferenceservice): add spec.modelCache.persistence for an ephemeral model cache

### LINKED ISSUE #1451: [FEATURE] Per-InferenceService opt-out of the model cache PVC (ephemeral mode)
## Feature Description

A per-InferenceService way to opt out of the model cache PVC and download into an
ephemeral `emptyDir` instead.

`ModelCacheSpec` currently carries only `ClaimName`, which redirects the cache to
a user-managed PVC. There is no way to say "this service does not want a cache
volume at all". The only opt-out is global: unset `--model-cache-path` (chart
`modelCache.enabled=false`), which turns caching off for every workload in the
cluster.

## Problem Statement

> As an operator running a mixed fleet, I want to skip the cache PVC on specific
> InferenceServices so that a workload can run on a node where dynamic
> provisioning is unavailable, or where re-pulling from a local origin is cheap
> enough that a persistent cache is not worth a volume.

Two cases, both real:

**1. Nodes that cannot get a dynamically provisioned volume.** microk8s-hostpath
provisions by spawning a helper pod pinned to the PVC's selected node, and that
helper carries no tolerations and exposes no configuration. Any node with a
`NoSchedule` taint therefore blocks provisioning forever, silently: no error, no
retry, no event naming a cause. Encountered on GPU nodes tainted
`nvidia.com/gpu=present:NoSchedule`. There is a storage-side fix (a provisioner
whose helper tolerations can be set), but LLMKube should not require a specific
provisioner to place a workload on a tainted node. Today the only escapes are
pre-creating a PV by hand or turning caching off cluster-wide.

**2. A fast local origin makes the cache optional.** The cache earns its keep
when the origin is slow and remote. With an in-cluster or on-LAN object store it
often does not: measured on this fleet, the same 15.65 GB GGUF took 4m22s at
61 MB/s from Hugging Face and 2m52s at 93 MB/s from a local MinIO over 2.5GbE.
At that rate a cold pull on pod restart is cheaper than carrying a 100Gi
per-service volume on every GPU node, particularly since `perService` caches are
node-local and are not shared anyway.

This is squarely the air-gapped / on-prem story: a user who has done the work to
stand up a local model origin should be able to treat node storage as
disposable.

## Proposed Solution

Add a mode to `ModelCacheSpec`, mutually exclusive with `claimName`:

```yaml
kind: InferenceService
metadata:
  name: qwopus-fusion-dgx
spec:
  modelRef: qwopus-fusion-q4km-s3
  modelCache:
    # Skip the cache PVC; download into an emptyDir and re-pull on restart.
    mode: Ephemeral      # Cached (default) | Ephemeral
```

The plumbing largely exists. `constructDeployment` already falls back to an
`emptyDir` when there is no effective cache key, and `modelNeedsCachePVC` is
already the single gate on provisioning the volume. This would add an explicit
spec-driven reason to take that path rather than only the implicit ones (local
source, fingerprint not yet landed, caching disabled globally).

Points worth deciding during implementation:

- `warnIgnoredModelCacheClaim` should treat `mode: Ephemeral` together with
  `claimName` as a conflict. CEL validation on the CRD is probably better than an
  event, since it is a static contradiction.
- The InferenceService status should say plainly that weights are ephemeral, so
  a slow first request after a restart is legible rather than mysterious.
- Worth confirming the choice is visible in `kubectl describe`, since the cost
  (re-download per restart) is paid at a time far removed from the decision.

## Alternatives Considered

**`spec.modelCache.claimName` pointing at a hand-made PVC.** Works today and is
what we are doing, but it needs a hand-written PV per service (each carrying a
PVC UID) and it inverts the intent: you create a volume in order to express that
you do not want the operator to manage one.

**Turning caching off globally.** Too blunt. It penalises every workload whose
origin genuinely is slow, to accommodate the ones whose origin is not.

**Fixing it purely in storage.** Necessary but not sufficient. It makes the
volume provisionable; it does not let a user say the volume is not wanted.

**A Model-level rather than InferenceService-level field.** Rejected: whether a
cache is worth a volume is a property of where the workload runs, not of the
model. The same Model may deserve a cache on one node and not on another.

## Additional Context

- Related: #1449 (`s3://` sources reaching Ready), #1098 (`s3://` support)
- `ModelCacheSpec` is in `api/v1alpha1/inferenceservice_types.go`
- The gate is `modelNeedsCachePVC` and the existing emptyDir fallback is in
  `internal/controller/model_storage.go`

Defilan:
## What

Adds `spec.modelCache.persistence` to InferenceService. `Ephemeral` declines the
model cache PVC and downloads into an `emptyDir` instead; `Cached` (the default,
and what an empty value resolves to) is today's behaviour.

```yaml
spec:
  modelRef: qwopus-fusion-q4km-s3
  modelCache:
    persistence: Ephemeral
```

## Why

Fixes #1451

`ModelCacheSpec` carried only `ClaimName`, which redirects the cache to a
user-managed PVC. There was no way to say "this service does not want a cache
volume at all". The only opt-out was global: unset `--model-cache-path`, which
disables caching for every workload in the cluster.

Two situations need it.

**Nodes where dynamic provisioning is unavailable.** A provisioner whose
volume-creation helper cannot tolerate the node's taints will never bind the
claim. microk8s-hostpath provisions by spawning a helper pod pinned to the PVC's
selected node; that helper carries no tolerations and the image exposes no args
and no ConfigMap, so a `NoSchedule` node blocks provisioning permanently and
silently: no error, no retry, no event naming a cause. A Pod on a tainted GPU
node waits forever on a volume that will never arrive. There is a storage-side
fix, but LLMKube should not require a particular provisioner to place a workload
on a tainted node.

**A fast local origin, where the cache does not earn its keep.** Measured on one
fleet, the same 15.65 GB GGUF took 4m22s at 61 MB/s from Hugging Face and 2m52s
at 93 MB/s from an in-cluster MinIO. At that rate a cold pull on restart can
cost less than a 100Gi per-service volume on every GPU node, particularly since
`perService` caches are node-local and never shared, so their only value is
surviving a restart on the same node.

This is the air-gapped and on-prem story: a user who has stood up a local model
origin should be able to treat node storage as disposable.

## How

**Named `persistence`, not `mode`.** The issue proposed `mode`, and I changed it
while implementing. The operator already has `--model-cache-mode`
(`shared` / `perService`) selecting WHICH cache PVC is used; this field answers
whether there is one at all. Two adjacent fields both called "mode" on different
axes is a footgun, and the CRD field docs would have had to spend a paragraph
undoing the confusion.

**Both gates now derive from one predicate.** The provisioning side
(`modelNeedsCachePVC`, which creates the claim) and the mount side (the
`useCache` value the deployment builder passes to `buildModelStorageConfig`)
were independent expressions of the same question. `modelWantsCacheVolume` is
now the mount-side answer and `modelNeedsCachePVC` is defined in terms of it, so
they cannot drift. Drift is the failure worth preventing: a mount without a
claim leaves the Pod Pending on a volume nobody creates, and a claim without a
mount leaves an orphaned ISVC-owned PVC. A test asserts their agreement directly
rather than trusting the two call sites to stay in sync.

The `pvc://` exclusion stays only on the provisioning side, because those
sources dispatch to `buildPVCStorageConfig` and never consult the mount gate.

**Only an explicit `Ephemeral` opts out.** A nil isvc, an absent `modelCache`
block, and an empty `persistence` all resolve to `Cached`. `Ephemeral` also does
not override the existing exclusions in either direction: caching disabled on
the operator stays disabled, and `pvc://` sources still take no cache PVC.

**Mutual exclusion with `claimName` is enforced by CEL,** not an event. One
names a cache volume to use and the other declines to use any, which is a static
contradiction, so it belongs at admission rather than in a warning the user
finds later. `claimName` with `Cached`, or with `persistence` unset, stays legal:
that is the pre-existing behaviour spelled out.

## Testing

Unit tests over the predicates (`model_cache_ephemeral_test.go`), including a
control assertion in each direction so a test cannot pass because the fixture
was never cacheable, and a direct check that the two gates agree.

Envtest specs (`model_cache_validation_test.go`) prove the CEL rule and the enum
are enforced **by the apiserver**, not merely present in the generated schema.
Unit tests over the Go predicates cannot catch a schema regression here, since
the conflict is rejected at admission and never reaches the reconciler. Verified
these actually execute rather than being filtered: `Ran 6 of 684 Specs ...
6 Passed | 0 Failed`.

- `go test ./internal/controller/` with envtest: pass (254s)
- `make lint`: 0 issues
- `make manifests` + `make chart-crds`: both CRDs regenerated and in sync
- `helm template charts/llmkube`: renders

## Checklist

- [x] Tests added/updated

Docs: the CRD field documentation carries the guidance (including that the cost
is paid at a time far removed from the decision, so the default should be
preferred), and `config/samples/inferenceservice_ephemeral_cache.yaml` is a
worked example registered in the samples kustomization.

Assisted-by: Claude Code (the implementation, tests, and this description;
I reviewed the change, chose the field naming, and ran the envtest suite,
make lint, and the CRD regeneration myself)

### DIFF
--- a/api/v1alpha1/inferenceservice_types.go
+++ b/api/v1alpha1/inferenceservice_types.go
@@ -879,6 +879,13 @@ type InferenceResourceRequirements struct {
 	// download. Node disk capacity varies by an order of magnitude across
 	// distributions and machine types, so there is no default that is safe
 	// everywhere and none is applied.
+	//
+	// The pattern is the Kubernetes quantity format, so a malformed value is
+	// rejected at admission. The sibling quantity fields above predate this and
+	// carry no pattern; they reach resource.MustParse, which panics on
+	// malformed input and takes the controller down for every workload rather
+	// than failing the one object at fault.
+	// +kubebuilder:validation:Pattern=`^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$`
 	// +optional
 	EphemeralStorage string `json:"ephemeralStorage,omitempty"`
 
--- a/api/v1alpha1/inferenceservice_types.go
+++ b/api/v1alpha1/inferenceservice_types.go
@@ -115,18 +115,32 @@ type ModelCacheSpec struct {
 	// entirely: weights land in an emptyDir and are re-downloaded every time
 	// the Pod starts. No cache PVC is created, mounted, or required.
 	//
-	// Ephemeral is for two situations. First, nodes where dynamic provisioning
-	// is not available: a provisioner whose volume-creation helper cannot
-	// tolerate the node's taints will never bind the claim, and with
-	// microk8s-hostpath that failure is silent, so the Pod waits forever on a
-	// volume that will never arrive. Second, a fast local origin, where the
-	// cache does not earn its keep: pulling from an on-LAN object store can be
-	// quick enough that a per-service volume on every GPU node costs more than
-	// the occasional re-download.
+	// Ephemeral is for two situations.
 	//
-	// The trade is paid at a time far removed from the decision: the first
-	// request after any restart waits for a full re-download. Prefer Cached
-	// unless one of the situations above applies.
+	// A model origin fast enough that the cache does not earn its keep. Pulling
+	// from an in-cluster or on-LAN object store can be quick enough that a
+	// per-service volume on every GPU node costs more than the occasional
+	// re-download, particularly since perService caches are node-local and are
+	// never shared between services.
+	//
+	// Clusters where a cache PVC is impractical for this workload: no suitable
+	// StorageClass for the nodes it must run on, or a local-volume provisioner
+	// that creates volumes through a helper Pod pinned to the target node and
+	// so cannot serve a node whose taints that helper does not tolerate. That
+	// last case can fail silently, leaving the Pod waiting on a volume that
+	// will never bind. It does not arise on provisioners that create volumes
+	// through a control-plane API call rather than an on-node Pod, which is how
+	// most managed-cloud CSI drivers work.
+	//
+	// Two costs, both paid later than the decision. The first request after ANY
+	// restart waits for a full re-download, and with more than one replica every
+	// replica downloads its own copy. Prefer Cached unless one of the situations
+	// above applies.
+	//
+	// Pair this with resources.ephemeralStorage. Weights land on node local
+	// disk, and without a declared budget the scheduler cannot account for the
+	// download and the kubelet has no per-pod ceiling to enforce; the operator
+	// emits an UnboundedEphemeralCache warning when it is unset.
 	//
 	// Mutually exclusive with claimName, which names a cache volume to use.
 	// +optional
@@ -846,6 +860,27 @@ type InferenceResourceRequirements struct {
 	// which can lead to OOM kills after model load.
 	// +optional
 	HostMemory string `json:"hostMemory,omitempty"`
+
+	// EphemeralStorage is the node local-disk budget for this pod (e.g. "40Gi").
+	// Translated to BOTH requests and limits for ephemeral-storage, and, when
+	// modelCache.persistence is Ephemeral, to the sizeLimit of the emptyDir the
+	// weights download into.
+	//
+	// Set this whenever weights land on node disk rather than a cache PVC.
+	// Without it the scheduler has no visibility into a download that can run to
+	// tens of gigabytes and will place further pods on a node that is about to
+	// fill, and the kubelet has no per-pod ceiling to enforce: the node crosses
+	// its DiskPressure threshold instead, and eviction then proceeds by QoS
+	// class, which can remove unrelated workloads before this one. The request
+	// makes the download visible to scheduling; the limit keeps an overrun
+	// charged to this pod.
+	//
+	// Size it above the model on disk with room for the partial file during
+	// download. Node disk capacity varies by an order of magnitude across
+	// distributions and machine types, so there is no default that is safe
+	// everywhere and none is applied.
+	// +optional
+	EphemeralStorage string `json:"ephemeralStorage,omitempty"`
 
 	// GPUMemory is recorded on the object and has no effect. It sets no pod
 	// resource request, does not influence scheduling, and is not validated;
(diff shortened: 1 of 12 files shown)


Defilan:
Pushed a second commit making this safe across distributions. Two problems with
the original, both raised in review.

**The emptyDir was unbounded and invisible to the scheduler.** Verified: the
volume was built as `EmptyDirVolumeSource{}` with no `SizeLimit`, there were zero
occurrences of `ephemeral-storage` anywhere in the operator, and
`InferenceResourceRequirements` exposed no knob for it, so a user could not fix
it either. Weights land on node local disk, so `Ephemeral` on a node with a small
boot disk meant the scheduler kept placing pods on a node about to fill, and the
kubelet had no per-pod ceiling: the node crossed its DiskPressure threshold and
eviction proceeded by QoS class, which can remove unrelated workloads before the
one that overran. Fine on a node with 780G free, dangerous on a 20G one.

`spec.resources.ephemeralStorage` now drives both halves: the ephemeral-storage
request and limit on the container, and the `sizeLimit` of the emptyDir. The
request is what the scheduler reserves; the limit is what keeps an overrun
charged to this pod.

No default is applied. None is safe across the range of node disk capacities in
play, and a guess under the real size would evict pods for exceeding a number
nobody chose. The model's size on disk is not reliably known either: `Status.Size`
comes from the GGUF metadata range read, which only runs for remote http(s)
sources and so is absent for `s3://`. Instead the operator emits an
`UnboundedEphemeralCache` warning when `Ephemeral` is selected without a budget.
A warning rather than a hard failure, since refusing to reconcile would take down
a workload that runs perfectly well on a node with room.

**The field documentation was written for one distribution.** It led with a
failure mode specific to local-volume provisioners that create volumes through a
helper Pod pinned to the target node, and named an implementation. That case does
not arise at all on provisioners that provision through a control-plane API call,
which is how most managed-cloud CSI drivers work, so it advertised a benefit most
users cannot collect.

Reordered: the universal reason (a model origin fast enough that the cache does
not earn its keep) now leads, the provisioner case is described as a class rather
than a product, and the cost of `Ephemeral` with `replicas > 1` (every replica
downloads its own copy) is written down, which the original left unstated.

Re-verified after the change: full controller suite with envtest passes (254s),
`make lint` 0 issues, both CRDs regenerated and in sync, `helm template` renders.

PR #1456, opened 2026-08-08
feat(foreman): advisory two-site parity signal in the coder gate

### LINKED ISSUE #1418: [BUG] Coder gate has no two-site parity signal: a one-site fix leaves an identical sibling function silently unchanged
## Bug Description

The coder gate has no signal for the two-site parity failure: a change lands correctly in one function and leaves an identically-named sibling function in another package untouched.

`AGENTS.md` states this rule for runtime argument construction, but nothing enforces or even surfaces it. `checkTestPresence` (`pkg/foreman/agent/mutation_gate.go:294`) collects only net-new function names and deliberately excludes body-modified functions, so a one-site fix produces no finding at all.

This is a narrow slice of #1391 (proposal 2 only). The other two proposals on that issue change blocking-tier behaviour and are not covered here.

## Steps to Reproduce

1. A coder fixes `appendModeArgs` in `internal/controller/runtime_llamacpp_args.go`, adding a flag.
2. `pkg/agent/executor.go` contains a function with the same name and the same switch shape, and is left unchanged.
3. Every gate passes. Nothing mentions the second site.

This is not hypothetical: it is exactly the shape of #1406, where the fix was required at both sites. That run happened to get both because the dispatching intent named the parity rule explicitly. Without that prompt, the gate would not have caught a one-site fix.

## Expected Behavior

An **advisory** finding, in the same tier as the test-dilution gate (`tierAdvisory`, `coder_gate.go:338`), naming the sibling site that was not updated. Advisory only: it must never block a branch, because there are legitimate reasons for two same-named functions to diverge.

## Proposed detector

Mechanically checkable, no judgment required:

1. For each changed non-test Go file in the diff, collect the set of **string literals the diff adds** (for example `"--cache-ram"`).
2. For each function whose body the diff modified, take its name (for example `appendModeArgs`).
3. Search the repository's other non-test Go packages for a function declaration with the **same name**.
4. If such a sibling exists and does **not** contain the added literal, emit one advisory naming both sites:
   `appendModeArgs: added "--cache-ram" in internal/controller/runtime_llamacpp_args.go but the same-named function in pkg/agent/executor.go does not contain it`

Scope guards, so this stays quiet:

- Only fire when the sibling function name is an exact match. Do not fuzzy-match.
- Only consider string literals the diff **added**; ignore removals and identifier changes.
- Skip `_test.go` files entirely on both sides.
- Emit at most one advisory per (function name, literal) pair, and sort output so findings are deterministic.

## Must fire

The #1406 shape: `appendModeArgs` gains `"--cache-ram"` in `internal/controller/runtime_llamacpp_args.go` while the same-named function in `pkg/agent/executor.go` does not.

## Must stay silent

- Both siblings updated in the same diff.
- A changed function with no same-named sibling anywhere.
- A sibling that already contains the literal.
- Literals added only inside `_test.go` files.

## Prior art to mirror

`pkg/foreman/agent/caller_impact_gate.go` already has occurrence-scanning machinery over the workspace, including identifier-boundary handling at :196-210 so a match inside a longer identifier is rejected. The test-dilution gate (`pkg/foreman/agent/test_dilution_gate.go`) is the closest structural model for a `tierAdvisory` detector, and `test_dilution_gate_test.go` shows the `dilutionRunner` fake-command-runner pattern that makes these testable without a real git tree.

## Environment

Not applicable: found by inspection against `main` at `929f0a87` while reviewing the #1406 fix. This is coder-gate behaviour, not a runtime or cluster defect.

## Additional Context

Refs #1391, #1332. Sibling advisory signals already shipped: removed/relocated coverage (#1332), assertion-value churn (#1347).

Defilan:
## What

Adds a two-site parity signal to the coder gate: a change that lands in one
function while an identically-named sibling in another package stays untouched
is now surfaced.

## Why

Fixes #1418

`AGENTS.md` states the two-site rule for runtime argument construction, but
nothing enforced or even surfaced it, so a one-site fix passed the gate silently.

## How

Extends the existing gate machinery in `pkg/foreman/agent/mutation_gate.go`
rather than adding a separate stage or binary.

**Advisory, not blocking** (`tierAdvisory`). This heuristic will produce false
positives, since a genuinely one-sided change is often correct; it informs the
verdict and surfaces to the reviewer rather than failing the gate. Tests pin both
directions: a real two-site miss is caught, and a legitimate one-site change is
not flagged.

## Verification

Build, vet, and `go test ./pkg/foreman/agent/` all pass in a clean worktree
against current main. The new test fails without the change (it references
`funcBodyContainsLiteral`, which does not exist on main), so it genuinely
exercises the new code.

Assisted-by: Claude Code (branch produced by the Foreman agent harness on a self-hosted model; I reviewed the diff, ran build, vet and the package test suites in a clean worktree, and verified the specifics called out above).

### DIFF
--- a/pkg/foreman/agent/coder_gate.go
+++ b/pkg/foreman/agent/coder_gate.go
@@ -350,6 +350,17 @@ func gateCheckRegistry(issueText, evidenceBaseSHA string, evidence []grounding.E
 			tier: tierAdvisory,
 			fn:   checkCommandStringTestDilution,
 		},
+		{
+			// two-site-parity (#1418): advisory. Surfaces to the reviewer when
+			// a changed function adds a string literal (e.g. a command-line
+			// flag) but a same-named sibling function in another package does
+			// not contain that literal. Advisory only: never blocks, because
+			// there are legitimate reasons for two same-named functions to
+			// diverge.
+			name: "two-site-parity",
+			tier: tierAdvisory,
+			fn:   checkTwoSiteParity,
+		},
 	}
 }
 
--- a/pkg/foreman/agent/mutation_gate.go
+++ b/pkg/foreman/agent/mutation_gate.go
@@ -505,6 +505,265 @@ func rangeIntersects(lines map[int]bool, start, end int) bool {
 // one gate run, so a sprawling change cannot blow the loop's time budget.
 const maxNeuterPackages = 5
 
+// checkTwoSiteParity is a tierAdvisory gate check (#1418). For each changed
+// non-test Go file it collects the string literals the diff adds and the names
+// of functions whose bodies were modified. It then searches the repository's
+// other non-test Go packages for a function declaration with the same name.
+// If such a sibling exists and does not contain the added literal, an advisory
+// naming both sites is emitted. Advisory only: it must never block a branch,
+// because there are legitimate reasons for two same-named functions to diverge.
+//
+// Scope guards:
+//   - Only fire when the sibling function name is an exact match.
+//   - Only consider string literals the diff added; ignore removals.
+//   - Skip _test.go files entirely on both sides.
+//   - Emit at most one advisory per (function name, literal) pair.
+//   - Output is sorted for determinism.
+func checkTwoSiteParity(ctx context.Context, workspace string, run commandRunner) (bool, string) {
+	changed := changedNonTestGoFiles(ctx, workspace, run)
+	if len(changed) == 0 {
+		return false, ""
+	}
+
+	// Collect added string literals per changed file.
+	type fileInfo struct {
+		path      string
+		addedLits map[string]bool
+		modFuncs  map[string]bool
+	}
+	var infos []fileInfo
+	for _, f := range changed {
+		added := addedStringLiterals(ctx, workspace, f, run)
+		modNames := modifiedFuncNames(ctx, workspace, f, run)
+		mod := map[string]bool{}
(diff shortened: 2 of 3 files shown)


joryirving (approved):
Hey Chris - this looks good. The advisory-not-blocking design is the right call given the false-positive rate, and the AST-based body inspection (rather than string matching) is the right approach for correctness. Tests cover both directions well: the one-site miss fires, and legitimate one-site changes stay silent.

A couple of minor notes, neither blocking: first, \`findSiblingMissingLiteral\` walks the entire workspace and reads every non-test Go file from disk for each (func name, literal) pair, so on a large change with several literals this is O(pairs × workspace_files) filesystem reads + AST parses. Fine for an advisory check, but if it ever gets slow, indexing func names to file paths once per run would kill the redundancy. Second, \`hasFuncDecl\` uses a regex that won't match generic function declarations (\`func Foo[T any](\`), so same-named generic functions would be missed as siblings. Probably rare enough not to matter, but worth knowing. And \`extractStringLiterals\` will happily extract strings from comment lines in the diff (\`+// "--cache-ram"\`), which could produce a spurious advisory — not a big deal at advisory tier, just slightly noisy.

None of that is worth blocking on. The design is solid, the test coverage is thorough, and the parity signal genuinely fills a gap that \`AGENTS.md\` defined but nothing enforced.

Issue #1434, opened 2026-08-06
[FEATURE] fetch_pull_request tool: coders cannot read the PR they are asked to fix


joryirving:
## Summary

Coder agents can read an issue (`fetch_issue`) but cannot read a **pull request**. On the PR-fix path that means the coder is asked to respond to review feedback it has no way to look at — it only sees whatever text the orchestrator chose to paste into `payload.prompt`.

Requesting a read-only `fetch_pull_request` tool, mirroring `FetchIssueTool`.

## Why the workaround doesn't hold

`GITHUB_TOKEN` is in `defaultBashEnvAllowlist` (`pkg/foreman/agent/tools/bash.go`), so in principle the coder could call the API from bash. In practice the coder image ships **no `curl`, `gh`, `wget`, or `jq`** — only `python3` and `git` (verified in a running `llmkube-coder:0.9.14` pod). So "read the PR yourself" means prompting the model to hand-write `urllib` requests against the GitHub REST API and parse the JSON inline, per attempt. That is exactly the kind of incidental work a model gets subtly wrong, and it puts the raw token into shell command strings.

`fetch_issue` exists for this reason. Its own registration comment in `cmd/foreman-agent/main.go` reads:

> read-only GitHub issue surface for the reviewer. The same token the foreman-agent already loads at startup (via `repo.TokenFromEnvOrFile`) reaches GitHub through one bounded Go-side call instead of being inherited by every bash subprocess via `$GH_TOKEN`. Closes #580.

Every word of that applies to pull requests.

## What the coder gets today

For a PR-fix task, `payload.prompt` is a digest assembled by the orchestrator: a type header, a reason line, and feedback strings. The coder cannot see:

- other reviews on the PR (only the event that triggered this task)
- inline review comments, with their file and line anchors
- check-run conclusions beyond a pre-trimmed log excerpt
- its own prior attempt's diff as the reviewer saw it

Three cases from one day of operation:

1. An **approving** review carried four inline findings (including `Minor (bug)` on inconsistent exception propagation, and two on silently swallowed errors). Ingestion now captures them, but the coder still receives them as flattened strings rather than being able to read the PR.
2. A PR-fix loop spent **all three attempts** reasoning from a log excerpt that had anchored on a `kubectl describe pod` dump instead of the pytest failures. The excerpt bug is fixed downstream, but the coder had no way to go look at the PR's checks itself and notice the excerpt was noise.
[...]

PR #1458, opened 2026-08-08
feat(foreman): read-only fetch_pull_request tool

Defilan:
## What

Adds a read-only `fetch_pull_request` tool so a coder asked to fix a PR can
actually read it, mirroring `FetchIssueTool`.

## Why

Refs #1434

On the PR-fix path the coder previously saw only whatever text the orchestrator
pasted into `payload.prompt`.

## How

New `pkg/foreman/agent/githubprfetch` package plus the tool wrapper, following
the existing `fetch_issue` construction, registration and test shape.

**Read-only is enforced, not just intended:** every request is
`http.MethodGet`. Nothing posts, pushes, approves, merges or mutates.

## Defect found in review, needs fixing before merge

`fetchCheckRuns` requests `/repos/{owner}/{repo}/pulls/{n}/check-runs`, which is
not a GitHub endpoint. Verified against the live API:

```
GET /repos/defilantech/LLMKube/pulls/1452/check-runs   -> 404 Not Found
GET /repos/defilantech/LLMKube/commits/{sha}/check-runs -> total_count=24
```

Check runs are addressed per commit ref, so this needs the PR's `head.sha` (which
the code already fetches) rather than the PR number. The unit tests do not catch
it because they answer any URL from a mock server.

The error is also discarded at the call site
(`pr.CheckRuns, _ = c.fetchCheckRuns(...)`), so this fails silently forever:
`CheckRuns` is simply always empty against real GitHub. Non-fatal, but it means
the coder never sees CI status, and nothing ever says why.

The rest of the tool (title, body, state, reviews, comments) is correct and
useful as-is. Marked `Refs` not `Fixes` until the check-runs path is fixed.

## Verification

Build, vet, and `go test ./pkg/foreman/agent/tools/ ./pkg/foreman/agent/githubprfetch/`
all pass in a clean worktree against current main. The new tests fail without the
change.

Assisted-by: Claude Code (branch produced by the Foreman agent harness on a self-hosted model; I reviewed the diff, ran build, vet and the package test suites in a clean worktree, and verified the specifics called out above).

### DIFF
--- a/cmd/foreman-agent/main.go
+++ b/cmd/foreman-agent/main.go
@@ -72,6 +72,7 @@ import (
 	"github.com/defilantech/llmkube/pkg/foreman/agent/codehost"
 	"github.com/defilantech/llmkube/pkg/foreman/agent/githubissue"
 	"github.com/defilantech/llmkube/pkg/foreman/agent/githubpr"
+	"github.com/defilantech/llmkube/pkg/foreman/agent/githubprfetch"
 	"github.com/defilantech/llmkube/pkg/foreman/agent/mcp"
 	"github.com/defilantech/llmkube/pkg/foreman/agent/repo"
 	foremantools "github.com/defilantech/llmkube/pkg/foreman/agent/tools"
@@ -810,6 +811,16 @@ func makeRegistryFactory(
 			// Closes #580.
 			&foremantools.FetchIssueTool{
 				Fetcher: githubissue.NewClient(),
+				Token:   repo.TokenFromEnvOrFile,
+			},
+			// fetch_pull_request: read-only GitHub PR surface for the
+			// coder. The same token the foreman-agent already loads
+			// at startup (via repo.TokenFromEnvOrFile) reaches GitHub
+			// through one bounded Go-side call instead of being
+			// inherited by every bash subprocess via $GH_TOKEN.
+			// Closes #1434.
+			&foremantools.FetchPullRequestTool{
+				Fetcher: githubprfetch.NewClient(),
 				Token:   repo.TokenFromEnvOrFile,
 			},
 			// run_integrate: deterministic tool for a sliced Workload's
--- a/pkg/foreman/agent/githubprfetch/fetch.go
+++ b/pkg/foreman/agent/githubprfetch/fetch.go
@@ -1,0 +1,429 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package githubprfetch fetches a single GitHub pull request's details,
+// reviews, review comments, and check runs so a coder Agent can read the
+// PR it is asked to fix. The model needs to know WHAT it is fixing; reading
+// the PR body and review feedback is the obvious way to learn that, and the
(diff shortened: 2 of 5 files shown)


joryirving (changes requested):
Hey Chris — requesting changes on this one. 

There's a blocking bug beyond the check-runs issue you flagged in the PR description: the `json:\"head_ref\"` and `json:\"base_ref\"` struct tags don't match the real GitHub API shape. GitHub returns `head` and `base` as nested objects (`head.ref`, `base.ref`, `head.sha`), not top-level fields, so HeadRef and BaseRef will always deserialize to empty strings in production. 

The mock server returns them as top-level fields, which is why the tests pass — but the contract is wrong. 

The fix is to parse the nested `head`/`base` objects and extract `.ref` and `.sha`, which also gives you the head SHA you need to fix the check-runs endpoint (`/commits/{sha}/check-runs` instead of `/pulls/{n}/check-runs`). 

One other thing on the sub-fetches: all three swallow errors silently with `_ =`, and the package doc says \"a failed fetch logs\" but the code doesn't actually log anything — a one-line log per failed sub-fetch would help operators figure out why the coder is missing context. 

Minor note: no pagination on reviews/comments, so a PR with 50+ inline comments gets silently truncated — fine for v1 but worth a TODO. Everything else looks good: read-only by construction, clean token handling, good error taxonomy, and the best-effort design is the right call. The core issue is just the mock contract diverging from the real API, not the test design itself.

Defilan:
All three confirmed against the live API before I touched anything. Pushed as 037ad151.

**head/base.** You're right that this never worked. There is no top-level `head_ref` or `base_ref` on the response at all:

```
$ gh api repos/defilantech/LLMKube/pulls/1458
top-level head_ref present: False
top-level base_ref present: False
head.ref = foreman/night-dgx-1434-fetchpr/issue-1434
head.sha = 791823ff748f
base.ref = main
```

Now parses the nested objects and keeps `head.sha`, which as you say is what the check-runs fix needed.

The confirmation was accidental and worth recording: I fixed `fetch.go` before touching the mocks, and the happy-path tests immediately failed with `HeadRef: want 'fix/something' got ""`. Exactly the symptom you predicted, which is the evidence the contract was wrong rather than the parsing.

**Check runs.** Endpoint moved to `/commits/{head sha}/check-runs`. `/pulls/{n}/check-runs` 404s, verified.

**The mocks were the actual bug, and that is the part I care most about.** They invented the flat PR shape and answered whatever URL they were given, which is why two wrong endpoints passed a full suite and a review. The check-runs handler now matches only `/commits/{sha}/check-runs`, so a request to the wrong path fails the test the way it fails against GitHub. Added `TestClient_UsesRealGitHubContract`, which asserts the request the client actually makes rather than trusting a permissive handler.

**Swallowed errors.** Fixed, and I went a different way than you suggested, so push back if you disagree. Rather than a log line, each failed sub-fetch is recorded in `PullRequest.Warnings` and surfaced in the tool result. The reasoning: the coder is what acts on this context, and a controller log does not reach it. Without this it cannot tell "no review comments" from "review comments could not be read", which is the same ambiguity that let the check-runs endpoint stay broken and silent. Happy to add a log line as well if you want it in the operator's view too.

The package doc did claim a failed fetch logs. It did not. It now says what actually happens.

**Pagination.** Took the cheap half: `per_page=100` is the API maximum, which moves the ceiling from 30 to 100. Link headers are still not followed, so the ceiling is real and it is now a stated TODO at both call sites rather than silent truncation.

joryirving (approved):
Hey Chris — pulled the updated diff, everything checks out. The nested head/base parsing is correct, HeadSHA is extracted and flows through to the check-runs endpoint properly, and \`TestClient_UsesRealGitHubContract\` is exactly the right guard — asserting the request path rather than trusting a permissive handler is what actually prevents this class of bug from recurring. The Warnings field is a better call than a log line for the reasoning you gave: the coder is the consumer that needs to distinguish empty from errored, and controller logs don't reach it. per_page=100 with stated TODOs for Link headers is the right cheap half. Approving.

Issue #1467, opened 2026-08-09
[FEATURE] Headless render-and-measure tool so coders can verify a page renders


Defilan:
## Feature Description

A render-and-measure tool for coder agents: load a local page in a headless
browser, then return console errors and objective frame statistics as text.

Not a screenshot for a human. Numbers the model can act on.

## Problem Statement

> As an operator running a coder on front-end work, I want the agent to find out
> whether what it built actually renders, without a human opening a browser.

Driving a four-stage Three.js build through Foreman, every visual defect was
caught by a human looking at the page, and none by the pipeline:

- A `ReferenceError` thrown once per frame from the render loop. Syntax checks
  pass; it is a runtime error, so nothing in the gate saw it.
- Post-processing that clipped the lower half of the frame to solid white,
  making the geometry the same commit had just built invisible. The diff looks
  entirely reasonable.

The agent pod also cannot do this itself today: it is Debian with `python3` and
`curl`, and no node, npx or chromium.

## Proposed Solution

An MCP server (matching how context7 and perplexity are already wired, so no
browser goes into the coder image) exposing roughly:

```
render_page(path, viewport, wait_ms) -> {
  consoleErrors: [...],
  clippedWhitePct: 38.2,
  nonBlackPct: 61.0,
  meanLuminance: 214,
  distinctColours: 3,
  firstFrameMs: 120
}
```

Interactions matter as much as the load: dispatching mousemove, drag, click,
wheel and touch before sampling is what surfaces a handler that throws, which a
plain page load will not.

Thresholds should be reported, not enforced. "38% of the frame is pure white" is
actionable; a hard pass/fail invites gaming and cannot know what the page is
supposed to look like.

## Alternatives Considered

**Bake a browser into the coder image.** Simpler to reach from `bash`, but adds
hundreds of megabytes to an image used by every task, most of which never render
anything.

**Vision review (#1466).** Complementary, not a substitute. Numbers catch
blowouts and blank canvases deterministically; eyes judge whether it looks good.
This one is far cheaper and would have caught both real defects above.

## Additional Context

- Related: #1466 (image content in the chat path)
- The coder model's own card describes it root-causing its own bugs from
  Playwright error reports, so the text-report shape is known to work for it.

Issue #556, opened 2026-05-26
[BUG] foreman/oai: turn 1 of reviewer-on-Devstral fails with HTTP 400 'All non-assistant messages must contain content'


Defilan:
## Bug

Foreman's OAI client sends non-assistant messages (tool / user / system) without a `content` field in some cases. Carnice tolerates this; Devstral 24B (Mistral-family) rejects with:

```
chat: oai: status 400: {"error":{"code":400,"message":"All non-assistant messages must contain 'content'","type":"invalid_request_error"}}
```

## Steps to reproduce (live tonight, 2026-05-26)

1. Stand up Devstral 24B (Mistral) via LLMKube metal-agent on Mac Studio
2. Apply a reviewer Agent CR pointing at that InferenceService
3. Run a Workload with this Agent in `reviewerAgentRefs`
4. The review task lands on the FleetNode, then turn 1 of the loop fails with HTTP 400

## Expected

The OAI client should always populate `content` on non-assistant messages (empty string when the message has none), matching the OpenAI chat-completions schema strictly. Models other than llama.cpp-permissive (Carnice / Qwen) will refuse otherwise.

## Where the fix likely goes

`pkg/foreman/agent/oai/client.go` request builder: when constructing tool / user / system messages, default `Content: ""` instead of omitting the field. Probably a 5-line fix + a regression test driven from a fake-OAI that 400s on missing content.

## Why it matters

This is the v0.2 model-diversity unlock. Without it, the cross-model reviewer architecture only works against llama.cpp-permissive backends (Qwen, MiniMax). Mistral / Mixtral / DeepSeek / many others enforce stricter schemas.

## Environment

- LLMKube + Foreman at v0.2-day4 (post #553, post #555-#524)
- Coder: Qwen3.6-35B-A3B Carnice (works fine; same OAI client, lenient backend)
- Reviewer: Devstral 24B Q6_K via llama-cpp-turboquant on Mac Studio (rejects)

PR #1471, opened 2026-08-09
feat(foreman): image input, so a vision-capable coder can see what it rendered

### LINKED ISSUE #1466: [FEATURE] Image content in the agent chat path, so a vision-capable coder can review what it rendered
## Feature Description

Support image content in the Foreman agent's chat path, so a vision-capable
model can be shown a screenshot, a chart, or a rendered page.

Today messages are text-only end to end: `grep -rn 'image_url|ImageURL'` over
`pkg/foreman` returns zero matches. There is no content-parts array and no
tool-result type that can carry an image.

## Why now

The blocker used to be the model. It is not any more.

The coder this fleet already runs, `Qwopus3.6-27B-Fusion`, ships a companion
projector (`Qwopus3.6-27B-Fusion-mmproj-Q8_0.gguf`, ~0.6 GB) and its model card
states image input is supported. The merge notes explicitly preserve vision
tensors. And LLMKube already serves projectors: `Model.spec.mmproj` exists and
`appendMmprojArgs` wires `--mmproj` into llama.cpp.

So the model can see, and the operator can serve it seeing. Only the harness
cannot pass it a picture.

## The concrete gap this leaves

Observed while driving a four-stage build of a Three.js page through Foreman:
every visual defect was caught by a human rendering the page and looking at it,
never by the pipeline.

- Stage 2 shipped a `ReferenceError` thrown once per frame. `node --check`
  passed; it is a runtime error.
- Stage 4 shipped post-processing that clipped the lower half of the frame to
  solid white. The reviewer read the diff and returned GO-shaped output; the
  defect is invisible in a diff.

A text reviewer cannot catch either. The second one is not even expressible in
text without someone looking first.

## Proposed Solution

1. Content-parts on user/tool messages: `[{type: text}, {type: image_url}]`,
   with base64 data URLs so no external fetch is required.
2. A tool-result shape that can return an image alongside text, so a render
   tool can hand back the frame it produced.
3. A per-Agent opt-in, since sending images to a text-only model wastes context
   and may error.

## Alternatives Considered

**Objective visual gates instead of vision.** Render headlessly and return
numbers as text: percentage of clipped pixels, mean luminance, distinct colour
count. Cheaper, deterministic, no vision model, and it would have caught both
defects above. Strictly better as a first step, and worth doing regardless.

It does not subsume vision, though: "38% of the frame is pure white" catches a
blowout but says nothing about whether a scene is beautiful, whether a layout is
balanced, or whether the leaves look like leaves. Those need eyes.

**Keep a human in the loop.** What happens today. It works, and it does not
scale to an overnight batch.

## Additional Context

- `pkg/foreman/agent` (request builder, tool results)
- Model card: https://hf.co/KyleHessling1/Qwopus3.6-27B-Fusion-GGUF
- Serving a projector from an object store is currently blocked by #1465

Defilan:
## What

`oai.Message` can now carry OpenAI content parts, and an AgenticTask payload can
attach images to the first user message. A vision-capable coder can be shown a
screenshot of what it built.

## Why

Fixes #1466

Messages were text-only end to end: `Content` is a string and the content-parts
form had no representation, so there was no way to put an image in front of a
model however capable it was.

That gap is why every visual defect on this fleet has been caught by a human. A
coder shipped post-processing that clipped half the frame to solid white and a
reviewer reading the diff returned a GO-shaped verdict, because the defect is
not expressible in a diff.

Both other halves were already in place: the fleet's coder ships a projector,
and the operator already serves projectors via `Model.spec.mmproj`. Verified on
hardware before writing this, with the same model the coders run:

```
srv load_model: loaded multimodal model, Qwopus3.6-27B-Fusion-mmproj-Q8_0.gguf
```

Shown a screenshot of its own blown-out render, it answered:

> "The lower half of the image displays a brightly illuminated floor area filled
> with plants and tables, all heavily overexposed by intense light. Yes, a large
> central portion of this area is pure white due to the extreme bloom effect
> washing out the details."

Accurate, including the cause. Only the harness could not pass the picture.

## How

**Wire format.** `Message.Parts []ContentPart`. When set it replaces `Content`
on the wire, since the schema allows a string or an array and never both. A
separate field rather than widening `Content` to `any`, so no existing call site
changes and the #556 always-emit-content rule survives. A test pins both
together, because that rule is load-bearing against strict backends.

The emitted shape is byte-identical to the request I verified by hand against
the live server:

```json
{"role":"user","content":[{"type":"text","text":"describe this"},
 {"type":"image_url","image_url":{"url":"data:image/png;base64,..."}}]}
```

**Entry point.** `AgenticTaskPayload.images`, workspace-relative paths rather
than inline data: an AgenticTask is a Kubernetes object and base64 in etcd would
bloat every list of it. The executor reads and encodes at dispatch.

**Opt-in.** `Agent.spec.vision.enabled`, off by default. Declared rather than
inferred, because an Agent points at a base URL and the harness cannot
introspect an arbitrary endpoint to learn whether a projector is loaded.
Guessing wrong sends image parts to a text-only model, which is an HTTP 400
rather than a graceful degrade.

**Caps.** 4MiB per image and 4 per message by default, measured after base64
since that is what travels and what the context window pays for. A 947KB
screenshot encoded to 484KB in practice.

**Degradation.** Missing file, unreadable file, oversized image and
workspace-escaping path all warn and skip. A task whose prompt describes the
problem in words is still worth running without the picture. The workspace
confinement is enforced rather than assumed: a task that can read arbitrary host
files and ship them to an inference endpoint is an exfiltration primitive.

No images means nil parts and the plain string path, so the request is
byte-identical for every existing agent.

## Testing

Wire format (parts vs string, #556 rule, assistant omitempty, parts-win-over-string)
and attachment (opt-in gating, both caps, missing file, path traversal). Full
`pkg/foreman/...` and `api/...` suites pass (17), `internal/controller` with
envtest passes (246s), `make lint` 0 issues.

## Not in this change

Tool-returned images, which is the other natural entry point and depends on a
render tool existing (#1467). Transcripts store the path rather than the
payload; a base64 blob in a ConfigMap transcript would be unpleasant.

Assisted-by: Claude Code (design, implementation and tests; I verified the
model half on hardware first, confirmed the emitted wire format matches a
request I had already exercised against the live server, and ran the suites and
lint myself)


Defilan:
Pushed a follow-up: the workspace confinement was lexical, and `filepath.Join` resolves nothing, so a symlink sitting *inside* the workspace and pointing outside it passed the prefix check and got read and base64'd into the request.

Reachable rather than theoretical, which is why it matters here: the coder has a bash tool and writes into this workspace, and cloned repositories routinely carry symlinks. The check exists to stop exactly the "read an arbitrary host file and ship it to an inference endpoint" primitive, so it has to hold against the coder itself.

Now compares fully resolved paths, resolving the root too (it is commonly a link, `/tmp -> /private/tmp` on macOS) and reading the resolved path so a link swapped in after the check cannot be followed. Tests cover both directions, since refusing all symlinks would break ordinary repositories.

Residual and deliberate: a parent directory replaced between the check and the read would still win. Closing that needs `openat2`/`RESOLVE_BENEATH` and is Linux-only, which is not a trade worth making for a workspace the agent owns.

Defilan:
Tested on the fleet, deployed as `0.9.15-pr1471.gf0a995a` on `foreman-agent` and
`foreman-operator`. The chart CRDs rode along, so the API server accepts
`agents.spec.vision` and `agentictasks.spec.payload.images`.

Pinned the run so it could not land on a stale agent: the two
`foreman-agent-multirepo` pods still ran released `v0.9.15`, which would have
ignored `payload.images` silently and produced a text-only run that looked like
a feature failure. Scaled to 0 for the test, restored after.

## Setup

An AgenticTask against a small Three.js repo, with the blown-out render attached
as `payload.images: ["screenshots/render.png"]`. The prompt asks two questions,
one of which has an answer that cannot be reached by reading code: transcribe
the HUD panel. `Pollen` is `particleSystem.activeCount`, assigned at runtime in
`js/ui.js:71`, so its value exists nowhere in the repository.

## A: with the image

The first user message went out as a content-parts array:

```
text part:  745 chars
image part: image_url | data:image/png;base64,iVBORw0KGgoAAAANSU... | 484,166 chars
```

356KB PNG in, 484KB of base64 on the wire, which is the inflation the cap
accounts for.

Transcription against ground truth:

| line | actual | model |
|---|---|---|
| 1 | `Health80%` | `Health88%` |
| 2 | `Pollen26` | `Pollen26` |
| 3 | `Wind0 m/s` | `Wind0 m/s` |
| 4 | `Night6%` | `Night66%` |

Two of four exact, including the runtime-only `Pollen26`. The strongest single
piece of evidence is in the reasoning that preceded any tool call, where it read
line 2 as `Pain10%`. That is a misread of the pixels in `Pollen`. No amount of
source reading produces the string "Pain", so the image is demonstrably what it
was working from.

It also returned a defect nobody asked it for:

> HUD labels and values are concatenated without any spacing between them (e.g.
> "Health88%" instead of "Health 88%"). The adjacent `<span>` elements for label
> and value have no CSS margin/padding to separate them.
>
> file: `js/ui.js`, parameter: innerHTML construction of stats-row elements

Verified, that is real. The spans are emitted back to back with no separator,
and it is visible in the render and invisible in a diff.

## B: control, same task without the image

Byte-identical task with `payload.images` omitted:

```
first user content type: plain string
```

The plain string path, so the no-image case is unchanged on the wire, which is
what the nil-parts return exists to guarantee. Verdict:

> ERROR: No screenshot image was attached to the conversation; unable to inspect
> the rendered output visually.

It refused to answer rather than reconstructing something plausible from source,
which is the behaviour the system prompt asked for and the right failure mode.

## Limitations worth recording

Fine text at 768px is not reliably legible to this model: `80` read as `88` and
`6` as `66`, both trailing-digit doubling. Values transcribed from a screenshot
should not be trusted as exact.

More interesting, it did not flag the overexposure, which is the larger defect
in that frame and the one this feature was motivated by. It anchored on the HUD
question and answered question two from there. The capability is present (in the
hand test before this PR the same model described the blowout and its cause
unprompted); what this run shows is that attention is promptable and a
review-oriented prompt will need to ask about the whole frame.

## Change since the last push

Added a success log line. Previously only the skip paths logged, so the sole
positive signal that an image reached the model was the absence of a warning,
which reads exactly like a run that never attached anything. Confirming this
test worked meant pulling the transcript ConfigMap and grepping for `data:image`
rather than reading the agent log. It now logs the count and the encoded size.

PR #1474, opened 2026-08-09
fix(dashboard): retarget amd-gpu-observability off amdgpu_*-only queries

### LINKED ISSUE #1357: [BUG] amd-gpu-observability dashboard ships default-on but queries amdgpu_* metrics no shipped component emits
Chart 0.9.13, `grafana.dashboards.enabled: true`, `only: []`.

`llmkube-amd-gpu-observability` is published on every install (the auto-mode runtime gate covers only `sglang-dashboard` and `vllm-dashboard`). Every panel is blank on a real AMD cluster because the metric names have no producer in this repo:

| Dashboard metric | Present in cluster |
|---|---|
| `amdgpu_gpu_busy_percent` | no |
| `amdgpu_gtt_used_bytes` | no |
| `amdgpu_vram_used_bytes` / `amdgpu_vram_total_bytes` | no |
| `amdgpu_sclk_mhz` | no |

Checked against VictoriaMetrics on a cluster serving SGLang on an AMD R9700 plus two llama.cpp services on Radeon 660M iGPUs. The names that do exist come from drm-exporter: `drm_engine_utilization_ratio`, `drm_memory_used_bytes{pool="vram"}`, `drm_memory_total_bytes`, `drm_temperature_celsius`, `drm_power_watts`, `drm_frequency_hertz`, `drm_fan_speed_rpm`, plus `node_drm_gpu_busy_percent` from node-exporter.

So the dashboard presupposes the exporter from #1185 / #1187, which is not shipped yet, while being published by default today. Either gate it on the exporter shipping, or point it at metric names that exist (drm-exporter and node-exporter are already common on AMD Kubernetes nodes).

One gap worth keeping when this is retargeted: the GTT panel. On AMD iGPUs GTT is the pool that actually exhausts host memory and it is invisible to `kubectl top`, so a working GTT series is the most valuable panel on the board.

Workaround in use: `grafana.dashboards.only: [llmkube-inference, sglang-dashboard]`.

Tanguille:
## What

Retarget `amd-gpu-observability.json`'s 8 `amdgpu_*`-only queries to `or`-chain
across amdgpu-sysfs, drm-exporter, and (for GTT) node-exporter's DRM
collector.

## Why

Fixes #1357

8 of the dashboard's 16 panels hardcode `amdgpu_*`, so it's permanently blank
on any cluster without the amdgpu-sysfs exporter specifically. Verified live
on a cluster running drm-exporter + node-exporter (SGLang on an R9700, two
llama.cpp services on Radeon 660M iGPUs): `amdgpu_*` has zero series there.
The Temperature and Power panels already key off `node_hwmon_*` and were
already vendor-correct — this only touches Busy %, Clock, GTT, and VRAM
Used/Total.

GTT needed a third source: neither amdgpu-sysfs's `amdgpu_gtt_used_bytes` nor
drm-exporter (only `pool="vram"`/`pool="smem"` on this cluster) expose it,
it comes from node-exporter's own DRM collector,
`node_drm_memory_gtt_used_bytes`, already what this cluster's own
`NodeGTTMemoryHigh` alert uses.

## How

Same `or`-chain pattern #1372 already established for the alert rules
(`llmkube.gpuMetricNames` in `_helpers.tpl`), `amdgpu_*` kept first in each
chain so nothing regresses for whoever runs that exporter. Dashboards are raw
JSON (`.Files.Glob "dashboards/*.json"`, no Helm templating applied inside),
so the helper itself isn't reachable from here — each `or`-chain is written
directly into the panel's `expr` string instead.

| Panel | New expr |
|---|---|
| GPU Busy % (+ History) | `amdgpu_gpu_busy_percent{card="card0"} or drm_engine_utilization_ratio{engine="gpu"} * 100` |
| GPU Clock (SCLK) | `(amdgpu_clock_hertz{card="card0", sensor="sclk"} or drm_frequency_hertz{domain="sclk", kind="actual"}) / 1000000` |
| GTT Used (+ History) | `amdgpu_gtt_used_bytes{card="card0"} or node_drm_memory_gtt_used_bytes` |
| VRAM Used (+ History) | `amdgpu_vram_used_bytes{card="card0"} or drm_memory_used_bytes{pool="vram"}` |
| VRAM Total | `amdgpu_vram_total_bytes{card="card0"} or drm_memory_total_bytes{pool="vram"}` |

All 5 new expressions verified live against the reference cluster's
VictoriaMetrics. Each returns non-empty series (GTT specifically now
returns all 3 GPU nodes).

Independent of #1446: that PR removes the dashboard's 4 runtime-agnostic
llama.cpp panels and fixes its datasource placeholder, this one fixes the
AMD-specific panels' vendor lock-in. Same file, non-overlapping line ranges
as of this branch's base. Flagging in case #1446 lands first and this needs
a rebase.

## Checklist

- [x] Tests added/updated — n/a, static JSON data change; `helm unittest` (72/72) and `python3 -m json.tool` cover the file
- [ ] `make test` passes locally — not run; no Go changed by this PR (`make fmt`/`vet`/`lint` all pass)

### DIFF
--- a/charts/llmkube/dashboards/amd-gpu-observability.json
+++ b/charts/llmkube/dashboards/amd-gpu-observability.json
@@ -234,7 +234,7 @@
             "type": "prometheus",
             "uid": "${datasource}"
           },
-          "expr": "amdgpu_gpu_busy_percent{card=\"card0\"}",
+          "expr": "amdgpu_gpu_busy_percent{card=\"card0\"} or drm_engine_utilization_ratio{engine=\"gpu\"} * 100",
           "legendFormat": "GPU busy",
           "refId": "A"
         }
@@ -295,7 +295,7 @@
             "type": "prometheus",
             "uid": "${datasource}"
           },
-          "expr": "amdgpu_clock_hertz{card=\"card0\", sensor=\"sclk\"} / 1000000",
+          "expr": "(amdgpu_clock_hertz{card=\"card0\", sensor=\"sclk\"} or drm_frequency_hertz{domain=\"sclk\", kind=\"actual\"}) / 1000000",
           "legendFormat": "SCLK",
           "refId": "A"
         }
@@ -377,7 +377,7 @@
             "type": "prometheus",
             "uid": "${datasource}"
           },
-          "expr": "amdgpu_gtt_used_bytes{card=\"card0\"}",
+          "expr": "amdgpu_gtt_used_bytes{card=\"card0\"} or node_drm_memory_gtt_used_bytes",
           "legendFormat": "GTT used",
           "refId": "A"
         }
@@ -446,7 +446,7 @@
             "type": "prometheus",
             "uid": "${datasource}"
           },
-          "expr": "amdgpu_vram_used_bytes{card=\"card0\"}",
+          "expr": "amdgpu_vram_used_bytes{card=\"card0\"} or drm_memory_used_bytes{pool=\"vram\"}",
           "legendFormat": "VRAM used",
           "refId": "A"
         }
@@ -506,7 +506,7 @@
             "type": "prometheus",
             "uid": "${datasource}"
           },
-          "expr": "amdgpu_vram_total_bytes{card=\"card0\"}",
+          "expr": "amdgpu_vram_total_bytes{card=\"card0\"} or drm_memory_total_bytes{pool=\"vram\"}",
           "legendFormat": "VRAM total",
           "refId": "A"
         }
@@ -949,7 +949,7 @@
             "type": "prometheus",
             "uid": "${datasource}"
           },
-          "expr": "amdgpu_gpu_busy_percent{card=\"card0\"}",
+          "expr": "amdgpu_gpu_busy_percent{card=\"card0\"} or drm_engine_utilization_ratio{engine=\"gpu\"} * 100",
           "legendFormat": "GPU busy",
           "refId": "A"
         }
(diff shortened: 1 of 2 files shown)


Defilan (approved):
Approving. Verified all 8 retargeted queries, and the idiom matches what #1372
established in `llmkube.gpuMetricNames` down to the
`drm_engine_utilization_ratio{engine="gpu"} * 100` ratio-to-percent conversion,
so the dashboard and the alert rules now agree on how to read a non-sysfs AMD
fleet. Keeping `amdgpu_*` first in each chain preserves existing behaviour on
Strix.

The `"drm_"` line in `externalPrefixes` is registering a new external exporter
alongside `amdgpu_` and `node_`, not relaxing the check, which is the right way
to do that.

Coordination note: this edits the same file as #1446, also open. There is no
textual conflict, and I applied both to confirm the semantics survive rather
than assuming: #1446 then #1474 gives 14 panels, `DS_PROMETHEUS` only, zero bare
`${datasource}` references, zero `llamacpp` references, and all 8 or-chains
intact. They compose in either order. I will take #1446 first since this one is
written against the 20-panel version.

Two things I noticed, neither blocking:

`NodeGTTMemoryHigh` is not in this repo, so that justification describes your
cluster rather than something here. `node_drm_memory_gtt_used_bytes` is a new
external dependency for us, which is fine, but worth naming as such.

The alert rules centralize the or-chain in a helper while the dashboard now
hardcodes it in 8 places, so a fourth exporter means updating two mechanisms.
Not worth solving here, since the dashboards ship as plain JSON, but it is the
kind of thing that drifts.

PR #1485, opened 2026-08-10
fix(modelpool): guard the nil HTTPClient that panics every swap

Defilan:
## What

Guards the nil `HTTPClient` that panics on every ModelPool swap.

## Why

**ModelPool cannot complete a single swap in 0.9.16.**

`cmd/main.go:476` constructs `ModelPoolReconciler` with `Client`, `Scheme` and
`Recorder` and never sets `HTTPClient`, so it is nil in the running operator.
`memberIdle` passed that nil straight into `IdleProbe`, and
`net/http.(*Client).do` dereferences it:

```
net/http.(*Client).do(0x0, ...)
runtime_llamacpp.go:236      llamaCppSlotsIdleProbe.func1
modelpool_controller.go:366  memberIdle
modelpool_controller.go:145  Reconcile
```

The panic takes the reconcile down, so the pool wedges in `Swapping`. Every
request for a non-resident member then blocks for the full `swapBudget` and
returns `503 model pool activation did not complete within the request budget`.

Observed on hardware, not inferred. A two-member pool on a DGX Spark held a
request for exactly 300s and never swapped, with the incumbent still resident:

```
status:
  phase: Swapping
  residentMember: s5-pool-fusion
  conditions:
    SlotAllocated  False  Swapping      No member resident yet
    SwapDeferred   False  NotDeferred   No swap is being deferred
```

Note `SwapDeferred: NotDeferred`. That is the second half of the problem: the
panic also defeats the fail-closed contract `memberIdle` documents. The probe
cannot defer a swap if it panics before it can return an error, so the pool
reports nothing wrong while nothing works.

## How

Same guard the rollout drain has used all along
(`drain_before_rollout.go:138`), with the same 5s default, so both users of this
probe behave alike.

Guarding in the reconciler rather than only wiring `main.go` is deliberate: the
field is optional on the struct, and a future construction site that forgets it
would reintroduce a panic rather than a degraded probe.

## Why the suite missed it

Every other ModelPool test injects the `IdleCheck` hook, which returns before
`HTTPClient` is ever read, so no existing test touches the real probe path. The
regression test here deliberately omits that hook and constructs the reconciler
in exactly the shape `main.go` uses.

It asserts two things: that it does not panic, and that an unreachable member is
never reported idle, since reporting idle would unload a possibly-busy model.

I verified the test fails with the guard removed and passes with it, rather than
assuming a new test must be catching something:

```
--- FAIL: TestMemberIdle_NilHTTPClientDoesNotPanic
    memberIdle panicked with a nil HTTPClient: runtime error: invalid memory
    address or nil pointer dereference
```

## Testing

`internal/controller` envtest suite passes (255s), `make lint` 0 issues.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (if user-facing change) — no doc describes the field

Assisted-by: Claude Code (found while running ModelPool on real hardware for a
GPU-sharing benchmark; wrote the fix and the regression test, and ran the suite,
the lint and the fails-without-the-fix check)

### DIFF
--- a/internal/controller/modelpool_controller.go
+++ b/internal/controller/modelpool_controller.go
@@ -363,7 +363,23 @@ func (r *ModelPoolReconciler) memberIdle(ctx context.Context, isvc *inferencev1a
 		}
 		baseURL = fmt.Sprintf("http://%s.%s.svc.cluster.local:%d", sanitizeDNSName(isvc.Name), isvc.Namespace, port)
 	}
-	return detector.IdleProbe(isvc, r.HTTPClient)(ctx, baseURL)
+	// HTTPClient is optional on the struct and is not wired at any construction
+	// site, so it is nil in the running operator. Passing nil straight through
+	// panics inside net/http.(*Client).do the first time a swap needs an idle
+	// check, which takes the whole reconcile down and wedges the pool in
+	// Swapping: every request against it then burns the full swapBudget and
+	// returns 503.
+	//
+	// It also defeats the fail-closed contract this function documents. The
+	// probe cannot defer a swap if it panics before it can return an error.
+	//
+	// Same guard the rollout drain already uses (drain_before_rollout.go), and
+	// the same default timeout, so both users of this probe behave alike.
+	httpClient := r.HTTPClient
+	if httpClient == nil {
+		httpClient = &http.Client{Timeout: 5 * time.Second}
+	}
+	return detector.IdleProbe(isvc, httpClient)(ctx, baseURL)
 }
 
 // metalMembers returns the names of members whose referenced Model targets a
--- a/internal/controller/modelpool_nil_client_test.go
+++ b/internal/controller/modelpool_nil_client_test.go
@@ -1,0 +1,76 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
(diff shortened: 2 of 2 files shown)


Defilan:
Verified on hardware. Two DGX Sparks, a two-member ModelPool
(Qwopus3.6-27B-Fusion Q4_K_M and Mellum2-12B Q4_K_M) fronted by a ModelRouter
with `defaultRouteStrategy: BackendNameMatch`.

**Before the fix**, on `llmkube-controller:0.9.16`, a request for the
non-resident member was held for the full budget and failed:

```
http=503  wall=300.0s
{"error":{"code":503,"message":"model pool activation did not complete within the request budget; retry"}}
resident after: s5-pool-fusion     # unchanged
```

with the panic in the controller log on every reconcile, and the pool stuck
reporting `SwapDeferred: NotDeferred` while nothing was deferring.

**After the fix**, same request, same pool:

```
http=200  SWAP+SERVE wall=122.0s
resident before: s5-pool-fusion
resident after:  s5-pool-mellum
s5-pool-fusion  Stopped
s5-pool-mellum  Ready
```

The response came from the Mellum GGUF, so the swap completed rather than
silently serving the incumbent.

The router-proxy metrics corroborate the wall clock independently. The
hold-duration histogram put its single observation in `le="204.8"` with nothing
in `le="102.4"`, bracketing the measured 122.0s:

```
llmkube_modelpool_hold_duration_seconds_bucket{...,le="102.4"} 0
llmkube_modelpool_hold_duration_seconds_bucket{...,le="204.8"} 1
llmkube_modelpool_held_requests{...} 0
```

**Anti-thrash coalescing also verified.** Two concurrent requests for the same
non-resident member both returned 200, and produced one swap, not two:

```
llmkube_modelpool_coalesced_total{member="s5-pool-fusion",...} 2
llmkube_modelpool_swaps_total{from="s5-pool-fusion",to="s5-pool-mellum",...} 1
llmkube_modelpool_swaps_total{from="s5-pool-mellum",to="s5-pool-fusion",...} 1
llmkube_modelpool_resident{member="s5-pool-fusion",...} 1
llmkube_modelpool_resident{member="s5-pool-mellum",...} 0
```

Swaps in both directions, one each, and the resident gauge agrees with the
cluster.

For context on the cost this measures: **122s to switch members**, against a
time-sliced arm on the same hardware that pays no switch penalty but shares
throughput four ways. That is the real trade between the two mechanisms, and it
was not measurable at all until this fix.

One incidental note for anyone reproducing: `spec.default` only warms a cold
pool and is a no-op once an owner is picked, so a swap has to be driven by
demand through the router. That is documented on the field, but it is easy to
reach for first.

PR #1486, opened 2026-08-10
fix(foreman): one tool list, so a tool cannot ship unusable again

### LINKED ISSUE #1482: fetch_pull_request missing from webhook tool catalog — shipped unusable in v0.9.16
## Bug

v0.9.16 ships `fetch_pull_request` (#1458) in a state where no Agent CR can enable it. The tool is registered in the runtime registry (`cmd/foreman-agent/main.go:822`), but `pkg/foreman/agent/tools/catalog/catalog.go` — the allow-list the Agent admission webhook validates `spec.tools` against — was not updated. Since `spec.tools` is a mandatory whitelist (`minItems: 1`), adding the tool to an Agent is rejected at admission:

```
admission webhook "vagent.foreman.llmkube.dev" denied the request:
Agent.foreman.llmkube.dev "coder" is invalid: spec.tools[6]: Unsupported value:
"fetch_pull_request": supported values: "bash", "fetch_issue", "grep",
"read_file", "run_gate_job", "str_replace", "submit_result", "write_file"
```

…and NOT adding it means the runtime `Filter()` never surfaces it to the model. Either way the feature is unreachable. (In a GitOps setup this also turns the webhook rejection into a failed Kustomization dry-run, blocking the whole path until reverted.)

## Why the drift guard missed it

`TestCatalogNamesMatchConstructors` compares the catalog against `canonicalToolSet()` in `catalog_drift_test.go` — which is itself a hand-maintained mirror of `makeRegistryFactory`. #1458 added the tool to `main.go` only; the two mirrors stayed stale *together*, so the test kept passing. Three copies of the same list, and the guard only checks the two that nobody changed.

## Fix

1. Add `"fetch_pull_request"` to `canonicalToolNames` in `pkg/foreman/agent/tools/catalog/catalog.go` (sorted position) and to `canonicalToolSet()` in the drift test.
2. Close the guard hole so this cannot recur: derive the drift test's constructor set from the same code path production uses (export the tool list from `makeRegistryFactory`, or move registration into `pkg/foreman/agent/tools` where the test can enumerate constructors mechanically). A drift test whose reference side is hand-maintained guards nothing.

Found on v0.9.16, operator + agent both at that version; reproduced with a minimal Agent CR whose `tools` list appends `fetch_pull_request`.

Defilan:
## What

Collapses the agent's tool list to one place, so a tool cannot ship registered
but unusable again.

## Why

Fixes #1482

**The drift is wider than reported.** Three tools shipped in v0.9.16 registered
at runtime but rejected by admission, not one:

| Tool | Runtime registry | Catalog allow-list |
|---|---|---|
| `fetch_pull_request` | yes | **missing** |
| `run_integrate` | yes | **missing** |
| `run_reconcile` | yes | **missing** |

Each is unreachable in both directions: naming it in `spec.tools` fails the
webhook, and omitting it means `Filter()` never surfaces it to the model.
`run_integrate` and `run_reconcile` had simply not been tried yet.

@joryirving's diagnosis of the cause is exactly right. The same list lived in
three places, and the guard compared the two nobody had edited.

## How

`tools.BuildAll` becomes the single construction site and `main.go` calls it, so
there is no second list to drift from. 49 lines of inline construction in
`cmd/foreman-agent/main.go` become one call.

`catalog` deliberately stays a zero-import leaf. The operator's Agent admission
webhook links it to validate `spec.tools`, and must not pull in Kubernetes
clients, GitHub clients or the slicer, so the dependency has to run
`tools -> catalog` and not the reverse. That rules out having catalog derive
itself from the constructors, which is why the invariant is enforced by a test
rather than by the type system.

## The guards, and how I know they work

Three, each verified by deliberately breaking it rather than assumed:

**Names must match catalog, both directions.** Removing `fetch_pull_request`
from catalog, which is precisely the #1482 shape:

```
--- FAIL: TestBuildAllMatchesCatalog
    BuildAll exposes 11 tools, catalog allows 10
```

Adding a name with no constructor fails too.

**Names must be unique**, so one tool cannot shadow another while another goes
missing and the count still matches.

**Every Tool implementation must be wired into BuildAll**, parsed from the
package source. This is the case the first two cannot see: a new tool type
wired nowhere is consistent with catalog, because both omit it. Removing
`RunReconcileTool` from `BuildAll`:

```
--- FAIL: TestBuildAllWiresEveryToolType
    the package declares 11 Tool implementations but BuildAll wires 10
```

The old `canonicalToolSet()` in `catalog_drift_test.go` now returns
`BuildAll(ToolDeps{})`, so its reference side is production code rather than a
copy of it. A drift test whose reference is hand-maintained guards nothing,
which is the lesson of this bug.

## Testing

`pkg/foreman/...` passes, `make lint` 0 issues.

Not added: the startup assertion I floated in discussion (agent refuses to boot
if a registered tool is missing from catalog). It is now redundant, since the
condition cannot reach a build, and it would fail the fleet at boot for a
problem CI already catches.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (if user-facing change) — no doc enumerates the tool set

Assisted-by: Claude Code (verified the issue's claims against the source, found
the two additional missing tools, wrote the refactor and the guards, and ran the
break-it checks and suites)

### DIFF
--- a/cmd/foreman-agent/main.go
+++ b/cmd/foreman-agent/main.go
@@ -72,7 +72,6 @@ import (
 	"github.com/defilantech/llmkube/pkg/foreman/agent/codehost"
 	"github.com/defilantech/llmkube/pkg/foreman/agent/githubissue"
 	"github.com/defilantech/llmkube/pkg/foreman/agent/githubpr"
-	"github.com/defilantech/llmkube/pkg/foreman/agent/githubprfetch"
 	"github.com/defilantech/llmkube/pkg/foreman/agent/mcp"
 	"github.com/defilantech/llmkube/pkg/foreman/agent/repo"
 	foremantools "github.com/defilantech/llmkube/pkg/foreman/agent/tools"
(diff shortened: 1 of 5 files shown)


joryirving (approved):
LGTM - Two quick notes/minor nits, neither blocking:

1. The AST guard counts, and can't see embedded methods. toolTypeNames finds types with a declared Name() string method, so a future Tool implemented via struct embedding (promoted Name()) would be invisible to it, producing a spurious built > declared failure. And count-equality gives weaker diagnostics than a set-diff; comparing AST type names against reflect.TypeOf(tool).Elem().Name() from BuildAll would say which type is unwired. Worth a follow-up, not a re-spin.

2. Stale doc fragment: the catalog package-level comment still describes the list as "a hand-maintained mirror" checked by the old mechanism. It's technically still hand-maintained, and the claim "fails the build if the two ever diverge" is finally true now, but it doesn't mention BuildAll - the updated var-level comment does. One-line touch-up.

Issue #1491, opened 2026-08-10
foreman metrics: operator-emitted task outcome counters; addendum: foreman.crs.namespace for the CRS ConfigMap


joryirving:
## Problem

The operator exports only stock controller-runtime metrics. Task outcomes — the data an operator of the *fleet* needs — live solely on AgenticTask/Workload CRs, which are deleted with their Workload on every retry/recreate. There is no durable record of per-Agent verdict rates, task duration, or outcome mix, so questions like "is my self-hosted coder's GO-rate worse than my hosted one's?" or "did the last prompt change increase NO-GOs?" are unanswerable without external scraping.

Downstream we've worked around it with kube-state-metrics `customResourceState` (exporting `status.phase`/`status.verdict` labeled by `spec.agentRef.name`), which works but only captures states a 30s scrape happens to see, and can't carry rich outcome data (`result.extra.outcome`, elapsed, turns).

## Proposal

Emit task lifecycle metrics from the operator (it already runs the metrics endpoint on `:8081`):

```
foreman_task_completed_total{agent, kind, verdict, outcome}   # counter: GO / NO-GO / INCOMPLETE / ERROR × PUSH-FAILED / ALREADY-RESOLVED / MODEL-DECIDED / ...
foreman_task_duration_seconds{agent, kind, verdict}           # histogram, from result.elapsedSec
foreman_task_turns{agent, kind}                               # histogram, from result.extra.turnCount
foreman_workload_completed_total{outcome}                     # completed-with-PR / completed-no-pr / failed
```

Labels are bounded (Agent names, task kinds, verdict/outcome enums) so cardinality stays trivial. Counters emitted once per task at terminal-phase transition in the existing reconcile path — survives CR deletion by construction.

## Why the operator and not the agent

Job-based agents are ephemeral (one pod per task) and can't hold counters; the operator observes every terminal transition regardless of execution mode, and already owns the CR lifecycle.

---

## Addendum: `foreman.crs.namespace` for the existing CRS config

Until (and independent of) operator-emitted counters, the shipped CRS config (#1086) has a consumption gap: kube-state-metrics can only mount ConfigMaps from **its own** namespace, and the chart renders `<fullname>-crs` in the foreman release namespace. An operator whose KSM lives in an observability namespace must either maintain a verbatim copy of the CRS definitions in their KSM values (drifts on every chart bump) or run a dedicated CRS-only KSM next to foreman (works, but is an extra deployment some clusters won't want).

[...]

PR #1492, opened 2026-08-10
feat(chart): foreman.crs.namespace — render the CRS ConfigMap where kube-state-metrics can mount it

joryirving:
## What

Adds `foreman.crs.namespace`: when set, the CRS ConfigMap renders into that namespace instead of the release namespace. Empty default preserves existing behavior byte-for-byte.

## Why

Refs #1491 (addendum). The values comment tells operators to point their kube-state-metrics at this ConfigMap, but pods can only mount ConfigMaps from their own namespace — and kube-state-metrics typically lives in an observability namespace, not Foreman's. Today that leaves consumers two bad options: maintain a verbatim copy of the CRS definitions in kube-state-metrics values (drifts on every chart bump), or run a dedicated CRS-only kube-state-metrics deployment next to Foreman. Helm creates cross-namespace resources without complaint; only the pod mount is namespace-bound, so a one-line namespace override closes the gap.

Scope note: only `crs-configmap.yaml` honors the value. The PrometheusRule is discovered by label selectors cross-namespace, and the Grafana dashboard ConfigMap serves sidecar-based setups where placement needs differ — both left as-is deliberately.

## Testing

New `charts/foreman/tests/crs_test.yaml`: default renders into the release namespace, override renders into the target namespace, and the override is independent of the chart-wide `namespace` value. `helm unittest charts/foreman` — 35/35 across both suites; `helm lint` clean with the override set.

## Checklist

- [x] Tests added/updated
- [x] AI assistance disclosed: Claude Code wrote the change and tests from my design and ran `helm unittest` (35/35) and `helm lint` in my environment
- [x] Human review: diff walked through and approved
- [x] Documentation: the new value is documented in values.yaml alongside the existing crs comment block

### DIFF
--- a/charts/foreman/templates/crs-configmap.yaml
+++ b/charts/foreman/templates/crs-configmap.yaml
@@ -10,7 +10,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ include "foreman.fullname" . }}-crs
-  namespace: {{ include "foreman.namespace" . }}
+  namespace: {{ .Values.foreman.crs.namespace | default (include "foreman.namespace" .) }}
   labels:
     {{- include "foreman.labels" . | nindent 4 }}
 data:
--- a/charts/foreman/tests/crs_test.yaml
+++ b/charts/foreman/tests/crs_test.yaml
@@ -1,0 +1,37 @@
+suite: CRS ConfigMap namespace override (#1491)
+templates:
+  - crs-configmap.yaml
+
+tests:
+  - it: renders into the release namespace by default
+    set:
+      foreman.crs.enabled: true
+    release:
+      namespace: llm
+    asserts:
+      - equal:
+          path: metadata.namespace
+          value: llm
+
+  - it: foreman.crs.namespace renders the ConfigMap into that namespace
+    set:
+      foreman.crs.enabled: true
+      foreman.crs.namespace: observability
+    release:
+      namespace: llm
+    asserts:
+      - equal:
+          path: metadata.namespace
+          value: observability
+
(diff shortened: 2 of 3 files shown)


Defilan (approved):
Approving. Verified the two claims that matter here.

**Default behaviour really is unchanged.** Rendered both revisions with
`foreman.crs.enabled=true` and compared the CRS ConfigMap rather than eyeballing
the template (a raw `helm template` diff is useless here, since the self-signed
webhook certs are regenerated on every render):

```
main : name=fm-foreman-crs ns=llm dataSHA=592589700184c138
PR   : name=fm-foreman-crs ns=llm dataSHA=592589700184c138
```

**The override is correctly scoped.** With `foreman.crs.namespace=observability`
set, exactly one resource moves and the data is untouched:

```
namespace=llm            resources=11
namespace=observability  resources=1     (fm-foreman-crs, same dataSHA)
```

That matches the scope note about the PrometheusRule and the dashboard being
left alone. `helm unittest charts/foreman` is 35/35 here too, and the foreman
chart has no `values.schema.json` (only the llmkube chart does), so there is no
schema to keep in step.

One non-blocking note for the values comment. Helm will not create the target
namespace, and the failure is at apply time rather than render time:

```
$ kubectl apply --dry-run=server -f cm-in-missing-ns.yaml
Error from server (NotFound): namespaces "..." not found
```

In practice the namespace exists, since the whole point is that
kube-state-metrics already lives there. But an operator who sets this before
standing up the namespace gets a failed release rather than an obvious message,
and the existing comment is the natural place to say so. Something like "the
namespace must already exist; Helm will not create it" would close it. Happy to
merge without it.

PR #1494, opened 2026-08-10
fix(chart): heartbeat alert uses time(), not the nonexistent time_seconds()

### LINKED ISSUE #1493: chart: CRS alert ForemanFleetNodeHeartbeatStale uses nonexistent time_seconds() — enabling alerts fails the whole Helm upgrade
## Bug

`foreman.crs.alerts.enabled: true` cannot install on any cluster running prometheus-operator's rule validation. `ForemanFleetNodeHeartbeatStale` calls `time_seconds()`, which is not a PromQL function (the function is `time()`):

```
Helm upgrade failed for release llm/foreman with chart foreman@0.9.16:
server-side apply failed for object llm/foreman-alerts monitoring.coreos.com/v1,
Kind=PrometheusRule: admission webhook "prometheusrulevalidate.monitoring.coreos.com"
denied the request: Rules are not valid
```

Because the rule is applied as part of the release, the **whole Helm upgrade fails and rolls back** — enabling the alerts takes down the upgrade path, not just the alerts.

The other two shipped exprs validate fine; it's just the heartbeat rule.

## Fix

```diff
- expr: foreman_fleetnode_heartbeat_timestamp_seconds > 0 and (time_seconds() - foreman_fleetnode_heartbeat_timestamp_seconds) > {{ .Values.foreman.crs.alerts.staleHeartbeatSeconds }}
+ expr: foreman_fleetnode_heartbeat_timestamp_seconds > 0 and (time() - foreman_fleetnode_heartbeat_timestamp_seconds) > {{ .Values.foreman.crs.alerts.staleHeartbeatSeconds }}
```

Like #1482, this shipped unexercised: helm-unittest checks the template renders, not that the PromQL parses. A cheap guard would be a chart test (or CI step) that runs each rendered expr through `promtool check rules` / the Prometheus parse endpoint.

Found on v0.9.16, kube-prometheus-stack with the default rule-validation webhook; each expr validated individually against a live Prometheus to isolate the failing one.

joryirving:
## What

One token in the `ForemanFleetNodeHeartbeatStale` alert expression: `time_seconds()` → `time()`.

## Why

Fixes #1493

`time_seconds()` is not a PromQL function, so prometheus-operator's rule-validation webhook rejects the PrometheusRule — and because the rule applies with the release, `foreman.crs.alerts.enabled: true` fails the **entire Helm upgrade** and rolls it back. The alerts feature has been un-installable on any cluster running rule validation since it shipped.

## How

Straight substitution; `time()` returns the evaluation timestamp in seconds, which is exactly what the expression's `(now - heartbeat_timestamp) > threshold` arithmetic assumes. Verified by rendering the rule and parsing every expression through a live Prometheus query API: 2/3 valid before the change (heartbeat rejected), 3/3 after. Like #1482, this shipped unexercised because helm-unittest proves templates render, not that PromQL parses — a `promtool check rules` CI guard is suggested in #1493 but deliberately out of scope for a one-token fix.

## Checklist

- [ ] Tests added/updated — none added: a one-token PromQL fix has no template-shape change for helm-unittest to assert; the durable guard (promtool over rendered rules in CI) is proposed in #1493
- [x] `make lint` passes locally (0 issues; `make test-chart` also passes, 105/105)
- [ ] Documentation updated (if user-facing change) — no docs surface changes; values.yaml comments already describe the alert and are unaffected

Assisted-by: Claude Code (diagnosed the failing webhook admission, made the change, and ran `make test` / `make lint` / `make test-chart` plus live-Prometheus validation of all three rendered expressions; I reviewed the diff and approved submission)

### DIFF
--- a/charts/foreman/templates/prometheusrule.yaml
+++ b/charts/foreman/templates/prometheusrule.yaml
@@ -47,7 +47,7 @@ spec:
     # fires after `staleHeartbeatDuration` of silence, which must be >
     # 90s to avoid flapping on a single missed beat.
     - alert: ForemanFleetNodeHeartbeatStale
-      expr: foreman_fleetnode_heartbeat_timestamp_seconds > 0 and (time_seconds() - foreman_fleetnode_heartbeat_timestamp_seconds) > {{ .Values.foreman.crs.alerts.staleHeartbeatSeconds }}
+      expr: foreman_fleetnode_heartbeat_timestamp_seconds > 0 and (time() - foreman_fleetnode_heartbeat_timestamp_seconds) > {{ .Values.foreman.crs.alerts.staleHeartbeatSeconds }}
       for: {{ .Values.foreman.crs.alerts.staleHeartbeatDuration }}
       labels:
         severity: critical


Defilan (approved):
Approving. I reproduced the failure and the fix independently rather than
taking the description on faith, by rendering the chart at both revisions and
parsing the rules through promtool:

```
=== promtool check rules (main) ===
  FAILED: group "foreman-fleetnode", rule 1, "ForemanFleetNodeHeartbeatStale":
  could not parse expression: 1:56: parse error: unknown function with name "time_seconds"

=== promtool check rules (this PR) ===
  SUCCESS: 3 rules found
```

Also checked the arithmetic rather than just the parse, since `time()` is only
the right substitution if the other operand is a Unix epoch. The CRS config
declares it as one:

> `foreman_fleetnode_heartbeat_timestamp_seconds` (gauge = unix seconds of
> LastHeartbeatTime, 0 when absent)

So `time() - heartbeat > threshold` is correct, and the `> 0` guard covers the
absent-heartbeat case. Confirmed this was the only `time_seconds()` in the tree.

Agreed on keeping the CI guard out of scope for a one-token fix. Worth noting
for #1493 that the check is cheap, because the whole guard is what I ran above:
render with `foreman.crs.enabled=true` and `foreman.crs.alerts.enabled=true`,
pull `spec.groups` out of the PrometheusRule into a plain rules file, and run
`promtool check rules` over it. It catches this exact class, and the existing
"Test Prometheus Integration" job is already the natural home for it.

Good catch on the blast radius being the whole Helm upgrade rather than just the
alert.

PR #1505, opened 2026-08-11
feat: fetch_pull_request: include check-run output (title/summary/text) for failed runs (#1481)

### LINKED ISSUE #1481: fetch_pull_request: include check-run output (title/summary/text) for failed runs
## Problem

`fetch_pull_request` (#1458) reports check runs as `{name, conclusion, details_url}` only — `pkg/foreman/agent/githubprfetch/fetch.go` drops the `output` object at unmarshal time. A coder on a CI-failure fix task can see *which* check failed but not *why*, and `details_url` is a dead end for it (task pods intentionally ship no curl/gh, and the URL is an HTML Actions page anyway).

The failure content is already in the response foreman makes: `GET /repos/{o}/{r}/commits/{sha}/check-runs` returns per-run `output.title`, `output.summary`, and `output.text`, which for common CI setups (Go test annotations, pytest summaries, linters, scanners) carry the actual error message and annotated lines. Dropping them forces the fix loop to depend on whatever failure text the re-dispatcher scraped into the task payload, with no way for the model to dig deeper when that summary is thin.

## Proposal

Extend `CheckRun` with the output fields, populated only when they add signal:

```go
type CheckRun struct {
	Name          string `json:"name"`
	Conclusion    string `json:"conclusion"`
	DetailsURL    string `json:"details_url"`
	OutputTitle   string `json:"output_title,omitempty"`
	OutputSummary string `json:"output_summary,omitempty"`
	OutputText    string `json:"output_text,omitempty"`
}
```

- Populate `output_*` for non-success conclusions only (`failure`, `timed_out`, `cancelled`, `action_required`); successful runs stay three fields so a green PR does not bloat the tool result.
- Truncate each field the same way the PR body is capped (rune-safe, per-field cap on the order of a few KiB; `output.text` can be tens of KiB on chatty runners).
- No new API calls, no new failure modes: the fields come from the response `fetchCheckRuns` already reads, and the existing warnings mechanism covers the endpoint erroring as before.

Job log retrieval (`actions/runs/.../logs`) is deliberately out of scope — that is extra calls and zip handling for marginal gain over `output.text`.

## Why it matters for the fix loop

CI-failure PR-fix tasks are dispatched with a scraped summary of the failure. When that summary is thin ("check failed"), the coder's only recourse today is to re-run the whole suite locally and hope it reproduces — expensive on runners that differ from the task pod. With `output_summary`/`output_text` the model can quote the exact assertion or lint line it is fixing, the same grounding pattern the tool's own description recommends for review comments.

Defilan:
## What

`fetch_pull_request` now returns each check run's `output.title`, `output.summary` and `output.text`, so a coder asked to fix a failing PR can read why CI failed rather than only that it did. Output is populated only for non-success conclusions, and each field is capped at 4 KiB.

## Why

A coder handed a red PR previously saw `{name, conclusion, details_url}`. The conclusion says "failure"; the reason lives in the check run's output, which was never fetched. The only recourse was `details_url`, which an agent cannot open.

Fixes #1481

## How

- `CheckRun` gains `OutputTitle` / `OutputSummary` / `OutputText`, all `omitempty`.
- `fetchCheckRuns` decodes the `output` object the API already returns on the same request, so this adds no extra API call.
- Gated on conclusion: `success`, `skipped` and `neutral` are treated as success and carry no output, so a green PR does not bloat the tool result.
- `DefaultOutputCap` is 4 KiB per field, with a `... [output truncated]` marker so a clipped field stays distinguishable from a genuinely short one.

**Known limitation, deliberately not fixed here.** `truncateOutput` slices bytes, so a cut can land mid-rune. That matches `truncateBody` directly above it, so this PR stays consistent with its neighbour rather than fixing one of two. The grep tool's `clipMatchText` already counts runes; making both truncators in this package rune-safe is worth a separate change.

**Not verified:** the live GitHub API, and a coder consuming the enriched output in a real loop. The decode is exercised against a fixture of the documented response shape.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (if user-facing change) — n/a, internal tool surface

`make lint`: 0 issues. `make test`: 34 packages ok, 0 failures.

The tests were confirmed to actually fail on a regression rather than merely pass: with the success-gating removed (`if !isSuccessConclusion(...)` forced true), `TestCheckRun_OutputFields` goes red with `lint output should be empty for success`. A test that cannot fail is not coverage.

Assisted-by: LLMKube Foreman (Qwopus-Fusion 27B coder on Strix Halo) generated the implementation and tests; reviewed by Nemotron-49B, then verified by hand before submission: read the diff, checked the tests assert the success-vs-failure gating rather than shape, ran `make test` and `make lint`, and confirmed the tests fail when the gating is broken.

### DIFF
--- a/pkg/foreman/agent/githubprfetch/fetch.go
+++ b/pkg/foreman/agent/githubprfetch/fetch.go
@@ -51,6 +51,12 @@ import (
 // with a marker so the model knows there is more.
 const DefaultBodyCap = 16 * 1024
 
+// DefaultOutputCap bounds each check-run output field (title, summary, text)
+// so a chatty CI runner does not bloat the tool result. 4 KiB per field is
+// enough for the actual error message and annotated lines from common CI
+// setups (Go test annotations, pytest summaries, linters, scanners).
+const DefaultOutputCap = 4 * 1024
+
 // DefaultTimeout caps a single fetch. The GitHub API is fast; if the
 // network is slow we'd rather skip the fetch than hold up the loop.
 const DefaultTimeout = 10 * time.Second
@@ -96,11 +102,15 @@ type ReviewComment struct {
 	Body         string `json:"body"`
 }
 
-// CheckRun is a single check run result.
+// CheckRun is a single check run result. Output fields are populated only
+// for non-success conclusions so a green PR does not bloat the tool result.
 type CheckRun struct {
-	Name       string `json:"name"`
-	Conclusion string `json:"conclusion"`
-	DetailsURL string `json:"details_url"`
+	Name          string `json:"name"`
+	Conclusion    string `json:"conclusion"`
+	DetailsURL    string `json:"details_url"`
+	OutputTitle   string `json:"output_title,omitempty"`
+	OutputSummary string `json:"output_summary,omitempty"`
+	OutputText    string `json:"output_text,omitempty"`
 }
 
 // Fetcher is the seam between the executor and the GitHub API. The
@@ -376,6 +386,11 @@ func (c *Client) fetchCheckRuns(
 			Name       string `json:"name"`
 			Conclusion string `json:"conclusion"`
 			DetailsURL string `json:"details_url"`
+			Output     struct {
+				Title   string `json:"title"`
+				Summary string `json:"summary"`
+				Text    string `json:"text"`
+			} `json:"output"`
 		} `json:"check_runs"`
 	}
 	if err := json.NewDecoder(resp.Body).Decode(&raw); err != nil {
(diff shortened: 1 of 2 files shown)


joryirving (approved):
Hey Chris - approving this one. The check-run endpoint and response shape are correct: it fetches against the PR head commit and decodes GitHub's nested output object. Gating output to non-success conclusions keeps successful PRs compact, and the per-field cap with a truncation marker is a sensible guard against noisy CI output.

The tests cover successful, failed, and timed-out runs, plus truncation. I also ran the focused githubprfetch and fetch_pr tests locally; they pass. The broader tools-package failures are unrelated test-harness issues around blocked direct pushes and missing FETCH_HEAD. Nothing blocking here.

Issue #744, opened 2026-06-18
[BUG] Reviewer issueAsk-verbatim rail demotes correct in-scope reviews to NO-GO even when scope-overlap confirms no drift


Defilan:
## Bug Description

The reviewer integrity rails produce a **false NO-GO on a correct, in-scope review** when the reviewer model paraphrases the issue ask instead of quoting it verbatim, *even when the computable scope-overlap rail confirms there is no drift*.

`enforceReviewerIssueAsk` (`pkg/foreman/agent/executor_native.go`) demotes a `GO` to `NO-GO` whenever `issueAskVerified == false` (the model's `issueAsk` is not a `strings.Contains` substring of the fetched issue body). `enforceReviewerScopeOverlap` (`pkg/foreman/agent/scope_overlap.go`, #647) runs **after** and, per its own comment, "annotates, not re-demotes" — so a correct, positive scope signal (`scopeDriftDetected=false` with a non-empty `scopeMatched`) **cannot rescue** an already-demoted verdict.

Net effect: an honest reviewer that accurately *paraphrases* the issue and reviews an in-scope diff is demoted to NO-GO with no recovery path. The verbatim-quote requirement was added to catch confabulation, but it also punishes accurate paraphrase. The maintainer comment at `executor_native.go` (the `reconcileReviewerIssueAsk` block) already notes verbatim-quoting drove a false NO-GO once before (#526).

## Steps to Reproduce

1. Run the Foreman coder on a real, in-scope issue and produce a branch (e.g. issue #710 -> branch `foreman/issue-710`, a focused `model_controller.go` fix + tests). Coder verdict: GO.
2. Run a `review` AgenticTask on that branch **against its own issue** (`payload.issue: 710`, `branch: foreman/issue-710`), with any local reviewer (reproduced with both Devstral and Gemma4-26B).
3. Observe `status.result.verdict`.

## Expected Behavior

Final verdict `GO` (the change is in-scope and correctly reviewed). The scope-overlap rail computes `scopeDriftDetected=false` with `scopeMatched=[internal/controller/model_controller.go]` — a positive, deterministic in-scope signal — which should prevent the verbatim-quote demotion.

## Actual Behavior

Final verdict `NO-GO`. Both reviewers' model `reviewOutcome` was `APPROVE`; both were demoted with:

```
demotionReason = issueAsk could not be verified as a verbatim quote of the fetched issue body; review verdict is untrusted
verdictClaimed = GO
issueAskVerified = false
[...]

PR #1516, opened 2026-08-12
fix: extract path refs the extension gate was discarding

### LINKED ISSUE #1515: Scope-overlap vouch never fires for issues citing files as path:line, bare build filenames, or .txt/.hcl
## Problem

`extractIssuePathRefs` discards any file reference that carries a line number, so an issue that cites its evidence as `main.go:135` — the conventional form, and the one our audit-generated issues use throughout — extracts **zero** path refs. With no refs, `enforceReviewerScopeOverlap` takes its documented observe-only path, `scopeMatched` stays empty, and `enforceReviewerIssueAsk` then demotes an otherwise-good `GO` to `NO-GO` because `scopeVouches` requires `len(scopeMatched) > 0`.

The net effect is that the scope-overlap rescue added in #744 cannot fire for issues written in `path:line` style, and completed work gets discarded.

## Root cause

`isPathRef` derives the extension with `path.Ext` after trimming only `.:!?` from the token ends:

```go
tok = strings.Trim(tok, ".:!?")                              // scope_overlap.go:91
ext := strings.ToLower(strings.TrimPrefix(path.Ext(tok), ".")) // scope_overlap.go:126
if !pathRefExtensions[ext] && !extraExts[ext] {
    return false
}
```

`path.Ext("main.go:135")` returns `.go:135`, not `.go`. The trailing digits are part of the final dot-suffix, so the trim never reaches them and the extension lookup misses.

## Reproduction

Running the upstream functions unmodified:

```
isPathRef, no extra extensions:
  main.go:135                  -> false
  main.go:49-63                -> false
  main.go:80                   -> false
  main.go                      -> true
  bridge/claim.py:149-156      -> false
  AGENTS.md                    -> true
  metrics.go                   -> true

extractIssuePathRefs(body, nil)          = []string(nil)
extractIssuePathRefs(body, []string{".go"}) = []string(nil)
```

where `body` is an issue citing `main.go:135`, `main.go:49-63` and `main.go:80`. Declaring `.go` in `GateProfile.SourceExtensions` does not help, because the miss is in the extension parse, not the extension set.

## Impact

On our fleet this is concentrated in issues whose evidence sections cite line numbers. In one repo, 11 of 16 open issues extract zero refs and therefore cannot produce a vouch.

A worked example: an issue asking for graceful shutdown cited `main.go` three times, always with a line number. The coder implemented it, added a regression test, and returned `GO`; the reviewer also returned `GO`. `enforceReviewerIssueAsk` demoted it to `NO-GO` and the branch was orphaned with no PR. The implementation builds clean and its tests pass — it was correct work, discarded because the citation style defeated the extractor.

The vouch itself is healthy: it fires normally (79 times in a 48h window here) for issues that cite bare filenames. This is specifically about the `path:line` form.

## Suggested fix

Strip a trailing `:<line>` / `:<start>-<end>` from the token before the extension lookup, and return the cleaned path so it still matches the diff by exact-path or basename equality. Happy to send a PR.

---

## Update: the same gate misses two more real reference forms

Widening this issue rather than filing separately, since all three are the
same defect — `isPathRef`'s extension gate rejecting a token that names a
real file — and a single fix covers them.

**Bare build filenames.** `Dockerfile`, `Makefile` and friends carry no
extension at all, so `path.Ext` returns `""` and the lookup can never admit
them. An issue whose subject *is* the Dockerfile cannot name its own subject.

**`.txt` and `.hcl` are absent from `pathRefExtensions`.** Both are ordinary
source-adjacent files that issues cite by name: `requirements.txt` for a
Python dependency set, `docker-bake.hcl` for a build definition. Neither can
be declared away through `GateProfile.SourceExtensions` for a repo on the
`generic` preset, which supplies no extensions.

Observed together on our fleet: after fixing every issue body we could, two
remained unfixable by any amount of editing — one citing `requirements.txt`,
one citing `Dockerfile` — because no wording gets past the extension gate.

Suggested shape: add `txt` and `hcl` to `pathRefExtensions`, and admit a
small allowlist of extensionless build entrypoints by basename. An allowlist
rather than a "capitalized token" heuristic, so ordinary prose cannot extract
as a file reference.

joryirving:
## What

Make `extractIssuePathRefs` recognize three forms of file reference it was silently discarding: line citations (`main.go:135`), bare build filenames (`Dockerfile`), and `.txt` / `.hcl`.

## Why

All three are the same defect — `isPathRef`'s extension gate rejecting a token that names a real file — and each has the same consequence. With no refs, `enforceReviewerScopeOverlap` takes its documented observe-only path and `scopeMatched` stays empty, so `enforceReviewerIssueAsk` computes `scopeVouches := !scopeDriftDetected && len(scopeMatched) > 0` as false. The #744 vouch cannot fire and a correct `GO` is demoted to `NO-GO`.

- **Line citations.** `path.Ext("main.go:135")` returns `.go:135`, and `strings.Trim(tok, ".:!?")` never reaches the digits.
- **Bare build filenames.** `Dockerfile`, `Makefile` and friends have no extension, so `path.Ext` returns `""`. An issue whose subject *is* the Dockerfile could not name its own subject.
- **`.txt` / `.hcl`.** Simply absent from `pathRefExtensions`, despite being cited routinely as `requirements.txt` and `docker-bake.hcl`. Neither can be supplied through `GateProfile.SourceExtensions` for a repo on the `generic` preset, which declares none.

Declaring extensions per-repo does not help any of these: the miss is in the extension parse, not the extension set.

Fixes #1515

## How

Two commits, one per defect class.

**Line citations** — `lineSuffixPattern` (`:\d+(?:-\d+)?$`) applied in `extractIssuePathRefs` after the existing punctuation trim, before the `seen` check. The stripped form is what gets returned, because refs are matched against the diff by exact path or basename and neither survives a `:135` suffix. Dedup runs on the stripped form, so repeated cites of one file at different lines collapse to a single ref.

**Bare filenames and extensions** — `bareSourceFilenames` is consulted by `path.Base` when the extension lookup misses; `txt` and `hcl` join `pathRefExtensions`.

The bare-filename set is an allowlist rather than a "capitalized token" heuristic on purpose. A general rule would admit ordinary prose as file references, and a wrong ref is worse than no ref here — it satisfies "refs exist, none matched," which demotes deterministically. Both changes narrow what gets rejected rather than widening what gets accepted.

## Checklist

- [x] Tests added/updated

Five regression tests, each run against unmodified upstream first to confirm it fails there:

- `TestExtractIssuePathRefs_LineCitations` — returned `[]` before.
- `TestExtractIssuePathRefs_LineCitationNonNumericUnaffected` — `foo.go:bar`, `12:30` stay rejected.
- `TestEnforceReviewerScopeOverlap_LineCitedRefVouches` — end-to-end: a diff touching the line-cited file keeps `GO` with non-empty `scopeMatched`.
- `TestExtractIssuePathRefs_BareBuildFilenames` — returned `[]` before.
- `TestExtractIssuePathRefs_ProseIsNotAPathRef` — guards the allowlist against prose.

`Assisted-by: Claude Code (diagnosed the bug and wrote the fix and tests; I reviewed the change and ran make fmt, make vet, make lint and make test locally before submitting)`

### DIFF
--- a/pkg/foreman/agent/scope_overlap.go
+++ b/pkg/foreman/agent/scope_overlap.go
@@ -19,12 +19,19 @@ package agent
 import (
 	"fmt"
 	"path"
+	"regexp"
 	"strings"
 
 	"github.com/go-logr/logr"
 
 	foremanv1alpha1 "github.com/defilantech/llmkube/api/foreman/v1alpha1"
 )
+
+// lineSuffixPattern matches the trailing line citation on a file reference
+// (`main.go:135`, `main.go:49-63`). path.Ext reads such a token's extension
+// as ".go:135", so the suffix has to come off before the extension lookup
+// or every line-cited reference is invisible to the extractor.
+var lineSuffixPattern = regexp.MustCompile(`:\d+(?:-\d+)?$`)
 
 // pathRefExtensions are the file extensions a token must carry to count
 // as a concrete path reference in an issue body. The set is deliberately
@@ -89,6 +96,12 @@ func extractIssuePathRefs(body string, sourceExtensions []string) []string {
 	seen := map[string]bool{}
 	for _, tok := range strings.Fields(normalized) {
 		tok = strings.Trim(tok, ".:!?")
+		// Strip the line citation before the extension lookup, and keep the
+		// bare path: refs are matched against the diff by exact path or
+		// basename, neither of which a ":135" suffix would survive. Dedup
+		// runs on the stripped form so repeated cites of the same file at
+		// different lines collapse to one ref.
+		tok = lineSuffixPattern.ReplaceAllString(tok, "")
 		if tok == "" || seen[tok] {
 			continue
 		}
--- a/pkg/foreman/agent/scope_overlap.go
+++ b/pkg/foreman/agent/scope_overlap.go
@@ -50,7 +50,17 @@ var lineSuffixPattern = regexp.MustCompile(`:\d+(?:-\d+)?$`)
 var pathRefExtensions = map[string]bool{
 	"go": true, "md": true, "yaml": true, "yml": true, "sh": true,
 	"json": true, "mod": true, "sum": true, "tmpl": true, "proto": true,
-	"toml": true, "mk": true,
+	"toml": true, "mk": true, "txt": true, "hcl": true,
+}
+
+// bareSourceFilenames are build files an issue cites by name because they
+// carry no extension, so path.Ext yields "" and the extension lookup can
+// never admit them. Kept to unambiguous, conventionally-capitalized build
+// entrypoints: a token like "Makefile" in an issue body is a file, not
+// prose, whereas a general "capitalized word" rule would admit sentences.
+var bareSourceFilenames = map[string]bool{
+	"Dockerfile": true, "Makefile": true, "Justfile": true,
+	"Containerfile": true, "Earthfile": true,
 }
 
 // extensionSet normalizes a GateProfile SourceExtensions list (".gd",
@@ -137,7 +147,7 @@ func hasSourceFile(paths []string, exts []string) bool {
 // language-agnostic pathRefExtensions base.
 func isPathRef(tok string, extraExts map[string]bool) bool {
 	ext := strings.ToLower(strings.TrimPrefix(path.Ext(tok), "."))
-	if !pathRefExtensions[ext] && !extraExts[ext] {
+	if !pathRefExtensions[ext] && !extraExts[ext] && !bareSourceFilenames[path.Base(tok)] {
 		return false
 	}
 	for _, seg := range strings.Split(tok, "/") {
--- a/pkg/foreman/agent/scope_overlap_test.go
+++ b/pkg/foreman/agent/scope_overlap_test.go
@@ -391,3 +391,45 @@ func TestEnforceReviewerScopeOverlap_PythonExtensions(t *testing.T) {
 		t.Error("should not claim demotion when no source files of configured type changed")
 	}
 }
+
+// Issue #1515: a reference carrying a line number (`main.go:135`) read as
+// extension ".go:135" and was discarded, so an issue whose evidence cites
+// lines — the conventional form — extracted no refs at all and could never
+// vouch. Verdicts on correct branches were demoted as a result.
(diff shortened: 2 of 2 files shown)


Defilan (approved):
Approving. Verified rather than taken on trust: I reverted only `scope_overlap.go` to upstream while keeping your tests, and four of the five fail there with exactly the `= []` symptom you describe. With the fix applied the full `pkg/foreman/agent` package passes, `GOOS=linux golangci-lint run` reports 0 issues, and gofmt is clean.

The diagnosis is the valuable part of this. `path.Ext("main.go:135")` returning `".go:135"` is the kind of thing that reads as correct until you print it, and tying it through to `scopeVouches` going false and demoting a correct GO is what makes it obviously worth fixing rather than a cosmetic parse nit.

I probed seven citation forms beyond your tests: plain, backticked, parenthesised, bracketed, trailing-comma, colon-then-prose, and range-only. All seven extract correctly. I expected `(main.go:138-157)` to slip through, since `strings.Trim(tok, ".:!?")` does not strip parens, but normalization handles it before the trim. Your fix is more robust than the tests demonstrate.

Agreed on the allowlist over a capitalized-token heuristic, and for the reason you give: a wrong ref satisfies "refs exist, none matched" and demotes deterministically, so a false positive costs more than a miss.

Two notes, neither blocking:

The description says the five tests were "each run against unmodified upstream first to confirm it fails there". `TestExtractIssuePathRefs_ProseIsNotAPathRef` passes both before and after, which is correct for a guard test but makes it a guard rather than a regression.

"Both changes narrow what gets rejected rather than widening what gets accepted" holds for the line-strip and the bare-filename allowlist, but adding `txt` and `hcl` does widen acceptance. That has a cost by your own argument: an issue citing a `.txt` the diff does not touch now produces "refs exist, none matched" where it previously produced no refs at all. It is the same tradeoff every extension in that set already carries and `requirements.txt` earns it, so I would keep it, but it is not free.

The failing `Run on Ubuntu (kind, default storage)` check is not yours. It died in `BeforeSuite` on `Failed to install CertManager`, a `kubectl apply` that could not reach GitHub. That is the third GitHub-network flake in this repo's CI today, alongside a syft download returning HTTP 000 and govulncheck getting a 403 from `vuln.go.dev`. Re-running should clear it.

PR #1518, opened 2026-08-12
fix(router-proxy): reject colliding listener addresses at startup

### LINKED ISSUE #1517: [BUG] router-proxy metrics listener: address collision is a crash-loop-or-silent-loss race, and --metrics-bind-address=0 does not disable
## Bug Description

Four loose ends around the router-proxy metrics listener, all verified against `main` at `12d59c2`. None of them affects the happy path that #1457 / #1473 / #1480 built — metrics serve correctly and the PodMonitor matches. They are about what happens when the address is wrong.

The first two are the substantive ones and share a cause: the metrics listener and the data-plane listener race for their ports, and the two losers are treated very differently.

### 1. `--listen` and `--metrics-bind-address` on the same address: usually a crash-loop, occasionally a silent metrics loss

Whichever listener binds first wins, and the outcome depends on which one loses:

- data plane loses → its error reaches `serverErr`, which is fatal → process exits 1 → crash-loop
- metrics loses → `logger.Error("metrics server failed", …)` and the proxy serves on happily with no metrics

That is a genuine race, not a fixed behaviour. 15 runs with both flags pointed at one address:

| outcome | count |
| --- | --- |
| data plane lost the port, process exited (crash-loop) | **14 / 15** |
| metrics lost the port, process kept serving without metrics | **1 / 15** |

```
{"level":"ERROR","msg":"server failed","error":"listen tcp :18135: bind: address already in use"}
```

The asymmetry is the defect: an operator-facing misconfiguration resolves either as a hard crash or as silent observability loss, and which one you get is scheduling. Since #1480 deliberately made metrics failures non-fatal, the data-plane path arguably wants the same treatment as a *port-conflict* case — or the two addresses want a startup check that rejects equality outright, before either listener binds.

### 2. `--metrics-bind-address=0` yields no metrics, and the last log line says the opposite

The guard is `if *metricsListen != ""` (`cmd/router-proxy/main.go:133`), and both the flag help (`:52`) and the comment (`:131`) document "Empty disables the metrics server." But `0` is the value this project's own manager uses for that purpose — `cmd/main.go:134` defaults `--metrics-bind-address` to `"0"` and documents "leave as 0 to disable the metrics service."

Carrying that value to the proxy produces:

```
{"msg":"metrics server listening"}                     <- printed before the bind is attempted
{"msg":"metrics server failed","error":"listen tcp: address 0: missing port in address"}
```

`/health` returns 200, the proxy serves normally, and there are no metrics. The log emits a positive "listening" line first, so the last reassuring signal an operator sees is untrue. Accepting `0` (and arguably `"none"`) in the same guard would make the two binaries agree:

```go
if addr := *metricsListen; addr != "" && addr != "0" {
```

### 3. `TestRouterDeploymentMetricsPort` cannot catch a wrong port constant

`internal/controller/router_builders_test.go:890` asserts the rendered port against `routerProxyMetricsPort` — the same constant the builder used — so the assertion holds for any value:

```go
if p.ContainerPort != routerProxyMetricsPort {
```

Mutation check: set `routerProxyMetricsPort = 8080`, colliding with `routerProxyPort`, and the test still passes (`ok internal/controller`). That mutation is exactly how you land in issue 1 above, which makes this the test that would have caught it. Pinning the literal and the distinctness closes it:

```go
if got := ports["metrics"]; got != 9090 {
    t.Errorf("metrics containerPort = %d, want 9090", got)
}
if ports["metrics"] == ports["http"] {
    t.Errorf("metrics port collides with the data-plane port (%d)", ports["http"])
}
```

Verified green as written and failing under the 8080 mutation with both messages.

### 4. `docs/grafana/README.md:36` still says the endpoint does not exist

> **Renders empty today:** `cmd/router-proxy` mounts no `/metrics` handler, so the `llmkube_router_*` collectors it registers are never scraped.

Both halves are now false: #1457 added the handler, #1480 pointed it at the right registry, and #1473 shipped the PodMonitor that scrapes it.

## Steps to Reproduce

```bash
# 1 — the race (repeat; the outcome varies)
router-proxy --config cfg.json --listen :18080 --metrics-bind-address :18080

# 2 — the manager's disable value
router-proxy --config cfg.json --listen :18081 --metrics-bind-address 0
curl -s localhost:18081/health      # 200, and no metrics anywhere

# 3 — the mutation
sed -i 's/routerProxyMetricsPort int32 = 9090/routerProxyMetricsPort int32 = 8080/' \
  internal/controller/router_common.go
go test ./internal/controller/ -run TestRouterDeploymentMetricsPort   # still passes
```

## Expected Behavior

Two listeners configured onto one address fail the same way every time, ideally with a startup error naming the conflict. `--metrics-bind-address=0` disables the listener, as it does on the manager. A wrong metrics-port constant fails a test. The dashboard catalog describes what ships.

## Actual Behavior

The collision is a coin flip between crash-loop and silent metrics loss. `0` produces a failed listener behind a log line claiming success. The port constant is unpinned. The doc describes the pre-#1457 world.

## Environment

**LLMKube Version:** `main` at `12d59c2`; all four checks run there. Items 1 and 2 from a locally built `cmd/router-proxy`, item 3 from `go test ./internal/controller/`, item 4 from the file.

Happy to send a PR for any subset — items 2, 3 and 4 are small and independent. Item 1 is a design call (reject equal addresses at startup, or make a port conflict non-fatal on both sides) and probably wants your preference first.

*Assisted-by: Claude Code (ran the reproductions and the mutation check quoted above; findings reviewed before filing).*

Defilan:
## What

Rejects `--listen` and `--metrics-bind-address` resolving to the same address, as a startup error raised before either listener binds. Also makes the metrics listener bind synchronously so its success line is only logged once the port is actually held.

Rebased onto `main` now that #1519 has landed. This PR is narrowed to **item 1 of #1517** only. Items 2, 3 and 4 (the `0` guard, the pinned port test, the dashboard catalog) all merged in #1519, and my earlier versions of those have been dropped rather than merged, so there is no duplicate coverage. Diff is 5 files down to 3.

## Why

Fixes #1517

Pointing both listeners at one address is a race, not a fixed failure. Whichever binds first wins, and the loser is treated very differently depending on which one it is:

- data plane loses, its error reaches `serverErr`, which is fatal, so the process crash-loops
- metrics loses, it is logged and the proxy serves on happily with no metrics

Over 15 runs with both flags on one address that came out **14 crash-loops to 1 silent metrics loss**. The same configuration produced two different outcomes, so an operator-facing misconfiguration resolved either as a hard crash or as silent observability loss, decided by scheduling.

## How

`listenConflict` compares both addresses before anything binds and exits 1 with a message that names the conflict.

Host comparison is deliberately conservative. Equal ports conflict when the hosts are equal or when either side is a wildcard, because a wildcard bind covers every specific address on that port. Two different specific hosts on the same port do **not** conflict and are left alone: that is legitimate on a multi-homed node, and refusing it would be worse than the bug being fixed.

Two smaller changes ride along because they are the same code path:

- **Synchronous bind.** The bind is the step that fails, so doing it outside the goroutine means `metrics server listening` is logged only once the port is held, and the log can name the *resolved* address. `--metrics-bind-address=:0` now reports the port it actually got instead of `:0`. A bind failure stays non-fatal, because killing the data plane over a scrape endpoint trades a monitoring gap for an outage.
- **`metricsDisabled`** replaces the inline `addr != "" && addr != "0"` guard from #1519 with a single helper, so the startup check and the serve path cannot drift. It additionally accepts `none` and trims whitespace, so a Helm value rendering as `"0 "` disables the listener instead of taking the bind-and-fail path.

### Verification

Behaviour, on a local build rather than by inspection:

| invocation | result |
| --- | --- |
| `--listen 127.0.0.1:18914 --metrics-bind-address 127.0.0.1:18914` | exit code **1**, one error naming the conflict, before either bind |
| `--metrics-bind-address 0` | `metrics server disabled` |
| `--metrics-bind-address ' 0 '` | `metrics server disabled` |
| `--metrics-bind-address 127.0.0.1:0` | `metrics server listening address=127.0.0.1:53745`, the resolved port |

Gate on this branch against current `main`:

- `go build ./...` clean
- `go vet ./cmd/...` clean
- `go test ./cmd/...` pass, including `cmd/router-proxy`
- `go test ./internal/controller/ -run TestRouterDeploymentMetricsPort` pass, so #1519's pinned-port test is unaffected
- `make lint` 0 issues, and `GOOS=linux golangci-lint run ./cmd/...` 0 issues
- `gofmt -l cmd/ internal/` clean

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (ran `go test ./cmd/...` plus the router builder test; no controller code is touched by this PR)
- [x] Documentation updated (if user-facing change) - the dashboard catalog line landed in #1519; the flag help text is updated here

*Assisted-by: Claude Code (wrote the commit, ran the gate and the behaviour table above; I reviewed the diff before pushing.)*

### DIFF
--- a/cmd/router-proxy/listenaddr.go
+++ b/cmd/router-proxy/listenaddr.go
@@ -1,0 +1,119 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"net"
+	"strings"
+)
+
+// metricsDisabled reports whether a --metrics-bind-address value means "do not
+// serve metrics".
+//
+// "" is this binary's documented sentinel, but "0" is what the manager uses
+// (cmd/main.go defaults --metrics-bind-address to "0" and documents "leave as 0
+// to disable"). Carrying the manager's value here used to produce a listener
+// that failed with `address 0: missing port in address` AFTER a log line
+// claiming it was listening, so the operator's last signal said the opposite of
+// what happened. Accepting both makes the two binaries agree.
+func metricsDisabled(addr string) bool {
+	switch strings.ToLower(strings.TrimSpace(addr)) {
+	case "", "0", "none":
+		return true
+	default:
+		return false
+	}
+}
+
+// listenConflict returns a non-nil error when two listen addresses would race
+// for the same port.
+//
+// Configuring both listeners onto one address is not a stable failure: whichever
+// binds first wins, and the loser's treatment differs by which one it is. The
+// data plane's error is fatal, so the process crash-loops; the metrics server's
+// is logged and execution continues without metrics. Measured over 15 runs that
+// came out 14 crash-loops to 1 silent metrics loss, so the same configuration
+// produced two different outcomes.
+//
+// Detecting it before either listener binds turns a coin flip into one
+// deterministic startup error that names the conflict.
+//
+// Host comparison is deliberately conservative. Equal ports conflict when the
+// hosts are equal or when either side is a wildcard, because a wildcard bind
+// covers every specific address on that port. Two different specific hosts on
+// the same port do not conflict and are left alone: that is a legitimate
+// configuration on a multi-homed node, and refusing it would be worse than the
+// bug being fixed.
+func listenConflict(dataPlane, metrics string) error {
+	dHost, dPort, err := splitListen(dataPlane)
+	if err != nil {
+		// A malformed data-plane address is the data plane's problem to report
+		// when it binds; this check declines to speak for it.
+		return nil
+	}
+	mHost, mPort, err := splitListen(metrics)
+	if err != nil {
+		return nil
+	}
+
+	if dPort != mPort {
+		return nil
+	}
+	if !hostsOverlap(dHost, mHost) {
+		return nil
+	}
+
+	return fmt.Errorf(
+		"--listen (%s) and --metrics-bind-address (%s) resolve to the same address; "+
+			"the two listeners would race for port %s and the loser's failure differs by "+
+			"which one loses (the data plane exits, the metrics server does not). "+
+			"Give them different ports, or disable metrics with --metrics-bind-address=0",
+		dataPlane, metrics, dPort)
+}
+
+// splitListen accepts the forms a Go listener accepts, including the bare
+// ":9090" that omits the host.
+func splitListen(addr string) (host, port string, err error) {
+	host, port, err = net.SplitHostPort(strings.TrimSpace(addr))
+	if err != nil {
+		return "", "", err
+	}
+	if port == "" {
+		return "", "", fmt.Errorf("no port in %q", addr)
+	}
+	return host, port, nil
+}
+
+// hostsOverlap reports whether two listener hosts can contend for one port.
+// The empty host, 0.0.0.0 and [::] all mean "every interface" and therefore
+// overlap with anything.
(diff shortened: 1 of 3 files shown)


brandonli-ybor (approved):

brandonliuli (approved):
lgtm

PR #1473, opened 2026-08-09
feat(chart): PodMonitor for router-proxy metrics

Tanguille:
## What

A `PodMonitor` for router-proxy pods, scraping the metrics listener #1457 adds.

## Why

Fixes #1472

#1457 adds a `--metrics-bind-address` listener (container port `metrics`,
default `:9090`) but nothing scrapes it: `inference-podmonitor.yaml` selects
`inference.llmkube.dev/service` on a port named `http`, and router-proxy pods
carry neither that label nor that port name. Without this, `model-router-dashboard.json`
still can't populate even once #1457 merges — its own PR body says so and
deliberately left #1427 open for that reason.

## How

`router-podmonitor.yaml` mirrors `inference-podmonitor.yaml`'s shape:

- Selector: `inference.llmkube.dev/model-router: Exists`, the stable label
  `routerProxySelectorLabels` (`internal/controller/router_common.go:68`)
  already puts on every router-proxy pod.
- `port: metrics`, `path: /metrics`.
- `job: llmkube-router` relabeling, so a future `RouterProxyDown` rule has a
  stable label to match — same pattern the inference/controller monitors use.
- Gated on `prometheus.routerPodMonitor.enabled` (default `false`, matching
  `inferencePodMonitor` and `serviceMonitor`'s existing default).

This PR is chart-only and depends on #1457 for the port to exist; opening as
draft until #1457 merges (port name/number final).

Assisted-by: OpenCode (helped draft this PR; I reviewed the final change and stand behind it.)

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally
- [x] `make lint` passes locally
- [x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)
- [x] All commits are signed off (`git commit -s`) per [DCO](https://developercertificate.org/)
- [x] AI assistance (if any) is disclosed above, per [CONTRIBUTING.md](../CONTRIBUTING.md#ai-assisted-and-agent-contributions)
- [ ] Documentation updated (if user-facing change)

PR #1519, opened 2026-08-13
fix(router-proxy): honour 0 as disable, pin the metrics port in tests, refresh the dashboard catalog

brandonliuli:
Addresses items 2, 3 and 4 of #1517. Three independent commits, any of which can be dropped.

**Item 1 is deliberately not here.** The address-collision race is the one that needs your preference — reject equal addresses at startup, or make a port conflict non-fatal on both sides — and I said in the issue it should wait for that. One observation while it is on your desk: with startup validation on equal addresses, the mutation the test in commit 2 now catches would surface as a clear startup error rather than as a race, so those two fixes reinforce each other.

## `0` now disables the metrics server, and the log stops claiming a listen that has not happened

The guard was `*metricsListen != ""`, so empty disabled the server but `0` did not — and `0` is what this project's own manager uses for that purpose (`cmd/main.go` defaults `--metrics-bind-address` to `"0"` and documents "leave as 0 to disable the metrics service"). Carrying the manager's value across produced a listener that failed to bind, no metrics, and a proxy that otherwise looked fine. The guard accepts both now, and the flag help and comment say so.

Separately, `logger.Info("metrics server listening", …)` fired *before* `ListenAndServe`, so it printed on the failing path too — the last reassuring line an operator saw was untrue. It now says `starting metrics server`, which is what is actually known at that point.

## The port test pins a literal instead of comparing against itself

`TestRouterDeploymentMetricsPort` asserted the rendered port against `routerProxyMetricsPort`, the same constant the builder had just used, so it held for any value. Setting the constant to `8080` — colliding with `routerProxyPort` — left the test green while the binary fails at startup with `bind: address already in use`.

Now it pins `9090` and adds the distinctness check, which is the assertion that actually matters: the two listeners cannot share an address. Strengthened in place rather than added alongside, so there is no duplicate coverage.

## The dashboard catalog no longer says the endpoint is missing

`docs/grafana/README.md` still described the router dashboard as empty because "`cmd/router-proxy` mounts no `/metrics` handler". #1457 added the listener, #1480 pointed it at the right registry, and #1473 shipped the PodMonitor.

## Verification

Gate, on this branch against current `main`:

- `helm unittest charts/llmkube` — 73/73
- `go test ./internal/controller/` with envtest — pass (265 s)
- `go test ./cmd/...` — pass, including `cmd/router-proxy`
- `make lint` — 0 issues
- `gofmt -l cmd/ internal/` — clean

Behaviour, on a local build rather than by inspection:

| invocation | result |
| --- | --- |
| `--metrics-bind-address 0` | data plane 200, no metrics listener, **no error line** (previously: failed listener behind a success log) |
| `--metrics-bind-address ''` | unchanged, still disables |
| `--metrics-bind-address :19152` | `/metrics` 200 with 18 `llmkube_router_*` series |

The test change is mutation-verified in both directions: green as written, and under `routerProxyMetricsPort = 8080` it fails with both `metrics containerPort = 8080, want 9090` and `metrics port collides with the data-plane port (8080)`.

*Assisted-by: Claude Code (wrote the three commits and ran the gate and the behaviour checks quoted above; I reviewed the diff and confirmed the mutation result before opening this.)*


Defilan (approved):
Approving. I checked the claims by running them rather than by reading the diff, and they hold.

## The mutation claim, both directions

I built a worktree at the PR head and exercised `TestRouterDeploymentMetricsPort`:

- as written: passes
- with `routerProxyMetricsPort = 8080`: fails with both messages, exactly as the description says

```
router_builders_test.go:898: metrics containerPort = 8080, want 9090
router_builders_test.go:901: metrics port collides with the data-plane port (8080); both listeners would bind the same address
```

- the same mutation against the pre-PR version of the test file: **passes**

That last run is the one that matters, and it confirms the premise. The old assertion compared the rendered port against the constant the builder had just used, so it was satisfied by any value including a colliding one. Strengthening it in place rather than adding a second test was the right call.

I also confirmed the new collision assertion is actually live: the builder names the ports `http` and `metrics` (`router_deployment_builder.go:175-181`), so `ports["http"]` resolves rather than silently returning the zero value and comparing against nothing.

## Runtime behaviour

On a local build:

| invocation | result |
| --- | --- |
| `--metrics-bind-address 0` | data plane 200, no metrics listener, and no metrics log line at all |
| `--metrics-bind-address 127.0.0.1:19155` (live backend) | `/metrics` 200 with exactly 18 `llmkube_router_*` series |

The 18 figure reproduces precisely. Worth noting for anyone repeating it: the series only appear once a request actually completes against a reachable backend. My first attempts pointed at a dead address, got a 502, and showed zero series, which is a property of labelled collectors rather than anything wrong here.

## The `0` parity argument checks out

`cmd/main.go:134-135` does default `--metrics-bind-address` to `"0"` and document "leave as 0 to disable the metrics service", so carrying that value into the proxy really did produce a failed listener behind a success log. Making the two binaries agree is right.

## Docs

Verified stale rather than taken on trust: `cmd/router-proxy/main.go:135` does mount the handler, `charts/llmkube/templates/router-podmonitor.yaml` exists, and all seven `Router*` collectors are in `AllCollectors` and registered from `internal/metrics`' `init()`, with that package in the router-proxy dependency graph. The old "renders empty by construction" line was wrong.

## Two non-blocking notes

1. `addr != "0"` does not trim, so a value that renders as `"0 "` with trailing whitespace still takes the bind-and-fail path. Trivial, and fine to leave.
2. The description does not use the repo's PR template headings (`## What / ## Why / ## How / ## Checklist`). Worth following next time; the content itself is more thorough than most.

## Overlap with #1518

Heads-up that #1518 currently covers items 2, 3 and 4 as well, so these two conflict on all three files. Since this PR is verified and self-contained, the plan is to land it and narrow #1518 down to item 1 only, the startup check for colliding addresses. Your read in the description is right that the two reinforce each other: with a startup rejection on equal addresses, the mutation this test now catches surfaces as a clear error instead of a race.

Thanks for splitting these into three droppable commits and for leaving item 1 for a decision rather than guessing at it.

PR #1527, opened 2026-08-13
fix: namespace MCP tools with __ so names match the OAI pattern

### LINKED ISSUE #1525: MCP tool names use '/' and violate the OpenAI function-name pattern, 400ing every request on validating providers
## Problem

`mcpTool.Name()` namespaces MCP tools as `mcp/<server>/<tool>`, but the OpenAI tool-call spec requires `function.name` to match `^[a-zA-Z0-9_-]+$`. The `/` separator is not in that set, so every request carrying an MCP tool is malformed.

Providers that validate reject the request at turn 1, before the model runs:

```
status 400: Invalid 'tools[3].function.name': string does not match pattern.
Expected a string that matches the pattern '^[a-zA-Z0-9_-]+$'.
```

Providers that do not validate accept it, which is why this stays invisible on a local-only fleet. `llama.cpp` and LM Studio pass the same payload through untouched; a strict hosted endpoint does not.

## Source

`pkg/foreman/agent/mcp/tool.go`:

```go
// tools.Tool, namespaced as "mcp/<server.Name>/<tool name>" so tools
func (t *mcpTool) Name() string {
	return "mcp/" + t.server + "/" + t.toolName
}
```

The upstream names are already valid on their own — context7 advertises `resolve-library-id` and `query-docs`. The namespacing introduces the violation.

## Impact

On our fleet this took an entire coder agent out of service without any signal that the model was never invoked. **14 of 22 errored coder pods** failed this way, each recording as a task error rather than a provider rejection, and the agent's 0-for-5 verdict record read as poor model quality for most of a day before we traced it.

Any Agent with `mcp.enabled: true` is unusable against a validating provider. The failure is total (turn 1, every request) and silent as to cause.

## Suggested fix

Use a separator inside the allowed set — `mcp__<server>__<tool>` keeps the same three-part structure and round-trips to `t.toolName` for the actual call, which already happens separately at `callTool(cctx, t.toolName, args)`.

Sanitising the server and tool segments would also guard against an MCP server advertising a name with a dot or slash in it.

Happy to send a PR.

joryirving:
## What

Namespace MCP tools as `mcp__<server>__<tool>` instead of `mcp/<server>/<tool>`, and sanitise each segment.

## Why

The OAI tool-call spec constrains `function.name` to `^[a-zA-Z0-9_-]+$`. `/` is not in that set, so every request carrying an MCP tool is malformed. Providers that validate reject it at turn 1, before the model runs:

```
status 400: Invalid 'tools[3].function.name': string does not match pattern.
Expected a string that matches the pattern '^[a-zA-Z0-9_-]+$'.
```

Providers that don't validate accept it unchanged, which is why this is invisible on a local-only fleet — `llama.cpp` and LM Studio pass the same payload through. It only surfaces when an Agent points at a strict hosted endpoint.

On our fleet it took a coder agent out of service completely: **14 of 22 errored coder pods** failed this way. Because the rejection records as a task error rather than a provider refusal, the agent's verdict record read as poor model quality — we spent most of a day attributing it to the model before tracing the 400.

The upstream names are already valid on their own; context7 advertises `resolve-library-id` and `query-docs`. The namespacing introduced the violation.

Fixes #1525

## How

`nameSeparator` is `"__"`, and `sanitizeNameSegment` maps anything outside the allowed set to `_` — so a server or tool advertising a dot, slash or space also can't produce a rejected name. The remote tool is still invoked by its bare name in `Execute` (`callTool(cctx, t.toolName, args)`), which is untouched.

The `mcp_types.go` doc comments carried the old form, so `make manifests`, `make chart-crds` and `make foreman-chart-crds` were re-run and the regenerated CRDs are included.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (if user-facing change) — CRD descriptions regenerated

Two regression tests, both run against unmodified upstream first to confirm they fail there:

- `TestToolNameMatchesOAIPattern` — asserts `Name()` matches `^[a-zA-Z0-9_-]+$` across four server/tool shapes. Before the fix: `Name() = "mcp/a b/c.d" ... does not match`.
- `TestToolNameSanitisesSegments` — `weird.server` + `tool/name` → `mcp__weird_server__tool_name`. Before: `"mcp/weird.server/tool/name"`.

The four existing tests that asserted the old separator were updated to the new one.

`Assisted-by: Claude Code (diagnosed the bug from a live 400, wrote the fix and tests; I reviewed the change and ran make fmt, make vet, make lint and make test locally before submitting)`

### DIFF
--- a/api/foreman/v1alpha1/mcp_types.go
+++ b/api/foreman/v1alpha1/mcp_types.go
@@ -43,10 +43,10 @@ type MCPConfig struct {
 }
 
 // MCPServer configures one MCP server the agent connects to. Its tools
-// are namespaced as mcp/<name>/<tool> so identically-named tools across
+// are namespaced as mcp__<name>__<tool> so identically-named tools across
 // servers do not collide in the model-facing tool whitelist.
 type MCPServer struct {
-	// Name namespaces this server's tools as mcp/<name>/<tool>.
+	// Name namespaces this server's tools as mcp__<name>__<tool>.
 	// +kubebuilder:validation:Required
 	Name string `json:"name"`
 
--- a/charts/foreman/templates/crds/agents.yaml
+++ b/charts/foreman/templates/crds/agents.yaml
@@ -309,7 +309,7 @@ spec:
                     items:
                       description: |-
                         MCPServer configures one MCP server the agent connects to. Its tools
-                        are namespaced as mcp/<name>/<tool> so identically-named tools across
+                        are namespaced as mcp__<name>__<tool> so identically-named tools across
                         servers do not collide in the model-facing tool whitelist.
                       properties:
                         allowedTools:
@@ -361,7 +361,7 @@ spec:
                             type: object
                           type: array
                         name:
-                          description: Name namespaces this server's tools as mcp/<name>/<tool>.
+                          description: Name namespaces this server's tools as mcp__<name>__<tool>.
                           type: string
                         transport:
                           description: Transport is the MCP transport. Phase 1 supports
--- a/config/crd/bases/foreman.llmkube.dev_agents.yaml
+++ b/config/crd/bases/foreman.llmkube.dev_agents.yaml
@@ -308,7 +308,7 @@ spec:
                     items:
                       description: |-
                         MCPServer configures one MCP server the agent connects to. Its tools
-                        are namespaced as mcp/<name>/<tool> so identically-named tools across
+                        are namespaced as mcp__<name>__<tool> so identically-named tools across
                         servers do not collide in the model-facing tool whitelist.
                       properties:
                         allowedTools:
@@ -360,7 +360,7 @@ spec:
                             type: object
                           type: array
                         name:
-                          description: Name namespaces this server's tools as mcp/<name>/<tool>.
+                          description: Name namespaces this server's tools as mcp__<name>__<tool>.
                           type: string
                         transport:
                           description: Transport is the MCP transport. Phase 1 supports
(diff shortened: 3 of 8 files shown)


Defilan (approved):
Approving. Checked the claims by running them rather than reading the diff.

## Mutation-verified

Reverted `tool.go` to `upstream/main` while keeping the new tests. `TestToolNameMatchesOAIPattern` fails on all four shapes with exactly the error you describe:

```
tool_test.go:439: Name() = "mcp/fake/echo" for server="fake" tool="echo"; does not match ^[a-zA-Z0-9_-]+$
tool_test.go:439: Name() = "mcp/context7/resolve-library-id" ... does not match
tool_test.go:439: Name() = "mcp/weird.server/tool/with slash" ... does not match
tool_test.go:439: Name() = "mcp/a b/c.d" ... does not match
```

So the regression test is real rather than vacuous, and the sanitiser is doing work beyond the separator swap. Package tests pass with the fix in place.

## The compatibility risk I went looking for is not there

My first concern was that changing the separator would invalidate every existing `allowedTools` list. It does not: `session.go:77` matches on `d.Name`, the bare discovered name, not the namespaced one. Confirmed against the live fleet, where every Agent CR uses bare upstream names:

```
coder-laguna   server=context7    allowedTools=['resolve-library-id', 'query-docs']
coder-glm      server=perplexity  allowedTools=['perplexity_ask', 'perplexity_search']
```

So no config migration is needed, and `Execute` still calling `callTool(cctx, t.toolName, args)` by bare name means the wire call is unchanged.

CRD regeneration is consistent across `config/crd/bases` and `charts/foreman/templates/crds`, so the manifests/chart-crds/foreman-chart-crds claim holds.

## One corroboration

Your "invisible on a local-only fleet" point matched something from today independently: a DeepSeek run on the Sparks loaded MCP fine (`serversConfigured 2, toolsAdded 4`) against llama.cpp, which passes the malformed name straight through. That is exactly the asymmetry you describe, and it is a good argument for the sanitiser rather than just the separator, since the next strict provider will not be so forgiving about a server name with a dot in it.

Good catch on the diagnosis too. Fourteen of twenty-two errored coder pods reading as model quality is the expensive kind of wrong.

PR #1529, opened 2026-08-13
fix: stop discarding a coder's diff when submit_result omits a message

### LINKED ISSUE #1526: A submit_result without a commit message discards the coder's completed diff
## Problem

`repo.Commit()` rejects an empty commit message, and the caller has no fallback — so a coder that edited files and reached `submit_result` without passing a commit message has its entire diff discarded and the task recorded as `NO-GO`.

```
FOREMAN-RESULT: {"verdict":"NO-GO","summary":"commit rejected",
  "branch":"foreman/wl-misospace-windowstead-333/issue-333",
  "extra":{"error":"Commit: Message is required"}}
```

Note `branch` is set and `commitSHA` is empty: the work existed, the branch was resolved, and nothing was committed.

## Source

`pkg/foreman/agent/repo/commit.go`:

```go
if strings.TrimSpace(opts.Message) == "" {
	return "", fmt.Errorf("Commit: Message is required")
}
```

## Impact

Four occurrences in a 40-hour window here, each after 1–2 hours of coder runtime:

```
2026-08-12T01:05  foreman/wl-misospace-miso-gallery-387/issue-387
2026-08-12T11:06  foreman/wl-misospace-miso-gallery-402/issue-402
2026-08-12T14:41  foreman/wl-misospace-foreman-dispatch-bridge-97/issue-97
2026-08-13T17:28  foreman/wl-misospace-windowstead-333/issue-333
```

The verdict is indistinguishable from a model declining the work, so it reads as a quality problem rather than a harness one. The retry then starts from scratch with no branch evidence, which is the cycle described in the branch-evidence issue.

## Suggested fix

Either of:

1. Mark `commitMessage` required in the `submit_result` tool schema, so the model cannot omit it and the provider enforces it.
2. Synthesise a fallback in the caller — the task carries repo, issue number and the model's summary, all of which make a serviceable message — and log that the model omitted one.

Option 2 alone is enough to stop discarding completed work; option 1 addresses the cause. Happy to send a PR for either.

joryirving:
## What

- `commitAndPush` synthesizes a commit message when `submit_result` carries none.
- `commit_message` becomes required in the `submit_result` schema.

## Why

`repo.Commit` rejects an empty message and `commitAndPush` had no fallback, so a coder that edited files and reached `submit_result` without a commit message had its entire diff discarded:

```json
{"verdict":"NO-GO","summary":"commit rejected",
 "branch":"foreman/wl-misospace-windowstead-333/issue-333",
 "extra":{"error":"Commit: Message is required"}}
```

`branch` is set and `commitSHA` is empty — the work existed, the branch was resolved, and nothing was committed. Because the verdict is a plain `NO-GO`, it is indistinguishable from the model declining the work, so it reads as a model-quality problem rather than a harness one.

Four occurrences in a 40-hour window on our fleet, each after 1–2 hours of coder runtime:

```
2026-08-12T01:05  miso-gallery#387
2026-08-12T11:06  miso-gallery#402
2026-08-12T14:41  foreman-dispatch-bridge#97
2026-08-13T17:28  windowstead#333
```

Fixes #1526

## How

**The fallback** follows the shape already used by the gate-failed preserve path a few hundred lines below, which synthesizes a WIP subject rather than losing the branch. It uses the model's own `Summary` as the body and `Refs #N` rather than `Fixes #N` — a message we invented should not auto-close an issue on merge.

**The schema change** is the actual cause. `commit_message` was declared but absent from `required`, so a model could legally omit it.

Both, deliberately: the schema stops it happening, the fallback stops it costing an hour when a model omits it anyway. Either alone leaves a gap — schema-only turns a recoverable case into a rejected tool call, fallback-only silently accepts sloppy submissions.

## Checklist

- [x] Tests added/updated

Two regression tests, both run against unmodified upstream first. `TestNativeExecutor_SynthesizesCommitMessageWhenModelOmitsIt` drives the real executor through `initBareWithSeed` + `fakeRegistry` and reproduces the production failure exactly before the fix:

```
verdict: want GO got NO-GO
Summary:commit rejected  Extra:[error:Commit: Message is required outcome:COMMIT-REJECTED ...]
```

It then asserts `commitSHA` is present and reads the message back off the bare remote to confirm it references the issue and does **not** contain `Fixes #`. `TestSubmitResultRequiresCommitMessage` covers the schema; before the change it reports `schema required is missing "commit_message"; got [verdict summary]`.

I first wrote the executor test against a copy of the synthesis logic, which would have passed with the production code broken, and replaced it with the harness above.

`Assisted-by: Claude Code (diagnosed the bug from four production occurrences, wrote the fix and tests; I reviewed the change and ran make fmt, make vet, make lint and make test locally before submitting)`

### DIFF
--- a/pkg/foreman/agent/executor_native.go
+++ b/pkg/foreman/agent/executor_native.go
@@ -1095,9 +1095,26 @@ func (e *NativeAgentLoopExecutor) commitPushAttempt(
 
 	envtestTouched = len(changedEnvtestPackages(ctx, workspace, execCommandRunner)) > 0
 
+	// A submit_result that carries edits but no commit message used to fail
+	// repo.Commit ("Commit: Message is required"), which discarded the whole
+	// diff and recorded NO-GO. The work is real and the branch is resolved by
+	// this point, so synthesize a subject the same way the gate-failed path
+	// below does rather than throwing it away. Refs (not Fixes) because a
+	// message we invented should not auto-close the issue on merge.
+	commitMessage := lr.Terminal.CommitMessage
+	if strings.TrimSpace(commitMessage) == "" {
+		commitMessage = fmt.Sprintf(
+			"fix: resolve issue #%d\n\n%s\n\nRefs #%d",
+			task.Spec.Payload.Issue,
+			strings.TrimSpace(lr.Terminal.Summary),
+			task.Spec.Payload.Issue)
+		log.Info("submit_result carried no commit message; synthesized one",
+			"issue", task.Spec.Payload.Issue)
+	}
+
 	sha, commitErr := repo.Commit(ctx, repo.CommitOptions{
 		Workspace: workspace,
-		Message:   lr.Terminal.CommitMessage,
+		Message:   commitMessage,
 		Author:    e.CommitAuthor,
 		Committer: e.CommitCommitter,
 	})
--- a/pkg/foreman/agent/executor_native_test.go
+++ b/pkg/foreman/agent/executor_native_test.go
@@ -2135,3 +2135,83 @@ func TestNativeExecutor_ModelNoGoAlreadyResolved_PromotesOutcome(t *testing.T) {
 		t.Errorf("modelExtra should keep the nested outcome for observability, got %v", res.Extra["modelExtra"])
 	}
 }
+
+// Issue #1526: a submit_result carrying edits but an empty CommitMessage used
+// to fail repo.Commit ("Commit: Message is required"), discarding the diff and
+// recording NO-GO "commit rejected" — indistinguishable from the model
+// declining the work. Four occurrences in a 40h window on our fleet, each
+// after 1-2 hours of coder runtime. The executor now synthesizes a message.
+func TestNativeExecutor_SynthesizesCommitMessageWhenModelOmitsIt(t *testing.T) {
+	gitOrSkip(t)
+	root := t.TempDir()
+	bare := initBareWithSeed(t, root)
+	oaiSrv := scriptedOAI(t, []string{submitGoBody})
+
+	agent, task := taskAndAgent("nomsg")
+	c := fake.NewClientBuilder().
+		WithScheme(newScheme(t)).
+		WithObjects(agent, task).
+		Build()
+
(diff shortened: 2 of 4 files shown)


Defilan (approved):
Approving. Doing both the schema change and the fallback is the right call, and your reasoning for it is the part I would have argued for anyway: schema-only turns a recoverable case into a rejected tool call, fallback-only silently accepts sloppy submissions.

Tests pass. The failure mode is well evidenced, and `branch` set with `commitSHA` empty is a good diagnostic for it.

## One real bug, filed rather than blocked: #1530

The synthesized message formats the issue number unconditionally:

```go
"fix: resolve issue #%d\n\n%s\n\nRefs #%d",
task.Spec.Payload.Issue, ..., task.Spec.Payload.Issue)
```

`Payload.Issue` is `int32` with `json:"issue,omitempty"`, and there are live tasks that have a repo but no issue number:

```
('freeform',  hasRepo=True, hasIssue=False) -> 4
('integrate', hasRepo=True, hasIssue=False) -> 3
('reconcile', hasRepo=True, hasIssue=False) -> 3
```

Those reach `commitPushAttempt`, because the `freeformGoResult` early return is gated on `needsRepo` and all of them have a repo. So one of those omitting a commit message produces `fix: resolve issue #0` with a `Refs #0` trailer pointing at nothing.

Not blocking, because it is cosmetic next to the bug this fixes, and a confusing commit subject beats a discarded diff by a wide margin. Filed as #1530 with the reproduction and the live task counts. Worth noting the existing regression test uses a task with an issue set, so it cannot catch this shape and a follow-up wants its own case.

## One design nit, not a blocker

`commit_message` is now `required` for every role, and `buildall.go` registers `submit_result` for all of them, so reviewers get it too (`nemotron-direct-reviewer` has `submit_result` in its tool list). A reviewer has nothing to commit, so it will now be told to produce a message for a commit that never happens.

Nothing breaks at the harness level, since `Execute` only enforces `verdict` and `summary`. It just adds noise to reviewer submissions. If it turns out to bother the reviewer models, the schema could be built per-role rather than shared, but I would not do that speculatively.

`Refs #N` rather than `Fixes #N` for a message we invented is exactly right.

PR #1532, opened 2026-08-13
fix: omit the issue reference when a task has no issue number

### LINKED ISSUE #1530: [BUG] synthesized commit message says "issue #0" when the task has no issue number
## Bug Description

Follow-up to #1529, which lands the fallback this concerns. Not present on `main` today; it becomes live when that PR merges.

The synthesized commit message formats the issue number unconditionally:

```go
commitMessage = fmt.Sprintf(
    "fix: resolve issue #%d\n\n%s\n\nRefs #%d",
    task.Spec.Payload.Issue,
    strings.TrimSpace(lr.Terminal.Summary),
    task.Spec.Payload.Issue)
```

`Payload.Issue` is `int32` with `json:"issue,omitempty"`, so a task that carries no issue number produces:

```
fix: resolve issue #0

<summary>

Refs #0
```

A subject naming a non-existent issue, and a `Refs #0` trailer pointing at nothing.

## Steps to Reproduce

1. Dispatch an AgenticTask with `payload.repo` set and `payload.issue` unset (kind `freeform`, `integrate` or `reconcile`).
2. Have the coder edit files and call `submit_result` without a `commit_message`.
3. Inspect the resulting commit on the task branch.

## Expected Behavior

With no issue number, the message omits the issue reference entirely rather than inventing `#0`. Something like `fix: <summary>` with no `Refs` trailer, or a subject keyed off the task name.

## Actual Behavior

Subject and trailer both reference issue `#0`.

## Environment

Reachable on the current fleet. Live task shapes that have a repo but no issue:

```
('freeform',  hasRepo=True, hasIssue=False) -> 4
('integrate', hasRepo=True, hasIssue=False) -> 3
('reconcile', hasRepo=True, hasIssue=False) -> 3
```

These do reach `commitPushAttempt`: the `freeformGoResult` early return is gated on `needsRepo`, and all of the above have a repo, so it returns nil and execution continues to the commit.

## Additional Context

Narrow and cosmetic next to the bug #1529 fixes, which discarded an entire diff after 1-2 hours of coder runtime. Not worth blocking that PR. Worth fixing because a `Refs #0` trailer is confusing in a git log and there is no issue for it to lead back to.

The conditional is small: build the trailer only when `Issue > 0`, and fall back to the summary alone for the subject otherwise. The gate-failed preserve path a few hundred lines below is the existing model for a message with no issue reference.

Worth a test alongside it, since the existing regression test for #1529 uses a task with an issue set and so cannot catch this shape.

joryirving:
## What

Extract the synthesized-commit-message logic into `synthesizedCommitMessage` and add the issue reference only when the task actually has one.

## Why

`Payload.Issue` is `int32` with `omitempty`, so a task carrying a repo but no issue number reads as `0`. The fallback landed in #1529 formatted it unconditionally:

```
fix: resolve issue #0

Tighten the SSRF allowlist

Refs #0
```

A subject naming a non-existent issue, and a trailer pointing at nothing.

Freeform, integrate and reconcile tasks have exactly that shape and do reach `commitPushAttempt` — `freeformGoResult`'s early return is gated on `needsRepo`, and those tasks have a repo.

Fixes #1530

## How

`synthesizedCommitMessage(task, summary)`:

- `Issue > 0` — unchanged: `fix: resolve issue #N` with the summary as body and a `Refs #N` trailer.
- otherwise — `fix: <summary>`, no issue reference and no trailer, matching the shape the gate-failed preserve path already uses for a message with no issue.
- summary also empty — falls back to the task name, so the message can never be blank and re-trip the `Commit: Message is required` failure #1529 fixed. The first line only is used as a subject, since `summary` is spec'd up to 280 chars.

## Checklist

- [x] Tests added/updated

`TestNativeExecutor_SynthesizedMessageOmitsIssueWhenTaskHasNone` drives the real executor with `Kind: Freeform` and `Issue: 0`. Against current `main` it reproduces the report exactly:

```
message must not reference issue #0:
  "fix: resolve issue #0\n\nTighten the SSRF allowlist\n\nRefs #0\n\nSigned-off-by: ..."
no issue number, so there should be no Refs trailer: (same)
```

It also asserts `commitSHA` is present, so the no-issue path still cannot discard a diff.

As #1530 notes, the existing #1529 regression test cannot catch this — `taskAndAgent` hardcodes `Issue: 9999`.

`Assisted-by: Claude Code (wrote the fix and test from the report; I reviewed the change and ran make fmt, make vet, make lint and make test locally before submitting)`

### DIFF
--- a/pkg/foreman/agent/executor_native.go
+++ b/pkg/foreman/agent/executor_native.go
@@ -1103,11 +1103,7 @@ func (e *NativeAgentLoopExecutor) commitPushAttempt(
 	// message we invented should not auto-close the issue on merge.
 	commitMessage := lr.Terminal.CommitMessage
 	if strings.TrimSpace(commitMessage) == "" {
-		commitMessage = fmt.Sprintf(
-			"fix: resolve issue #%d\n\n%s\n\nRefs #%d",
-			task.Spec.Payload.Issue,
-			strings.TrimSpace(lr.Terminal.Summary),
-			task.Spec.Payload.Issue)
+		commitMessage = synthesizedCommitMessage(task, lr.Terminal.Summary)
 		log.Info("submit_result carried no commit message; synthesized one",
 			"issue", task.Spec.Payload.Issue)
 	}
@@ -1154,6 +1150,29 @@ func (e *NativeAgentLoopExecutor) commitPushAttempt(
 // practice (the retry loop is unreachable for read-only reviewers), so no
 // reviewer rails run here. Pulled out as a method so runLLMPath stays under
 // the gocyclo ceiling, mirroring applyWorkClassPolicyForTask above.
+// synthesizedCommitMessage builds a message for a submit_result that carried
+// none, so a real diff is never discarded on "Commit: Message is required".
+//
+// Payload.Issue is int32 with omitempty, so a task with no issue number reads
+// as 0. Freeform, integrate and reconcile tasks carry a repo without an issue
+// and do reach the commit path, so the issue reference is added only when
+// there is one — otherwise the subject names issue #0 and the trailer points
+// at nothing (#1530).
+func synthesizedCommitMessage(task *foremanv1alpha1.AgenticTask, summary string) string {
+	subject := strings.TrimSpace(summary)
+	if i := strings.IndexByte(subject, '\n'); i >= 0 {
+		subject = strings.TrimSpace(subject[:i])
+	}
+	if subject == "" {
+		subject = "apply coder changes for " + task.Name
+	}
+	if issue := task.Spec.Payload.Issue; issue > 0 {
+		return fmt.Sprintf("fix: resolve issue #%d\n\n%s\n\nRefs #%d",
+			issue, strings.TrimSpace(summary), issue)
+	}
+	return "fix: " + subject
+}
+
 func (e *NativeAgentLoopExecutor) retryCoderTerminalResult(
 	ctx context.Context, log logr.Logger,
 	agent *foremanv1alpha1.Agent, task *foremanv1alpha1.AgenticTask,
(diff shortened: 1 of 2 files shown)


Defilan (approved):
Approving. Fast turnaround on #1530, and the diagnosis in the description is exactly right.

## Mutation-verified

Reverted `executor_native.go` to `upstream/main` while keeping the new test. It fails there with the exact output from the issue report, and both assertions fire rather than just one:

```
message must not reference issue #0:
  "fix: resolve issue #0\n\nTighten the SSRF allowlist\n\nRefs #0\n\n..."
no issue number, so there should be no Refs trailer: (same)
```

Your note that the #1529 regression test could not have caught this because `taskAndAgent` hardcodes `Issue: 9999` is correct, and worth having in the commit history.

Full `pkg/foreman/agent/...` passes with the fix in place.

## Behaviour probe

I exercised `synthesizedCommitMessage` directly across the shapes rather than reading it:

| issue | summary | result |
| --- | --- | --- |
| 42 | `Fix the thing` | `fix: resolve issue #42\n\nFix the thing\n\nRefs #42` |
| 42 | `First line\nSecond line` | full summary kept in the body |
| 0 | `Fix the thing` | `fix: Fix the thing` |
| 0 | `First line\nSecond line` | `fix: First line` |
| 0 | (blank) | `fix: apply coder changes for <task>` |

Subject extraction and the issue>0 path are both unchanged from #1529, which is the part I most wanted to confirm since that path is the common one.

## One consistency point, not blocking

`subject` is computed for every call, including the blank-summary fallback, but it is only *used* in the `issue == 0` return. On the `issue > 0` branch a blank summary produces a message with an empty body:

```
"fix: resolve issue #42\n\n\n\nRefs #42"
```

Your description frames the fallback as defence against a blank message re-tripping `Commit: Message is required`. That reasoning applies equally to both branches, so as written the defence only covers half the cases it was written for.

In practice it is unreachable: `submit_result.go:147-149` rejects an empty summary outright, and I grepped for another path that synthesizes a Terminal with one and found none. So this is belt-and-braces either way and I would not hold the PR for it. If you do touch it, using `subject` for the body on the issue>0 path too would make the guard symmetric and remove the "computed but conditionally used" smell.

## Minor

`gofmt` clean, DCO signed. The `Fixes #1530` reference is correct, and using the task name rather than a generic string in the fallback subject is a nice touch for anyone reading `git log` on a freeform branch later.

PR #1545, opened 2026-08-14
fix(foreman): honour Workload spec.repo for gate, push and PR, not just clone

### LINKED ISSUE #1464: [BUG] Workload.spec.repo ignored for clone and push when --git-remote-url is set: work pushes to the wrong repository and still reports GO
## Bug Description

When `--git-remote-url` is set on the agent, `Workload.spec.repo` is honoured for
fetching the issue but **ignored for clone and push**. The coder reads the right
issue, works against the right base, and then pushes the result into whatever
repository the flag names, reporting GO.

Work silently lands in an unrelated repository. Nothing in the pipeline notices.

## Root cause

`executor_native.go`, clone-target resolution:

```go
cloneURL := e.GitRemoteURL
if cloneURL == "" {
    cloneURL = resolveUpstream(task.Spec.Payload.Repo)
}
```

The static flag wins; `payload.repo` is only consulted when it is unset. The
comment immediately above states the opposite intent:

> Clone target: the statically configured fork remote when set, otherwise the
> task's own repo (#915). Deriving the clone+push target from payload.repo lets
> a single agent serve many repos instead of being pinned to one
> --git-remote-url

That goal is unreachable whenever the flag is set, which is the normal fleet
configuration.

## To Reproduce

1. Run an agent with `--git-remote-url=https://github.com/OWNER/repo-A.git`.
2. Create a Workload with `spec.repo: OWNER/repo-B` and an issue in repo-B.
3. Let the coder reach a GO.

## Expected Behavior

The branch is pushed to repo-B, the repository the Workload named. Or, if the
static flag is meant to win, the task fails fast with a clear conflict rather
than pushing somewhere the caller did not ask for.

## Actual Behavior

Observed 2026-08-08 with `spec.repo: Defilan/greenhouse-demo` on an agent
carrying `--git-remote-url=https://github.com/Defilan/LLMKube.git`:

- Issue body fetched correctly from greenhouse-demo (`bodyLen: 2243`).
- Workspace cloned so that the diff applied against greenhouse-demo's tree
  (`index.html modified +402/-52`, only sensible against that base).
- Live inspection of the running workspace:

```
WORKSPACE=/workspace/default/greenhouse-s1-rerun-code-1
origin  https://github.com/Defilan/LLMKube.git (fetch)
origin  https://github.com/Defilan/LLMKube.git (push)
BRANCH=foreman/greenhouse-s1-rerun/issue-1
```

- Verdict GO, `extra.commitSHA` recorded, and the branch present on
  **Defilan/LLMKube**, absent from greenhouse-demo:

```
$ gh api repos/Defilan/greenhouse-demo/commits/5d7e5d6e...  -> 422 No commit found
$ gh api repos/Defilan/LLMKube/branches/foreman/greenhouse-stage1-shell/issue-1
  FOUND  sha=5d7e5d6e
```

## Why nothing caught it

Every check the harness runs passed legitimately. It committed, `git push`
exited 0, and the branch existed on the remote it pushed to. No stage compares
the push target against `spec.repo`, so this is invisible to the GO verdict, to
the gate, and to a reviewer reading the diff.

This is a distinct failure class from the known "GO verdicts are not
trustworthy" problem. There the model overclaims; here the model is honest and
the harness retargets the work.

## Suggested fix

Invert the precedence so `payload.repo` wins when set, with `--git-remote-url`
as the fallback, which is what #915 describes. Two further guards worth
considering:

- Assert after push that the pushed remote matches `payload.repo` when both are
  known, and fail the task if not. A push to an unexpected repository should
  never be reportable as GO.
- Log the resolved clone and push URLs at task start. The only way I identified
  this was by exec'ing into the agent pod mid-run to read `git remote -v`; the
  logs never state where the work is going.

## Additional Context

- `pkg/foreman/agent/executor_native.go` (clone-target resolution), `pkg/foreman/agent/repo/push.go`
- Related: #915 (multi-repo intent)
- Workaround: unset `--git-remote-url`, or point it at the target repo.

Defilan:
## What

Resolves the clone/push target once in `Execute` and threads it through every downstream caller, so a Workload's `spec.repo` governs the gate, the post-push envtest, and PR opening — not just the clone.

## Why

Fixes #1464

An earlier change inverted precedence so `payload.repo` wins over the static `--git-remote-url`, but only the clone itself used the resolved value. Every other caller still read `e.GitRemoteURL` directly, so a Workload naming repo-B could clone repo-B and then gate, push and open a PR against repo-A. A partial fix here is arguably worse than none: the clone succeeding makes it look like the Workload's repo took effect.

## How

The resolved target is computed once and passed down three paths:

- `executeDeterministic` -> `buildDeterministicArgs` (gate Job clone URL)
- `runLLMPath` -> `postPushGateDecision` -> `evaluatePostPushEnvtest`
- `runLLMPath` / `retryCoderTerminalResult` -> `maybeOpenPullRequest` -> `openPullRequest`, which needs it for cross-fork head qualification

`openPullRequest` falls back to `e.GitRemoteURL` when a caller passes an empty clone URL, so the fork-detection path is unchanged for callers that do not supply one.

## Behaviour change

Called out here because it changes a failure mode, and a reader trying to understand that change should not have to find it in a commit message.

**Before:** `crossProjectRemoteMismatch` ran *before* the clone and failed fast whenever `--git-remote-url` named a different repository than `payload.repo`, with an explicit "refusing to clone and push work for one repo into another" message.

**After:** that pre-check is gone. The static remote is ignored when it is not a fork of `payload.repo`, and the task proceeds against `payload.repo`. The replacement guard, `pushedRemoteMismatch`, runs in `commitPushAttempt` *after* a successful push: it reads the workspace's actual `origin` and fails the task if it disagrees with `payload.repo`.

**Why the swap is an improvement:** checking the actual push outcome is a stronger test than predicting the target from configuration, because a config prediction can disagree with where git actually landed.

**Net effect, stated plainly:** a mismatching `--git-remote-url` no longer fails immediately. It is caught only if a push actually lands in the wrong repository. That is a deliberate trade, not an oversight.

## Also addressed in review round 2

- **Deleted `crossProjectRemoteMismatch` and its contract test.** After this PR nothing in production called it. A guard that reads as live but can no longer fire is worse than no guard, because the next person to touch this path assumes the protection is there.
- **Made the new guard actually testable.** `TestPushedRemoteMismatch` previously contained only the two cases where the guard must stay *quiet* (a local bare-repo path that is not `owner/repo` shaped, and an empty task repo), so it would have passed even if `pushedRemoteMismatch` unconditionally returned `""` — exactly the regression that would reintroduce #1464 silently. Added a positive case that points `origin` at a differently-named repo and asserts a non-empty message, plus a case confirming that a *fork* of the task repo is correctly not flagged.

## Verification

- `./pkg/foreman/...` suite on this branch: **pass** (agent 37s, repo 16.6s, all subpackages ok)
- `go build ./...` clean

Coverage was added at both ends: `executor_native_internal_test.go` for the threading, `repo/push_test.go` for the push target itself.

## Checklist

- [x] Tests added/updated - internal threading tests plus push-target coverage
- [x] `make test` passes locally - full `pkg/foreman/...` suite
- [x] Documentation updated (if user-facing change) - n/a, corrects existing documented behaviour

*Assisted-by: DeepSeek-V4-Flash (MXFP4) on two DGX Sparks via LLMKube, dispatched by Foreman; reviewed by Nemotron-49B on a Strix Halo node, which returned NO-GO on the first pass for a missing cloneURL fallback and executor parity, both addressed in the revision. Review round 2 was implemented by Qwen3.8-27B on the same fleet; its automated reviewer returned GO, but that verdict was not grounded in the diff (see #1570), so the three review items were checked against the diff by hand instead. I wrote the intent and reviewed the diff before opening this.*

### DIFF
--- a/pkg/foreman/agent/executor_native.go
+++ b/pkg/foreman/agent/executor_native.go
@@ -401,19 +401,19 @@ func (e *NativeAgentLoopExecutor) Execute(ctx context.Context, task *foremanv1al
 	}
 
 	branch := branchNameForTask(task)
+	// Resolve the clone+push target ONCE so every downstream caller — the
+	// clone itself, the deterministic gate path, the post-push envtest gate,
+	// and PR opening — uses the SAME repository the Workload asked for
+	// (#1464). Precedence: payload.repo wins when set (a single agent serves
+	// many repos instead of being pinned to one --git-remote-url); the
+	// statically configured fork remote is the fallback, and stays the target
+	// when it is a fork of payload.repo (same repo name, different owner): the
+	// fork deployment pushes to the fork, not upstream (#915).
+	cloneURL := ""
 	if needsRepo {
-		// Clone target: the statically configured fork remote when set,
-		// otherwise the task's own repo (#915). Deriving the clone+push
-		// target from payload.repo lets a single agent serve many repos
-		// instead of being pinned to one --git-remote-url; the branch
-		// still pushes to this clone's origin, which in that case is the
-		// task's repo itself.
-		cloneURL := e.GitRemoteURL
-		if cloneURL == "" {
-			cloneURL = resolveUpstream(task.Spec.Payload.Repo)
-		}
-		if msg := crossProjectRemoteMismatch(e.GitRemoteURL, task.Spec.Payload.Repo); msg != "" {
-			return e.failResult(start, foremanv1alpha1.FailureGitRemoteNotConfigured, msg), nil
+		cloneURL = resolveUpstream(task.Spec.Payload.Repo)
+		if cloneURL == "" || isForkOf(e.GitRemoteURL, task.Spec.Payload.Repo) {
+			cloneURL = e.GitRemoteURL
 		}
 		if cloneURL == "" {
 			return e.failResult(start, foremanv1alpha1.FailureGitRemoteNotConfigured,
@@ -452,13 +452,13 @@ func (e *NativeAgentLoopExecutor) Execute(ctx context.Context, task *foremanv1al
 	// first tool directly with the task payload as JSON arguments. The
 	// gate Agent uses this; M5+ reviewer agents use the LLM path.
 	if deterministic {
-		return e.executeDeterministic(ctx, task, &agent, branch, registry, start), nil
+		return e.executeDeterministic(ctx, task, &agent, branch, registry, cloneURL, start), nil
 	}
 
 	// 6+. LLM-driven path: extracted to keep Execute below the
 	// cyclomatic-complexity threshold. runLLMPath owns OAI + loop +
 	// transcript + commit/push.
-	return e.runLLMPath(ctx, task, &agent, endpoint, workspace, branch, registry, auth, needsRepo, start)
+	return e.runLLMPath(ctx, task, &agent, endpoint, workspace, branch, registry, auth, needsRepo, cloneURL, start)
 }
 
 // setupTaskBranch cuts the task's working branch in the freshly cloned
@@ -631,6 +631,7 @@ func (e *NativeAgentLoopExecutor) runLLMPath(
 	registry ToolRegistry,
 	auth *repo.Auth,
 	needsRepo bool,
+	cloneURL string,
 	start time.Time,
 ) (*Result, error) {
 	log := logf.FromContext(ctx).WithName("native-agent-loop").WithValues("task", task.Name, "ns", task.Namespace)
@@ -891,7 +892,7 @@ func (e *NativeAgentLoopExecutor) runLLMPath(
 		// record it under r.Extra. Coder-role only (reviewers are read-only).
 		e.maybePreserveGateFailedBranch(ctx, log, agent, task, workspace, branch, auth, loopRes, r)
 		e.maybeOpenPullRequest(ctx, log, task, auth, verdict, r,
-			workspace, reviewBase, reviewDiff)
+			workspace, reviewBase, reviewDiff, cloneURL)
 		// Attach the normalized failure reason from the model-to-CRD
 		// mapping (e.g. ERROR→INCOMPLETE + ModelReportedError for #649). Only
 		// set when the normalizer produced a reason AND the result does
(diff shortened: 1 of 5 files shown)


joryirving (changes requested):
Request changes: the threading fix is right and I have no issue with it, but this also swaps out a safety guard without saying so, leaves the old one inert in the tree, and the new one has no test that it ever fires.

Resolving the target once in `Execute` and passing it to the gate, the post-push envtest and PR opening is the correct shape, and the partial-fix argument in the description is the right way round: a clone that succeeds while the push goes elsewhere is worse than not honouring `payload.repo` at all.

The part the description does not mention is that this removes the `crossProjectRemoteMismatch` call at executor_native.go:415 and replaces it with `pushedRemoteMismatch` after the push. That is a real behaviour change, not a refactor. Before, a deployment whose `--git-remote-url` named a different repository than the task failed immediately with an explicit message about refusing to clone and push work for one repo into another. Now the static remote is simply ignored in that case and the task proceeds against `payload.repo`. I think that is defensible given the new precedence, and checking the outcome after the push is a better test than checking config beforehand, but it should be in the description because someone reading this to understand a failure mode change will not find it.

`crossProjectRemoteMismatch` itself is still in the file at executor_native.go:2495 with its doc comment at :2477, and after this PR nothing calls it. Main has three references, this removes one, and the other two are the definition and its comment. A guard that reads as live but can no longer fire is worse than no guard, because the next person to touch this path will assume the protection is there. Either delete it or keep the call.

On the new check, `TestPushedRemoteMismatch` only asserts the two cases where it must stay quiet: a local bare-repo path that is not owner/repo shaped, and an empty task repo. There is no case where the remote genuinely disagrees with `payload.repo` and the function is expected to return a message. As written the test passes if `pushedRemoteMismatch` unconditionally returns "", which is exactly the regression that would reintroduce #1464 silently. Worth adding a positive case with `SetRemote` pointing origin at a differently-named repo and asserting the message comes back.

`repo.RemoteURL` reading the actual origin after the push is a good addition, and I like that the guard is now an assertion about what happened rather than a prediction from config.

PR #1546, opened 2026-08-14
feat(foreman): add optional Agent.spec.maxConcurrentTasks to bound in-flight tasks

### LINKED ISSUE #1497: Agent: optional spec.maxConcurrentTasks to bound in-flight tasks per Agent
## Problem

There is no way to bound how many tasks run concurrently for a given Agent. `Agent.spec` carries `maxTurns`, `maxRetries`, `maxOutputTokens`, `maxEnvtestIterations` — all per-task limits — but nothing per-Agent-fleet.

That matters when an Agent's model is served by a backend shared with other consumers. Concretely: our `coder-strix` and `coder-revision` Agents both point at a local model whose inference server has 2 parallel slots, and that same server also serves a documentation cron job and a GitHub PR-review action. When several coder tasks run at once, Foreman consumes every slot and the non-Foreman consumers queue behind multi-minute coder turns.

The only levers today are indirect and blunt:

- The re-dispatcher's global in-flight cap throttles *every* Agent, including ones on completely different backends, to protect one.
- Routing ratios (which Agent gets which issue) change the Agent's *share*, not its absolute concurrency, so the bound moves with total load.

Neither can express "this Agent may have at most N tasks in flight", which is the actual constraint when a backend's capacity is shared.

## Proposal

An optional `spec.maxConcurrentTasks` on Agent. When set, the AgenticTask controller does not dispatch an (N+1)th task for that Agent until one completes; tasks simply wait in their existing pre-dispatch state, exactly as they do when no node matches.

- Unset = today's behavior (unbounded), so nothing changes on upgrade.
- Enforced at dispatch time in the controller, next to the existing node-reservation logic, so it composes with FleetNode reservation rather than duplicating it.
- Counting is per-Agent across the namespace (a simple list-and-count of non-terminal tasks referencing that `agentRef`), which is the same shape the controller already computes for other decisions.

This gives operators a direct expression of backend capacity: an Agent on a 2-slot server can be pinned to 1 or 2, leaving predictable headroom for other consumers of that server, without throttling Agents that run elsewhere.

Related: #1496 (Job-mode tasks reserving FleetNodes) — that one is about a reservation that should not exist; this is about a bound that cannot currently be expressed.

Defilan:
## What

Adds an optional `Agent.spec.maxConcurrentTasks` that bounds how many of an Agent's tasks may be in flight at once. Unset means unbounded, which is today's behaviour.

## Why

Fixes #1497

Without a bound, one Agent can occupy every available FleetNode. Observed live on 2026-08-14: four concurrent reviews consumed all four nodes in a pool while coder tasks sat Pending, leaving the inference box those coders target completely idle with work queued. Raising the replica count lifts the ceiling but does not change the ordering, so a burst of one task kind can still starve another.

## How

Enforced at dispatch in the AgenticTask reconciler, beside the existing node-reservation logic so it composes with FleetNode reservation rather than duplicating it. When the bound is set, the controller counts the Agent's non-terminal tasks in the namespace and declines to claim an (N+1)th, leaving the excess Pending for the next pass.

The task being reconciled is excluded from the count, since it is not yet in flight.

Per-Agent only. No controller-wide global bound: the failure this addresses is one Agent monopolising capacity, and a global cap would throttle unrelated Agents to fix it.

## Verification

- `./pkg/foreman/...`, `./internal/foreman/...`, `./api/foreman/...` on this branch: **pass**
- CRD regeneration confirmed complete: re-running `make manifests` and `make foreman-chart-crds` leaves the tree **clean**, which is what CI's CRD Sync Check enforces. Both `config/crd/bases/` and `charts/foreman/templates/crds/` are included in the commit
- `go build ./...` clean

`make foreman-chart-crds` is a separate target from `make chart-crds`; both generated copies are present.

## Checklist

- [x] Tests added/updated - 122 lines in `agentictask_controller_test.go` covering the bound, the unset case, and exclusion of the reconciled task
- [x] `make test` passes locally - foreman suites pass
- [x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/) - PR title carries the conventional prefix; the commit subject does not
- [x] Documentation updated (if user-facing change) - new CRD field is documented in its godoc, which `kubectl explain` surfaces

*Assisted-by: DeepSeek-V4-Flash (MXFP4) on two DGX Sparks via LLMKube, dispatched by Foreman; reviewed by Nemotron-49B on a Strix Halo node. I wrote the intent, ran the suites, and independently re-ran both CRD generators to confirm the commit was complete before opening this.*

### DIFF
--- a/api/foreman/v1alpha1/agent_types.go
+++ b/api/foreman/v1alpha1/agent_types.go
@@ -282,6 +282,15 @@ type AgentSpec struct {
 	// +optional
 	MaxOutputTokens int32 `json:"maxOutputTokens,omitempty"`
 
+	// MaxConcurrentTasks bounds how many of this Agent's tasks may be in
+	// flight at once. When set, the AgenticTask controller does not claim
+	// an (N+1)th task for this Agent until one of its in-flight tasks
+	// completes; excess tasks stay Pending for the next pass. Unset means
+	// unbounded (today's behavior).
+	// +kubebuilder:validation:Minimum=1
+	// +optional
+	MaxConcurrentTasks *int32 `json:"maxConcurrentTasks,omitempty"`
+
 	// ChatTemplateKwargs is a free-form map of template-specific keyword
 	// arguments forwarded verbatim on every chat-completions request as
 	// chat_template_kwargs. Each value is a raw JSON blob so booleans stay
--- a/api/foreman/v1alpha1/zz_generated.deepcopy.go
+++ b/api/foreman/v1alpha1/zz_generated.deepcopy.go
@@ -282,6 +282,11 @@ func (in *AgentSpec) DeepCopyInto(out *AgentSpec) {
 		*out = new(int32)
 		**out = **in
 	}
+	if in.MaxConcurrentTasks != nil {
+		in, out := &in.MaxConcurrentTasks, &out.MaxConcurrentTasks
+		*out = new(int32)
+		**out = **in
+	}
 	if in.ChatTemplateKwargs != nil {
 		in, out := &in.ChatTemplateKwargs, &out.ChatTemplateKwargs
 		*out = make(map[string]apiextensionsv1.JSON, len(*in))
--- a/charts/foreman/templates/crds/agents.yaml
+++ b/charts/foreman/templates/crds/agents.yaml
@@ -236,6 +236,16 @@ spec:
                     type: string
                 type: object
                 x-kubernetes-map-type: atomic
+              maxConcurrentTasks:
+                description: |-
+                  MaxConcurrentTasks bounds how many of this Agent's tasks may be in
+                  flight at once. When set, the AgenticTask controller does not claim
+                  an (N+1)th task for this Agent until one of its in-flight tasks
+                  completes; excess tasks stay Pending for the next pass. Unset means
+                  unbounded (today's behavior).
+                format: int32
+                minimum: 1
+                type: integer
               maxEnvtestIterations:
                 description: |-
                   MaxEnvtestIterations bounds how many times the executor re-runs a
--- a/config/crd/bases/foreman.llmkube.dev_agents.yaml
+++ b/config/crd/bases/foreman.llmkube.dev_agents.yaml
@@ -235,6 +235,16 @@ spec:
                     type: string
                 type: object
                 x-kubernetes-map-type: atomic
+              maxConcurrentTasks:
+                description: |-
+                  MaxConcurrentTasks bounds how many of this Agent's tasks may be in
+                  flight at once. When set, the AgenticTask controller does not claim
+                  an (N+1)th task for this Agent until one of its in-flight tasks
+                  completes; excess tasks stay Pending for the next pass. Unset means
+                  unbounded (today's behavior).
+                format: int32
+                minimum: 1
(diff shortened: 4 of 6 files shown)


joryirving (changes requested):
Request changes: this deadlocks as soon as more tasks are queued than the bound allows, which is the exact situation it exists to handle.

`agentAtConcurrencyLimit` in internal/foreman/controller/agentictask_controller.go counts every non-terminal task for the Agent, and `isTerminalPhase` treats only Succeeded and Failed as terminal. Pending is not terminal, so queued tasks that have not been claimed count against the bound they are queuing for. With `maxConcurrentTasks: 2` and five Pending tasks for one Agent, reconciling any one of them sees four non-terminal siblings, four is already at or above two, and it declines. Every task declines on every pass and nothing ever starts. The bound does not throttle the Agent, it stops it.

The tests do not catch this because each case has exactly one sibling: one Running task in the at-limit case, one Running in the under-limit case, one Succeeded in the terminal case. A case with three or four Pending tasks for the same Agent and a bound of two would fail today.

The fix is to count only tasks that actually hold a slot rather than everything that is not finished. Scheduled, Running and Verifying occupy capacity; Pending does not, by definition, since it is the state a task sits in while waiting for exactly this decision. Inverting `isTerminalPhase` into an explicit in-flight predicate would also read better against the field's godoc, which says the controller "does not claim an (N+1)th task until one of its in-flight tasks completes".

Separately, worth being explicit somewhere that the count-then-claim is not atomic the way the node reservation is. Two reconciles for the same Agent could each count N-1 and both proceed. It is safe today only because this controller does not set MaxConcurrentReconciles and so runs at the controller-runtime default of one. That is a real coupling to an unstated default, and if anyone raises it later the bound silently stops holding. A comment at the call site would be enough.

The shape of the feature is right and it is the knob I wanted. We hit the underlying problem the other night: reviews ran sixteen wide against a two-slot backend and I had to cap total agent replicas to get it under control, which throttled coders at the same time because they share the pool. Per-Agent rather than global is the correct call for that.

joryirving (approved):
Approving. The deadlock is gone and the test for it is exactly the case I was worried about.

`holdsInFlightSlot` counting only Scheduled, Running and Verifying is the right inversion. Pending is the state a task sits in while waiting for this decision, so counting it against the bound was the thing that turned a throttle into a stop. The godoc spelling out why is worth having, because "count what has not finished" is the intuitive implementation and it is wrong here in a way that only shows up under load.

The new case with four Pending siblings against a bound of two is the shape that would have failed before, and asserting the task comes back Scheduled rather than Pending is the right assertion. The earlier tests all had exactly one sibling, which is why they passed either way.

One small thing still outstanding, and it does not block. The count-then-claim is not atomic the way the node reservation is, so two reconciles for the same Agent could each count N-1 and both proceed. It holds today only because this controller does not set `MaxConcurrentReconciles` and runs at the controller-runtime default of one. A comment at the call site would be enough to stop someone raising that later and silently breaking the bound.

This is the knob I wanted. We hit the underlying problem directly: reviews ran sixteen wide against a two-slot backend, and capping total agent replicas throttled coders at the same time because they share the pool. Per-Agent is the correct granularity for that. All 25 checks are green.

PR #1558, opened 2026-08-14
fix(foreman): do not reserve a FleetNode for Job-mode AgenticTasks

### LINKED ISSUE #1496: Job-mode AgenticTasks reserve a FleetNode for the Job's lifetime, starving in-process reviewers/gates
## Bug

A `mode: Job` AgenticTask reserves a FleetNode for the Job's entire lifetime, even though the work executes in a separate Job pod. The claiming agent does nothing locally but wait, so every long-running Job-based coder permanently consumes one of the fleet's one-task-per-node slots.

With coders on `mode: Job` and gates/reviewers in-process, this starves the in-process roles outright:

```
foreman-agent-...-4l2lj  currentTask=llm/wl-...-438-code-438
foreman-agent-...-f4qd9  currentTask=llm/wl-...-420-code-420
foreman-agent-...-t8459  currentTask=llm/wl-...-35-code-35
```

```
INFO no free FleetNode matches; will retry  {"controller":"agentictask","task":"wl-...-34-review-34-0"}
```

Five review tasks sat `Pending`/unassigned for up to two hours on a 3-replica deployment while the three coder Jobs ran for hours in their own pods. Throughput collapses to `replicaCount` concurrent coders and zero reviewers, regardless of how much model capacity is available.

## Why this is the wrong reservation

`reserveNode` / `Status.CurrentTask` exists to enforce one-task-per-node because an in-process task consumes the agent's own CPU, memory and workspace. A Job-mode task consumes none of those on the agent — its resources are the Job pod's. The reservation models a constraint that no longer exists for that execution mode.

Observed cost: the agent pods holding these reservations measured 1–2m CPU and ~17–20Mi RSS while "busy".

## Proposal

Options, roughly in order of preference:

1. **Don't reserve for Job-mode tasks.** Track them by owner reference / Job status instead of `CurrentTask`; the concurrency limit that matters is the cluster's ability to schedule Jobs, not the agent's local capacity.
2. **Separate reservation pools by execution mode** — an agent can hold N Job-mode claims (bounded, since it only polls Job status) plus one in-process task.
3. **Per-node concurrency for Job-mode claims**, e.g. `agent.maxJobClaims` (default 1, so current behavior is preserved), letting operators size it without inflating replica counts.

## Workaround

Raise `agent.replicaCount` until there are more nodes than concurrent Job-based coders plus expected in-process reviews/gates. This works (replicas are cheap coordinators) but scales pod count with coder concurrency for no real reason, and it is not discoverable — the only symptom is reviews silently never scheduling.

Foreman 0.9.16; coder Agents `execution.mode: Job`, gates/reviewers in-process.

Defilan:
## What

Job-mode AgenticTasks no longer reserve the FleetNode they are assigned to. The task is still placed on an eligible node so its Job is created there, but the node's `CurrentTask` is left untouched.

## Why

Fixes #1496

A Job-mode task's work runs in an ephemeral Job pod, not on the claiming node. Stamping `Status.CurrentTask` held a one-task-per-node slot at the CONTROLLER level for the Job's entire lifetime while the node itself did nothing but wait. Reserving a resource you do not use is wrong on its own terms, and it is a prerequisite for ever running in-process work alongside a Job.

## What this does NOT do

**This alone does not enable concurrent execution.** @joryirving caught that the original version of this description overclaimed, and the correction is worth stating plainly rather than quietly editing away.

Emptying `CurrentTask` lets the controller schedule a second task onto the node. The agent still will not run it, because the node's `AgenticTaskWatcher` keeps one process-wide in-flight slot:

- `pkg/foreman/agent/watcher.go:251-254` returns from `pollOnce` as soon as `inflight != nil`.
- `pkg/foreman/agent/watcher.go:314` sets the slot; the deferred clear at `:323` only runs once `Executor.Execute` returns.
- For Job mode, `Execute` blocks: `executeCoderJob` calls `CoderJobSubmitter.Submit` (`RunCoderJob.Run`), which creates the Job and then polls in a `for` loop until `job.Status.Succeeded >= 1` or `job.Status.Failed >= 1`.

So a Job-mode task still occupies its node for the Job's whole lifetime. The watcher fix is tracked separately in #1559.

This is a necessary but not sufficient step. Both are needed for Job-mode concurrency, and they can land in either order.

## How

`reserveFirstFitNode` returns the chosen node without writing the reservation when the task is Job-mode:

```go
if jobMode {
    return n.Name, nil
}
```

Node selection and eligibility filtering are unchanged, so the Job still lands where it would have. In-process scheduling is untouched.

## Scope note

This is a **latent** fix rather than an observed one. No Agent in our deployment currently sets `execution.mode`, so nothing here runs in Job mode today and this changes nothing for us in practice.

The superficially similar starvation we hit on 2026-08-14, four concurrent reviews consuming every node while coder tasks sat Pending, was **in-process** contention and is not addressed by this change. That one needs a per-Agent concurrency bound, which is #1497 (and #1546).

Note that #1546's `maxConcurrentTasks` is also controller-side. It bounds how many tasks the controller schedules for an Agent and does not touch `watcher.go`, so it does not address #1559 either.

## Verification

- `./internal/foreman/...` with `KUBEBUILDER_ASSETS` set: **pass** (18.3s), matching unmodified `main` (17.0s)
- Merges cleanly onto current `main` (`6694d0b0`)
- `go build ./...` clean

**What the added test actually proves:** that a Job-mode task and an in-process task can both be marked `Scheduled` on the same node, and that the in-process reservation path is unchanged. It does **not** prove they execute concurrently, and cannot, because that is gated by the watcher (#1559). The test names should be read with that scope in mind.

Note for anyone running these locally: without `KUBEBUILDER_ASSETS` the suite fails in ~0.03s on envtest setup, which looks like a real assertion failure and is not.

The branch shares `agentictask_controller.go` and its test with the #1497 work, but `git merge-tree` confirms they touch different regions and do not conflict, so the two can land in either order.

## Checklist

- [x] Tests added/updated - 57 lines covering the Job-mode reservation path and the unchanged in-process path
- [x] `make test` passes locally - foreman suites pass with envtest assets
- [x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/) - PR title carries the conventional prefix; the commit subject does not

### DIFF
--- a/internal/foreman/controller/agentictask_controller.go
+++ b/internal/foreman/controller/agentictask_controller.go
@@ -327,6 +327,14 @@ func (r *AgenticTaskReconciler) allDepsSucceeded(ctx context.Context, task *fore
 // is an optimistic-lock status patch, so at most one wins; the loser (Conflict,
 // or a node already reserved for a live task) falls through to the next
 // candidate. Without this, every task funnels onto one node. See #977.
+//
+// In Job mode the task's work runs in an ephemeral Job pod, not on the claiming
+// node, so the node must NOT be reserved: stamping CurrentTask would hold a
+// one-task-per-node slot for the Job's whole lifetime while the node itself
+// does nothing but wait, starving in-process tasks that could run there. A
+// Job-mode task is still assigned to a node (so the Job is created there), but
+// the node's CurrentTask is left untouched and stays free for in-process work.
+// See #1496.
 func (r *AgenticTaskReconciler) reserveFirstFitNode(ctx context.Context, task *foremanv1alpha1.AgenticTask, required foremanv1alpha1.RequiredCapability, requiredModel string, jobMode bool) (string, error) {
 	var nodes foremanv1alpha1.FleetNodeList
 	if err := r.List(ctx, &nodes); err != nil {
@@ -344,6 +352,13 @@ func (r *AgenticTaskReconciler) reserveFirstFitNode(ctx context.Context, task *f
 		}
 		if !capabilitySatisfies(required, requiredModel, n, jobMode) {
 			continue
+		}
+		if jobMode {
+			// Job-mode work runs in an ephemeral Job pod, not on this node, so
+			// the node must not be claimed. Pick the first eligible node and
+			// leave its CurrentTask untouched so it stays free for in-process
+			// tasks. See #1496.
+			return n.Name, nil
 		}
 		reserved, err := r.reserveNode(ctx, n, key)
 		if err != nil {
--- a/internal/foreman/controller/agentictask_controller_test.go
+++ b/internal/foreman/controller/agentictask_controller_test.go
@@ -445,6 +445,63 @@ var _ = Describe("AgenticTaskReconciler scheduler", func() {
 			To(ConsistOf("default/spread-t1", "default/spread-t2"))
 	})
 
+	It("schedules a Job-mode task without reserving the node's CurrentTask", func() {
+		// Regression for defilantech/LLMKube#1496: a Job-mode task's work runs
+		// in an ephemeral Job pod, not on the claiming node, so the node must
+		// not be claimed via Status.CurrentTask. The node stays free so an
+		// in-process task can still be dispatched to it while the Job runs.
+		agent := newAgent("job-coder")
+		agent.Spec.Execution = &foremanv1alpha1.ExecutionSpec{
+			Mode: foremanv1alpha1.ExecutionModeJob,
+		}
+		Expect(k8sClient.Create(ctx, agent)).To(Succeed())
+		DeferCleanup(func() { _ = k8sClient.Delete(ctx, agent) })
+
+		task := newTask("job-target")
+		task.Spec.AgentRef = &corev1.LocalObjectReference{Name: agent.Name}
+		Expect(k8sClient.Create(ctx, task)).To(Succeed())
+		DeferCleanup(func() { _ = k8sClient.Delete(ctx, task) })
+		setPhase(task, foremanv1alpha1.AgenticTaskPhasePending)
+
+		node := newFleetNode("job-node")
+		Expect(k8sClient.Create(ctx, node)).To(Succeed())
+		DeferCleanup(func() { _ = k8sClient.Delete(ctx, node) })
+		setNodeReady(node, foremanv1alpha1.FleetNodeCapability{
+			Accelerator: foremanv1alpha1.FleetNodeAccelerator("metal"),
+		})
+
+		_, err := reconciler.Reconcile(ctx, reqFor(task))
+		Expect(err).NotTo(HaveOccurred())
+
+		var fresh foremanv1alpha1.AgenticTask
+		Expect(k8sClient.Get(ctx, nn(task), &fresh)).To(Succeed())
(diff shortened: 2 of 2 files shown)


joryirving (changes requested):
Request changes: this fixes the FleetNode reservation, but it does not actually enable the concurrency the PR claims.

The controller now leaves CurrentTask empty for Job-mode tasks at internal/foreman/controller/agentictask_controller.go:356, so it can schedule an in-process task onto the same node. However, the node's AgenticTaskWatcher still has a single process-wide inflight slot. pollOnce returns immediately whenever inflight is non-nil, and launchExecutor does not clear that slot until the executor returns. For a Job-mode task, the executor blocks while it submits and polls the ephemeral Job, so the watcher will not claim the in-process task until that Job finishes.

That means the added controller test proves only that both tasks can be marked Scheduled; it does not prove that both tasks can execute concurrently. In the live agent, a Job-mode task can still starve an in-process task for the entire Job lifetime despite CurrentTask being empty. Please either make the watcher concurrency-aware for Job-mode tasks, or narrow the PR's claim and add an explicit follow-up/guard so this change is not presented as freeing the node for concurrent in-process execution. The relevant watcher path is pkg/foreman/agent/watcher.go:248-285 and pkg/foreman/agent/watcher.go:310-342.

CI checks are otherwise green, and I found no separate correctness issue in the changed reservation branch.

Defilan:
Confirmed, on all three points. I traced each against `main` at `6694d0b0` rather than take it on trust:

- `pollOnce` returns as soon as `inflight != nil` (`watcher.go:251-254`).
- `launchExecutor` sets the slot at `:314` and the deferred clear at `:323` only runs once `Executor.Execute` returns.
- `Execute` does block in Job mode: `executeCoderJob` calls `CoderJobSubmitter.Submit`, which is `RunCoderJob.Run`, and that creates the Job then polls in a `for` loop until `job.Status.Succeeded >= 1` or `job.Status.Failed >= 1`.

So the slot is held for the Job's whole lifetime, and emptying `CurrentTask` only moves the bottleneck from the controller to the agent. The description was wrong, not just imprecise.

Your reading of the test is right too. It proves both tasks can be marked `Scheduled` on one node and that the in-process path is unchanged. It cannot prove concurrent execution, because the watcher gates that.

Taking your second option rather than the first. I have:

- Filed #1559 for making the watcher concurrency-aware, with the three call sites and the `// Exactly one task at a time in v0.1` comment that documents the constraint.
- Rewritten this PR's description to state that it is necessary but not sufficient, with a "What this does NOT do" section naming `watcher.go:251-254` / `:314` / `:323` and pointing at #1559, plus an explicit note on what the test does and does not establish.

The reason for splitting rather than fixing the watcher here: this PR is a latent fix (no Agent in our fleet sets `execution.mode`, so it changes nothing for us today), while making the watcher concurrency-aware touches the claim races, the `watchTaskLiveness` watchdog, and `patchTerminal`, all of which assume a single in-flight task. That did not seem like something to bolt onto a two-line reservation change.

One thing worth flagging since it is adjacent: #1546's `maxConcurrentTasks` does not help here either. It is controller-side only and never touches `watcher.go`, so it bounds the ceiling while #1559 is the floor being stuck at one.

Happy to fold the watcher work in here instead if you would rather see them land together.

Issue #1548, opened 2026-08-14
[EPIC] Make agent verdicts verifiable: ground the method, not just the summary


Defilan:
## Feature Description

Make a Foreman `GO` verifiable rather than asserted. Today every stage reports
that it did its job; nothing checks that it did. This epic tracks the mechanisms
that convert stage judgments into auditable facts.

## Problem Statement

> As anyone relying on an autonomous pipeline, I want `GO` to mean a specific,
> checkable set of things were true, so that accepting it does not require
> re-doing the review by hand.

The gap is **not** capability, tooling or instructions:

- The reviewer Agent already has `bash`, `grep`, `read_file` and `fetch_issue`.
- The reviewer system prompt already carries a ten-section rubric, including
  section F, "Regression check (read `main` for every touched file)", complete
  with a worked calibration case about a fix that deleted a fallback path
  nobody asked to remove.
- The gate already runs the project's real checks.

All three stages can pass a change that a human reviewer rejects in minutes,
because a stage that skips a rubric section is indistinguishable from one that
performs it and finds nothing.

A run of ten issues produced concrete instances of each failure mode. Every one
was caught by a human doing something the pipeline had the tools to do:

| what was missed | who passed it | what would have caught it |
| --- | --- | --- |
| Fix deleted an existing bounded fix for the same issue, reintroducing the regression its comment warned about | coder GO, gate PASS, reviewer GO | pre-flight staleness check; deleted-reference guard; evidence for rubric F |
| Fix handled the enabled path, left the disabled path (named in the issue) still broken | coder GO, gate PASS, reviewer GO | expected-behaviour clause coverage |
| Coder returned GO with a detailed summary of edits it never committed | coder GO, gate PASS | cross-stage contradiction detection |
| Reviewer reported "branch has no commits" against a branch carrying a real commit, killing the workload | reviewer NO-GO | contradiction detection; evidence required for emptiness claims |

The last two are the important pair: the same stage produced both a false
negative and a false positive within one workload, and the pipeline resolved
each **positionally** (whoever spoke last won) rather than treating the
disagreement as signal.

## Proposed Solution

Sub-issues, ordered by catch-rate per unit of effort:

1. Cross-stage contradiction detection
2. Pre-flight issue-staleness check
3. Evidence-bearing rubric compliance
4. Require evidence for emptiness / unreadability claims (route to `ERROR`)
5. Deleted-reference guard
6. Expected-behaviour clause coverage
7. Mutation check in the gate

[...]

PR #1571, opened 2026-08-16
fix(foreman): guard issue rendering in reviewer prompt when issue is unset

### LINKED ISSUE #1570: [BUG] Reviewer returns GO without ever diffing the branch: verdicts uncorrelated with the code they approve
## Bug Description

An in-process reviewer can return **GO** without ever obtaining the diff of the branch it is reviewing. In two consecutive runs the reviewer navigated to the workspace, read an arbitrary file that the change never touched, and approved. Neither transcript contains a single line of the branch's diff.

This is more severe than a shallow or lenient review. A lenient reviewer at least read the change. Here the verdict is uncorrelated with the code it approves, so a GO carries no information at all, while every downstream consumer (PR opening, escalation routing, the fix-cycle) treats it as verification.

Related in spirit to #1454 (a GO that asserts what the reviewer did not do) and to epic #1548 (ground the method, not just the summary). It is arguably the strongest case in that epic, because it invalidates the verdict rather than merely weakening its justification.

## Steps to Reproduce

1. Dispatch a `Workload` using `spec.pipeline` with a `review` step whose `payload` carries `repo`, `branch`, `baseBranch` and `branchStrategy` but **no `issue` key**.
2. Let the coder produce a branch and the gate pass.
3. Inspect the review task's stored transcript ConfigMap `foreman-transcript-<review-task-name>` and grep it for `diff --git` and for any identifier the diff introduced.

## Expected Behavior

A reviewer that never diffs the branch under review cannot return GO. Either the loop makes the diff unavoidable, or a rail fails the verdict when the transcript shows the branch was never diffed.

## Actual Behavior

Two runs on 2026-08-15, both `GO`, both ungrounded.

**Run A — `wl-1454-revise-review-1454-r2`** (3 turns, 684s)

Verdict `GO`. Summary:

> Dockerfile follows best practices with non-root user, minimal base image, and proper dependency handling.

The branch changed exactly three Go files (`pkg/foreman/agent/unverified_claim_gate.go`, its test, and 6 lines of `executor_native.go`) and no Dockerfile. The transcript shows the reviewer ran `ls -l` at the workspace root and then read `Dockerfile.foreman-agent`. Grepping the full 29KB transcript:

| token | occurrences |
|---|---|
| `unverified_claim_gate` | 0 |
| `phraseCountsAsUnverified` | 0 |
| `diff --git` | 0 |
| `Dockerfile` | 28 |

**Run B — `wl-1545-revise-review-1545-r2`** (4 turns, 749s)

Verdict `GO`. Summary:

> Code contains multiple TODOs and FIXMEs indicating incomplete work

The diff adds **zero** TODO or FIXME lines. The reviewer ran `git branch --show-current`, then grepped the repository and surfaced 36 pre-existing TODOs and 18 pre-existing FIXMEs as though they were findings against the change. Grepping its transcript:

| token | occurrences |
|---|---|
| `pushedRemoteMismatch` | 0 |
| `crossProjectRemoteMismatch` | 0 |
| `diff --git` | 0 |

Run B also **states a defect and returns GO in the same breath**. Whatever one thinks of the finding's accuracy, "incomplete work" and approval are not consistent, and nothing in the pipeline noticed.

## Observation, not a proven cause

Both review prompts rendered:

> You are reviewing the branch the coder produced for issue **#0** of defilantech/LLMKube.

`issue: 0` is what a `spec.pipeline` review payload with no `issue` key produces. The workload in the same session that DID pass an issue number (top-level `spec.issues`) produced a summary that at least described the right change. That correlates, it is not established as causal, and it should be checked rather than assumed.

Independently of the cause, rendering `issue #0` into a reviewer prompt is worth fixing on its own: it tells the model to reason about an issue that cannot exist.

## Suggested fix

Two layers, the second being the load-bearing one:

1. **Make the diff unavoidable.** Put the `git diff <base>...<branch>` output into the reviewer's context up front rather than relying on the model to fetch it via Step 1 of the system prompt. The existing rails already compute the changed-line set (`reviewerGroundedChangedLines`), so the data is at hand.
2. **Add a rail: no diff, no verdict.** Mirror the existing rail chain in `pkg/foreman/agent/executor_native.go`. If the reviewer's transcript contains no evidence it obtained the branch diff, the verdict cannot stand as GO. The natural target is `INCOMPLETE` with an infrastructure-style reason rather than `NO-GO`, since the reviewer did not reject the work, it failed to review it, and per #1552 a wrong NO-GO destroys good work.

A cheaper interim guard, in the spirit of #1166 grounded-findings: require at least one reviewer tool call whose output overlaps the set of files the diff touched, and fail the verdict when the intersection is empty. In Run A the intersection was `{}` against a three-file diff.

Also worth fixing: `spec.pipeline` review payloads should either require the issue number or omit the sentence entirely rather than emitting `#0`.

## Environment

**LLMKube Version:** chart `foreman-0.9.17`, agent image `0.9.17-main.g6694d0b`

**Kubernetes Version:** `v1.36.2`

**Cluster Type:**
- [x] Other: MicroK8s, 4 nodes (1 AMD Strix Halo, 2 DGX Spark GB10, 1 x86 infra)

**GPU (if applicable):**
- [x] Other: reviewer served on AMD Strix Halo (gfx1151, Vulkan); coder on 2x DGX Spark GB10 (CUDA sm_121)

Reviewer model: Nemotron-49B, `contextWindowTokens: 30000`, `maxTurns: 40`, in-process (`ExecutionModeInProcess`).

## Logs

Transcripts are the evidence here rather than pod logs:

```bash
kubectl -n default get cm foreman-transcript-wl-1454-revise-review-1454-r2 -o jsonpath='{.data.transcript\.json}' | grep -c 'diff --git'   # 0
kubectl -n default get cm foreman-transcript-wl-1545-revise-review-1545-r2 -o jsonpath='{.data.transcript\.json}' | grep -c 'diff --git'   # 0
```

## Additional Context

Both branches under review were, on independent human inspection, **correct** and did what their revision prompts asked. So the reviewer approved good work for entirely wrong reasons. The failure is silent precisely because the outcome happened to be right; had the coder produced something broken, these same two GOs would have shipped it.

Assisted-by: Claude Opus 5

snowyukitty:
## Problem
When a pipeline dispatch payload omits the `issue` key (or passes `issue: 0`), `buildUserPrompt` on `AgenticTaskKindReview` rendered `You are reviewing the branch the coder produced for issue #0 of <repo>` and `- issue: 0`. This instructed reviewer models to evaluate an issue number that does not exist (#1570).

## Solution
- In `pkg/foreman/agent/executor_native.go`, check `p.Issue > 0` before emitting issue references in the review prompt header; when `Issue` is 0 or unset, render `You are reviewing the branch <branch> of <repo>.` without referencing issue `#0`.
- Add unit test `TestBuildUserPrompt_ReviewerCase_ZeroIssue_OmitsIssueLine` in `pkg/foreman/agent/executor_native_internal_test.go` verifying that unset or zero issues omit issue lines.

Fixes #1570

### DIFF
--- a/pkg/foreman/agent/executor_native.go
+++ b/pkg/foreman/agent/executor_native.go
@@ -2651,11 +2651,18 @@ func buildUserPrompt(task *foremanv1alpha1.AgenticTask) string {
 		// it, but parity across the local reviewer fleet matters.
 		// Empirical: rerun-7 review-510-1 (devstral) failed turn 1
 		// with that exact 400 before this case existed.
-		fmt.Fprintf(&b, "You are reviewing the branch the coder produced for issue #%d of %s.\n\n",
-			p.Issue, p.Repo)
-		fmt.Fprintf(&b, "- repo: %s\n", p.Repo)
-		fmt.Fprintf(&b, "- issue: %d\n", p.Issue)
-		fmt.Fprintf(&b, "- branch: %s\n", p.Branch)
+		if p.Issue > 0 {
+			fmt.Fprintf(&b, "You are reviewing the branch the coder produced for issue #%d of %s.\n\n",
+				p.Issue, p.Repo)
+			fmt.Fprintf(&b, "- repo: %s\n", p.Repo)
+			fmt.Fprintf(&b, "- issue: %d\n", p.Issue)
+			fmt.Fprintf(&b, "- branch: %s\n", p.Branch)
+		} else {
+			fmt.Fprintf(&b, "You are reviewing the branch %s of %s.\n\n",
+				p.Branch, p.Repo)
+			fmt.Fprintf(&b, "- repo: %s\n", p.Repo)
+			fmt.Fprintf(&b, "- branch: %s\n", p.Branch)
+		}
 		b.WriteString("\nFollow Step 1 of your system prompt to navigate to ")
 		b.WriteString("the branch under review before forming any judgment, ")
 		b.WriteString("then apply the Step 2 review checklist, then call ")
--- a/pkg/foreman/agent/executor_native_internal_test.go
+++ b/pkg/foreman/agent/executor_native_internal_test.go
@@ -966,6 +966,39 @@ func TestBuildUserPrompt_ReviewerCaseProducesNonEmptyContent(t *testing.T) {
 	}
 }
 
+// TestBuildUserPrompt_ReviewerCase_ZeroIssue_OmitsIssueLine guards against
+// rendering "issue #0" into reviewer prompts when the workload pipeline
+// omits the issue key (#1570).
+func TestBuildUserPrompt_ReviewerCase_ZeroIssue_OmitsIssueLine(t *testing.T) {
+	task := &foremanv1alpha1.AgenticTask{
+		Spec: foremanv1alpha1.AgenticTaskSpec{
+			Kind: foremanv1alpha1.AgenticTaskKindReview,
+			Payload: foremanv1alpha1.AgenticTaskPayload{
+				Repo:   "defilantech/LLMKube",
+				Issue:  0,
+				Branch: "foreman/custom-pipeline-branch",
+			},
+		},
+	}
+	got := buildUserPrompt(task)
+	if got == "" {
+		t.Fatal("reviewer user prompt must not be empty")
+	}
+	if strings.Contains(got, "#0") || strings.Contains(got, "issue: 0") {
+		t.Errorf("reviewer prompt must not mention issue #0 when issue is unset, got:\n%s", got)
+	}
+	for _, want := range []string{
+		"You are reviewing the branch foreman/custom-pipeline-branch of defilantech/LLMKube.",
+		"- repo: defilantech/LLMKube",
+		"- branch: foreman/custom-pipeline-branch",
+		"Step 1 of your system prompt",
+	} {
+		if !strings.Contains(got, want) {
+			t.Errorf("reviewer prompt missing %q in:\n%s", want, got)
+		}
+	}
+}
+
 // TestFileListsEqual_TypeShapes covers the two ways the model's
 // claim shows up in extra: as []string (the executor's own writes)
 // or as []any (the standard shape after a json.Unmarshal pass over


snowyukitty:
Before this is merged, a correction to my own description.

**`Fixes #1570` is wrong.** #1570 is "Reviewer returns GO without ever diffing the branch: verdicts uncorrelated with the code they approve". This PR does not address that — it stops `issue #0` and `- issue: 0` being rendered into the reviewer prompt when the dispatch payload carries no `issue`. That is a separate defect, not the bug #1570 reports.

#1588 ("rail detecting a GO from a reviewer that never diffed", merged 2026-08-18) uses `Refs #1570`, and #1570 is still open. The closing keyword here is in both this description and the commit message, so merging as written would close it. I wanted to correct that before merge so #1570 stays open.

The defect this PR does fix is still present on `main`: in `pkg/foreman/agent/executor_native.go`, the `AgenticTaskKindReview` branch of `buildUserPrompt` still emits both issue lines unconditionally, while `repoMapQuery` in the same file already uses the `if p.Issue > 0` guard.

Happy to reword the description and the commit message to `Refs #1570`, and to rebase and re-run the tests, if either would be useful.

Defilan (changes requested):
Thanks for this, and welcome. The code change is correct and I want it: guarding on `p.Issue > 0` before emitting the issue references is the right shape, the fallback wording reads well, and it comes with a test. My concerns are about what the PR claims and how it is filed, not about the change itself.

## Blocking: this does not fix #1570, and merging it would close that issue

#1570 reports that an in-process reviewer returns **GO without ever obtaining the diff of the branch it is reviewing**. Its Expected Behavior is explicit:

> A reviewer that never diffs the branch under review cannot return GO. Either the loop makes the diff unavoidable, or a rail fails the verdict when the transcript shows the branch was never diffed.

This PR does neither. After it merges, a reviewer can still read an unrelated file, never diff the branch, and return GO. The verdict stays uncorrelated with the code it approves, which is the whole of that bug.

I think I can see how this happened. Step 1 of #1570's reproduction says to dispatch a review step "whose `payload` carries `repo`, `branch`, `baseBranch` and `branchStrategy` but **no `issue` key**." That missing key is the *trigger condition* I used to reproduce the ungrounded review, not the defect. Rendering `issue #0` is a real and separate bug that you found while reading the same code.

Because the body says `Fixes #1570`, merging this auto-closes that issue. A severe correctness bug would vanish from the backlog while remaining entirely present in the code, and #1570 is arguably the strongest case in epic #1548. That is the one thing here I cannot merge past.

**Please change `Fixes #1570` to `Refs #1570`**, and open a separate issue for the `issue #0` rendering so the fix has its own tracking. Happy to review that issue quickly.

## Blocking: the PR template is required

This repository has a mandatory PR template at `.github/PULL_REQUEST_TEMPLATE.md`, and this description does not use it. None of the What / Why / How / Checklist sections are present and the checklist is absent, so I cannot see the state of the DCO sign-off, the local `make test` and `make lint` runs, or the documentation question.

Your Problem / Solution write-up is genuinely clearer than a lot of what the template produces, so please keep that content. It maps cleanly: Problem becomes Why, Solution becomes How. What you would be adding is the checklist and a one-line What.

While filling it in, please answer the AI-assistance item explicitly. `CONTRIBUTING.md` asks that if any part of a contribution was generated or substantially assisted by an AI tool, the PR body says which tool, what it produced, and what you verified before submitting. We welcome AI-assisted contributions and hold them to the same bar as any other; we just need the disclosure, and a human owning the review conversation. If no tool was involved, say that and we move on.

I am raising it because the scope mismatch above is the shape a tool produces when pointed at an issue: a correct, well-formed change to something adjacent to the reported defect, described as fixing it. That is a normal failure mode and not a criticism of the work, but it is exactly why the disclosure matters, since it tells a reviewer where to look hardest.

## Non-blocking, on the change itself

**Worth covering the same guard on the coder path, or saying why not.** This fixes the `AgenticTaskKindReview` branch of `buildUserPrompt`. If the coder branch renders issue references the same way, an issueless pipeline step hits the identical `#0` there. Either extend the guard or note in the PR that the coder path always has an issue and cannot reach it.

**One question about the test.** Does `TestBuildUserPrompt_ReviewerCase_ZeroIssue_OmitsIssueLine` fail if the guard is removed? Worth confirming rather than assuming: this codebase has shipped several tests that passed whether or not the code under test worked, and the cheap check is to delete the `if p.Issue > 0` guard, run the test, and see it go red. If it does, say so in the PR and I will take your word for it.

**Negative assertions are the easy ones to get wrong here.** If the test asserts the output does *not* contain `issue #0`, make sure it also asserts the fallback line it *should* contain. A test that only checks for an absence passes against an empty string too.

## To summarise

Two things to change before merge: `Fixes` to `Refs` with a separate issue for the real defect, and the PR template with the AI-assistance item answered. Then this is a good small fix and I will merge it.

Sorry to hand a first contribution a request for changes over bookkeeping rather than code. The code is fine. It is the issue link that would do damage.

Defilan:
Closing due to inactivity. If you wish to resubmit, please do. We're always looking for more contributors.

Issue #1447, opened 2026-08-08
[BUG] Reviewer scope-overlap makes test-coverage issues unwinnable (diff touches X.test.ts, issue names X.ts)


joryirving:
## Bug Description

The reviewer's scope-overlap demotion requires the diff to touch the files the issue names literally. For a "add tests for X" issue that is unsatisfiable: the issue names `X.ts`, the correct change creates `X.test.ts`, and the check reports the diff "touches none of them". The task then burns its full attempt budget and the frontier escalation, every time, and no PR is ever opened.

## Evidence

`misospace/dispatch#654` ("[P3] Add tests for high-risk untested modules"). The coder returned **GO twice** with typecheck and lint passing — 52 unit tests, then 2 more branch tests, 1104 lines across three new files. The reviewer demoted both:

```
demotionReason: scope drift: the issue names 6 file(s)
  (automation-sync.ts, groomer-lock.ts, resolve-actor.ts,
   src/lib/automation-sync.ts, src/lib/groomer/groomer-lock.ts, src/lib/resolve-actor.ts)
  and the diff touches none of them

issueAsk: Add unit tests covering the main execution paths and error conditions
          for these three modules.
```

The diff creates exactly the right files:

```
src/lib/automation-sync.test.ts      | 738 +++
src/lib/groomer/groomer-lock.test.ts | 218 +++
src/lib/resolve-actor.test.ts        | 148 +++
```

`issueAsk` is extracted correctly — it says "add unit tests". The check then rejects the work for not editing the modules under test.

## Two defects

**1. Test files are not mapped to the modules they cover.** A change to `foo.test.ts` satisfies an issue naming `foo.ts`. Without that, the entire class of test-coverage issues is unwinnable, and "add tests for X" is a very common audit finding — the same fleet has several open right now.

**2. The file list double-counts.** The issue names **three** modules; the check reports **six**. It mentions each once bare and once with its path (`automation-sync.ts` and `src/lib/automation-sync.ts`), and both are counted as distinct files. Cosmetic next to (1), but it inflates the denominator of any overlap ratio computed from that list.

## Expected

A diff of `X.test.ts` (or `X_test.go`, `test_X.py`, `X.spec.ts` — whatever the GateProfile's conventions imply) counts as touching `X` for scope-overlap purposes. Extracted file references should be deduplicated by resolved path.

## Why this is expensive

[...]

PR #1573, opened 2026-08-16
fix(foreman): map test files to their modules in scope-overlap, and dedupe issue file refs

Defilan:
## What

Makes the reviewer's scope-overlap rail recognise test files as touching the modules they cover, and stops the extracted issue-file list from double-counting the same file.

## Why

Refs #1447

The rail demotes a reviewer GO when the diff touches none of the files the issue names. That makes an entire class of issue unwinnable: an "add tests for X" issue names `X.ts`, the correct change creates `X.test.ts`, and the check reports the diff "touches none of them". The task then burns its full attempt budget and the frontier escalation every time, and no PR is ever opened.

The issue records a real instance: a coder returned GO twice with typecheck and lint passing, 1104 lines across three correct new test files, and was demoted both times.

Second, smaller defect: `extractIssuePathRefs` returned both the bare filename and the full path as separate entries, so an issue naming three modules reported six files. That inflates the denominator of any overlap ratio and makes the demotion message misleading.

## How

Two changes in `pkg/foreman/agent/scope_overlap.go`:

- **`testTargetsForPath`** maps a diff path to the module path(s) it could be testing: `X.test.ts` and `X.spec.ts` to `X.ts` (extension preserved, so `.tsx`/`.js`/`.jsx` work too), and `X_test.go` to `X.go`. Direct matches are unaffected: a diff touching `X.ts` still satisfies `X.ts`.
- **`dedupRefs` / `samePathRef`** collapse references that resolve to the same file, so `automation-sync.ts` and `src/lib/automation-sync.ts` count once.

`enforceReviewerScopeOverlap`'s signature and its call site in `executor_native.go` are unchanged, and no other rail was touched.

## Verification

Run on this branch, not inherited from an agent verdict:

- `go test ./pkg/foreman/agent/ -count=1` -> **ok, 29.6s**
- **Falsifiability proved.** Applying this branch's `scope_overlap_test.go` to unmodified `main` fails to build with `undefined: testTargetsForPath`, so the new tests cannot pass without the implementation.

Coverage includes the issue's concrete example (`TestEnforceReviewerScopeOverlap_Issue654ConcreteExample` with `automation-sync`, `groomer-lock` and `resolve-actor`), the dedup case, the Go `_test.go` mapping, and a regression guard (`TestEnforceReviewerScopeOverlap_RealDriftStillBitesAfterFix`) proving genuine scope drift is still demoted. That last one matters: the fix must not pass by disabling the rail.

## A reviewer NO-GO on this branch was wrong

Recorded rather than quietly dropped, because it is evidence for #1570.

The automated reviewer returned NO-GO twice, stating the changes "lack tests for the issue's concrete example files (automation-sync.ts, groomer-lock.ts, etc.)". The branch contains a test named `TestEnforceReviewerScopeOverlap_Issue654ConcreteExample` using exactly those files. The claim is false on its face, and it triggered a retry cycle that regenerated already-correct work.

This is the mirror image of the failure in #1570, where two reviews returned GO without ever reading the diff. Same root cause, opposite direction. Per #1552, an unsupported NO-GO destroys good work, and here it did.

The diff was verified by hand instead. Reviewers should judge the code, not this note.

## Checklist

- [x] Tests added/updated - 9 new cases, including the issue's concrete example and a drift regression guard
- [x] `make test` passes locally - `./pkg/foreman/agent/` suite
- [x] `make lint` passes locally - `go vet` clean
- [ ] All commits are signed off (`git commit -s`) per [DCO](https://developercertificate.org/) - **bot-only sign-off, see below**
- [x] Documentation updated (if user-facing change) - n/a, internal rail behaviour

## Sign-off

Commits carry `Signed-off-by: Foreman Bot`. Per CONTRIBUTING.md a human sign-off is required before merge:

```
git commit --amend -s --no-edit && git push --force-with-lease
```

Not added on the maintainer's behalf: the DCO attests that a person takes responsibility for the change.

*Assisted-by: Qwen3.8-27B via the Foreman harness, on two DGX Sparks behind one LLMKube InferenceService. Its automated reviewer returned NO-GO; that verdict was checked against the diff and found false (see above), so it was not acted on. Reviewed-by: Claude, which read the full diff, confirmed both defects are addressed, ran the package suite, and proved the tests fail against unmodified main. The maintainer scoped this batch, reviewed the findings in conversation (including the false NO-GO recorded above), and directed this PR to be opened. DCO sign-off is still outstanding, per the checklist.*

### DIFF
--- a/pkg/foreman/agent/scope_overlap.go
+++ b/pkg/foreman/agent/scope_overlap.go
@@ -120,7 +120,49 @@ func extractIssuePathRefs(body string, sourceExtensions []string) []string {
 			refs = append(refs, tok)
 		}
 	}
-	return refs
+	return dedupRefs(refs)
+}
+
+// dedupRefs collapses refs that resolve to the same file so the count of
+// issue-named files is not inflated (#1447). A bare filename that is a path
+// suffix of a fuller ref (automation-sync.ts vs src/lib/automation-sync.ts)
+// names the same module and must not be counted as two distinct files; the
+// more specific (longer) ref is kept. Refs that are not suffix-related (e.g.
+// a/foo.ts and b/foo.ts) are distinct and are all preserved.
+func dedupRefs(refs []string) []string {
+	out := make([]string, 0, len(refs))
+	for _, r := range refs {
+		replaced := false
+		for i, k := range out {
+			if samePathRef(k, r) {
+				// Prefer the more specific (longer) of the two.
+				if len(r) > len(k) {
+					out[i] = r
+				}
+				replaced = true
+				break
+			}
+		}
+		if !replaced {
+			out = append(out, r)
+		}
+	}
+	return out
+}
+
+// samePathRef reports whether two refs resolve to the same file: either they
+// are equal, or one is a path suffix of the other aligned on a path boundary
+// (a bare filename that ends a fuller path, or vice versa). The "/" boundary
+// guard keeps, e.g., x-automation-sync.ts from matching automation-sync.ts.
+func samePathRef(a, b string) bool {
+	if a == b {
+		return true
+	}
+	full, short := a, b
+	if len(full) < len(short) {
+		full, short = short, full
+	}
+	return len(short) < len(full) && strings.HasSuffix(full, "/"+short)
 }
 
 // hasSourceFile reports whether any path in paths ends with one of the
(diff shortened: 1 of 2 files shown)


joryirving (approved):
Approving. This is the right fix and the test-file mapping is the piece that makes the rail usable: `testTargetsForPath` folding `X.test.ts`, `X_test.go` and `test_X.py` back to their modules, indexed under both full path and basename, means an "add tests for X" issue finally matches the diff that adds the tests. That whole class of issue was unwinnable before, burning the full attempt budget and the frontier escalation every time.

`samePathRef` is careful in the way it needs to be. Anchoring the suffix match on `"/"` keeps `x-automation-sync.ts` from collapsing into `automation-sync.ts`, and `dedupRefs` preferring the longer ref means `src/lib/foo.ts` survives over a bare `foo.ts` rather than the other way round. Refs that are not suffix-related stay distinct, so `a/foo.ts` and `b/foo.ts` are still two files.

One gap worth knowing about, not a blocker since unrecognised paths return nil and just fall through to the old behaviour. The mapping assumes tests live beside their subject, because it preserves the directory and only rewrites the basename. That holds for Go, and usually for JS/TS, but not for languages with a parallel test tree. Elixir is the one that bites us: `test/pinchflat/boot/config_helpers_test.exs` covers `lib/pinchflat/boot/config_helpers.ex`, so both the directory and the extension differ, and `.exs` is not in the recognised set either. An "add tests" issue in an Elixir repo still hits the original problem. Same shape for Java's `src/test` vs `src/main` and Rust's top-level `tests/`. Worth a follow-up that lets a GateProfile declare its test-layout convention, in the way `SourceExtensions` already declares its source ones.

Issue #1568, opened 2026-08-15
[BUG] PR bodies are truncated to 280 bytes: the reviewer is asked for a PR description in a one-sentence field


joryirving:
## Bug Description

The reviewer is asked to write a full PR description into `submit_result.summary`, a field whose own schema says one sentence and whose handler truncates at 280 bytes. Every foreman-opened PR body is therefore cut off, usually mid-sentence, and the fragment is what a human sees.

The two contracts are directly incompatible.

`submit_result` advertises (`pkg/foreman/agent/tools/submit_result.go`):

```json
"summary": {"type": "string", "description": "One-sentence outcome summary (1-280 chars)."}
```

and enforces it:

```go
const MaxSubmitSummaryLen = 280
if len(a.Summary) > MaxSubmitSummaryLen {
    a.Summary = truncateRuneSafe(a.Summary, MaxSubmitSummaryLen)
}
```

The reviewer system prompt asks for a complete PR body in that same field: `## What`, `## Why`, `Fixes #N`, `## How`, an `Assisted-by:` line, and in the fork variant a seven-item checklist. That is several hundred bytes at minimum.

`openPullRequest` then renders whatever survived as the description verbatim (`pkg/foreman/agent/executor_native.go:1966`).

## Steps to Reproduce

1. Let any issue reach a reviewer GO so a PR is opened.
2. Read the PR body.

## Expected Behavior

The PR description is the document the reviewer was asked to write: complete, structured, grounded in the diff it just read.

## Actual Behavior

280 bytes of it, then `…`. Six recent foreman-opened PRs across four repositories, total body length including the `Fixes #N` and provenance footer:

```
miso-gallery#412              341 chars   no ellipsis (fit by luck)
alert-triage#53               323 chars   truncated
pinchflat#35                  302 chars   truncated
pinchflat#36                  363 chars   truncated
dispatch#770                  333 chars   truncated
foreman-dispatch-bridge#153   377 chars   truncated
```

Five of six truncated, all clustering at the cap. Two concrete examples, verbatim and complete:

> The new park_for_human implementation in bridge/main.py posts the escalation comment with @-mention and adds the needs-human label, matching the issue's core ask, but the behaviour change carries no test (tests only stub park_for_human with lambdas) and the comment omit…

> ## Verdict: Request changes — this branch will not merge against current origin/main
>
[...]

Issue #1541, opened 2026-08-14
[FEATURE] Foreman-opened PRs should honour the target repository's PR template


Defilan:
## Feature Description

When Foreman opens a pull request, it should honour the **target repository's**
PR template if that repository has one, instead of always emitting Foreman's own
fixed body shape.

## Problem Statement

> As a maintainer of any repository Foreman contributes to, I want agent-opened
> PRs to arrive in the shape my repo asks contributors for, so that reviewing an
> agent's PR costs no more than reviewing a human's.

`NativeAgentLoopExecutor.openPullRequest`
(`pkg/foreman/agent/executor_native.go:1926-1936`) builds every PR body the same
way regardless of destination:

```go
if s := strings.TrimSpace(summaryBody); s != "" {
    bodyB.WriteString(s)
    bodyB.WriteString("\n\n")
}
fmt.Fprintf(&bodyB, "Fixes #%d\n\n_Opened by foreman on review GO (workload %s)._",
    p.Issue, task.Labels["foreman.llmkube.dev/workload"])
```

The reviewer's summary, an issue link, a provenance line. Nothing else, for
every repo.

**Why this is invisible rather than merely imperfect:** GitHub applies a PR
template only in the browser flow. A PR created through the API must supply the
body itself, and an API client that never reads the template silently bypasses a
convention the repository relies on. Foreman creates PRs through the API, so a
repo can have a mandatory template, a CONTRIBUTING policy pointing at it, and
still receive agent PRs that skip both, with nothing failing to signal it.

The cost lands on the maintainer, who must either reconstruct the missing
context or hand-edit each agent PR. Repos with strict contribution policies are
exactly the ones most likely to adopt an agent, and exactly the ones this
inconveniences most.

## Proposed Solution

1. **Discover** the template in the workspace the coder already has checked out,
   following GitHub's own resolution order and locations:
   `.github/PULL_REQUEST_TEMPLATE.md`, `PULL_REQUEST_TEMPLATE.md`,
   `docs/PULL_REQUEST_TEMPLATE.md` (case-insensitive, `.txt` accepted), plus
   `.github/PULL_REQUEST_TEMPLATE/` for multi-template repos (pick the default
   or first deterministically).

2. **Fill it with the model that already read the diff.** The reviewer has read
   the full diff against the issue to reach GO and its summary is already
   diff-grounded (#1411). Handing it the discovered template and asking it to
   fill the sections is the only approach that generalises: Foreman cannot know
[...]

PR #1575, opened 2026-08-16
fix(foreman): let the reviewer author an uncapped PR body via extra.prBody

Defilan:
## What

Gives the PR description its own home. `openPullRequest` now prefers `extra["prBody"]` when present and non-empty, falling back to the 280-byte `summary` when it is absent.

## Why

Refs #1568

The reviewer is asked to write a full PR description into `submit_result.summary`, a field whose own schema says "One-sentence outcome summary (1-280 chars)" and whose handler enforces that:

```go
const MaxSubmitSummaryLen = 280
if len(a.Summary) > MaxSubmitSummaryLen {
    a.Summary = truncateRuneSafe(a.Summary, MaxSubmitSummaryLen)
}
```

`openPullRequest` then renders whatever survived as the description verbatim. Every foreman-opened PR body is therefore cut off, usually mid-sentence, and that fragment is what a human sees.

The two contracts are directly incompatible: the reviewer prompt asks for `## What`, `## Why`, `Fixes #N`, `## How` and a provenance line, which is several hundred bytes at minimum.

## How

Deliberately **not** a larger cap. `summary` is genuinely useful as a short outcome line for logs, status and audit records, and 280 is a reasonable bound for that. The problem was that a second, much larger artifact was being smuggled through the same field.

```go
body := summaryBody
if pb, ok := extra["prBody"].(string); ok && strings.TrimSpace(pb) != "" {
    body = pb
}
body = githubpr.PRBody(githubpr.FindTemplate(workspace), body, ...)
```

- `extra` already passes through `submit_result` unvalidated and untruncated, so this needs **no schema change** and no change to `MaxSubmitSummaryLen`.
- The fallback is what keeps every existing agent working unchanged, so it is required rather than a nicety.
- The chosen body still flows through `PRBody`, so the target repository's PR template (#1541) is still honoured.

The reviewer system prompt is **not** changed here. Telling the reviewer it is authoring a PR description is a separate concern, and bundling it would make this diff impossible to review against the issue. This change makes the capacity exist; a follow-up can use it.

## Verification

- `go test ./pkg/foreman/agent/ -count=1` -> **ok, 30.0s**
- **Falsifiability proved.** Applying this branch's `executor_pr_test.go` to unmodified `main` fails to build on the signature mismatch, so the new tests cannot pass without the implementation.

Cases: `prBody` present wins over summary; absent falls back; empty string falls back; whitespace-only falls back; wrong type falls back; nil extra map does not panic; and a body well over 280 bytes survives intact, which is the point of the change.

## Checklist

- [x] Tests added/updated - 6 cases including the >280-byte survival case and the nil-map guard
- [x] `make test` passes locally - `./pkg/foreman/agent/` suite
- [x] `make lint` passes locally - `go vet` clean; `lll` line-length violations fixed
- [ ] All commits are signed off (`git commit -s`) per [DCO](https://developercertificate.org/) - **bot-only sign-off, see below**
- [x] Documentation updated (if user-facing change) - n/a, the behaviour change is described here

## Sign-off

Commits carry `Signed-off-by: Foreman Bot`. Per CONTRIBUTING.md a human sign-off is required before merge:

```
git commit --amend -s --no-edit && git push --force-with-lease
```

*Assisted-by: Qwen3.8-27B via the Foreman harness. Reviewed-by: Claude, which read the full diff, confirmed the fallback covers empty/whitespace/wrong-type/nil-map, verified the body still routes through PRBody so #1541's template support survives, ran the package suite, and proved the tests fail against unmodified main. The maintainer scoped this batch, reviewed the findings in conversation (including the false NO-GO recorded above), and directed this PR to be opened. DCO sign-off is still outstanding, per the checklist.*

### DIFF
--- a/config/foreman/agents/claude-sonnet-cloud-reviewer.yaml
+++ b/config/foreman/agents/claude-sonnet-cloud-reviewer.yaml
@@ -452,6 +452,28 @@ spec:
                                  operator to inspect; chronic drift here is
                                  a model-quality signal.
 
+          - `prBody`          :: the full pull-request description, in
+                                 Markdown, with no length limit. Set it
+                                 whenever you APPROVE; it is ignored
+                                 otherwise. Write for someone who will read the
+                                 PR WITHOUT reading the diff: what changed, why
+                                 it was needed, and what a maintainer should
+                                 know. You have just read the diff and the
+                                 issue, so you are the best-placed step to
+                                 write it.
+
+                                 `summary` is capped at 280 characters and is
+                                 the wrong home for this. When `prBody` is
+                                 absent the executor falls back to `summary`,
+                                 so the PR description becomes a truncated
+                                 fragment of a single sentence.
+
+                                 This does not make you the PR author: the
+                                 executor still opens the PR and applies the
+                                 target repository's template. You supply the
+                                 text, you do not perform the action the Hard
+                                 rules below forbid.
+
           Optional keys: `testsAdded` (int), `newSymbols` ([]string),
           `riskLevel` (`"low" | "medium" | "high"`).
 
--- a/config/foreman/agents/devstral-24b-reviewer.yaml
+++ b/config/foreman/agents/devstral-24b-reviewer.yaml
@@ -398,6 +398,28 @@ spec:
                                  operator to inspect; chronic drift here is
                                  a model-quality signal.
 
+          - `prBody`          :: the full pull-request description, in
+                                 Markdown, with no length limit. Set it
+                                 whenever you APPROVE; it is ignored
+                                 otherwise. Write for someone who will read the
+                                 PR WITHOUT reading the diff: what changed, why
+                                 it was needed, and what a maintainer should
+                                 know. You have just read the diff and the
+                                 issue, so you are the best-placed step to
+                                 write it.
+
+                                 `summary` is capped at 280 characters and is
+                                 the wrong home for this. When `prBody` is
+                                 absent the executor falls back to `summary`,
+                                 so the PR description becomes a truncated
+                                 fragment of a single sentence.
+
+                                 This does not make you the PR author: the
+                                 executor still opens the PR and applies the
+                                 target repository's template. You supply the
+                                 text, you do not perform the action the Hard
+                                 rules below forbid.
+
           Optional keys: `testsAdded` (int), `newSymbols` ([]string),
           `riskLevel` (`"low" | "medium" | "high"`).
 
--- a/config/foreman/agents/qwen36-35b-a3b-reviewer.yaml
+++ b/config/foreman/agents/qwen36-35b-a3b-reviewer.yaml
@@ -433,6 +433,28 @@ spec:
                                  operator to inspect; chronic drift here is
                                  a model-quality signal.
 
+          - `prBody`          :: the full pull-request description, in
+                                 Markdown, with no length limit. Set it
+                                 whenever you APPROVE; it is ignored
+                                 otherwise. Write for someone who will read the
+                                 PR WITHOUT reading the diff: what changed, why
+                                 it was needed, and what a maintainer should
+                                 know. You have just read the diff and the
+                                 issue, so you are the best-placed step to
+                                 write it.
+
+                                 `summary` is capped at 280 characters and is
+                                 the wrong home for this. When `prBody` is
+                                 absent the executor falls back to `summary`,
+                                 so the PR description becomes a truncated
+                                 fragment of a single sentence.
+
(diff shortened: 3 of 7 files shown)


joryirving (changes requested):
Request changes, though the code itself is right and I want this landed. `openPullRequest` taking `extra` and preferring `extra["prBody"]` with a trimmed non-empty check, falling back to `summaryBody`, and still routing through `PRBody` so the #1541 template path is preserved — that is exactly the shape I asked for in #1568, and the fallback keeps existing agents working.

The problem is that nothing produces the field. `git grep prBody` across main returns only this PR's own changes. This repo ships the canonical reviewer prompt at `config/foreman/system-prompts/reviewer.md`, and line 300 still documents the terminal payload as `summary :: one-sentence outcome, 280 characters or fewer` with no mention of `prBody`. `submit_result`'s schema describes `extra` only as "Structured extra fields the executor may surface". So a stock reviewer never sets `prBody`, `openPullRequest` falls back to `summaryBody`, and the uncapped body in the PR title never happens for anyone who has not hand-written their own prompt.

It is still an improvement — one clean sentence beats one truncated mid-word — but the feature ships unreachable. Could you add the `prBody` bullet to `reviewer.md`'s terminal-payload section and mention it in the `extra` description in `submit_result.go`? That is the half that makes the plumbing do something. If you would rather do it as an immediate follow-up I will not hold the PR, but I do not want it to land and be forgotten, because from the outside it will look fixed.

Defilan:
Verified all three before changing anything, and they hold: `git grep prBody` on main matches only this PR, reviewer.md line 300 documents the payload as `summary :: one-sentence outcome, 280 characters or fewer` with no prBody, and submit_result describes extra only as "structured extra fields the executor may surface." Nothing produces the field, so the fallback always wins and the uncapped path never runs.

Fixed here rather than as a follow-up, since you're right that from the outside it looks fixed. 0e814251:

- reviewer.md documents prBody in the extra key list: set on APPROVE, Markdown, no length limit, written for someone who will read the PR without reading the diff. It also says why summary is the wrong home, so the next person doesn't re-litigate it.
- submit_result's extra description names prBody, so a model reading the tool definition alone can find it without the system prompt.
- `make sync-reviewer-prompts` propagated it into the three reviewer Agent manifests, and `check-reviewer-prompts` reports no drift.

One thing I watched for: the prompt has a Hard rule saying don't call `gh pr create`, PR authorship is a separate pipeline step. Asking the same agent for a PR body reads like a contradiction, so the new entry states that it supplies the text while the executor still opens the PR and applies the template. Didn't want a reviewer to hit both rules and stall.

Worth flagging since it affects when you'll actually see this working: it only takes effect for reviewers whose systemPrompt comes from those manifests. The nemotron-reviewer Agent running on my Strix box was created by hand and has no systemPrompt field, so it won't pick this up until I recreate it from a synced manifest. The plumbing is reachable now, but my own fleet won't exercise it until I do that.

joryirving (approved):
Approving. Everything I asked for landed, and doing it here rather than as a follow-up was the right call.

`reviewer.md` now documents `prBody` in the extra key list, `submit_result.go` names it in the `extra` description so a model reading only the tool definition can find it, and the sync propagated it to all three reviewer manifests with no drift. The part I did not think to ask for is the better half: explaining in the prompt why `summary` is the wrong home, so the next person does not re-litigate it, and stating that the reviewer supplies the text while the executor still opens the PR. That second one is a real catch. A reviewer reading the Hard rule against `gh pr create` and then being asked for a PR body has a genuine contradiction to resolve, and resolving it by stalling would have been a plausible failure I would not have predicted from the diff.

CI is red on lint, which I assume you have seen: `pkg/foreman/agent/tools/submit_result.go:127` is 286 characters against the 120 limit, from the description string you just added. Substance is unaffected, it wants the string split.

The note about this only reaching reviewers whose `systemPrompt` comes from the manifests is worth more than it might look. Mine are in the same position as your Strix box: `reviewer` and `reviewer-fork` are hand-written with inline prompts, so they will not pick this up either and I have my own edit to make before any of it runs here. Anyone who hand-wrote an Agent gets the plumbing and none of the instruction, which is the same shape as the problem this PR fixes, one layer out. Might be worth a line in the release notes rather than leaving people to notice that their PR bodies are still truncated.

For what it is worth, an external conformance bench measured the contract this fixes while this PR was open: the current shape caps a GO at 0.929 because the structured-body requirement and the 280 character cap cannot both be satisfied, and both failure directions cost the same, which is the signature of an unsatisfiable rule rather than a hard one. After this lands it reaches 1.000. So the truncation was not theoretical and it was costing every approval.

Issue #1605, opened 2026-08-19
[BUG] Scope-overlap rail degrades open silently: an unchecked verdict is indistinguishable from a checked one


Defilan:
## Bug Description

The scope-overlap rail degrades open silently. When it cannot run, it returns the model's verdict unchanged and records nothing, so a verdict that was **never checked** is indistinguishable afterwards from one that **was checked and passed**.

This is the sibling of the defect #1576 fixed in `grounded_finding_gate.go`. That PR applied item 2 of the #1570 diagnosis ("make the degrade-open observable") to the grounding rails only. `scope_overlap.go` was not touched and still vanishes without a trace.

## Four silent fail-open paths

**One at the call site** (`executor_native.go:879`):

```go
if reviewDiffErr == nil {
    verdict = enforceReviewerScopeOverlap(...)
} else {
    log.Info("reviewer scope: ground-truth diff unavailable; skipping scope check", "err", ...)
}
```

A log line is not a record. Nothing lands in `extra`, so nothing downstream can tell the check was skipped.

**Three inside the function** (`scope_overlap.go:294`):

```go
if extra == nil || issueBody == "" || len(diffFiles) == 0 {
    return verdict
}
```

- `issueBody == ""` fires when the reviewer never called `fetch_issue`, since the body comes from `extractFetchIssueBody(loopRes.Transcript)`, not from the API.
- `len(diffFiles) == 0` fires when the diff succeeded but returned nothing, which is the same starvation condition described in #1570.
- `extra == nil` cannot record by definition, but the caller can.

Every one of these is the shape #1570 named: **the safety net is disabled by the same condition that creates the hazard.**

## Evidence

Replaying the real issue bodies against the real shipped diffs of two merged PRs:

```
#1570 / PR #1588
  refs    = [Dockerfile, pkg/foreman/agent/unverified_claim_gate.go,
             pkg/foreman/agent/executor_native.go]
  matched = []          drift = true    verdict -> NO-GO

#1549 / PR #1589
  refs    = [pkg/foreman/agent/summary_grounding.go,
             pkg/foreman/agent/grounding/groundtruth.go,
             internal/foreman/controller/workload_controller.go]
  matched = []          drift = true    verdict -> NO-GO
```

Both PRs opened and merged on a GO. Since the rail demotes to NO-GO when given the real diff, it cannot have run with the real diff on either task. One of the fail-open paths above was taken, silently, and no annotation survives to say which.

[...]

Issue #1603, opened 2026-08-19
[FEATURE] Dead-code guard: fail CI when production code is referenced only by its own tests


Defilan:
## Feature Description

Fail CI when production code is referenced only by its own tests.

`golangci-lint`'s `unused` linter already detects this. It never gets the chance, because the main `.golangci.yml` run analyses test files, so a symbol that only its own tests call counts as used. Running the same linter with test files dropped from the graph reports it.

## Problem Statement

> As a maintainer, I want a rail that is merged but never called to fail CI, so that "tests pass, lint clean, gate green" cannot mean "this code does nothing."

This is not hypothetical. **Nine files in `pkg/foreman/agent` are on `main` right now with full unit coverage and zero production callers**: `cross_stage.go`, `deleted_reference_gate.go`, `issue_clauses.go`, `mcp_audit.go`, `mutation_check.go`, `reviewer_diff_gate.go`, `staleness_check.go`, `test_layout.go`, and `workclass.go`.

Every one passed unit tests, `make lint`, and the gate. The tests are what kept `unused` quiet: a well-tested dead function is invisible to every check we have.

Six of the nine are children of #1548, "Make agent verdicts verifiable." The harness's own fix for shipping functionally inert work shipped functionally inert. #1391 already names this failure mode from the model side ("5/5 GO verdicts, 0/5 complete fixes, 1/5 functionally inert"); this issue mechanises the detection so it does not depend on a reviewer noticing.

## Proposed Solution

A second lint config that enables only `unused`, with test files excluded from the graph:

```yaml
# .golangci-deadcode.yml
version: "2"
run:
  tests: false   # test-only references must not count as uses
linters:
  default: none
  enable:
    - unused
```

Exposed as `make lint-deadcode` and run as a blocking step in the lint workflow, reusing the pinned `golangci-lint-action` and version.

**Verification.** The guard must be proven against the pattern, not argued for. Injecting a new function plus a passing test and no caller gives:

| Check | Result |
|---|---|
| `go test ./pkg/foreman/agent/` | pass |
| `make lint` | 0 issues (blind) |
| `make lint-deadcode` | reported |

**Suppression registers.** Adopting this blocking on day one needs the known cases recorded. Two registers live in the config, both of which should shrink to empty:

- **Register A**: the nine merged-but-unwired files, each tagged with its tracking issue. The entry is deleted by the PR that wires the rail up, because that deletion is the proof the code is reachable.
[...]

Issue #1581, opened 2026-08-18
[BUG] MCP tool-name rename in #1527 left the coder grounding rail and system prompt on the old `mcp/` prefix


Defilan:
## Bug Description

#1527 renamed MCP tool names from `mcp/<server>/<tool>` to `mcp__<server>__<tool>` (correct: the OAI tool-call spec constrains `function.name` to `^[a-zA-Z0-9_-]+$`). Two consumers of that name were not updated, so both have been silently wrong since it merged.

**1. The coder grounding rail matches a prefix nothing produces.**

`pkg/foreman/agent/coder_grounding_gate.go:83` still tests `strings.HasPrefix(name, "mcp/")`. Every MCP tool result now arrives as `mcp__...`, so `context7Evidence` returns an empty corpus, `applyCoderGroundingRail` returns early at its `len(evidence) == 0` guard, and the rail records nothing. It is wired live on the coder path at `pkg/foreman/agent/executor_native.go:1005`.

The rail is advisory (it writes `Extra["coderGroundingViolations"]` and never changes the verdict), so no verdict was corrupted. The signal is simply gone: a coder that queries context7 and then writes an identifier absent from the retrieved docs is no longer flagged.

**2. The coder system prompt names tools that cannot dispatch.**

`config/foreman/system-prompts/coder.md` (lines 59-60, 104-105) instructs the coder to call `mcp/perplexity/perplexity_ask`, `perplexity_search` and `mcp/context7/*`. Those names no longer exist in the registry, so a model that follows the prompt literally emits a tool call that cannot be dispatched.

## Steps to Reproduce

Feed `context7Evidence` the name the registry actually builds:

```go
tr := []oai.Message{
    {Role: oai.RoleAssistant, ToolCalls: []oai.ToolCall{
        {ID: "c1", Type: "function", Function: oai.ToolCallFunction{Name: "mcp__context7__query-docs"}}}},
    {Role: oai.RoleTool, ToolCallID: "c1", Name: "mcp__context7__query-docs",
        Content: "prometheus_http_requests_total"},
}
context7Evidence(tr)
```

Observed on `main` (827cad8c):

```
tool name "mcp__context7__query-docs"  -> evidence items: 0
tool name "mcp/context7/query-docs"    -> evidence items: 1
```

## Expected Behavior

`context7Evidence` collects every MCP tool result, and the system prompt names tools the registry can dispatch.

## Actual Behavior

[...]

PR #1606, opened 2026-08-19
fix(foreman): record why a reviewer rail could not run

Defilan:
## What

Make every reviewer rail record when it could not run, and why, through one shared `railsSkipped` key.

## Why

Refs #1605

A single `repo.DiffNameOnly` call at `executor_native.go:848` feeds **four** rails:

| Line | Rail | Before |
|---|---|---|
| 855 | `runEmptyClaimRail` | silent |
| 865 | `enforceReviewerGroundedFindings` | recorded (#1576) |
| 873 | `enforceReviewerVerdictFromFindings` | silent |
| 880 | `enforceReviewerScopeOverlap` | silent |

When that call errors or returns nothing, all four lose their input simultaneously. This is the mechanism #1570 named, *"the safety net is disabled by the same condition that creates the hazard."* That diagnosis described two rails. It is four, and #1576 instrumented one of them.

**Two fail in the dangerous direction.** `verdict-from-findings` promotes GO to NO-GO when a grounded blocking finding exists, so without the diff **a GO carrying a real blocker stands and opens a PR**. `empty-claim` is the #1552 rail that remaps unsupported emptiness claims to ERROR, so the rail built because *an unverifiable claim should not be a NO-GO* was itself failing open on exactly that input.

**Evidence.** Replaying two merged PRs' real issue bodies against their real shipped diffs:

```
#1570 / PR #1588   refs=[..., executor_native.go]        matched=[]  -> NO-GO
#1549 / PR #1589   refs=[..., workload_controller.go]    matched=[]  -> NO-GO
```

Both opened and merged on a GO, so the rail cannot have run with the real diff, and nothing survived to say so. Those two are among the nine rails now on `main` with no production caller (#1603).

## How

One shared key, `railsSkipped`, holding `"<rail>: <reason>"` entries, written by `recordRailSkipped`. One key rather than a marker per rail, because the operational question is *"did anything fail to run on this verdict?"* and answering it should not require knowing every rail's name. Repeated entries collapse; a nil map is tolerated.

`groundingUnavailable` is preserved unchanged, since #1576's tests assert it; the grounded rail now also joins the shared list so the list is not misleading by omission.

**No verdict changes anywhere.** Demoting because a check could not run destroys good work, which is the #1552 lesson observed live on `wl-1447-review-1447-0` where a false NO-GO burned a coder slot regenerating correct work. Whether an unverifiable review should return `INCOMPLETE` is item 3 of the #1570 diagnosis and deserves its own decision. Item 4, the diff-acquisition bug underneath, is still open and unticketed; this is the backstop, not the cause.

## Scope note

This PR originally covered scope-overlap only. An audit of every rail found the same defect in two more fed by the same call, so they were folded in rather than split across three PRs with two sitting unfixed. Findings deliberately **not** included here, worth their own issue: `enforceReviewerIssueAsk` returns silently when `issueAskVerified` is absent; `applyCoderGroundingRail` returns silently on empty evidence (this is what let it sit dark for weeks after the #1527 rename, until #1581); and the three env kill-switches disable rails with zero annotation.

## Verification

- `go test ./pkg/foreman/agent/... -count=1` → ok, agent package 53.2s, all subpackages
- `go build ./...` → ok
- `GOOS=linux golangci-lint run ./pkg/foreman/agent/...` → 0 issues, fresh cache
- `rail_skip.go` passes the dead-code guard from #1604 with 0 findings, so this change does not add to the problem it documents
- **Falsifiability proved.** With the constants present but no rail wired, the assertions fail:
  ```
  --- FAIL: TestRailSkipped_VerdictFromFindingsRecordsMissingDiff  want verdict-from-findings recorded, extra=map[]
  --- FAIL: TestRailSkipped_EmptyClaimRecordsMissingDiff           want empty-claim recorded, extra=map[]
  --- FAIL: TestRailSkipped_GroundedFindingJoinsSharedList         want grounded-finding in the shared list
  ```

Twelve cases across two files: each rail records its missing input; each rail that **ran** is unmarked (four of the cases exist solely to hold that line, since a marker on every review is worthless); `groundingUnavailable` survives; entries dedupe; nil `extra` does not panic.

## Checklist

- [x] Tests added/updated - 12 cases, watched fail before implementing
- [x] `make test` passes locally - `./pkg/foreman/agent/...`
- [x] `make lint` passes locally - 0 issues, fresh lint cache
- [x] Documentation updated (if user-facing change) - n/a, internal rail annotations

`Assisted-by: Claude Opus 5` (traced the fail-open while reviewing #1447, audited every rail to find the other two, replayed the merged PRs against their real diffs to establish the rails would have demoted them, wrote the tests and the fix).

### DIFF
--- a/pkg/foreman/agent/empty_claim_gate.go
+++ b/pkg/foreman/agent/empty_claim_gate.go
@@ -142,7 +142,15 @@ func enforceReviewerEmptyClaim(
 ) (foremanv1alpha1.AgenticTaskVerdict, foremanv1alpha1.AgenticTaskFailureReason) {
 	if emptyClaimDisabled() ||
 		verdict != foremanv1alpha1.AgenticTaskVerdictNoGo ||
-		extra == nil || changedLines == nil {
+		extra == nil {
+		return verdict, ""
+	}
+	// Without the diff this rail cannot separate an unsupported emptiness
+	// claim from a supported one, so the NO-GO it exists to remap stands.
+	// Record it (#1605): the rail built because an unverifiable claim should
+	// not be a NO-GO must not itself fail open silently.
+	if changedLines == nil {
+		recordRailSkipped(extra, railEmptyClaim, skipReasonNoDiff)
 		return verdict, ""
 	}
 	if !assertsEmptyBranch(emptyClaimTexts(summary, extra)...) {
--- a/pkg/foreman/agent/executor_native.go
+++ b/pkg/foreman/agent/executor_native.go
@@ -884,6 +884,10 @@ func (e *NativeAgentLoopExecutor) runLLMPath(
 				scopeDriftDetected, _ = loopRes.Terminal.Extra["scopeDriftDetected"].(bool)
 				scopeMatched, _ = loopRes.Terminal.Extra["scopeMatched"].([]string)
 			} else {
+				// A log line is not a record (#1605). Mark the skip in extra so a
+				// verdict produced without the scope check is distinguishable
+				// afterwards from one that earned it.
+				recordRailSkipped(loopRes.Terminal.Extra, railScopeOverlap, skipReasonNoDiff)
 				log.Info("reviewer scope: ground-truth diff unavailable; skipping scope check",
 					"err", reviewDiffErr.Error())
 			}
--- a/pkg/foreman/agent/grounded_finding_gate.go
+++ b/pkg/foreman/agent/grounded_finding_gate.go
@@ -164,6 +164,7 @@ func enforceReviewerGroundedFindings(
 	// Silently returning the model's verdict is what made #1570 hard to see.
 	if changedLines == nil {
 		extra["groundingUnavailable"] = true
+		recordRailSkipped(extra, railGroundedFinding, skipReasonNoDiff)
 		extra["groundingUnavailableReason"] =
 			"branch diff unavailable; grounded-finding rail could not evaluate this NO-GO"
 		log.Info("reviewer grounded-finding: branch diff unavailable; verdict left unchecked",
--- a/pkg/foreman/agent/rail_skip.go
+++ b/pkg/foreman/agent/rail_skip.go
@@ -1,0 +1,52 @@
+package agent
+
+// Rail-skip observability (#1605).
+//
+// A single `repo.DiffNameOnly` call feeds four reviewer rails. When it fails
+// or returns nothing, every one of them loses its input at once, which is the
+// mechanism #1570 named: the safety net is disabled by the same condition that
+// creates the hazard. Two of the four already returned the model's verdict
+// silently, so a verdict nothing checked was indistinguishable afterwards from
+// one that earned it.
+//
+// These helpers make that visible. They never change a verdict: demoting
+// because a check could not run destroys good work, which is the #1552 lesson.
+
+// railsSkippedKey is the single extra key recording every rail that could not
+// run on a task, as "<rail>: <reason>" entries. One key rather than a marker
+// per rail, because the operational question is "did anything fail to run on
+// this verdict?" and answering it should not require knowing every rail's name.
+const railsSkippedKey = "railsSkipped"
+
+// Rail names used in railsSkipped entries.
+const (
+	railScopeOverlap        = "scope-overlap"
+	railVerdictFromFindings = "verdict-from-findings"
+	railEmptyClaim          = "empty-claim"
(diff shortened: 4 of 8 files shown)


joryirving (approved):
Approve: the split is correct, the invariant that a check which ran is never marked is genuinely held by two of the tests, and the mutation check confirms they are load-bearing. One gap worth closing, ideally here.

`len(refs) == 0` at `pkg/foreman/agent/scope_overlap.go:331` still returns unrecorded. That is the same class as the three cases this PR now records: the issue named no extractable paths, so the rail could not answer, and nothing survives to say so. By this PR's own argument a verdict produced through that branch is indistinguishable afterwards from one that earned it. It is also likely the most frequent of the four, since it fires for any issue that cites no file paths at all, which hand-written issues routinely do not. Four lines and a `scopeSkipNoPathRefs` constant would close it, and without it the claim that the degrade-open is now observable is only three quarters true.

Two smaller things. `scopeSkipDiffError` is referenced only at its declaration and at `executor_native.go:890`, so the one constant covering the call-site path has no test behind it. I understand why, since it lives inside `runLLMPath`, but it is worth knowing that the path you added at the call site is the one the suite does not exercise. Separately, the two branches that detect drift and then decline to demote, `!hasSourceFile` at line 66 of the function and the already-non-GO case just below it, remain log-only. They do leave `scopeDriftDetected` behind so they are not silent in the way the four you fixed were, but reading `extra` you cannot tell which of the two happened. Same "a log line is not a record" reasoning, lower stakes.

Worth flagging that this does not touch the failure I hit live tonight, in case it reads that way. On `misospace/pr-reviewer-action#438` the rail ran and answered: `scopeDriftDetected: true` with `scopeMatched: []`, because the issue named the module under test and the diff correctly edited a differently-named existing test file. That is #1447 territory rather than a degrade-open, so the vouch was unavailable for a real reason rather than a silent one.

Verified on the branch rather than from the diff: `go build ./...` clean, the five new tests pass, and neutering `recordScopeCheckSkipped` fails two of them. I did not run the full agent package to completion, it exceeded my timeout at around seven minutes, so I am taking your 48s figure on trust there.

Defilan:
Both gaps closed in 1f2f575.

**`len(refs) == 0`** now records `scope-overlap: no-path-refs-in-issue`. You are right that it is the same class, and the frequency argument is what decided it: `pathRefExtensions` is deliberately small and non-Go languages only get their extensions via `GateProfile.sourceExtensions`, so that branch fires for a large share of real issues. Leaving it out did make the observable claim three quarters true.

**The two decline branches** now set `scopeDriftNotDemoted` to either `diff-has-no-source-file` or `verdict-already-non-go`. I deliberately did not route those through `recordRailSkipped`. The rail ran and answered in both cases, so filing them as skips would conflate "could not check" with "checked, and chose not to act on the answer". Separate key, same "a log line is not a record" reasoning you applied.

**On the untested call-site constant, you are right, and it stays untested here.** The clean fix is extracting the diff-available branch into something testable, but that block is also where #1607 threads the resolved `TestLayout` through, and the two PRs already collide on those exact lines. Restructuring it in this PR makes that conflict materially worse for no behavioural gain. Worth doing once one of them lands and the other rebases.

**One thing to re-check against the right diff.** Your notes reference `recordScopeCheckSkipped`, `scopeSkipDiffError`, and "the five new tests". None of those exist on `9b59f8a6`, the commit the review is attached to: it has `recordRailSkipped` (11 uses), `skipReasonNoDiff` (16), and 11 test functions across two files. I force-pushed the audit expansion shortly before you submitted, so I think you read the pre-fold diff and GitHub re-attached the review to the new head. Everything substantive in your review survived that rename and I have treated it as valid, but the expansion is worth a look: the same defect turned out to be in `enforceReviewerVerdictFromFindings` and `enforceReviewerEmptyClaim` too, both fed by the same `repo.DiffNameOnly` call, so it is four rails from one failure rather than one.

**On `misospace/pr-reviewer-action#438`.** Agreed that is #1447 territory and not a degrade-open, and it is worth separating clearly: the rail answered honestly there. It is also a case neither #1573 nor #1607 can fix. Both are **name-based**: they fold `X.test.ts` or `src/test/java/FooTest.java` back to a subject derived from the filename. A diff that correctly edits an existing test file whose name does not derive from the module under test is invisible to any amount of name mapping. Closing that needs a content signal, something like "this test file imports or references the named module", which is a different mechanism and a different issue. If you have the workload record handy I will file it with your case as the reproduction, since a real one beats a constructed example.

Issue #1579, opened 2026-08-16
[FEATURE] Let GateProfile declare a test-layout convention for parallel test trees


Defilan:
## Feature Description

Let a `GateProfile` declare its repository's test-layout convention, the way it already declares `SourceExtensions`, so the scope-overlap rail can map a test file back to its subject in repositories where tests do not sit beside the code they cover.

## Problem Statement

> As someone queueing "add tests for X" work against a non-Go repository, I want the scope-overlap rail to recognise my project's test layout, so the same class of issue is not unwinnable there.

#1573 fixed the common case: `testTargetsForPath` folds `X.test.ts`, `X_test.go` and `test_X.py` back to their modules, so a diff that adds `X.test.ts` now satisfies an issue naming `X.ts`. Before that, every "add tests for X" issue burned its full attempt budget and the frontier escalation.

**It assumes tests live beside their subject.** The implementation takes `path.Dir(p)` and rewrites only the basename:

```go
func testTargetsForPath(p string) []string {
	base := path.Base(p)
	ext := path.Ext(base)
	...
	dir := path.Dir(p)
```

That holds for Go, and usually for JS/TS. It does not hold for languages with a parallel test tree, where both the directory and the extension differ:

| language | test file | subject |
|---|---|---|
| Elixir | `test/pinchflat/boot/config_helpers_test.exs` | `lib/pinchflat/boot/config_helpers.ex` |
| Java | `src/test/java/com/x/FooTest.java` | `src/main/java/com/x/Foo.java` |
| Rust | `tests/integration.rs` | `src/lib.rs` |

`.exs` is also absent from the recognised extension set entirely (`grep -c exs` on `scope_overlap.go` returns 0), so the Elixir case fails on both axes.

Not a regression: unrecognised paths return nil and fall through to the previous behaviour. The consequence is simply that an "add tests" issue in one of these repositories still hits the original #1447 problem.

## Proposed Solution

`GateProfile` already declares what counts as a source file for a repository. Extend the same idea to where tests live. Roughly:

```yaml
gateProfile:
  sourceExtensions: [".ex", ".exs"]
  testLayout:
    # A test path matching `pattern` maps to a subject built from `subject`.
    - pattern: "test/(?P<rest>.*)_test\\.exs"
      subject: "lib/${rest}.ex"
```

Whatever the exact shape, the properties that matter:

- **Directory rewriting, not just basename rewriting.** This is the part the current helper cannot express.
[...]

Issue #1593, opened 2026-08-18
[BUG] Helm install: Controller manager attempts to list Jobs at cluster scope but does not have permission


rouke-broersma:
## Bug Description

2026-08-18T07:29:14Z	ERROR	controller-runtime.cache.UnhandledError	Failed to watch	{"reflector": "pkg/mod/k8s.io/client-go@v0.36.3/tools/cache/reflector.go:343", "type": "*v1.Job", "error": "failed to list *v1.Job: jobs.batch is forbidden: User \"system:serviceaccount:llmkube:llmkube-controller-manager\" cannot list resource \"jobs\" in API group \"batch\" at the cluster scope"}

## Steps to Reproduce

1. Deploy model with prefetch enabled

## Expected Behavior

Prefetch job gets created

## Actual Behavior

Controller Manager has permission errors on listing job cluster-wide

## Environment

**LLMKube Version:**
```bash
v0.9.18
```

**Kubernetes Version:**
```bash
1.36.3
```

**Cluster Type:**
- [x] Other: Talos

## Logs

**Controller Logs:**
```bash
2026-08-18T07:24:46Z	ERROR	controller-runtime.cache.UnhandledError	Failed to watch	{"reflector": "pkg/mod/k8s.io/client-go@v0.36.3/tools/cache/reflector.go:343", "type": "*v1.Job", "error": "failed to list *v1.Job: jobs.batch is forbidden: User \"system:serviceaccount:llmkube:llmkube-controller-manager\" cannot list resource \"jobs\" in API group \"batch\" at the cluster scope"}
k8s.io/apimachinery/pkg/util/runtime.logError
	/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.36.3/pkg/util/runtime/runtime.go:252
k8s.io/apimachinery/pkg/util/runtime.handleError
	/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.36.3/pkg/util/runtime/runtime.go:243
k8s.io/apimachinery/pkg/util/runtime.HandleErrorWithContext
	/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.36.3/pkg/util/runtime/runtime.go:229
k8s.io/client-go/tools/cache.DefaultWatchErrorHandler
	/home/runner/go/pkg/mod/k8s.io/client-go@v0.36.3/tools/cache/reflector.go:227
k8s.io/client-go/tools/cache.(*Reflector).RunWithContext.func1
[...]

PR #1607, opened 2026-08-19
feat(foreman): let GateProfile declare a test-layout convention

### LINKED ISSUE #1447: [BUG] Reviewer scope-overlap makes test-coverage issues unwinnable (diff touches X.test.ts, issue names X.ts)

Defilan:
## What

Let a `GateProfile` declare where its repository's tests live, so the reviewer's scope-overlap rail can fold a test file back to the module it covers when tests are not beside the code.

## Why

Refs #1579, Refs #1447

#1573 fixed #1447 for conventions that keep the subject's stem behind a dot or underscore marker. Five language families still reproduced @joryirving's original symptom unchanged, because their tests live in a **separate tree** *and* decorate the basename differently:

| Test file | Issue names | Before |
|---|---|---|
| `spec/foo_spec.rb` | `lib/foo.rb` | demoted |
| `src/test/java/FooTest.java` | `src/main/java/Foo.java` | demoted |
| `tests/FooTests.cs` | `src/Foo.cs` | demoted |
| `test/FooSpec.scala` | `src/Foo.scala` | demoted |
| `tests/FooTest.php` | `src/Foo.php` | demoted |

For those the coder writes the right tests, the rail reports the diff touches none of the files the issue names, and the task burns its attempt budget plus the frontier escalation with **no PR opened**. The failure is indistinguishable from "the model could not do it". That cost was measured on `misospace/dispatch#654`, where 1104 lines of finished tests sat on an abandoned branch until someone audited why the workload was Failed.

## How

**The folding logic already existed and could not run.** `test_layout.go` shipped in #1592 with `TestLayout{TestRoot, SourceRoot}` and a doc comment naming the Maven case verbatim. It had no production caller, and `TestLayout` appeared nowhere outside its own file, so there was also no way to configure it. Unreachable twice over.

Three changes:

- **`GateProfile.testLayout`** (`testRoot` / `sourceRoot`) on the CRD, resolved through `ResolvedGate` exactly as `sourceExtensions` already is, including preset seeding and profile overlay
- **`enforceReviewerScopeOverlap`** takes the resolved layout and folds diff paths through `testTargetsWithLayout` instead of `testTargetsForPath`
- **`stripTestDecoration`** learns `_spec` (Ruby RSpec) and the CamelCase suffix forms `Tests` and `Spec`, alongside the `Test` suffix it already handled. Longest-first ordering so `FooTests` is not left as `FooTest`

A zero layout delegates to the beside-the-code path unchanged, so a repository that declares nothing behaves exactly as before.

```yaml
gateProfile:
  language: generic
  sourceExtensions: [".java"]
  testLayout:
    testRoot: src/test/java
    sourceRoot: src/main/java
```

## Verification

- All five rows above flip from demote to match, as five subtests
- **@joryirving's #1447 case stays green**, asserted both with a layout declared and without one, so this cannot regress the fix it builds on
- **Real drift still demotes** with a layout declared, which is the rail's entire purpose and the thing a permissive change would break
- `stripTestDecoration` table covers the three new conventions plus the four pre-existing ones as regression guards
- 19 existing call sites in `scope_overlap_test.go` updated for the new parameter, all still passing
- `go test ./pkg/foreman/agent/... ./api/... -count=1` → ok
- `GOOS=linux golangci-lint run ./...` → 0 issues
- `make generate manifests` and `make foreman-chart-crds` run; `testLayout` present in `config/crd/bases` and `charts/foreman`

**`test_layout.go` is no longer dead.** Under the #1604 guard it now reports 0 findings, where it previously reported 3. That is the acceptance criterion for this issue, and it is checkable rather than asserted.

## Merge-order note

#1604 lists `pkg/foreman/agent/test_layout.go` in Register A of `.golangci-deadcode.yml` as merged-but-unwired. This PR wires it, so that entry becomes stale. Whichever of the two merges **second** should delete that line. I have not touched it here because the file does not exist on `main` yet.

## Checklist

- [x] Tests added/updated - 5 language subtests + 4 guards, watched fail before implementing
- [x] `make test` passes locally - `./pkg/foreman/agent/...` and `./api/...`
- [x] `make lint` passes locally - 0 issues, fresh lint cache
- [x] Documentation updated (if user-facing change) - CRD field documented inline; the API doc comment carries the Maven and Ruby examples

`Assisted-by: Claude Opus 5` (established the five-language coverage gap while reviewing #1447, found the folding logic already existed but was unreachable, wrote the CRD field, wiring, decorations and tests).


Defilan:
Live-verification status: the first fleet run is **invalid as an A/B** but caught two real defects, one of which is now fixed on this branch.

**Run setup.** Fixture issue #1608 names `scripts/relnotes.rb`; branch `test/1607-fixture-specs` adds only `spec/relnotes_spec.rb` on top of a base branch carrying the module, so the review diff is exactly the #1447 shape in a decoration (`_spec`) that only this PR's layout path folds. Two review-only pipeline steps on the same branch and issue, differing only in `gateProfile.testLayout`.

**Why the run was invalid.** The reviewer's diff-base resolver fetches `baseBranch` from the workload repo (`defilantech`), and the fixture base lived only on the fork: `couldn't find remote ref`, fallback to local `main`, so the diff contained the module file itself and scope-overlap would match trivially in both arms. Fixed by pushing the base branch upstream; base branches must exist on `spec.repo`, only the head branch resolves from the agent's git remote.

**What the invalid run caught anyway.**

1. The reviewer returned a **confabulated GO in 5 turns**: its submit describes "E2E tests cover core functionality with proper RBAC and reconciliation checks" and cites #1593 — none of which exists on a branch adding one Ruby minitest file. It never called `fetch_issue` and never diffed. #1570's class, reproduced on demand. The rail built to catch exactly this, `reviewer_diff_gate.go`, is one of the nine unwired Register A files (#1603) and could not fire.

2. Its published `result.extra` was exactly `[outcome, transcriptRef, turnCount]` — the #1570 starved-extra signature — **on the build that records every rail skip**. Root cause is upstream of all of it: `submit_result` copies the model's optional `extra` verbatim, so a model that omits it publishes `Extra: nil`, and every rail nil-guards silently. The #1605/#1606 observability is downstream of this map existing. Fixed in 32142df75: the submit handler initializes the map, so the rails can always speak regardless of the model's submit verbosity. TDD'd, watched fail first.

Re-run of the corrected A/B is pending a fleet redeploy of this branch head; results will follow in this thread with the task records.

joryirving (approved):
Approve: the folding is correct, it is backward compatible, and it makes real dead code reachable. Two things worth knowing before it changes anything in practice.

I verified the five families against the branch rather than reading the table. `spec/foo_spec.rb` folds to `lib/foo.rb`, `src/test/java/FooTest.java` to `src/main/java/Foo.java`, `tests/FooTests.cs` to `src/Foo.cs`, `test/FooSpec.scala` to `src/Foo.scala`, and `tests/FooTest.php` to `src/Foo.php`. The prefix trap the comment names holds too: with `TestRoot: "tests"`, a path under `testsuite/` is not swallowed. And a zero layout still delegates, so `pkg/a/foo_test.go` and `tests/test_foo.py` behave exactly as before. The longest-first ordering in `stripTestDecoration` is load-bearing and correct, since `Tests` ahead of `Test` is what stops `FooTests` stranding as `FooTest`.

The dead-code claim checks out. On the merge base, `TestLayout` and `testTargetsWithLayout` appear nowhere outside `test_layout.go`, so it was unreachable twice over as you say. Removing the `unused` exclusion in `.golangci-deadcode.yml` is a nice touch, because the linter now enforces that it stays reachable rather than the exclusion hiding a regression.

The CRD field is `+optional` with `omitempty` on both the wrapper and the two roots, so existing GateProfiles are unaffected. Resolving `resolvedGate` once instead of calling `Resolve()` twice in `executor_native.go` is a small bonus.

The first thing worth flagging is scope. `submit_result.go` gaining a nil-`Extra` guard is a genuinely good fix, and the reasoning about the starved-extra signature is right, but it has nothing to do with declaring a test layout. It is eight lines that silently disable every rail when they are absent, which is the kind of change that deserves its own PR and its own visibility in the log rather than riding along in a feature about path folding.

The second is that this changes nothing until an operator declares a layout. In my deployment the profiles live in the bridge's `GATEPROFILE_MAP` and none of them carry `testLayout`, so merging this is inert for me until I add entries. That is the right design, but the PR reads as though the five families are fixed, when what is fixed is the ability to declare them.

Worth being explicit that this does not close the tail of #1447 I hit live. On `misospace/pr-reviewer-action#438` the diff was `tests/test_platform_gh_api_forgejo.py` against an issue naming `pr_reviewer/platform.py`. With `TestRoot: tests, SourceRoot: pr_reviewer` this folds to `pr_reviewer/platform_gh_api_forgejo.py`, which still does not match, because the test is named after a feature rather than the module. Same on `misospace/miso-gallery#365`, where `tests/test_dedup.py` covers functions in `app.py`. Those are a different shape from a parallel test tree and are not something a layout declaration can express.

Checked on the branch: `go build ./...` clean, `api/foreman/v1alpha1` and `pkg/foreman/agent/tools` tests pass. I did not run the full agent package to completion, it exceeds my timeout, so the 48s figure is on trust.

Defilan:
**Live verification complete. The A/B is conclusive.**

Setup: fleet at `0.9.18-pr1607.g32142df`, fixture issue #1608 naming `scripts/relnotes.rb`, branch `test/1607-fixture-specs` whose diff against its base is exactly one file, `spec/relnotes_spec.rb`. Two review-only pipeline steps on the same branch, base, issue, and reviewer (`nemotron-49b-strix`), differing **only** in `gateProfile.testLayout`. Ruby `_spec` chosen deliberately: it is the one decoration class where control and treatment can diverge, since basename matching makes the `.test`/`_test` conventions fold even without a layout.

| | control (no layout) | treatment (`testLayout: {spec, scripts}`) |
|---|---|---|
| task | `wl-1607-live-review-nolayout` | `wl-1607-live-review-layout` |
| `scopeRefs` | `[scripts/relnotes.rb]` | `[scripts/relnotes.rb]` |
| `scopeMatched` | `[]` | `[scripts/relnotes.rb]` |
| `scopeDriftDetected` | `true` | `false` |
| model verdict | GO | GO |
| final verdict | **NO-GO** (demoted) | **GO** (stands) |
| `demotionReason` | scope drift: the issue names 1 file(s) (scripts/relnotes.rb) and the diff touches none of them | — |
| `railsSkipped` | absent (rail ran) | absent (rail ran) |

The control arm is #1447 reproduced on demand: correct work, the right spec file for the right module, demoted because nothing folds `_spec` without a declared layout. The treatment arm is this PR working: the identical diff folds `spec/relnotes_spec.rb` back to `scripts/relnotes.rb`, matches the issue ref, and the GO stands. Incidentally this also live-verifies #1606: both arms show the rail ran rather than skipped, and attempt 2 of this exercise produced a recorded `scope-overlap: no-issue-body` skip when the reviewer lacked the issue.

Getting here surfaced three real defects, each documented as found:

1. `submit_result` published `Extra: nil` when the model omitted the optional field, silently disabling every rail (fixed in 32142df on this branch; likely the mechanism behind #1570's original starved extra).
2. The live `nemotron-reviewer` Agent CR was hand-created with no `systemPrompt` and no `fetch_issue` tool, so every review it produced since creation was uninstructed and structurally unable to read the issue. Patched from the canonical `config/foreman/agents` definition; config drift between repo and cluster deserves its own issue.
3. Fixture rule for future live tests: `payload.baseBranch` must exist on the workload repo; only the head branch resolves from the agent's git remote.

Fixture teardown follows. Closing #1608 as completed by this verification.

Defilan:
All three points land, and each now has a home outside this branch:

- **Scope**: agreed the nil-`Extra` guard deserved its own line in the log, and under this repo's squash-only merging it will not get one; that is a fair hit. It rode along because the live verification physically required it on the deployed build. The squash message will name it, and the full story is in #1605 and this thread.
- **"Fixes the ability, not the five families"**: correct, and worth restating plainly: nothing changes until a `GateProfile` declares a layout. A docs follow-up with the Maven and Ruby examples is queued so operators can actually find the field.
- **The feature-named tail**: filed as #1610 with both of your reproductions. You are right that no layout declaration can express `test_platform_gh_api_forgejo.py` covering `platform.py`; the proposal there is a deterministic content signal (import references), applied only where name-folding failed, with content-vouched matches annotated distinctly in the record.

The verification thread above also answers the "on trust" note: the full agent package ran to completion on the fleet build during the live A/B, alongside the 48s local figure.

Issue #438, opened 2026-05-12
feat: external provider backends via LiteLLM passthrough


Defilan:
Part of epic #437. **Phase 1: MVP**.

## Scope

- Extend `internal/router/backend.go` to support `RouterBackend.External` entries.
- Resolve `credentialsSecretRef` to environment variables in the proxy Pod (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `LITELLM_MASTER_KEY`).
- Default LiteLLM URL when `provider: litellm` is configurable, supporting deployments that already run a LiteLLM proxy as their cloud-provider abstraction.
- Map provider-specific request/response quirks to OpenAI-compatible shape where needed. Strategy: when `provider: litellm`, the upstream already speaks OpenAI; for direct provider calls (`anthropic`, `openai`), use the provider SDK or hand-rolled HTTP.
- Add `backend.tier` (local|cloud) to all audit log lines and metrics labels.

## Success criteria

- e2e test with a stub LiteLLM container routes a request and returns its streamed response intact.
- A test ModelRouter with one `external: {provider: anthropic, model: claude-opus-4-7}` backend, given a valid Anthropic key in a Secret, can dispatch a request end-to-end (against the live API or recorded fixtures).
- Setting `credentialsSecretRef` to a missing Secret surfaces a clear error in `.status.backends[*].message`.
- Audit log entries identify `backend.tier=cloud` for external dispatches.

## Dependencies

Depends on #428.

PR #1614, opened 2026-08-19
feat(foreman): content-based test-coverage vouch for feature-named test files

### LINKED ISSUE #1610: [FEATURE] Scope-overlap: content-based test-coverage vouch for feature-named test files (#1447 tail)
## Feature Description

Let the scope-overlap rail recognise a test file as covering module X when the file's **content** references X, not only when its **name** derives from X. This is the remaining tail of #1447 that no name-based mapping can express.

## Problem Statement

> As someone queueing "add tests for X" work, I want a correct test file that is named after a feature or behaviour to satisfy the scope check for X, so the issue is not unwinnable simply because the repository names tests by what they exercise rather than what they cover.

#1573 fixed name-derivable tests beside the code (`X.test.ts`), and #1607 fixed name-derivable tests in parallel trees (`spec/x_spec.rb`, `src/test/java/XTest.java`). Both fold the **filename**. A test named after a feature defeats both, and two live cases from @joryirving's fleet show the shape:

- `misospace/pr-reviewer-action#438`: issue names `pr_reviewer/platform.py`; the correct diff adds `tests/test_platform_gh_api_forgejo.py`. Even with `testLayout: {tests, pr_reviewer}` this folds to `pr_reviewer/platform_gh_api_forgejo.py`, which matches nothing: the test is named for the feature (Forgejo API support), not the module.
- `misospace/miso-gallery#365`: issue names `app.py`; the correct diff adds `tests/test_dedup.py`, named for the behaviour under test.

The cost is #1447's: the rail reports the diff touches none of the named files, a legitimate GO demotes, and the task burns its attempt budget plus the frontier escalation with the finished work stranded. The failure reads as "the model could not do it".

## Proposed Solution

A deterministic content signal, applied only to diff-**added** test files that name-folding failed to match: the file satisfies ref X when its content imports or references X's module.

For the observed cases this is an import check. `tests/test_platform_gh_api_forgejo.py` contains `from pr_reviewer.platform import ...` (or `import pr_reviewer.platform`); `tests/test_dedup.py` imports from `app`. Per-language reference patterns can live beside the decoration table #1607 already established:

- Python: `import X` / `from X import`
- Go: the module's package import path
- JS/TS: `import ... from './X'` / `require('./X')`
- Ruby: `require`/`require_relative 'X'`

Ordering: name-based matching first (unchanged), content check only for still-unmatched refs against still-unmatched added test files. The rail stays a pure function of (issue refs, diff, file contents); no model judgment enters.

Record the vouch distinctly, e.g. `scopeMatched` entries annotated `via=content`, so a content-vouched match is distinguishable from a path match in the task record, in keeping with #1605/#1606.

## Alternatives Considered

- **Ignore directories and match basenames loosely.** Rejected in #1579 already: `a/foo.test.ts` satisfying an issue naming `b/foo.ts` trades this problem for false vouching, and the rail exists to catch drift.
- **Let the model judge coverage.** Rejected: review judgment is demonstrably stochastic (#647, #1570); the rail's value is that it is computable.
- **Status quo.** The class stays unwinnable, and it is common: test-by-behaviour naming is the dominant style in pytest codebases.

## Additional Context

- Related issues: #1447 (fixed for name-derivable cases), #1579 / PR #1607 (parallel trees; its review thread contains both reproductions), #654-class cost measurement in #1447's report.
- The content check reads files the reviewer's workspace already has at the branch's checkout; no network surface is added.

## Priority

- [x] Medium - Nice to have

## Willingness to Contribute

- [x] Yes, I can submit a PR

Defilan:
## What

Content-based test-coverage vouch in the scope-overlap rail: a diff-**added** test file whose content imports the module an issue names now satisfies the scope check for that module, even when the test's *name* derives from a feature or behaviour rather than the module.

## Why

Fixes #1610

Name-based folding (#1573 beside-the-code, #1607 parallel trees) cannot express a test named for a feature or behaviour rather than its module. For that class, a correct "add tests for X" branch demotes for scope drift and burns its attempt budget with no PR.

**Scope, corrected per review:** of the two live reproductions from @joryirving's fleet, this fixes the diff-**added** shape — `misospace/miso-gallery#365` (`tests/test_dedup.py` covering `app.py`), verified vouching on this branch with `scopeMatchedViaContent: [app.py]`. The other, `misospace/pr-reviewer-action#438`, **appended** to an existing test file; the added-files gate deliberately excludes modified files (soundness: a modified file may have imported the module all along), so that shape remains out of scope here and is tracked in #1616 with added *lines* as the discriminator.

## How

`contentReferencesModule` matches per-language import forms (Python, Go, JS/TS, Ruby) against the named module's path stem; `testTargetsWithLayout`-style ordering is preserved — the content check runs **only** for refs name-matching left unmatched, against **only** diff-added test files. Content-vouched matches are recorded distinctly under `scopeMatchedViaContent`. A file whose content cannot be read is not a vouch: a read failure never flips a ref from unmatched to matched. Distinguishing added files required exporting a small `repo` diff helper, covered by its own tests.

## Provenance and verification

Authored by the Foreman coder (`wl-1610-content-vouch`, task records on the fleet) from the issue's spec; pipeline: coder GO → gate GATE-PASS → reviewer APPROVE with the verdict standing under the deterministic rails. Independently judged before this PR was opened:

- Full `pkg/foreman/agent` suite passes on the branch
- **Mutation checks bite at both layers**: forcing `contentReferencesModule` to `false` fails the unit table *and* the integration test; severing the wiring call fails the integration test — the wiring is load-bearing, not decorative
- `make lint-deadcode` → 0 issues (the CI guard added in #1604 fails on unwired code by construction)
- `GOOS=linux golangci-lint` → 0 issues
- Non-match guards present: a test importing a different module must not vouch; prefix overlap must not vouch; `scopeMatchedViaContent` absent when no content vouch fired

## Checklist

- [x] Tests added/updated - 9 test functions including negative cases
- [x] `make test` passes locally - full agent package on the branch
- [x] `make lint` passes locally - 0 issues
- [x] Documentation updated (if user-facing change) - rail-internal; the language-gates doc gains a cross-reference in the #1610 follow-up if wanted

Foreman-authored per the AI-assisted contribution policy; human-reviewed and accountable (this PR's review conversation is owned by the maintainer).

### DIFF
--- a/pkg/foreman/agent/executor_native.go
+++ b/pkg/foreman/agent/executor_native.go
@@ -883,10 +883,19 @@ func (e *NativeAgentLoopExecutor) runLLMPath(
 			var scopeMatched []string
 			if reviewDiffErr == nil {
 				resolvedGate := task.Spec.GateProfile.Resolve()
+				// The content-based vouch (#1610) reads diff-ADDED test files
+				// from the workspace checkout. A failure to compute the added
+				// set or to read a file degrades the vouch to name-only — the
+				// rail must never demote on a read error, so this stays best-effort.
+				scopeAdded, _ := repo.DiffAdded(ctx, workspace, reviewBase)
+				scopeReadFile := func(relPath string) ([]byte, error) {
+					return os.ReadFile(filepath.Join(workspace, relPath))
+				}
 				verdict = enforceReviewerScopeOverlap(log, loopRes.Terminal.Extra,
 					extractFetchIssueBody(loopRes.Transcript), reviewDiff, verdict,
 					resolvedGate.SourceExtensions,
-					testLayoutFrom(resolvedGate.TestLayout))
+					testLayoutFrom(resolvedGate.TestLayout),
+					scopeAdded, scopeReadFile)
 				scopeDriftDetected, _ = loopRes.Terminal.Extra["scopeDriftDetected"].(bool)
 				scopeMatched, _ = loopRes.Terminal.Extra["scopeMatched"].([]string)
 			} else {
--- a/pkg/foreman/agent/repo/diff.go
+++ b/pkg/foreman/agent/repo/diff.go
@@ -55,6 +55,57 @@ func DiffNameOnly(ctx context.Context, workspace, base string) ([]string, error)
 	return parseNameOnly(out), nil
 }
 
+// DiffAdded returns the subset of the branch's diff whose status is ADDED
+// (git status "A"), using `git diff --name-status base...HEAD`. The three-dot
+// form matches DiffNameOnly, so the two agree on which files the branch
+// touched; this only narrows to the additions.
+//
+// The content-based test-coverage vouch (#1610) reads the body of a diff-ADDED
+// test file to check it imports the module it covers. Reading only added files
+// keeps the vouch scoped to what this branch created — a test the branch added
+// is the coverage we are vouching for — and bounds the file reads to the diff
+// rather than the whole tree. An empty result is not an error.
+func DiffAdded(ctx context.Context, workspace, base string) ([]string, error) {
+	if workspace == "" {
+		return nil, fmt.Errorf("DiffAdded: workspace is required")
+	}
+	if base == "" {
+		return nil, fmt.Errorf("DiffAdded: base ref is required")
+	}
+	out, err := runGit(ctx, workspace, baseEnv(), "diff", "--name-status", base+"...HEAD")
+	if err != nil {
+		return nil, err
+	}
+	return parseNameStatusAdded(out), nil
+}
+
+// parseNameStatusAdded keeps only the ADDED ("A") paths from tab-separated
+// `git diff --name-status` output. Rename/copy rows ("R..", "C..") are not
+// additions of a new path and are excluded: a renamed test file's coverage was
+// already present on the base, so it is not the new coverage the vouch targets.
+// Returns nil (not []string{}) when nothing was added, so a DeepEqual against
+// a nil expectation works in callers and tests.
+func parseNameStatusAdded(out string) []string {
+	if out == "" {
+		return nil
+	}
+	var added []string
+	for _, line := range strings.Split(out, "\n") {
+		line = strings.TrimRight(line, "\r")
+		if line == "" {
+			continue
+		}
+		f := strings.Split(line, "\t")
+		if len(f) < 2 {
+			continue
+		}
+		if f[0] == "A" {
+			added = append(added, f[len(f)-1])
+		}
+	}
+	return added
+}
+
 // parseNameOnly splits a `git diff --name-only` output into a clean
 // slice of paths. Strips empty lines and trims surrounding whitespace
 // off each entry. Returns nil (not []string{}) when the output yields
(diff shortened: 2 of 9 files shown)


joryirving (approved):
Approve with one correction to the claim: this fixes one of the two reproductions it cites, not both.

`misospace/miso-gallery#365` is fixed. I ran the rail on the branch with that issue's body and its real diff and it vouches, recording `scopeMatchedViaContent: [app.py]`. `misospace/pr-reviewer-action#438` still demotes, because its diff **modified** `tests/test_platform_gh_api_forgejo.py` rather than adding it: the file already existed on main at 20,317 bytes and the coder appended 25 lines. Feeding the same content through as an added file vouches correctly and yields `scopeMatchedViaContent: [pr_reviewer/platform.py]`, so the matching logic handles that case perfectly and only the `addedFiles` gate stands between it and a pass.

That gate is worth more discussion than the comment gives it. The stated reason is that reading only added files "bounds the work to the diff", which is a cost argument, and on a normal diff the cost of also reading modified test files is negligible. The stronger argument is soundness, and it is not the one written down: a modified test file may have imported the module before this branch touched it, so vouching on whole-file content would let a coder earn a pass by editing any test that happens to import the named module, without adding coverage for it. That is a real false-positive class and a legitimate reason to restrict, but as written the code reads as an optimisation rather than a deliberate trade.

If you want the modified case without reopening that hole, the discriminator is the added lines rather than the file. A test file that gains `from pr_reviewer.platform import ...` in this diff is evidence; one that merely already contained it is not. `repo/diff.go` currently exposes `DiffNameOnly` and a name-status `DiffAdded`, so this would need an added-lines variant, `git diff -U0` and collecting `+` lines. More work than this PR, and reasonable as a follow-up rather than a change here.

The rest holds up under direct testing rather than reading. All nine of the guards I tried behave: `pr_reviewer.platform_utils` does not vouch for `pr_reviewer/platform.py`, a bare mention of the word platform in a comment does not vouch, a Go import of a different last element does not vouch, and the Python, Go, JS and Ruby positive forms all do. Recording content matches separately under `scopeMatchedViaContent` is the right call, since it keeps a name-earned match distinguishable from a content-earned one in the task record.

My only ask is that the PR body and the `testContentReferencesModule` comment stop claiming #438 as fixed, since someone hitting that exact shape will otherwise conclude the rail is broken rather than that the case is out of scope.

Defilan:
All three points taken, and each is now landed somewhere concrete:

- **The over-claim is retracted.** The PR body now states the corrected scope: #365's added-file shape is fixed (with your verification quoted); #438's appended-to-existing shape is explicitly out of scope here.
- **The gate comment now argues soundness, not cost** (pushed as the top commit): a modified test file may have imported the module before the branch touched it, so whole-file content on modified files would let editing any importing test earn a pass. Your framing, essentially verbatim, because it was the right one and the code should carry it.
- **The modified-file case is filed as #1616** with your discriminator — added lines via `git diff -U0`, an added-lines variant beside `DiffNameOnly`/`DiffAdded` in `repo/diff.go` — and your direct-testing evidence cited, including that the matching logic already handles #438's content when fed as added.

The comment push dismisses your approval under stale-review; nothing else on the branch changed. Your nine adversarial probes surviving direct testing is the part of this review I'd like every future rail to get.

Issue #1621, opened 2026-08-20
[BUG] PrefetchJob does not support setting Tolerations


rouke-broersma:
## Bug Description

My GPU nodes are tainted so regular pods won't schedule on it. Most LLMKube resources support setting tolerations so this is fine. The prefetch job has no tolerations configuration options and also does not inherit tolerations from either controller or inferenceservices. Since I use local storage, the prefetch job must run on the same node as the rest of LLMKube.

## Steps to Reproduce

1. Taint node
2. Create local storage/localpv on the tainted node
3. Deploy LLMKube with tolerations
4. Enable prefetch on a model

## Expected Behavior

Prefetch job runs with controller tolerations or inferenceservice tolerations. Inferenceservice(s) tolerations is probably simplest as these are already available to the controller.

## Actual Behavior

Jobs fail to schedule

## Environment

**LLMKube Version:**
```bash
v0.9.19
```

**InferenceService:**
```yaml
# Paste your InferenceService YAML here
```

## Additional Context

Add any other context about the problem here (screenshots, error messages, etc.).

Defilan:
Confirmed against the source: `buildPrefetchJob` emits a pod spec with no tolerations at all, while the serving deployment builder threads both the GPU-derived and `spec.tolerations` sets. Good catch, and the repro is exactly right.

One design divergence from the suggestion, with reasoning: inheriting the InferenceService's tolerations is undefined at the moment prefetch runs, because prefetch exists precisely to complete **before** the first InferenceService does. So the fix adds `Model.spec.tolerations`, threaded onto the Job. Your deploy-time flow becomes: the same tolerations you already set elsewhere, declared once more on the Model that prefetches.

Fix is up in #1622 with the field shape in the PR body. It is unit-verified; if you can confirm it schedules on your tainted node once merged, that closes this with live evidence.

Defilan:
Already shipped. `d0db838a6` ("fix(controller): carry Model tolerations onto the prefetch Job (#1622)") added `spec.prefetchTolerations`, applied at `internal/controller/model_prefetch.go:277`:

```go
Tolerations:    model.Spec.PrefetchTolerations,
```

Closing as resolved. Found while staleness-checking the backlog before queuing agent work — the issue stayed open after its fix merged.

PR #1622, opened 2026-08-20
fix(controller): carry Model tolerations onto the prefetch Job

Defilan:
## What

`Model.spec.prefetchTolerations`, applied to the prefetch Job's pod. Prefetch-scoped by name after review discussion: an unscoped `tolerations` on Model would be ambiguous to consumers, since storage topology is not part of the Model spec.

## Why

Refs #1621

@rouke-broersma's report: GPU nodes are tainted, every other LLMKube resource can declare tolerations, but the prefetch Job's pod spec carries none, so on a fleet whose shared model cache lives on node-local storage pinned to those nodes, prefetch sits Pending forever.

## How

The report suggested inheriting the InferenceService's tolerations. That would be ill-defined at exactly the moment it matters: prefetch's purpose is to run **before** the first InferenceService exists, so there may be nothing to inherit from when the Job is built. Instead the Model declares `prefetchTolerations`, and `buildPrefetchJob` threads them onto the pod. No declaration means no tolerations emitted, byte-identical to today's spec.

`nodeSelector` was deliberately left out: for node-local storage the PV's own node affinity already pins the pod's placement at bind time; the taint was the only blocker.

## Verification

- Ginkgo specs, watched fail first: a Model with tolerations produces a Job carrying them; a Model without produces a clean spec
- Full `internal/controller` suite passes (324s, envtest)
- `make generate manifests chart-crds` run; the field is in both CRD copies
- `GOOS=linux golangci-lint` and the dead-code guard: 0 issues

`Refs` rather than `Fixes` per house discipline: unit-verified, not yet reproduced against a live tainted node. @rouke-broersma, if you can confirm on your cluster once this merges, that closes it with real evidence; the field is exactly:

```yaml
kind: Model
spec:
  source: https://...
  prefetch: true
  prefetchTolerations:
    - key: nvidia.com/gpu
      operator: Equal
      value: present
      effect: NoSchedule
```

## Checklist

- [x] Tests added/updated - watched fail before implementing
- [x] `make test` passes locally - full controller suite
- [x] `make lint` passes locally - 0 issues
- [x] Documentation updated (if user-facing change) - field documented on the API type

`Assisted-by: Claude Opus 5` (root-caused against buildPrefetchJob, chose the Model-field design over ISvc inheritance, implemented with TDD).

### DIFF
--- a/api/v1alpha1/model_types.go
+++ b/api/v1alpha1/model_types.go
@@ -103,6 +103,16 @@ type ModelSpec struct {
 	// +optional
 	Prefetch bool `json:"prefetch,omitempty"`
 
+	// PrefetchTolerations are applied to the prefetch Job's pod, and govern
+	// nothing else: they say where the download may run, not where the model
+	// will be served. A fleet whose GPU nodes are tainted needs the prefetch
+	// download to schedule onto those nodes when the shared cache lives on
+	// node-local storage (#1621). Inheriting from an InferenceService would
+	// be ill-defined here: prefetch runs before the first InferenceService
+	// exists by design.
+	// +optional
+	PrefetchTolerations []corev1.Toleration `json:"prefetchTolerations,omitempty"`
+
 	// Format specifies the model file format.
 	// "gguf" is used with the llama-server runtime; "mlx" is used with the oMLX runtime;
 	// "safetensors", "pytorch", and "custom" are used with the generic runtime.
--- a/api/v1alpha1/zz_generated.deepcopy.go
+++ b/api/v1alpha1/zz_generated.deepcopy.go
@@ -1482,6 +1482,13 @@ func (in *ModelSpec) DeepCopyInto(out *ModelSpec) {
 		*out = new(corev1.LocalObjectReference)
 		**out = **in
 	}
+	if in.PrefetchTolerations != nil {
+		in, out := &in.PrefetchTolerations, &out.PrefetchTolerations
+		*out = make([]corev1.Toleration, len(*in))
+		for i := range *in {
+			(*in)[i].DeepCopyInto(&(*out)[i])
+		}
+	}
 	if in.Hardware != nil {
 		in, out := &in.Hardware, &out.Hardware
 		*out = new(HardwareSpec)
--- a/charts/llmkube/templates/crds/models.yaml
+++ b/charts/llmkube/templates/crds/models.yaml
@@ -340,6 +340,53 @@ spec:
                   source remains the alternative there. Ignored for local and pvc://
                   sources, which need no remote fetch.
                 type: boolean
+              prefetchTolerations:
+                description: |-
+                  PrefetchTolerations are applied to the prefetch Job's pod, and govern
+                  nothing else: they say where the download may run, not where the model
+                  will be served. A fleet whose GPU nodes are tainted needs the prefetch
+                  download to schedule onto those nodes when the shared cache lives on
+                  node-local storage (#1621). Inheriting from an InferenceService would
+                  be ill-defined here: prefetch runs before the first InferenceService
+                  exists by design.
+                items:
+                  description: |-
+                    The pod this Toleration is attached to tolerates any taint that matches
+                    the triple <key,value,effect> using the matching operator <operator>.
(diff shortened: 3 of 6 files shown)


rouke-broersma:
While I understand the reasoning, I think Tolerations on Model is an awkward fit because Model is a download request and not a request for a workload resource. Tolerations on Model is an implementation detail of how Prefetch is implemented leaking through. Model does not actually care where it is stored, this is decided by the controller cache config + InferenceService. If implemented it begs the question if should also add a validation to make sure InferenceService and Model share the same Toleration since this becomes an implicit constraint. Or, if we need Toleration on InferenceService explicitly at all. We can also move that responsibility to Model in that case since InferenceService must copy it's referenced Models tolerations since the Model may be in that 'location'.

Defilan:
Appreciate you thinking this through past the surface fix, this is the right kind of pushback.

Where I agree: a top-level `tolerations` on Model does read a bit like "Model is a workload," and that framing matters. The fleet-level annoyance you're pointing at is real too, "our GPU nodes are tainted" ends up restated on every resource.

The shape was a deliberate choice though, for two reasons:

1. Model already is a pod-spawning resource. The prefetch Job is a real workload the operator runs on the Model's behalf (and it's the only one, the serve-time download runs inside the serving pod, which inherits the InferenceService's tolerations). Resources that cause pods growing scheduling knobs is pretty standard Kubernetes, Job and CronJob templates being the precedent.

2. The Model/ISvc coupling you'd want to validate or inherit only exists in one topology. With node-local cache storage, yes, both must reach the tainted node. But with shared RWX storage the prefetch pod can legally run anywhere with zero tolerations while serving carries GPU ones, so enforcing a match would reject valid configs. Same issue with flipping ownership to Model: multiple InferenceServices can reference one Model across heterogeneous nodes with different taints, so the ISvc has to own its own. The real constraint ("prefetch must be schedulable where the cache PV binds") is enforced by the PV's node affinity at bind time, tolerations just unlock the node.

If the top-level placement still feels wrong I'm open to scoping it under a prefetch-specific block before this merges, that's the cheap moment for it. And the version of your concern I think has legs long-term is fleet-level default tolerations for operator-spawned pods, so nobody restates this per-CR at all. Happy to file that as a follow-up if it'd help your setup.

rouke-broersma:
> The Model/ISvc coupling you'd want to validate or inherit only exists in one topology. With node-local cache storage, yes, both must reach the tainted node. But with shared RWX storage the prefetch pod can legally run anywhere with zero tolerations while serving carries GPU ones, so enforcing a match would reject valid configs.

In the current implementation yes this is absolutely true, but from a spec design it is not. If a Model declares that it can tolerate certain taints then InferenceService which depends on Model being available to consume must assume that it also has to tolerate those taints as it does not 'know' why those taints are tolerated and what the consequence is as model storage configuration is not part of the Model spec.

> Same issue with flipping ownership to Model: multiple InferenceServices can reference one Model across heterogeneous nodes with different taints, so the ISvc has to own its own.

InferenceService can absolutely supply it's own additional tolerations, I was thinking it would assume it _also_ requires whatever toleration is on the Model.

> If the top-level placement still feels wrong I'm open to scoping it under a prefetch-specific block before this merges, that's the cheap moment for it.

That avoids the above for me, it would clearly be coupled only to the prefetch workload requirements in that case. That said, all of this is nitpicking, I will be happy with any of the solutions, I would be unblocked by the current solution as well.

Defilan:
Went back and forth on this and your spec-semantics framing won: an unscoped `tolerations` on Model is ambiguous to consumers precisely because storage topology isn't part of the Model spec, so nothing tells a reader (human or controller) what the field governs. The doc comment carries that scoping in prose, and the field name should carry it structurally.

So before merge this becomes `spec.prefetchTolerations`, a flat field rather than a scheduling block: we deliberately excluded nodeSelector (the cache PV's node affinity owns placement, tolerations just unlock the node), so a block reserving room for knobs we've argued against would be self-contradictory. Flat field, self-describing name, scoped to exactly the one workload that exists.

Genuinely useful pushback, this is a better API than the one I opened with. Pushing the rename shortly.

PR #1624, opened 2026-08-20
feat(chart): alert when an engine is up but not serving

Tanguille:
`InferenceServiceDown` only catches a dead process. An engine can hold its port open, pass `/health` and report `Ready` while admitting almost no work: requests queue, KV cache sits partly free, prompt throughput reads zero. Nothing in the chart caught that.

Observed on a single-GPU vLLM deployment: three hours of `1/1 Running`, `/health` 200, no chart alert, while the engine sat at `Running: 1, Waiting: 5` with 36% KV free and 0.0 tok/s prompt throughput.

## New `llmkube-serving` group

| alert | expr basis | runtimes | default |
|---|---|---|---|
| `InferenceAdmissionStalled` | queued requests with none running | vllm + sglang | on |
| `InferenceQueueTimeHigh` | mean queue wait > threshold | vllm + sglang | on |
| `KVOffloadLoadSlow` | mean async KV-offload load > threshold | vllm | off |

`KVOffloadLoadSlow` names the likely cause (a slow, non-preemptible async KV load gates admission for everything behind it); `InferenceQueueTimeHigh` only reports the symptom. It's off by default since `OffloadingConnector` is upstream-experimental and its series are absent without a KV connector configured, same pattern as the existing `gpu.memoryPressure` toggle.

## Also fixes a recording rule that only covered SGLang

vLLM exports `vllm:request_queue_time_seconds`, which `charts/llmkube/dashboards/vllm-dashboard.json` already queries. Both queue-wait recording rules now use the same `or` alternation as their `e2e_request_latency` and `ttft_seconds` neighbours.

## Thresholds

Values-driven, set well above the healthy band. Measured on a healthy single-GPU vLLM deployment:

| | healthy | default threshold | headroom |
|---|---|---|---|
| mean queue wait | 1.4s | 30s | ~21x |
| mean async offload load | 0.9s | 10s | ~11x |

## Validation

- `helm unittest`: 83/83 passing, including 6 new cases.
- Rendered expressions run against a live VictoriaMetrics instance: all parse and return the expected series with `service` / `namespace` / `runtime` labels. Confirmed they aren't silent no-ops by lowering thresholds to 0 and checking they then match, the `#1223` failure mode.

## Caveat

Thresholds are backtested against one incident on one deployment. Metric names are stock and rule shapes are runtime-agnostic, but treat the numbers as reasonable defaults rather than fleet-validated. Happy to adjust, or gate `serving.enabled` off by default for a first release.

### DIFF
--- a/charts/llmkube/README.md
+++ b/charts/llmkube/README.md
@@ -142,7 +142,7 @@ The following table lists the configurable parameters of the LLMKube chart and t
 | `prometheus.serviceMonitor.interval` | Scrape interval | `30s` |
 | `prometheus.serviceMonitor.namespace` | ServiceMonitor namespace (defaults to release namespace) | `""` |
 | `prometheus.serviceMonitor.additionalLabels` | Additional labels for ServiceMonitor | See values.yaml |
-| `prometheus.prometheusRule.enabled` | Enable PrometheusRule for alerts | `false` |
+| `prometheus.prometheusRule.enabled` | Enable PrometheusRule for alerts. Also required by the queue-wait panels on the vLLM and SGLang dashboards, which read the `llmkube:inference:queue_time_seconds:p95_5m` recording rule; without it those panels render blank. | `false` |
 | `prometheus.prometheusRule.namespace` | PrometheusRule namespace | `monitoring` |
 | `prometheus.prometheusRule.rules.gpu.enabled` | Enable GPU alerts | `true` |
 | `prometheus.prometheusRule.rules.gpu.highUtilizationThreshold` | GPU high utilization threshold (%) | `90` |
--- a/charts/llmkube/README.md
+++ b/charts/llmkube/README.md
@@ -150,6 +150,10 @@ The following table lists the configurable parameters of the LLMKube chart and t
 | `prometheus.prometheusRule.rules.gpu.memoryPressureThreshold` | GPU memory pressure threshold (%) | `90` |
 | `prometheus.prometheusRule.rules.gpu.powerLimitThreshold` | GPU power limit threshold (W) | `250` |
 | `prometheus.prometheusRule.rules.inference.enabled` | Enable inference alerts | `true` |
+| `prometheus.prometheusRule.rules.serving.enabled` | Enable serving-quality alerts (also requires `prometheus.inferencePodMonitor.enabled`) | `true` |
+| `prometheus.prometheusRule.rules.serving.queueTimeSecondsThreshold` | Mean per-pod queue wait threshold (s) | `30` |
+| `prometheus.prometheusRule.rules.serving.kvOffload.enabled` | Enable the KV-offload-slow alert (requires an OffloadingConnector) | `false` |
+| `prometheus.prometheusRule.rules.serving.kvOffload.loadSecondsThreshold` | Mean async KV-offload load threshold (s) | `10` |
 
 #### Scrape labels
 
--- a/charts/llmkube/dashboards/sglang-dashboard.json
+++ b/charts/llmkube/dashboards/sglang-dashboard.json
@@ -12,7 +12,7 @@
       }
     ]
   },
-  "description": "LLMKube SGLang runtime observability. Throughput, TTFT / E2E / inter-token latency, queue depth, KV pool utilization, scheduler backpressure, prefix cache hit rate, queue wait, per-stage latency, new token ratio, prompt length distribution and request rate, grouped by service / namespace from the sglang: metrics the chart's inference PodMonitor scrapes. TTFT and end-to-end p95 / p50 read the chart's llmkube:inference: recording rules. Mamba, HiCache host and HiCache L3 rows are collapsed because those pools only exist on hybrid-attention models, with --hicache enabled, and with --hicache-storage-backend enabled respectively.",
+  "description": "LLMKube SGLang runtime observability. Throughput, TTFT / E2E / inter-token latency, queue depth, KV pool utilization, scheduler backpressure, prefix cache hit rate, queue wait, per-stage latency, new token ratio, prompt length distribution and request rate, grouped by service / namespace from the sglang: metrics the chart's inference PodMonitor scrapes. TTFT, end-to-end, and queue-wait p95 / p50 read the chart's llmkube:inference: recording rules, per pod. Mamba, HiCache host and HiCache L3 rows are collapsed because those pools only exist on hybrid-attention models, with --hicache enabled, and with --hicache-storage-backend enabled respectively.",
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 1,
@@ -459,13 +459,13 @@
         {
           "datasource": {"type": "prometheus", "uid": "${DS_PROMETHEUS}"},
           "expr": "llmkube:inference:queue_time_seconds:p95_5m{runtime=\"sglang\", service=~\"$service\"}",
-          "legendFormat": "{{service}} ({{namespace}}) p95",
+          "legendFormat": "{{service}}/{{pod}} p95",
           "refId": "A"
         },
         {
           "datasource": {"type": "prometheus", "uid": "${DS_PROMETHEUS}"},
           "expr": "llmkube:inference:queue_time_seconds:p50_5m{runtime=\"sglang\", service=~\"$service\"}",
-          "legendFormat": "{{service}} ({{namespace}}) p50",
+          "legendFormat": "{{service}}/{{pod}} p50",
           "refId": "B"
         }
       ],
--- a/charts/llmkube/dashboards/vllm-dashboard.json
+++ b/charts/llmkube/dashboards/vllm-dashboard.json
@@ -12,7 +12,7 @@
       }
     ]
   },
-  "description": "LLMKube vLLM runtime observability. Throughput, TTFT / E2E / inter-token latency, queue depth, KV cache utilization, prefix cache hit rate and scheduler backpressure (preemptions, aborts), grouped by service / namespace from the vllm: metrics the chart's inference PodMonitor scrapes. TTFT and end-to-end p95/p50 read the chart's llmkube:inference: recording rules. Request-phase timing (prefill / decode / queue) is in a collapsed row.",
+  "description": "LLMKube vLLM runtime observability. Throughput, TTFT / E2E / inter-token latency, queue depth, KV cache utilization, prefix cache hit rate and scheduler backpressure (preemptions, aborts), grouped by service / namespace from the vllm: metrics the chart's inference PodMonitor scrapes. TTFT, end-to-end, and queue-wait p95/p50 read the chart's llmkube:inference: recording rules, per pod. Request-phase timing (prefill / decode / queue) is in a collapsed row.",
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 1,
@@ -412,8 +412,8 @@
           "targets": [
             {
               "datasource": {"type": "prometheus", "uid": "${DS_PROMETHEUS}"},
-              "expr": "histogram_quantile(0.95, sum by (service, namespace, le) (rate(vllm:request_queue_time_seconds_bucket{service=~\"$service\"}[5m])))",
-              "legendFormat": "{{service}} ({{namespace}}) p95",
+              "expr": "llmkube:inference:queue_time_seconds:p95_5m{runtime=\"vllm\", service=~\"$service\"}",
+              "legendFormat": "{{service}}/{{pod}} p95",
               "refId": "A"
             }
           ],
--- a/charts/llmkube/templates/prometheusrule.yaml
+++ b/charts/llmkube/templates/prometheusrule.yaml
@@ -253,8 +253,10 @@ spec:
       labels:
         latency_kind: ttft
 
-    # Queue wait over 5m, covers vLLM and SGLang. By pod, not just service: a
-    # service-wide percentile would dilute one stalled replica out of it.
+    # Queue wait over 5m, covers vLLM and SGLang. By pod, not just service (unlike
+    # the request_latency/ttft_seconds records above): a service-wide percentile
+    # would dilute one stalled replica out of it, and InferenceQueueTimeHigh
+    # alerts straight off this one.
     - record: llmkube:inference:queue_time_seconds:p95_5m
       expr: histogram_quantile(0.95, sum by (service, namespace, runtime, pod, le) (rate(vllm:request_queue_time_seconds_bucket[5m]) or rate(sglang:queue_time_seconds_bucket[5m])))
       labels:
--- a/charts/llmkube/templates/prometheusrule.yaml
+++ b/charts/llmkube/templates/prometheusrule.yaml
@@ -86,15 +86,9 @@ spec:
   - name: llmkube-serving
     interval: 30s
     rules:
-    # TGI has no queued-count metric to pair with tgi_batch_current_size, and
-    # PersonaPlex/Generic/LlamaCppRouter export no queue metrics at all, so
-    # they're left out.
-    #
-    # A single scrape where running momentarily ticks above the ceiling
-    # resets an instantaneous `== 0` check's pending timer, so this averages
-    # over the window instead. "<= 1" is a small-constant ceiling, not zero,
-    # because a partial stall (e.g. running pinned at 1 while waiting queues
-    # up) is still a stall.
+    # TGI, PersonaPlex, Generic and LlamaCppRouter export no queue metrics, so
+    # they're left out. avg_over_time and a "<= 1" ceiling (not "== 0") catch
+    # a partial stall and survive a single scrape ticking above it.
     - alert: InferenceAdmissionStalled
       expr: |
         (avg_over_time(vllm:num_requests_waiting[10m]) > 0 and avg_over_time(vllm:num_requests_running[10m]) <= 1)
@@ -110,10 +104,8 @@ spec:
         summary: "Inference engine is queuing requests but running almost none"
         description: "{{`{{ $labels.service }}`}} in {{`{{ $labels.namespace }}`}} has {{`{{ $value }}`}} requests queued while running almost none. The process is up; it is not admitting work."
 
-    # The utilization series dropping out (not dropping to zero) reads as an
-    # idle-but-healthy engine, same silent no-op class as GPUMetricsMissing /
-    # ControllerMetricsMissing: a selector that stops matching or a job-label
-    # drift is indistinguishable from nothing running.
+    # Same silent no-op class as GPUMetricsMissing/ControllerMetricsMissing: a
+    # dead selector reads as an idle-but-healthy engine, not a gap.
     - alert: InferenceMetricsMissing
       expr: absent(vllm:num_requests_running) and absent(sglang:num_running_reqs) and absent(llamacpp:requests_processing)
       for: 15m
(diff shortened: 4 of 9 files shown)


Defilan:
Thanks for this, and for the writeup. The gap is real: `InferenceServiceDown` only catches a dead process, and an engine that holds its port open while admitting nothing is exactly the case the chart has been blind to. The recording-rule half of this PR is a clean fix and I want that regardless of what happens to the rest.

I reviewed the alert expressions closely, ran `helm unittest` (83/83 pass), rendered and parsed the templates, and checked every metric name against `internal/metrics/testdata/*-metrics.txt`. The blocking problem is that I do not think the two on-by-default alerts can fire for the incident described above.

## Blocking

**1. `InferenceAdmissionStalled` cannot match the motivating incident.**

The PR body states the observed state was `Running: 1, Waiting: 5`. The expression requires `vllm:num_requests_running == 0`. With one request running that operand is an empty vector, the `and` yields nothing, and the alert never enters pending. A partial stall, which is what three hours at `Running: 1` is, needs a ratio or a "waiting high while running pinned at or below a small constant for N minutes" form.

**2. `InferenceQueueTimeHigh` goes silent in exactly the total-stall case.**

`vllm:request_queue_time_seconds` and `sglang:queue_time_seconds` are observed at request *completion*. An engine that admits nothing completes nothing, so `increase(..._count[15m])` stays 0, `increase(..._sum[15m])` stays 0, and the expression evaluates `0/0` = NaN. NaN > 30 is false, no series exists, nothing fires. A three hour hard stall produces zero pages from this rule. That is the same silent-no-op class as #1223 that `GPUMetricsMissing` and `ControllerMetricsMissing` were added to close.

Taken together, neither new on-by-default alert detects the failure the PR was written for. Worth validating the corrected expressions against the recorded incident data if you still have it.

**3. The serving group defaults on, but depends on labels only a disabled-by-default PodMonitor injects.**

`values.yaml` has `prometheus.inferencePodMonitor.enabled: false` (line 257) while the new `rules.serving.enabled` is `true`. The three new rules group by `service` and `runtime`, which are applied only by `inference-podmonitor.yaml`'s relabelings. Under any foreign scrape config, `sum by (service, namespace, runtime)` collapses every vLLM pod in a namespace into one empty-label series (so the mean queue wait becomes a fleet average that hides the single stalled engine), and the descriptions render `<no value> in <namespace>`.

The chart already handles this for `ControllerMetricsMissing`, gated with `{{- if .Values.prometheus.serviceMonitor.enabled }}` and the comment "Gated on the chart owning the scrape target" (line 167). The serving group should follow that convention.

**4. No `absent()` companion for the new group.**

All three rules are presence-dependent: `X > 0 and Y == 0` and `increase(sum)/increase(count) > T` all produce an empty vector when the series vanish. If the PodMonitor selector stops matching after a pod label rename, or the `job` label drifts, or scraping just breaks, the operator sees no alert and it is indistinguishable from a healthy fleet. The chart established the counter-pattern twice for this reason (`GPUMetricsMissing` line 74, `ControllerMetricsMissing` line 171, both citing #1223/#1356). Something like `absent(vllm:num_requests_running) and absent(sglang:num_running_reqs)` is missing.

## Should fix before merge

**5. No minimum-sample guard, so one slow request pages `critical`.**

A freshly rolled vLLM pod queues its first request 45s behind weight load and CUDA graph capture, then serves nothing else. Over the next 15m, `increase(_sum[15m])` is about 45 and `increase(_count[15m])` is about 1, so the mean reads 45 > 30 continuously, satisfies `for: 15m`, and fires a critical alert on an idle healthy service. The usual guard (`and sum by (...) (increase(..._count[15m])) > N`) is absent from both this rule and `KVOffloadLoadSlow`, which has the same shape over 30m.

**6. `InferenceQueueTimeHigh` is service-aggregated while the alert above it is per-pod.**

With `replicas: 4`, one pod stalled at a 200s mean queue wait and three healthy at 1.4s, the ratio weights by completed-request count. The healthy pods complete nearly all requests, so the service-level mean lands around 2 to 3s, far under threshold, and the stalled replica is invisible. Adding `pod` to the `by` clause keeps per-replica resolution and matches `InferenceAdmissionStalled` directly above it.

**7. `for: 10m` over an instantaneous gauge resets on any single scrape.**

An engine that drains one queued request every few minutes but is otherwise wedged shows `num_requests_running == 1` at roughly one evaluation in six. Each one empties the vector and resets the pending timer, so the alert never reaches 10 continuous minutes. `avg_over_time(vllm:num_requests_running[10m]) < 1` or a `max_over_time` form survives the flap.

**8. `severity: critical` is inconsistent with the rest of the chart.**

Existing criticals are hard-down or damage conditions: `up == 0` for `InferenceServiceDown` and `ControllerDown`, and GPU over-temperature. A 30s mean queue wait is a saturation signal a busy but healthy batch deployment can hit legitimately. `KVOffloadLoadSlow` in the same new group is correctly `warning`. Combined with finding 5, this pages out of hours on an idle service.

**9. One test passes vacuously.**

`tests/prometheusrule_test.yaml:250`, "should omit the KV-offload alert by default". I ran that assertion with `prometheus.prometheusRule.rules.serving.enabled: false` so the JSONPath resolves to nothing, and it still passes. helm-unittest treats `notMatchRegex` over an empty path as satisfied, so if a future edit drops or renames the serving group the test stays green. The same file already solved this for the GPU group with `lengthEqual` (line ~137); the serving tests want `count: 2` by default and `count: 3` with kvOffload enabled.

## Worth addressing, non-blocking

**10. Alert expressions are the one PromQL surface in the repo with no name guard.** `internal/metrics/dashboard_sync_test.go` validates dashboards and marked doc regions against the metric fixtures precisely to stop the #786/#1223/#1226 class, but it only reads `- record:` lines from the chart, never `- alert:`. `vllm:kv_offload_lookup_async_delay_seconds` is real upstream (`kv_connector/v1/offloading/metrics.py`) but is in no fixture, so a typo or rename in any of the six names here ships a rule that silently never fires. The durable fix is extending that extractor to alert exprs; happy to take that as a follow-up rather than block you on it.

**11. Partial runtime coverage, unguarded.** `TestChartRestartRuleCoversEveryBackend` enforces that the restart recording rule matches every backend's `ContainerName()`. The new alerts enumerate vllm/sglang/llamacpp only; `PersonaPlexBackend`, `GenericBackend` and `LlamaCppRouterBackend` get no coverage and no comment explaining the omission (only TGI's is explained). A seventh runtime will silently go unmonitored.

**12. The queue-metric alternation is now written in two places.** The PR widens the p95/p50 recording rules to `rate(vllm:...) or rate(sglang:...)`, and twelve lines earlier hand-writes the same two-runtime alternation over `_sum`/`_count`. A third runtime means editing four expressions instead of two. Alerting on the existing `llmkube:inference:queue_time_seconds:p95_5m` recording rule collapses this to one runtime list and gives a percentile, which is also more robust to the single-slow-request case in finding 5.

**13. `charts/llmkube/README.md` values table is not updated.** Lines 145 to 152 document every `prometheus.prometheusRule.rules.*` parameter. The four new keys, including an on-by-default group and two tunable thresholds, are undocumented.

**14. The vLLM dashboard now duplicates the recording rule.** `vllm-dashboard.json:415` computes `histogram_quantile(0.95, ...rate(vllm:request_queue_time_seconds_bucket...))` at query time, while `sglang-dashboard.json:461` reads `llmkube:inference:queue_time_seconds:p95_5m`. Now that the recording rule covers vLLM, switching that panel closes the gap and drops the raw-bucket cost per refresh.

**15. The p50 recording rule has no test.** Both p95 and p50 changed, but "should cover both runtimes in the queue-wait recording rule" asserts only against p95. A later revert or typo in p50 leaves the suite green and the vLLM p50 panel empty.

## On the caveat

You offered to gate `serving.enabled` off by default for a first release. I would rather fix findings 1 and 2 and ship it on, since an alert group that ships off tends to stay off. The thresholds themselves I am not worried about; they are values-driven and the headroom is documented.

Findings 1, 2 and 9 are the ones I would want addressed before merge. Happy to pair on the expression rewrite if that is useful, and if you would rather split the recording-rule fix into its own PR I will merge that side quickly.

Tanguille:
Pushed 5b2f04a for 1-9, then 8d02c1b for 10/12/14.

1/7: swapped the instantaneous `running == 0` check for `avg_over_time(...[10m]) <= 1`. That directly covers the Running:1/Waiting:5 incident, and the averaging also kills the flap problem in 7 since a single scrape ticking above the ceiling no longer resets the pending timer.

2: with 1 fixed, InferenceAdmissionStalled now actually catches the total-stall case too (running <= 1 and waiting > 0), so between the two alerts the incident isn't missed anymore. Didn't touch QueueTimeHigh's own NaN-on-zero-completions behavior — that's just what a rate-over-completions rule does when nothing completes, and AdmissionStalled is the one that's supposed to own the no-completions case.

3: gated the whole serving group on `inferencePodMonitor.enabled` as well as `rules.serving.enabled`. Same move as ControllerMetricsMissing. Doesn't really cost us the "ship it on" goal since without that PodMonitor the metrics can't be scraped with the right labels anyway — turn it on and the rules just work.

4: added InferenceMetricsMissing.

5: min-sample guard (`count > 5`) on both QueueTimeHigh and KVOffloadLoadSlow.

6: QueueTimeHigh groups by pod now instead of service.

8: QueueTimeHigh dropped to warning. Left AdmissionStalled at critical, that one's a real non-admission condition not a saturation signal.

9: swapped the notMatchRegex test for lengthEqual rule-count checks (3 default, 4 with kvOffload on).

Also did the README table (13) and added the p50 test that was missing (15).

Went back and did 10, 12 and 14 too:

10: extended `dashboard_sync_test.go`'s metric-name validator to also read alert `expr:`, not just recording rules and dashboards. Turned up two real gaps in the allowlist that predate this PR: `up` (Prometheus's own synthesized series, used by InferenceServiceDown/ControllerDown/ControllerMetricsMissing) and `vllm:kv_offload_lookup_async_delay_seconds` (real upstream, just missing from the fixture). Both fixed.

12: InferenceQueueTimeHigh now alerts straight off `llmkube:inference:queue_time_seconds:p95_5m` instead of re-deriving the sum/sum ratio per runtime. A third runtime is a one-line record edit now, not four expressions.

14: switched the vLLM dashboard's queue-wait panel off its own `histogram_quantile` onto that same record, matching what the SGLang dashboard was already doing.

One tradeoff from doing 12: the record is now grouped by pod (it has to be, since the alert reads off it directly and per-pod resolution is what 6 was for), which makes it inconsistent with the sibling request_latency/ttft_seconds records that stay service-level, and permanently adds pod cardinality to it. Left those two alone since nothing depends on them being per-pod and it felt like scope creep to touch them for consistency's sake.

helm unittest (85/85), go test ./..., helm lint all green.

Defilan (changes requested):
Thanks for turning all fifteen around, and for going back for 10, 12 and 14 rather than leaving them. I re-checked every item against the branch, evaluated the rendered rules with `promtool` rather than reading them, and ran the gate. Fourteen of the fifteen are genuinely fixed. One of them broke something, and the fix I asked for is what broke it.

## Blocking

**`InferenceQueueTimeHigh` can no longer fire at all.**

Finding 12 asked you to alert off the p95 record instead of re-deriving the ratio, and you did. But that record stamps `latency_kind: queue_wait` via its `labels:` block, and the min-sample guard is `sum by (service, namespace, runtime, pod)` with no such label. PromQL `and` matches on the entire label set, so the join is empty and the alert is dead in every input.

Evaluated against a pod at p95 = 57.5s with 30 completions in 15m, which clears both the 30s threshold and the `> 5` guard:

- the record alone: `57.5`
- the guard alone: `30`
- the two joined by `and`: **empty**
- `alert_rule_test` at 60m and 120m: **no alerts**

The `5b2f04a` form worked, because both sides were `sum by (...)` and carried identical labels. `KVOffloadLoadSlow` uses the same idiom and is fine for the same reason, which is exactly what makes this invisible on a read-through.

The fix is one operator:

```
llmkube:inference:queue_time_seconds:p95_5m
and on (service, namespace, runtime, pod) (
  ...
) > 5
```

I ran that and it yields `57.5`, so the alert fires as intended.

Worth saying plainly: this is a consequence of my own review note, not something you got wrong. Finding 12 was a good change and it should stay; it just needs the explicit matcher.

## Should fix before merge

**`InferenceAdmissionStalled` pages `critical` on a healthy default llama.cpp.**

`pkg/agent/executor.go:77` omits `--parallel` when `ParallelSlots <= 1`, because one slot is llama-server's own default. With one slot, `llamacpp:requests_processing` is only ever 0 or 1, so `avg_over_time(...[10m]) <= 1` is permanently true, and any sustained queue satisfies the other operand. A healthy engine completing requests normally at `processing=1, deferred=3` fires this alert at `severity: critical`.

The vLLM arm does not have this problem, since `max_num_seqs` defaults to 256 and a real load climbs above 1. The llama.cpp arm needs a throughput term to separate "stalled" from "busy and saturated" — something like requiring `rate(llamacpp:tokens_predicted_total[10m])` to be flat alongside the queue condition, so a single-slot engine that is actually producing tokens does not page.

Related, and the reason I would not just raise the ceiling: with running oscillating 0, 1, 2, the 10m average lands near 1.05 / 1.00 / 0.95 on consecutive evaluations, so the condition flips each scrape and `for: 10m` is never satisfied. That happens to suppress this false positive, but it also means a real stall oscillating around 1 does not fire either.

**The exclusion comment is not accurate.** The new comment says TGI, PersonaPlex, Generic and LlamaCppRouter "export no queue metrics, so they're left out". `internal/controller/runtime_llamacpp_router.go:130` appends `--metrics` to llama-server and its `ContainerName()` is `llama-server`, so router pods are scraped and do emit `llamacpp:requests_*`. They are not excluded, they are covered, and they are exposed to the false positive above.

## Non-blocking

**The `up` allowlist entry widened the thing meant to catch typos.** `internal/metrics/dashboard_sync_test.go:50` went from `"up:"` to `"up"`, and `emitted()` matches with `strings.HasPrefix`, so it now allowlists every metric starting with `up` — `uptime_*`, `upstream_*`. An exact-match entry for `up` plus the existing `up:` prefix keeps the guard tight. Extending that validator to alert exprs was the right call and I am glad you took it.

**Finding 14 adds a prerequisite the panel does not state.** The vLLM queue-wait panel previously computed `histogram_quantile` from raw buckets and needed only the PodMonitor. It now reads the recording rule, which exists only when `prometheus.prometheusRule.enabled` is true, and that defaults false independently of `grafana.dashboards.enabled`. `values.yaml` already warns that a blank panel "is indistinguishable from an idle cluster". The SGLang panel had this coupling already; this extends it to vLLM. A line in the README values table, or a panel description, is enough.

## On two of your calls

**Leaving `InferenceQueueTimeHigh`'s NaN-on-zero-completions behaviour alone (finding 2): I accept it, with one caveat.** With finding 1 fixed, `AdmissionStalled` does own the total-stall case, so the split is coherent. The caveat is that a stall with requests wedged in-flight and no new arrivals — running above 1, waiting at 0 — is caught by neither: `AdmissionStalled` is silent because the average is above the ceiling, `QueueTimeHigh` is NaN because nothing completes. Not a blocker and not new to this PR, but the "AdmissionStalled owns it" argument is narrower than it reads.

**The pod-cardinality tradeoff on the p95 record: good call, and thank you for flagging it rather than letting me find it.** Per-pod is right precisely because a service-wide percentile dilutes one stalled replica, and you updated both dashboard legends to `{{service}}/{{pod}}` so nothing reads wrong. Leaving the sibling `request_latency`/`ttft_seconds` records service-level was also the right instinct; consistency for its own sake would have been scope creep.

**One thing I owe you.** Finding 3 asked for the `inferencePodMonitor.enabled` gate, and since that defaults false, the serving group now ships off in a default install — the exact outcome I said I wanted to avoid when I wrote "an alert group that ships off tends to stay off". That is my error, not yours; the gate is correct given the label dependency. If you want to make the PodMonitor default on in a follow-up, I would support it, but it does not belong in this PR.

## Gate

`helm unittest` 85/85, `helm lint` clean, `promtool check rules` on the rendered output SUCCESS with 17 rules, `go test ./internal/metrics/...` ok, and all twelve serving-group metric names present in the fixtures. The item-3 gating renders exactly as described: group absent at defaults, 3 alerts with the PodMonitor on, 4 with kvOffload on. The item-9 swap to `lengthEqual` is a real test, and the two remaining `notMatchRegex` gating tests are not vacuous.

Worth noting that the whole gate is green while `InferenceQueueTimeHigh` is dead: `helm unittest` regex-matches rendered YAML and `dashboard_sync_test.go` validates metric names, so nothing in the repo evaluates PromQL semantics. That is a gap in our tooling rather than anything you did, and it is what finding 10 was circling. If you felt like adding a small `promtool test rules` suite to CI as a follow-up, it would have caught this in seconds.

Defilan:
Pushed 8849553 to carry the remaining items over the line. All five are addressed; details and the evidence below.

**Blocking: `InferenceQueueTimeHigh`.** Fixed with the explicit matcher, exactly as suggested. I reproduced the dead alert before changing anything rather than taking the diagnosis on faith — at p95 = 60s with 30 completions in 15m:

| form | 60m | 120m |
|---|---|---|
| bare `and` | no alerts | no alerts |
| `and on (service, namespace, runtime, pod)` | fires, value 60 | fires, value 60 |

Reverting the operator makes that promtool test fail again, so it is not vacuous.

One thing worth knowing for next time: my first attempt put the explanation as a `#` comment *inside* the `expr: |` block, and `TestAlertExprsQueryEmittedMetrics` immediately failed with `alert InferenceQueueTimeHigh queries "stylistic"` and a dozen similar. The validator tokenizes the whole block scalar, so prose inside `expr:` becomes metric names. The comment now sits above the rule like the others. That validator earned its keep twice in this PR.

**`InferenceAdmissionStalled`.** The llama.cpp arm now gates on throughput rather than the running-count ceiling: a persistent queue with `rate(llamacpp:tokens_predicted_total[10m]) == 0`.

I dropped the `<= 1` term on that arm rather than adding to it, because the ceiling is doing no work there — with one slot `requests_processing` is only ever 0 or 1, so it is permanently true — and because of the oscillation problem raised in the review: a real stall hovering around 1 never satisfies `for: 10m` either. Zero predicted tokens while a queue persists is the honest definition of "not admitting work", and it holds for one slot and for many. vLLM and SGLang keep the ceiling, where `max_num_seqs` defaults to 256.

Verified both directions with promtool: a healthy single-slot engine at `processing=1, deferred=3` producing 120 tok/min no longer alerts; a real stall with the same queue and flat tokens still does; and the previous expression fires on the healthy case.

**Exclusion comment.** Corrected. Confirmed at `internal/controller/runtime_llamacpp_router.go:136` that the router appends `--metrics`, and `ContainerName()` is `llama-server`, so router pods emit `llamacpp:*` and are covered by the third arm rather than excluded by it.

**`up` allowlist.** Split into an exact-match `externalExact` for `up` plus the original `up:` prefix, so `uptime_*` and `upstream_*` are no longer silently allowlisted.

**Finding 14 prerequisite.** Documented in both `values.yaml` (beside the existing "indistinguishable from an idle cluster" note) and the README values table, since the queue-wait panels read the p95 recording rule and `prometheusRule.enabled` defaults false independently of `dashboards.enabled`.

**Gate:** `helm lint` clean, `helm unittest` 85/85, `promtool check rules` SUCCESS on 16 rules, `go test ./internal/metrics/...` ok, `gofmt` clean.

---

On the tooling gap noted in the review — that the whole gate stayed green while the alert was dead — the two promtool suites I used are what would have caught it. I have deliberately kept them out of this PR so it stays scoped to the review items, but they are short and I am happy to open the follow-up that wires `promtool test rules` into CI. Say the word and it is yours or mine, whichever you prefer.

Thanks for the thorough back-and-forth on this one, and sorry it sat for a few days.

Issue #1625, opened 2026-08-21
[BUG] Empty payload.repo silently defeats the upstream-base fetch: task branches cut from a stale fork HEAD


Defilan:
## Bug Description

When an AgenticTask's `payload.repo` is empty and the agent runs with a statically configured `--git-remote-url` (the normal single-repo fork setup), the task branch is cut from the **cloned fork's HEAD**, not the current upstream base. A fork whose `main` has drifted silently produces workspaces that are arbitrarily stale, and nothing surfaces it.

Observed live 2026-08-21: five overnight workloads (no `payload.repo`, agent cloning `Defilan/LLMKube`) all cut branches from `c95e06f0` (2026-07-23) while upstream main was `d0db838a6` (2026-08-20) — a month of drift. The pre-flight staleness check (#1550) caught it (`NO-GO`, "bug not present on this base"), which is the only reason no stale-based diff was produced. The ~8 GO'd branches from earlier runs this week predate that check's deployment and need staleness triage for the same reason.

## Where the gap is

- `pkg/foreman/agent/executor_native.go` (`setupTaskBranch`, ~line 608): `if upstreamURL := resolveUpstream(task.Spec.Payload.Repo); upstreamURL != ""` → `repo.CreateBranchFromUpstream` (the #813 fix, which does the right thing) — else falls through to `repo.CreateAndCheckoutBranch`, i.e. fork HEAD.
- `pkg/foreman/agent/codehost/codehost.go` (`ResolveCloneURL` contract, ~line 79-83 and 113-117): documented to return `""` for an empty slug "so callers fall back to the cloned fork's HEAD". The fallback is the hazard: for issue-fix tasks it silently defeats #813.
- `pkg/foreman/agent/claim_gate.go` (~line 80): assumes branches are "cut from the CURRENT upstream tip (repo.CreateBranchFromUpstream, #813)" — the empty-slug path violates that assumption, so the claim-evidence gate's evidenceBaseSHA and the actual branch base can diverge.

## Chosen fix (implement exactly this)

[...]

Issue #1297, opened 2026-07-27
[FEATURE] Pluggable code hosting and work tracking: Forgejo, GitLab, and a path for third-party providers


Defilan:
## Feature Description

Let a Foreman user run against the forge they actually have. Ship GitHub, a Gitea-family provider covering Forgejo, and GitLab as built-ins, behind a selection mechanism, with the interfaces kept clean enough that a third party can add a provider without forking.

This is the follow-up work #1158 anticipated. That issue deliberately scoped itself to extracting the boundaries and said "Forgejo, Linear, and Woodpecker integrations can be added in follow-up issues after the boundaries are agreed." The boundaries landed in 0.9.12. These are the follow-ups.

## Problem Statement

Three interfaces exist today: `CodeHost`, `WorkItems`, and `ChangePolicy`. All three are injectable in Go. None is selectable by a user, and two are bypassed at runtime, so a correct third-party implementation would compile, satisfy the interface, and still be routed around.

The git path is already provider-neutral and works. `repo/push.go` and `repo/clone.go` contain zero GitHub references, `agent.gitRemoteURL` accepts any git URL, and the token is used as an opaque credential through a `GIT_ASKPASS` shim. What is GitHub-bound is everything that talks to an API rather than to git.

## Proposed Solution

Five pieces, in dependency order. The first two are prerequisites for any provider work, because they would silently break an otherwise-correct adapter.

- [ ] #1298 Make the seams hold. Three consumers bypass them today.
- [ ] #1299 Repo identity beyond `owner/name`, so subgroups and nesting are expressible.
- [ ] #1300 Provider selection: a registry plus `forge` and `apiBaseURL`, defaulted by host.
- [ ] #1301 A Gitea-family adapter. One adapter covers Forgejo, Gitea, and Codeberg.
- [ ] #1302 Make `ChangePolicy` configurable.

#1298 and #1299 gate #1301: without them a correct adapter returns github.com clone URLs and rejects nested namespaces. #1302 is independent and can run in parallel. GitLab follows the same shape as #1301 once the registry exists.

### Design constraints worth agreeing before code

**Hand-roll the adapters. Do not adopt a multi-forge library.** Every candidate was evaluated against the four operations Foreman actually needs, and each fails on a different one:

| Library | Forgejo | All four ops? |
|---|---|---|
[...]

PR #1626, opened 2026-08-21
fix: Empty payload.repo silently defeats the upstream-base fetch: task branches cut from a stale fork HEAD (#1625)

Defilan:
## What

Repo-bearing task kinds (`issue-fix`, `verify`, `review`) with an empty or malformed `payload.repo` now fail with a configuration error at branch setup instead of silently basing the task branch on the cloned fork's HEAD. Freeform tasks keep the fork-HEAD fallback unchanged.

## Why

Refs #1625

The fork a deployment clones can lag upstream by an arbitrary amount (observed live: a month), and the empty-slug fallback silently defeated the #813 upstream-base fetch — five overnight workloads based on a July tree before the pattern was caught. The fix is the payload, not a workaround, so a missing slug is refused with a message naming the remedy.

## How

`setupTaskBranch` keys the refusal on `codehost.IsValidRepoSlug(payload.repo)`, not on the resolved URL: a valid slug whose resolver degrades to "" keeps its existing, deliberate fail-open path (covered by `TestNativeExecutor_WorkClassPolicy_FailsOpenWhenBaseUnresolved`). Doc comments in `codehost.go` (ResolveCloneURL contract) and `claim_gate.go` (stale-base assumption) updated to state the new invariant. Explicitly out of scope: fork-parent resolution, upstream config knobs, webhook validation (they overlap #1297/#1300).

## Verification

- Three new tests, all passing: issue-fix with empty repo refuses and creates no branch; freeform with empty repo falls back exactly as before; issue-fix with a valid slug still cuts from the upstream tip
- Full `pkg/foreman/...` suite green; `GOOS=linux golangci-lint` and the dead-code guard: 0 issues
- Pipeline: coder GO, gate GATE-PASS (fresh-clone envtest), reviewer GO; the branch itself was cut from current upstream main via the exact code path this PR hardens

## Checklist

- [x] Tests added/updated
- [x] AI assistance is disclosed below, per CONTRIBUTING.md
- [x] Documentation updated (doc comments carry the new invariant)

Assisted-by: Foreman (qwen38-coder) authored the change from the issue's mandated design; gate-verified, reviewer-approved (Nemotron), and human-reviewed pre-PR (diff, tests re-run, lints) by the maintainer's session.

### DIFF
--- a/pkg/foreman/agent/claim_gate.go
+++ b/pkg/foreman/agent/claim_gate.go
@@ -77,7 +77,9 @@ var claimDocsPathspec = []string{"*.md", "docs", "examples"}
 //   - Fork-lag false positive: in a fork-based deployment, --git-remote-url
 //     points origin at a fork that can lag the upstream project by an
 //     arbitrary amount, while setupTaskBranch always cuts the task branch
-//     from the CURRENT upstream tip (repo.CreateBranchFromUpstream, #813).
+//     from the CURRENT upstream tip (repo.CreateBranchFromUpstream, #813) —
+//     for repo-bearing kinds an empty payload.repo is now a configuration
+//     error (#1625), so the stale fork-HEAD fallback cannot produce one.
 //     Deriving the anchor from `git merge-base HEAD origin/<baseBranch>`
 //     inside the workspace landed on the STALE fork tip, so AddedLines swept
 //     the entire upstream lag delta into the diff and flagged other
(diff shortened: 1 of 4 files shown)


Defilan:
Pre-PR review record: branch fetched and diffed against current main, the three mandated tests plus the full `pkg/foreman` suite re-run locally (green), cross-arch lint and dead-code guard clean. One design note beyond the issue's mandate, verified correct: the refusal keys on slug validity rather than resolved-URL emptiness, which preserves the pre-existing deliberate fail-open path for a valid slug whose resolver degrades — the distinction matters and the coder both implemented and documented it unprompted.

Issue #1636, opened 2026-08-23
[BUG] issueAsk demotion opens a revision cycle whose feedback is an APPROVE with zero findings


Defilan:
## Bug Description

When the issueAsk integrity rail demotes an approved review to NO-GO, the
Workload reconciler treats that synthetic NO-GO as a real rejection and opens a
revision cycle. The feedback body handed to the revision coder is the reviewer's
own **APPROVE** summary followed by `- findings: []`.

The coder is told to address feedback that does not exist, on a branch the
reviewer approved. On a local 27B coder that is roughly an hour of GPU time per
occurrence, and the cycle can repeat: nothing about a re-push makes the next
issueAsk verification any more likely to pass.

This is distinct from #1454 (reviewer returns GO while stating it could not
verify) and #1618 (enforce reviewer steps by position). Both of those are about
the reviewer's own reasoning. This one is about the harness discarding a verdict
it already has the information to interpret correctly.

## Steps to Reproduce

1. Run a Workload whose issue body describes the problem in prose without naming
   the file paths the fix will touch. A chart-only fix for an issue written about
   runtime behaviour is the easy case.
2. Let the reviewer approve it. The reviewer will summarise the issue ask rather
   than quoting it verbatim.
3. Observe the review task land NO-GO, and a `-r1` revision cycle spawn.

Observed live on 2026-08-22 with `wl-1438-agent-rollout-values` against #1438.

## Expected Behavior

A demotion is a statement about verification confidence, not about the change.
The reconciler should distinguish the two. Either:

- do not open a revision iteration when the NO-GO came from a demotion with no
  findings, or
- render the actual demotion reason so the coder is told what to fix (the
  reviewer paraphrased the issue ask) rather than being pointed at an approval.

Re-running the coder cannot fix an unverifiable issueAsk, so the current
behaviour cannot converge.

## Actual Behavior

The revision prompt for `wl-1438-agent-rollout-values-code-1438-r1`, verbatim:

```
The reviewer rejected the previous attempt on this issue (verdict NO-GO).
Your workspace already contains that previous attempt: the working branch was
restored from this task's branch on the remote, so its files and history are
present. Do not rebuild the fix from scratch. Amend the existing work with the
smallest changes that address every point below, then push to the same branch.

Reviewer review-1438-0: APPROVE: diff exposes rollout strategy and grace period
per pool in Helm chart, aligning with #1438's request for drain-before-roll.
Changes are minimal, well-tested, and documented.
- findings: []

[...]

PR #1641, opened 2026-08-23
fix(foreman): do not open a fix iteration for an empty demoted NO-GO

Defilan:
## What

Stop the Workload reconciler opening a fix iteration for a NO-GO that a harness
rail manufactured and that carries no findings.

## Why

Refs #1636

`enforceReviewerIssueAsk` (`pkg/foreman/agent/executor_native.go`) rewrites a
reviewer's GO to NO-GO when it cannot verify that the reviewer's stated issue ask
covers the fetched issue body. That is a statement about verification confidence,
not about the change under review, but `noGoReviewRound` could not tell the two
apart.

The result is a fix iteration whose prompt says "the reviewer rejected the
previous attempt (verdict NO-GO) ... address this feedback", followed by the
reviewer's own approval and `- findings: []`. There is nothing to act on, and
re-running the coder cannot make an unverifiable issueAsk verify, so the loop has
no terminating condition.

Observed twice on one branch while working through the Tier 1 backlog. Each
occurrence costs a full coder run, which on a local 27B is roughly an hour of GPU
time. The second occurrence is recorded in the issue: two rails flipped the
verdict in sequence (NO-GO to GO to NO-GO) and the value that survived was
asserted by nobody.

## How

`isInertDemotion` reads `extra.verdictDemoted` and the findings already present
on the task, and `noGoReviewRound` skips a NO-GO that is both demoted and
empty-handed. The predicate keys off a new `verdictDemotedBy` marker that names the rail, because `verdictDemoted` alone is set by several rails: `enforceReviewerIssueAsk` already stamps
`verdictDemoted` and `demotionReason`, and nothing read them.

The predicate is deliberately narrow, and two of the three new test cases exist
to hold that line:

- a demotion **carrying findings** still iterates, because those findings are
  real work
- an **undemoted** NO-GO always iterates, even with no structured findings,
  because that is the reviewer's own judgement and the summary carries the
  feedback

Malformed result JSON is not suppressed either; it falls through and iterates as
before.

The non-conforming findings fallback mirrors `reviewFeedbackSection`, so anything
that would actually reach the coder as raw JSON counts as feedback rather than
inertness.

A drive-by from `unparam`: `noGoChild`'s `step` parameter always received the same
value, so it is now a shared const, removing the argument from four call sites.

## Verification

- `make test` passes (RC=0, 0 FAIL lines); `internal/foreman/controller` at 83.7%
- `make lint` and `make lint-all` both report 0 issues
- `make lint-deadcode` reports 0 issues
- Watched the new suppression case fail first, for the right reason (the
  iteration was produced when it should not have been), then pass

## Checklist

- [x] Tests added/updated
- [ ] Documentation updated (if user-facing change) — no user-facing surface changes

Assisted-by: Claude Code (wrote the tests and the implementation under TDD, and
reproduced the failure from the live audit records; I reviewed the change and ran
`make test`, `make lint`, `make lint-all` and `make lint-deadcode` before
submitting).

### DIFF
--- a/internal/foreman/controller/workload_gateless_test.go
+++ b/internal/foreman/controller/workload_gateless_test.go
@@ -162,7 +162,7 @@ func TestReviewIterationSteps_GatelessReviewDependsOnCode(t *testing.T) {
 		child("review-641-0", foremanv1alpha1.AgenticTaskPhaseSucceeded, foremanv1alpha1.AgenticTaskVerdictNoGo),
 	}
 
-	steps, iterated := reviewIterationSteps(w, children)
+	steps, iterated, _ := reviewIterationSteps(w, children)
 	want := []string{"code-641-r1", "review-641-0-r1"}
 	if len(steps) != len(want) {
 		t.Fatalf("got %d steps, want %d (%v)", len(steps), len(want), want)
--- a/internal/foreman/controller/workload_intent_test.go
+++ b/internal/foreman/controller/workload_intent_test.go
@@ -107,7 +107,7 @@ func TestReviewIterationSteps_CarriesIntent(t *testing.T) {
 	children := []foremanv1alpha1.AgenticTask{
 		child("code-641", foremanv1alpha1.AgenticTaskPhaseSucceeded, foremanv1alpha1.AgenticTaskVerdictGo),
 		child("verify-641", foremanv1alpha1.AgenticTaskPhaseSucceeded, foremanv1alpha1.AgenticTaskVerdictGatePass),
-		noGoChild("review-641-0", "needs work", "[]"),
+		noGoChild("needs work", "[]"),
 	}
 
 	steps, _ := reviewIterationSteps(w, children)
--- a/internal/foreman/controller/workload_intent_test.go
+++ b/internal/foreman/controller/workload_intent_test.go
@@ -110,7 +110,7 @@ func TestReviewIterationSteps_CarriesIntent(t *testing.T) {
 		noGoChild("needs work", "[]"),
 	}
 
-	steps, _ := reviewIterationSteps(w, children)
+	steps, _, _ := reviewIterationSteps(w, children)
 
 	var found bool
 	for _, s := range steps {
--- a/internal/foreman/controller/workload_iteration.go
+++ b/internal/foreman/controller/workload_iteration.go
@@ -40,6 +40,14 @@ import (
 // chained behind it.
 const conditionTypeReviewIterationTriggered = "ReviewIterationTriggered"
 
+// conditionTypeReviewIterationSuppressed reports that a terminal NO-GO did
+// NOT open a fix iteration because the issueAsk rail manufactured it out of
+// the reviewer's own GO and it carries no findings (#1636). The demoted
+// NO-GO still counts as incomplete in the rollup, so without this the
+// operator sees a Failed Workload on a branch the reviewer approved and
+// nothing anywhere says why no iteration ran.
+const conditionTypeReviewIterationSuppressed = "ReviewIterationSuppressed"
+
 // defaultMaxReviewIterations is the per-issue fix-iteration bound when
 // Workload.spec.maxReviewIterations is unset. One iteration mirrors
 // what a human does with a single request-changes round; an explicit 0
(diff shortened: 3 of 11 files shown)


joryirving (changes requested):
I don't think this fix can work as written, so requesting changes on one blocker plus two narrowness problems that only bite once the blocker is fixed.

The blocker: `isInertDemotion` reads `extra.verdictDemoted` (`internal/foreman/controller/workload_iteration.go:162`), but the rails never write there. `enforceReviewerIssueAsk` stamps into `loopRes.Terminal.Extra` (`pkg/foreman/agent/executor_native.go:951` passing the map, keys set at `:3130`-`:3157`), and `modelDecidedResult` nests that whole map one level down as `extra.modelExtra` (`pkg/foreman/agent/executor_native.go:2160`-`:2166`), promoting only `outcome`, `unverified` and `resolvedBy` to the top level (`promoteTerminalOutcome`, `:2191`). I ran the real path (rail plus `modelDecidedResult`, marshalled) and got `{"extra":{"modelExtra":{"demotionReason":"...","verdictDemoted":true,"verdictClaimed":"GO",...},"outcome":"MODEL-DECIDED","turnCount":5}}`. So in production the predicate returns false at `:170` every time and #1636 reproduces exactly as before. `pkg/foreman/audit/record.go:97`-`:107` is the confirmation from the other direction: every sibling key the same rails set (`issueAskVerified`, `scopeDriftDetected`, `scopeMatched`, `reviewOutcome`) is read from `extra.modelExtra`, and `workload_coder_escalation.go:44` does the same for the coder-gate case.

The reason CI is green is the fixture. `demotedNoGoChild` (`workload_iteration_test.go:56`-`:62`) puts `verdictDemoted` as a sibling of `modelExtra`, a shape the executor never emits, while `noGoChild` two functions down nests correctly. I confirmed the test does fail if the predicate call is reverted, so it is a real assertion, just against invented JSON. What I'd do: read `extra.modelExtra.verdictDemoted` (same struct you already use for the findings), and build the fixture by marshalling an actual `agent.Result` from `modelDecidedResult` rather than hand-writing the envelope, so the shape can't silently drift again.

Second, `verdictDemoted` is not the issueAsk rail's private flag, and once the path is fixed the predicate will fire on cases you didn't intend. `enforceReviewerScopeOverlap` sets it too (`pkg/foreman/agent/scope_overlap.go:505`) when the issue names files and the diff touches none of them. That demotion is a GO with empty findings, so it satisfies your predicate exactly, and suppressing it is wrong on the merits: scope drift is actionable feedback ("you edited files the issue doesn't name"), and unlike an unverifiable issueAsk, re-running the coder genuinely can fix it, so the non-convergence argument in the doc comment does not transfer.

Third, and this is the one I'd worry about most, `enforceReviewerIssueAsk` also stamps `verdictDemoted=true` on a verdict it did not change: `executor_native.go:3130`-`:3138` marks an unverified non-GO review as untrusted and returns the reviewer's own NO-GO. Nothing was demoted there, so a genuine summary-only rejection gets classified inert and the fix iteration is dropped, with the reviewer's prose in `summary` thrown away. I probed the predicate with that envelope shape and it returns `inert=true`. Your own third test case ("genuine NO-GO with no findings still iterates") asserts the opposite principle, and it only passes because the fixture omits the marker. That path is reachable through the grounded-finding rail's REJECT exemption (`grounded_finding_gate.go:135`-`:138`), which is precisely how a findings-less NO-GO survives to the controller. Concretely, I'd gate on `verdictClaimed == "GO"` so only an actual rewrite qualifies, and I'd have the rail identify itself (something like `extra["verdictDemotedBy"] = "issueAsk"` at `:3156`, `"scope-overlap"` at `scope_overlap.go:505`) and check that, rather than inferring the source from a flag three rails share.

Last thing, smaller: when suppression fires, nothing anywhere records it. `isInertDemotion` and `noGoReviewRound` have no logger, no condition, no event, and the demoted NO-GO still lands in the `incomplete` bucket at `workload_controller.go:668` and rolls the Workload to Failed. So the operator gets a Failed Workload on a branch the reviewer approved and nothing that says the iteration was deliberately skipped. That is the same gap `emitReviewIterations` already refuses to accept twenty lines further down ("No silent cap: report why the fix iteration did not run", `workload_iteration.go:565`) and that `recordRailSkipped` exists for on the agent side ("a log line is not a record (#1605)", `scope_overlap.go:496`). I'd set a condition or emit an event naming the task and the `demotionReason`, which also answers the observability question the issue's second option was reaching for.

Two corrections to the prose while you're in there: `applyIssueAskIntegrity` does not exist. The function is `enforceReviewerIssueAsk` (`executor_native.go:3115`), and the name appears in the doc comment at `workload_iteration.go:143`, in the test comment at `workload_iteration_test.go:51` and in the PR body. Grep for it and you find only your own comments.

The parts I think are right: not touching `escalationSteps` is the correct call, because the demotion's stated purpose at `executor_native.go:3103` is routing to a bigger model, and `workload_escalation.go:91` still sees the NO-GO, so narrowing only the iteration path preserves that. Letting malformed JSON fall through to iterate is the right default. The `noGoChild` step-parameter cleanup is fine.

Defilan:
You're right, and this was a no-op in production. I verified it before fixing: `promoteTerminalOutcome` lifts only `outcome`, `unverified` and `resolvedBy`, so `verdictDemoted` never leaves `extra.modelExtra` and the predicate returned false every time. The function was already inconsistent with itself, reading findings from `modelExtra` and the flag from `extra` on adjacent lines, and I missed it.

Pushed `1b0a6035b`. I took your `verdictDemotedBy` suggestion rather than just repointing the read, for the reason you gave: repointing alone fixes the blocker but leaves the predicate keyed on a flag several rails share, so it would then start suppressing scope-drift demotions, which should iterate. Both rails now name themselves and the predicate requires the issueAsk marker, `verdictClaimed == "GO"`, and no findings.

One deliberate difference from your suggested placement. You pointed at the GO to NO-GO branch; I stamp the marker where `verdictDemoted` itself is set, so it covers the untrusted-non-GO path too. If the marker only existed on the true-demotion branch, the `verdictClaimed == "GO"` check would be unreachable, and I wanted that check to be the thing doing the work rather than dead code. It also keeps the invariant that anything finding `verdictDemoted` can ask who set it.

New cases for both paths you identified: scope-overlap demotion iterates, untrusted-non-GO iterates. Mutations confirm each bites, including reverting the read to `extra` (3 failures) and pointing the marker at `scope-overlap` (3 failures, with scope drift wrongly suppressed).

`modelDecidedResult` is unexported so the controller test package genuinely cannot reach it. The fixtures marshal a real `agent.Result` with the same nesting and a comment naming the source, and `TestIssueAskDemotionLandsUnderModelExtra` closes the drift hole from the agent side instead.

Suppression is now recorded rather than silent, and I fixed `applyIssueAskIntegrity` in the code, comments and the PR body. It only ever existed in my own prose.

joryirving:
Re-reviewed at `1b0a603`. All four points are addressed, and two of them better than I suggested.

The predicate now reads `envelope.Extra.ModelExtra` rather than `extra` directly, so it matches what `modelDecidedResult` actually emits. The comment recording that reading from `extra` "never matched anything the executor emits" is worth keeping: that failure was invisible because the fixture was hand-written into the shape the predicate expected rather than produced by the executor.

`verdictDemotedBy` landed as a shared constant in `pkg/foreman/agent/reviewer/demotion.go`, which is the right home — the producer in `pkg/foreman/agent` and the consumer in `internal/foreman/controller` need the same strings and neither imports the other. The doc comment also states the distinction that matters: the issueAsk rail's demotion is about verification confidence and re-running the coder cannot change it, while scope-overlap's is about the diff and the coder can act on it. That is the reason keying off the bare flag was wrong, written where the next person will find it.

The `verdictClaimed == GO` gate closes the case I was most worried about, where `enforceReviewerIssueAsk` stamps the demotion fields on a path that returns the reviewer's own NO-GO without rewriting anything. A genuine summary-only rejection is no longer discarded.

Two things you added that I did not ask for and that improve it. The non-conforming findings check mirrors `reviewFeedbackSection`'s fallback, so a findings blob that `ParseFindings` cannot read but which still reaches the coder as raw JSON is correctly not treated as inert — I had said "no findings" without noticing that empty-after-parse is not the same as nothing-reaches-the-coder. And the suppression now emits both a `ReviewIterationSuppressed` condition with the per-step demotion reasons and a Warning event, with change-detection so it does not churn the condition on every reconcile. I asked for one or the other.

Good to merge from my side.

Issue #1637, opened 2026-08-23
[BUG] Foreman gate's DefaultGateChecks is a strict subset of CI, so GATE-PASS does not imply CI-green


Defilan:
## Bug Description

`DefaultGateChecks` in `pkg/foreman/agent/tools/run_gate_job.go:68` is a strict
subset of what CI blocks on. Five checks that fail a pull request in
`.github/workflows/lint.yml` are never run by the Foreman gate, so a branch can
earn `GATE-PASS` and then fail CI.

The gate template already states the principle this violates. From
`pkg/foreman/agent/tools/gate_job_template.yaml:105-107`:

> Versions are pinned to CI's so the gate and CI validate charts with the same
> tooling; a gate that disagrees with CI is worse than no gate.

That reasoning was applied to helm and helm-unittest versions but not to the
check list itself.

## Actual Behavior

The gate runs these make targets (`run_gate_job.go:68`):

```go
var DefaultGateChecks = []string{
	"fmt", "vet", "lint", "test",
	"manifests", "chart-crds", "foreman-chart-crds", ChartCheck,
}
```

`lint.yml` runs on `pull_request` and additionally blocks on five checks absent
from that list:

| Check | Guards | In gate image already? |
|---|---|---|
| `lint-deadcode` (as `--config=.golangci-deadcode.yml`, `lint.yml:34`) | production code referenced only by its own tests | yes, golangci-lint |
| `generate` | DeepCopy codegen drift (gate runs `manifests` but not `generate`) | yes, controller-gen |
| `check-helm-rbac` | chart RBAC covers every kubebuilder-generated rule (#379) | yes, via `manifests` |
| `check-reviewer-prompts` | reviewer `systemPrompt` drift from `reviewer.md` (#804) | yes, `go run ./cmd/sync-reviewer-prompts --check` |
| `validate-samples` | `config/samples` against CRD schemas | **no**, needs python3 + jsonschema |

Verified that no other code path runs them: nothing under `pkg/foreman` or
`internal/foreman` references any of these targets.

## Expected Behavior

`GATE-PASS` should mean the branch is expected to pass CI. Today it means a
subset passed, and the subset is not stated anywhere the operator or the harness
can see.

## Steps to Reproduce

1. Dispatch a Workload whose change introduces a symbol referenced only by its
   own tests, or that edits a reviewer prompt, or that changes an RBAC marker.
2. Observe the verify task return `GATE-PASS`.
[...]

Issue #1602, opened 2026-08-19
[FEATURE] Triage four pre-existing dead production symbols (dead-code guard Register B)


Defilan:
## Feature Description

The dead-code guard (`make lint-deadcode`) reports four symbols in production packages that nothing outside their own tests calls. They are recorded in Register B of `.golangci-deadcode.yml` so the guard could go blocking immediately, but none of them has had a delete-or-wire decision. This issue is that decision.

| Symbol | Location | Open question |
|---|---|---|
| `isNeedsVerificationCoder` | `internal/foreman/controller/workload_coder_escalation.go:114` | Its own comment says "Used by `shouldEscalateCoder` to skip escalation", but `shouldEscalateCoder` does not call it. Either the comment is stale or a call was dropped. |
| `metalReadyEndpoints` | `internal/controller/inferenceservice_controller.go:866` | Method on `InferenceServiceReconciler` for metal endpoint readiness. Superseded by another path, or never wired? |
| `parseMaxUSD` | `internal/router/budget.go:262` | Budget parsing helper. ModelRouter budgets are still roadmap (#434, #440), so this may be deliberate staging. |
| `taskSucceeded` | `pkg/cli/dispatch.go:384` | CLI dispatch helper. |

`isNeedsVerificationCoder` is the one to look at first. A helper whose documentation names a caller that does not exist is more likely a dropped call than intentional staging, and this one sits on the coder escalation path.

## Problem Statement

> As a maintainer, I want every dead symbol either deleted or wired up, so the guard's suppression register shrinks to empty and the tree carries no code that pretends to be reachable.

An unbounded suppression list is how a guard stops meaning anything. Register B needs to reach zero.

## Proposed Solution

For each symbol: delete it, or wire it and remove its Register B entry from `.golangci-deadcode.yml`. This issue closes when Register B is empty.

Removing the entry is the deliverable, not an afterthought: it is what proves the symbol is reachable or gone.

## Alternatives Considered

- **Delete all four unexamined.** Rejected. `parseMaxUSD` may be deliberate staging for #434 / #440, and `isNeedsVerificationCoder` may indicate a real dropped call on the escalation path rather than dead weight. Deleting that one would hide a bug rather than fix it.
- **Leave them suppressed indefinitely.** Rejected. The register is debt, and a permanent suppression list defeats the guard.

## Additional Context

- Related issues: #1391 (functionally inert work passing gates, the class this guard mechanises)
[...]

PR #1642, opened 2026-08-23
fix(foreman): gate the make-invoked CI checks, and pin them against the workflows

Defilan:
## What

Add the make-invoked CI checks the Foreman gate was not running, and add a drift test so
the gate's check list cannot silently fall behind CI again.

## Why

Refs #1637

`DefaultGateChecks` was a strict subset of what CI blocks on, so a branch could
earn `GATE-PASS` and still fail its pull request. Missing: the dead production
code guard, DeepCopy codegen drift, the chart RBAC cross-check, and reviewer
prompt drift.

A `GATE-PASS` is not only read by humans. It feeds the verdict and escalation
machinery, and a coder self-gate uses the same path, so a partial check set makes
every downstream signal weaker than it appears.

The gate template already states the principle, for its helm pins:

> Versions are pinned to CI's so the gate and CI validate charts with the same
> tooling; a gate that disagrees with CI is worse than no gate.

It had never been applied to the check list itself.

This surfaced concretely on #1602, whose entire deliverable is that
`make lint-deadcode` still passes once a suppression register is emptied. The
gate returned `GATE-PASS` without ever running that guard, so it structurally
could not verify the change it was gating.

## How

Added `lint-deadcode` and `check-reviewer-prompts` to `DefaultGateChecks`; these
are the two real gaps. `generate` is added explicitly too, though it already ran
transitively via `test: manifests generate fmt vet setup-envtest`. `check-helm-rbac`
is NOT gated: `scripts/check-helm-rbac.sh` hard-exits without PyYAML and the gate
image is a plain `golang` image with no `pip3`, so gating it would fail every gate
run; it is in `gateExemptCIChecks` with that reason. These run on tooling the gate
image already fetches
(golangci-lint and controller-gen), so this adds no bootstrap cost.

`validate-samples` is deliberately left out and recorded as such: it hard-requires
python3 and the `jsonschema` package and exits non-zero without them, so adding it
naively would fail **every** gate run rather than none.

The durable half is `TestDefaultGateChecksCoverCI`. It scans every workflow for
make invocations and fails when CI gains a check the gate does not run.
Deliberate omissions live in `gateExemptCIChecks` **with a stated reason**, and
`TestGateExemptionsAreLive` fails when an exemption names a CI step that no longer
exists, so the list cannot decay into a blanket waiver.

Two things worth calling out about the extractor, both of which are why the test
is trustworthy rather than decorative:

- It handles `make` invoked inside a `run: |` block, not just `run: make ...`.
  My first version matched only the single-line form and missed `make test` in
  `test.yml`. A matcher that under-reports is precisely the failure mode this
  change is about, so it is worth being explicit that it was caught and fixed.
- It skips comment lines, because CI comments mention targets in prose
  (`# make test writes cover.out`, `Run 'make generate ...'`).

The issue description attributes all five gaps to `lint.yml`; that was wrong and
is corrected in a comment there. Only two block in `lint.yml`; the others are in
`helm-chart.yml` and `test.yml`, which is why the test scans all workflows rather
than one.

## Verification

- `make test` passes (RC=0, 0 FAIL lines)
- `make lint` and `make lint-all` both report 0 issues
- `make lint-deadcode` reports 0 issues
- Tree is clean after `make test` runs `generate`/`manifests`, so the newly gated
  codegen checks pass on current `main`
- Watched the drift test fail first, listing the gaps, and the
  liveness test fail on speculative exemptions until they were pruned to what CI
  actually runs

## Checklist

- [x] Tests added/updated
- [ ] Documentation updated (if user-facing change) — no user-facing surface changes

Note for reviewers: this makes the gate stricter, so the first gate run on any
branch with stale codegen or a dead symbol will now fail where it previously
passed. That is the intent, but it is a behaviour change for in-flight branches.

Assisted-by: Claude Code (wrote the tests and the implementation under TDD; I
reviewed the change, corrected the extractor after finding it under-reported, and
ran `make test`, `make lint`, `make lint-all` and `make lint-deadcode` before
submitting).

---

**Scope correction after review (thanks @joryirving).** The original title claimed the
gate runs *every* CI check. It does not, and the test cannot see the difference: a check
added as a bare `run:` step rather than a make target is invisible to it, and
`security.yml`'s `govulncheck` is exactly that. The doc comment on `DefaultGateChecks` now
claims only what the test can verify. Also fixed: the workflow glob missed `.yaml` files
entirely, a dead `.golangci.yml` config entry, and a new assertion fails when a target is
both gated and exempt, which nothing previously caught.

### DIFF
--- a/pkg/foreman/agent/tools/run_gate_job.go
+++ b/pkg/foreman/agent/tools/run_gate_job.go
@@ -65,9 +65,17 @@ const MaxLogTailBytes = 32 * 1024
 // DefaultGateChecks is the make-target list every gate run executes
 // when the caller does not override it. Mirrors what the autofix gate
 // pipeline ran across hundreds of coder-to-verifier runs.
+//
+// It must remain a SUPERSET of what CI blocks on. A GATE-PASS is read by
+// operators, and by the verdict and escalation machinery, as "this branch
+// is expected to pass CI"; that claim only holds while this list covers
+// every check a pull request has to clear. TestDefaultGateChecksCoverCI
+// pins that against .github/workflows, and gateExemptCIChecks records the
+// deliberate omissions with their reasons (#1637).
 var DefaultGateChecks = []string{
-	"fmt", "vet", "lint", "test",
-	"manifests", "chart-crds", "foreman-chart-crds", ChartCheck,
+	"fmt", "vet", "lint", "lint-deadcode", "test",
+	"generate", "manifests", "chart-crds", "foreman-chart-crds",
+	"check-helm-rbac", "check-reviewer-prompts", ChartCheck,
 }
 
 // ChartCheck is the make target that lints and unit-tests the Helm charts.
--- a/pkg/foreman/agent/tools/run_gate_job.go
+++ b/pkg/foreman/agent/tools/run_gate_job.go
@@ -66,16 +66,23 @@ const MaxLogTailBytes = 32 * 1024
 // when the caller does not override it. Mirrors what the autofix gate
 // pipeline ran across hundreds of coder-to-verifier runs.
 //
-// It must remain a SUPERSET of what CI blocks on. A GATE-PASS is read by
-// operators, and by the verdict and escalation machinery, as "this branch
-// is expected to pass CI"; that claim only holds while this list covers
-// every check a pull request has to clear. TestDefaultGateChecksCoverCI
-// pins that against .github/workflows, and gateExemptCIChecks records the
-// deliberate omissions with their reasons (#1637).
+// A GATE-PASS is read by operators, and by the verdict and escalation
+// machinery, as "this branch is expected to pass CI". This list is what
+// backs that reading, and it must remain a superset of the make-invoked
+// subset of CI, plus the golangci configs a workflow names explicitly
+// with --config=. TestDefaultGateChecksCoverCI pins exactly that against
+// .github/workflows, and gateExemptCIChecks records the deliberate
+// omissions from it with their reasons (#1637).
+//
+// The claim stops there. A CI check added as a direct `run:` step is NOT
+// covered by that test: security.yml runs `govulncheck ./...` on every
+// pull request and the test does not see it, and most of helm-chart.yml
+// drives helm and ct the same way. Widening the gate to such a check is
+// a manual decision; nothing here will prompt for it.
 var DefaultGateChecks = []string{
 	"fmt", "vet", "lint", "lint-deadcode", "test",
 	"generate", "manifests", "chart-crds", "foreman-chart-crds",
-	"check-helm-rbac", "check-reviewer-prompts", ChartCheck,
+	"check-reviewer-prompts", ChartCheck,
 }
 
 // ChartCheck is the make target that lints and unit-tests the Helm charts.
--- a/pkg/foreman/agent/tools/run_gate_job_ci_sync_test.go
+++ b/pkg/foreman/agent/tools/run_gate_job_ci_sync_test.go
@@ -27,9 +27,12 @@ import (
 // workflowDir is the CI definition this gate is meant to mirror.
 const workflowDir = "../../../../.github/workflows"
 
-// workflowGlob matches both spellings GitHub honours. A workflow file
-// named .yaml runs exactly like a .yml one, so a glob that saw only
-// *.yml would let an entire file of CI checks pass these tests unread.
+// workflowGlob is deliberately *.y*ml and not the tidier *.yml. GitHub
+// honours .yaml and .yml equally, so a workflow named .yaml runs its
+// checks exactly like any other. Narrowing this pattern does not fail
+// anything: it makes that entire file invisible to the tests below, which
+// keep passing while covering less CI than they claim to. If you are here
+// to simplify it, that is the failure you would be introducing.
 const workflowGlob = "*.y*ml"
 
 // gateExemptCIChecks are make targets CI runs that DefaultGateChecks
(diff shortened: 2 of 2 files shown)


joryirving (changes requested):
Needs one change before this can merge, and it is the kind that turns the gate off rather than tightening it: `check-helm-rbac` will fail every gate run. `scripts/check-helm-rbac.sh:36` hard-exits 2 unless python3 can `import yaml`, and the gate image is `golang:1.26` (`api/foreman/v1alpha1/gateprofile_types.go:155`). I ran it: that image ships python3 3.13.5 but no `yaml` module and no `pip3`. CI only gets away with the target because `helm-chart.yml` has a dedicated "Install PyYAML" step ahead of it at line 539; `gate_job_template.yaml` has no equivalent, and the checks loop is `make {{ . }} || rc=1`, so the missing module becomes `rc=1` and `GATE FAIL` on every branch, clean ones included. That is exactly the failure mode you reasoned about for `validate-samples`, and the exemption reason you wrote for it ("the gate image is a plain golang image and the target hard-exits when they are absent, so adding it would fail every gate run") applies verbatim to `check-helm-rbac`. I would move `check-helm-rbac` into `gateExemptCIChecks` with that reason, and if you want it genuinely gated, bake pyyaml into a gate image rather than installing it at gate time. While you are in there, the `validate-samples` reason is slightly off: python3 is present, it is only the third-party modules that are not.

Worth noting that the template already has the right pattern for this two dozen lines lower down. A missing actionlint or shellcheck prints "not installed, changed workflows unchecked" and does not touch `rc`, and the bite check degrades to `GATE-WARN`. A check that cannot run in the image is an infra condition, not a coder failure, and the loop's flat `|| rc=1` has no way to say so.

The `generate` addition is not closing a real gap. `Makefile:110` is `test: manifests generate fmt vet setup-envtest`, so `generate` and `manifests` were both already running transitively, and `gate_job_template.yaml` already fails on `if ! git diff --quiet` after the checks with "tree dirty after checks (unformatted or stale codegen)". DeepCopy drift was caught before this PR. Adding it explicitly is cheap and I do not object, but the body's "four CI checks the gate was not running" is really two real ones (`lint-deadcode`, `check-reviewer-prompts`), one no-op, and one that breaks the gate.

On the title, the gate does not run every CI check and the test does not pin it to every CI check. `security.yml:33` is a bare `run: govulncheck ./...`, it blocks pull requests, it is not a make target, and `TestDefaultGateChecksCoverCI` passes today without noticing. Same for most of `helm-chart.yml`, which drives helm and ct directly. What the test actually pins is "the make-invoked subset of CI, plus two golangci configs", which is a genuinely useful thing to pin, but the doc comment at `run_gate_job.go:69-74` claims the stronger property ("must remain a SUPERSET of what CI blocks on ... TestDefaultGateChecksCoverCI pins that against .github/workflows") and a future reader will trust it. I would narrow that comment to what the test can see, and say plainly that a check added as a direct `run:` step is not covered.

Four smaller things:

The glob at `run_gate_job_ci_sync_test.go:76` is `*.yml` only. GitHub honours `.yaml` equally. I dropped a `newcheck.yaml` containing `run: make brand-new-blocking-check` into `.github/workflows` and the test passed; renaming it to `.yml` failed it immediately. `*.y*ml` fixes it.

`golangciConfigTargets[".golangci.yml"]` is dead. `lint.yml:23` invokes the action with no `args:`, so there is no `--config=` for `golangciRe` to match, and nothing keeps that map honest the way `TestGateExemptionsAreLive` keeps the exemptions honest.

`release-please.yml:337` runs `make chart-crds`, so a release-only target counts as "CI blocks on" here. The direction is safe (it can only force the gate wider) but it means a future release-only step gets pushed into every gate run with no way to say "release, not a branch check" other than an exemption whose reason will read oddly.

`makeTargetsIn` stops at the first non-target token, which is right for `make docker-build IMG="${IMG}"`, but it also means `make fmt && make vet` silently yields only `fmt`, and `cd sub && make x` does not match `makeLineRe` at all. Neither form is in the tree today, so this is a note rather than a finding.

The parts I would keep exactly as they are: `lint-deadcode` and `check-reviewer-prompts` are real gaps, verified at `lint.yml:34` and `lint.yml:37`. The test bites; I reverted `DefaultGateChecks` to its old value on your head commit and it failed naming all four targets with the workflow each came from. And `TestGateExemptionsAreLive` is load-bearing rather than decorative in a way that is easy to miss: `docker-build-foreman-operator` and friends are only reachable through the `run: |` block form, so if that matcher ever regressed the liveness test would fail rather than the coverage test silently passing. That is a good property, and worth a line in the test comment so nobody "simplifies" the matcher later.

Defilan:
Confirmed and fixed. I reproduced your finding before changing anything: `check-helm-rbac.sh:36` hard-exits without PyYAML, the gate image is `golang:1.26`, `gate_job_template.yaml` installs nothing, and only `helm-chart.yml:535` adds it. Gating it would have turned every gate run red on clean branches, which is the opposite of what the PR was for. It is now in `gateExemptCIChecks` with that reason, and I corrected the `validate-samples` reason too, since python3 is present and only the modules are not.

Pushed `b4b227325` and `504893c29`.

Your glob catch was real: `*.y*ml` now. I reproduced it the way you did, with a `.yaml` workflow that the old pattern ignored. Dropped the dead `.golangci.yml` entry and added a liveness assertion so a dead entry cannot come back.

One thing your review made me add that neither of us listed. Nothing caught the inverse mistake: a target in both `DefaultGateChecks` and `gateExemptCIChecks`, the two lists silently contradicting each other, every test green. That is the same shape as the bug you found, so `TestGateChecksAndExemptionsAreDisjoint` now fails naming the target. Putting `check-helm-rbac` back while leaving the exemption in place fails only that new test and none of the other three.

You are right about the title and the doc comment. Both claimed the gate runs every CI check; it runs the make-invoked subset, and `govulncheck` as a bare `run:` step is exactly the blind spot. Retitled, and the comment now claims only what the test can verify.

Keeping `generate` despite it being a no-op today, since it is explicit and free, but I corrected the body's "four CI checks" to say plainly that it is two real gaps, one no-op and one that would have broken the gate.

joryirving (approved):
Verified all of it against `b4b227325`/`504893c29` and this clears my block.

`check-helm-rbac` is in `gateExemptCIChecks` with the PyYAML reason and gone from `DefaultGateChecks`, so the gate no longer fails clean branches. `workflowGlob` is `*.y*ml` with a comment saying why it is not the tidier form. The dead `.golangci.yml` key is gone, leaving only `.golangci-deadcode.yml`, and `TestGolangciConfigTargetsAreLive` now keeps it honest. I ran the four sync tests on your head and they pass.

`TestGateChecksAndExemptionsAreDisjoint` is the better catch of the two, and it is yours, not mine. A target sitting in both lists is the same silent-contradiction shape as the original gap and nothing covered it. Naming `check-helm-rbac` as the live example in the comment is what stops someone re-adding it in a year and reading the exemption as stale.

Keeping `generate` is fine by me now that the body says plainly it is a no-op today rather than counting it as a gap.

Issue #1648, opened 2026-08-23
[FEATURE] Stable FleetNode identity via StatefulSet agent pools (follow-up to #1640)


Defilan:
## Feature Description

Evaluate moving the in-cluster foreman-agent pools from Deployments to
StatefulSets, so FleetNode identity is stable across rollouts
(`foreman-coder-agent-0`) instead of minted fresh from each pod's random suffix.

This is the deliberate follow-up to the #1640 decision. That decision settled
*what* a FleetNode identifies (the agent process, not the machine). This issue
is about how *stable* that identity should be — orthogonal, and deliberately not
folded into #1640 so it gets its own evaluation rather than riding along.

## Problem Statement

> As an operator, I want an agent pool's FleetNodes to survive a routine image
> bump, so a rollout does not mint a new fleet and orphan the old one, and so
> per-node state has somewhere durable to live.

Because identity is the pod name today, **every rollout mints a new generation
of FleetNodes and abandons the previous one**. Observed live on 2026-08-22: a
chart upgrade left eight FleetNodes in `Draining` until the orphan reaper
deleted them. The reaper (#979/#980) exists largely to clean up after this
churn.

Churn also has a second-order cost: it makes `FleetNode.status` a poor home for
any durable per-node state. That is exactly where the supervision-budget
follow-up from #1635 wants to put its bound ("a process-local int is a strange
place for capacity to live when the cluster's model is a single `currentTask`
string") — and with identities that die on every rollout, moving the budget to
status would just lose it on every upgrade.

## Proposed Solution

StatefulSet per agent pool, ordinal-named pods, `--fleet-node-name` derived
from the stable pod name (which `os.Hostname()` already yields; the change is
that the name stops churning).

Expected wins, to be validated rather than assumed:

- Rollouts reuse FleetNodes: no Draining generation, no reaping, no window
  where `kubectl get fleetnodes` shows double the fleet.
- `FleetNode.status` becomes viable for durable per-node state
  (supervision budget from #1635; anything else that wants a per-node ledger).
- First-fit-alphabetical scheduling (`reserveFirstFitNode`, deterministic by
  name) becomes genuinely deterministic across rollouts instead of
  random-suffix roulette. Note this cuts both ways: it also concentrates load
  on ordinal 0 more predictably, which intersects with the placement-spread
  discussion in #1634.

## Costs and open questions to evaluate

- Chart surgery: per-pool StatefulSet, and how `updateStrategy` interacts with
[...]

Issue #1643, opened 2026-08-23
[BUG] foreman.crs.enabled produces no metrics at all: four defects in the CustomResourceState config


Defilan:
## Bug Description

`foreman.crs.enabled` has never produced a single metric. Enabling it yields no
`foreman_*` series at all, which means every panel in the Grafana dashboard
shipped alongside it (`charts/foreman/templates/grafana-dashboard.yaml`, "Foreman
Fleet & Tasks") renders empty, and the three alerts in
`charts/foreman/templates/prometheusrule.yaml` can never fire.

Four separate defects, each reproduced against a live kube-state-metrics.

## Actual Behavior

**1. The ConfigMap is not a complete document.**

`crs-configmap.yaml` emits a bare `resources:` mapping. kube-state-metrics'
`--custom-resource-state-config-file` expects a whole
`kind: CustomResourceStateMetrics` / `spec:` document. Given the bare form it
parses to an empty config, logs **nothing**, and serves no custom metrics. That
is indistinguishable from a healthy install that simply has no CRs, which is
presumably how this survived.

**2. `gauge.valueFrom` is the wrong type.**

Every gauge writes:

```yaml
gauge:
  valueFrom:
    path: [status, phase]
```

The schema is `[]string`. Once defect 1 is fixed so the config is actually read,
kube-state-metrics crashloops:

```
failed to parse Custom Resource State metrics: yaml: unmarshal errors:
  line 46: cannot unmarshal !!map into []string
  ...10 occurrences
```

**3. `gauge.valueFrom` is relative to `gauge.path`, and `path` is required.**

With an absolute `valueFrom` and no `path` the config parses cleanly and emits
nothing, silently. Measured with a three-way probe on one cluster:

| shape | series |
|---|---|
| `valueFrom: [status, lastHeartbeatTime]` (shipped) | **0** |
| `path: [status]` + `valueFrom: [lastHeartbeatTime]` | **10** |
| `type: StateSet` over `status.phase` | **40** (10 nodes x 4 states) |

**4. Enum fields use `type: Gauge` plus a `labelsMapping` block.**

`labelsMapping` is not a kube-state-metrics field. It is silently ignored. And a
Gauge's value must be a number, boolean or RFC3339 string, so a Gauge over
`status.phase` (`"Ready"`) emits nothing. The construct for enums is `StateSet`.

## Expected Behavior

Enabling `foreman.crs.enabled` produces the metrics the shipped dashboard and
alert rules query.

[...]

PR #1649, opened 2026-08-23
feat(foreman): report the Kubernetes node on FleetNode.status, not as identity

Defilan:
## What

Implement the #1640 decision: FleetNode identity is the agent **process**, and
the Kubernetes node it runs on becomes a reported **property** —
`status.kubernetesNode`.

## Why

Refs #1640

The chart has set `FLEET_NODE_NAME` from the downward API (`fieldRef:
spec.nodeName`) since M4c, but no Go code ever read it. The agent falls back to
`os.Hostname()` — the pod name — so every replica registers its own FleetNode.
The decision on #1640 established that fallback as the correct model:
`spec.nodeName`'s contract was always worker-scoped, and every
`FleetNode.status` mechanism (one `currentTask` reservation, heartbeat-driven
claim expiry and the Draining reaper, `phase`) assumes a single writer. Wiring
the env var into identity would break all of them at once and reintroduce the
two-agents claim race.

What was genuinely missing is the machine as a *property*: nothing recorded
which physical node an in-cluster agent pod runs on. On a fleet of two GB10s, a
Strix and metal Macs, "which box is this agent on?" was unanswerable from the CR.

## How

- **`status.kubernetesNode`**, stamped on heartbeat alongside `os`/`arch`, fed
  from `FLEET_NODE_NAME` — the previously-dead env var, repurposed rather than
  deleted. Off-cluster agents leave it empty (covered by a test).
- **`spec.nodeName` godoc corrected** to state it is the worker's own name (the
  pod name in-cluster) and that the Kubernetes node lives at
  `status.kubernetesNode`. The field is **not** renamed — that is a CRD break,
  deferred per the decision.
- **`resolveFleetNodeName` extracted from `main()` and tested**, pinning the
  identity-source contract (#1640 ask 3): the `--fleet-node-name` flag wins,
  otherwise the sanitized hostname; `FLEET_NODE_NAME` is deliberately not an
  input.
- CRDs regenerated: `make manifests`, `chart-crds`, `foreman-chart-crds`;
  `make check-helm-rbac` green.

The kube-state-metrics CustomResourceState label for the new field rides with
the #1643 CRS fix, where that config file is already being rewritten — adding it
here would guarantee a conflict between the two branches.

## Follow-up

#1648 (StatefulSet stable identity) is the named follow-up for the churn cost
this decision accepts.

## Verification

- Watched both registrar tests and the resolver test fail first (missing field /
  missing function), then pass
- `go test ./pkg/foreman/agent/ ./cmd/foreman-agent/ ./api/foreman/...` green
- `make lint`, `GOOS=linux` cross-arch lint, `make lint-deadcode`: 0 issues
- `make check-helm-rbac`: green; `helm unittest charts/foreman`: 39/39

## Checklist

- [x] Tests added/updated
- [ ] Documentation updated (if user-facing change) — CRD godoc is the user-facing surface here

Assisted-by: Claude Code (implemented under TDD from the #1640 decision I made
after its architecture study; I reviewed the change and ran the tests, both
lints, the RBAC check and the chart suite before submitting).

### DIFF
--- a/api/foreman/v1alpha1/fleetnode_types.go
+++ b/api/foreman/v1alpha1/fleetnode_types.go
@@ -95,8 +95,13 @@ type FleetNodeCapability struct {
 // FleetNode object. Most of the resource's interesting fields live in Status,
 // which the FleetAgent updates on every heartbeat.
 type FleetNodeSpec struct {
-	// NodeName is the human-readable identity of the worker. Conventionally
-	// matches metadata.name; required for the scheduler to address it.
+	// NodeName is the human-readable identity of the WORKER — the agent
+	// process, not the machine it runs on (#1640). Conventionally matches
+	// metadata.name; required for the scheduler to address it. For an
+	// in-cluster agent this is the pod name (the agent falls back to
+	// os.Hostname()); off-cluster agents set it via --fleet-node-name. The
+	// Kubernetes node a pod is scheduled on is reported separately at
+	// status.kubernetesNode.
 	// +kubebuilder:validation:Required
 	NodeName string `json:"nodeName"`
 
@@ -182,6 +187,15 @@ type FleetNodeStatus struct {
 	// (runtime.GOARCH). Used alongside OS for platform artifact selection.
 	// +optional
 	Arch string `json:"arch,omitempty"`
+
+	// KubernetesNode is the Kubernetes node the agent pod is scheduled on,
+	// reported by the agent on heartbeat from the FLEET_NODE_NAME downward-API
+	// env var (fieldRef: spec.nodeName). It is a PROPERTY of the worker, not
+	// its identity: FleetNode identity is the agent process (#1640), and
+	// spec.nodeName holds the worker's own name. Off-cluster agents (metal
+	// Macs) have no Kubernetes node and leave this empty.
+	// +optional
+	KubernetesNode string `json:"kubernetesNode,omitempty"`
 
 	// UpdateRequest is written by the AgentReleaseReconciler when a new
 	// agent version is ready for this node to install. The agent reads
--- a/api/foreman/v1alpha1/fleetnode_types.go
+++ b/api/foreman/v1alpha1/fleetnode_types.go
@@ -220,6 +220,7 @@ type FleetNodeStatus struct {
 // +kubebuilder:printcolumn:name="Accelerator",type=string,JSONPath=`.status.capability.accelerator`
 // +kubebuilder:printcolumn:name="RAM",type=integer,JSONPath=`.status.capability.availableRAMGB`
 // +kubebuilder:printcolumn:name="Current Task",type=string,JSONPath=`.status.currentTask`
+// +kubebuilder:printcolumn:name="K8s Node",type=string,JSONPath=`.status.kubernetesNode`
 // +kubebuilder:printcolumn:name="Heartbeat",type=date,JSONPath=`.status.lastHeartbeatTime`
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
 
--- a/charts/foreman/templates/crds/fleetnodes.yaml
+++ b/charts/foreman/templates/crds/fleetnodes.yaml
@@ -69,8 +69,13 @@ spec:
             properties:
               nodeName:
                 description: |-
-                  NodeName is the human-readable identity of the worker. Conventionally
-                  matches metadata.name; required for the scheduler to address it.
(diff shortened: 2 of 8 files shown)


joryirving (commented):
This is right, and it is the right shape: identity stays the agent process, the machine becomes a property. I checked the premise rather than taking it, and it holds. `charts/foreman/templates/agent-deployment.yaml:112` does set `FLEET_NODE_NAME` from `spec.nodeName`, and before this PR nothing in Go read it, so the pod-name fallback at the old `main.go:279` really was the effective behaviour. The `resolveFleetNodeName` extraction is behaviour-preserving (same flag-wins ordering, same error on empty or failed hostname, same sanitize-only-when-user-supplied log line), and `sanitizeName` already trims, truncates to 63 and falls back, so the "DNS-1123 label" claim in the new godoc is honest. Both CRD copies match the Go comments verbatim and there are only two copies, so the regen story is complete. Off-cluster is real too, not aspirational: `deployment/macos/com.llmkube.foreman-agent.plist:23` and `deployment/linux/llmkube-foreman-agent.service:32` both pass `--fleet-node-name`.

One test does not test what its comment says. `TestRegistrar_PatchHeartbeat_OmitsKubernetesNodeWhenUnset` (`pkg/foreman/agent/fleetnode_test.go:602`) claims "an empty value must not be stamped as anything", but it cannot distinguish the guard from its absence: I deleted `if r.KubernetesNode != ""` at `pkg/foreman/agent/fleetnode.go:253` and made the assignment unconditional, and both new tests still pass. On a freshly created node an unconditional `""` writes `""` either way. The case that separates them is the one that actually matters: a FleetNode whose status already carries a non-empty `kubernetesNode`, patched by a Registrar with the field empty. With the guard the old value sticks forever; without it, `MergeFrom` plus `omitempty` clears it. I would add that case (Upsert, PatchHeartbeat with `KubernetesNode` set, then a second Registrar with it empty, assert the value you actually want) so the sticky semantics are a decision on the record rather than an accident. Keeping the guard for consistency with `OS` and `Arch` is fine by me; it just needs to be the tested behaviour.

Second, the field is currently write-only. Nothing in the repo reads `status.kubernetesNode`, and it is not a printer column, so the motivating question ("which box is this agent on?") is still only answerable with `-o yaml`. Adding a `+kubebuilder:printcolumn` for it next to `api/foreman/v1alpha1/fleetnode_types.go:222` costs nothing and is not a CRD break, and it is the fix that matches the stated motivation. The kube-state-metrics label deferral checks out (#1650 does carry `kubernetes_node: [status, kubernetesNode]` plus a chart test asserting that pattern), so worth noting the ordering: if #1650 lands first its config labels a field the CRD does not have yet. Harmless, just less confusing merged the other way round.

Third, the docs the issue was about are still wrong. `docs/site/foreman/README.md:58` says the foreman-agent "on each host self-registers", and `:116` describes the chart as registering "a FleetNode for the gate-runner role on the Linux/K8s host". Under this decision it is one FleetNode per agent process, so a pool with `replicaCount: 2` gets two on one host, which is exactly what #1640 showed live. That is a two-line fix and it belongs in this PR rather than behind the CRD-godoc justification in the checklist, because the whole point here is that a documented invariant was inverted.

On #1635: no conflict, and this helps it. I test-merged `fix/agent-jobmode-slot` into this branch and all four shared files (`fleetnode_types.go`, both CRD YAMLs, `main.go`) auto-merged clean, since #1635 touches `currentTask`'s godoc and the watcher wiring while this touches the `Arch` neighbourhood and the resolve block. More usefully, #1635's `Recreate` comment and its new README section both assume identity is per agent process, so nailing that down as design here removes the risk that #1640 gets closed the other way and breaks them.

Defilan:
All three applied, pushed `fd09bd709`.

The test finding was the useful one, and I reproduced your method: deleting the guard leaves both original tests green, because on a freshly created node an unconditional empty write is indistinguishable from no write. The new case is the one you described, a node already carrying a non-empty `kubernetesNode` patched by a Registrar with the field empty, and it fails when the guard is removed. Sticky is now the tested decision rather than an accident, consistent with `OS` and `Arch`.

Added the `K8s Node` printcolumn. You are right that a write-only field does not answer the question the issue was about. Worth noting for anyone else touching a `foreman.llmkube.dev` CRD: `make chart-crds` is the wrong target and quietly no-ops, it needs `make foreman-chart-crds`.

Fixed both README lines. That one stings a bit, since the whole point of the issue was that a documented invariant had been inverted, and I left the inverted documentation in place.

On ordering with #1650: agreed, this should land first so the CRD has the field before the kube-state-metrics config labels it.

Issue #1654, opened 2026-08-23
[FEATURE] Archive Foreman task audit records and transcripts before they are garbage-collected


Defilan:
## Feature Description

Durably archive each Foreman task's audit record and transcript as an immutable bundle on disk, behind an opt-in `--archive-dir` flag. This is capture only. Dataset building, training and evaluation are separate follow-ups.

## Problem Statement

> As an operator running Foreman, I want a durable record of what each coder run actually did, so that runs can be analysed after the fact rather than only while their objects still exist.

The data with the most signal is the data with the shortest life, and nothing preserves it today:

| Artifact | Lifetime today | Contains |
|---|---|---|
| Transcript ConfigMap | Deleted with its AgenticTask | Full trajectory: turns, tool calls, the gate output the coder saw |
| Audit ConfigMap | 7 days | Verdict, gate/scope/issueAsk/reviewer outcomes, branch, commit SHA, issue |
| Git | Permanent | Final merged diff |

Transcripts carry `OwnerReferences: ownerRefForTask(task)` (`pkg/foreman/agent/transcript.go`), so Kubernetes garbage-collects them the moment the AgenticTask is deleted. Nothing reaps them separately. Audit records are deliberately owner-unbound, but `--audit-retention` defaults to seven days (`cmd/foreman-operator/main.go:99`).

Git keeps the destination but not the journey. A merged diff cannot show that the first attempt failed `go vet`, what the coder did next, or why a reviewer demoted a GO. That exists only in the two perishable artifacts above.

## Proposed Solution

Hook the existing `audit.RecordTerminal` call in `internal/foreman/controller/agentictask_controller.go:132`. That instant is the only one where the verdict is final, the audit record is fully populated, `TranscriptRef` still resolves, and the git join keys are set. No finalizer, no race with garbage collection.

Write one immutable bundle per terminal task:

```
<archive-dir>/<repo>/<issue>/<taskName>-<recordedAt>/
  audit.json
  transcript.json   # when one exists; deterministic runs write none
  meta.json
```

Opt-in via a single flag, defaulting off:

```yaml
foreman:
  archive:
    enabled: false
    dir: /var/lib/foreman/archive
```

The operator writes files and nothing else. No S3 client, no credentials, no endpoint. Getting bundles off-node is a separate optional CronJob using the `mc` / `curl --aws-sigv4` idiom this repo already uses for model fetching.

[...]

PR #1655, opened 2026-08-24
feat(foreman): opt-in archival of task audit records and transcripts

Defilan:
## What

Opt-in archival of Foreman task data. When an AgenticTask reaches a terminal state, the operator writes its audit record and agent transcript to disk as an immutable bundle:

```
<archive-dir>/<repo>/<issue>/<taskName>-<recordedAt>/
  audit.json        # the audit.Record verbatim
  transcript.json   # when one exists
  meta.json         # completion sentinel
```

Off unless `--archive-dir` is set. A disabled install renders byte-identically to before: no flag, no volume, no mount, no directory, no API read, no metric series.

## Why

The data with the most signal has the shortest life, and nothing preserved it:

| Artifact | Lifetime | Contains |
|---|---|---|
| Transcript ConfigMap | deleted with its AgenticTask | the full trajectory: turns, tool calls, the gate output the coder saw |
| Audit ConfigMap | 7 days (`--audit-retention`) | verdict, gate/scope/issueAsk/reviewer outcomes, branch, commit SHA |
| Git | permanent | the final merged diff |

Git keeps the destination but not the journey. A merged diff cannot show that the first attempt failed `go vet`, what the coder did next, or why a reviewer demoted a GO.

Refs #1654. Design: `docs/proposals/1654-foreman-trajectory-archival.md`, committed here and reconciled against what shipped.

## How

`audit.RecordTerminal`'s call site is the only moment where the verdict is final, the record is fully populated, the transcript still resolves, and the git join keys are set. Archival hooks there, after the node reservation is released.

Decisions worth calling out, several of which came out of review rather than design:

- **Containment is symlink-aware, not lexical.** The root is resolved with `EvalSymlinks`, containment uses `filepath.Rel` rather than a string prefix, and it is **re-verified after `MkdirAll`** because a pre-create check cannot see a symlink planted on a segment it has not created yet. Containment runs before any bytes are written. Note this is *stronger* than `resolveInside` in `pkg/foreman/agent/tools`, which skips symlink resolution for paths that do not exist yet; unifying the two must go upward.
- **`meta.json` is a completion sentinel**, written last. A directory without it is debris from an interrupted write and is removed and rewritten, not skipped. Refusing would make the debris permanent, which is the same data loss by another route. Only a *completed* bundle is immutable.
- **The bundle key falls back to `Task.UID`** when `RecordedAt` is empty, which happens whenever `Status.FinishedAt` is nil. Without it the key is constant across retries and the first attempt wins permanently.
- **Archival runs on every terminal reconcile**, not only the first, so a failed write retries rather than being lost.
- **`Status.TranscriptRef` was never assigned anywhere in production.** One read, one test fixture, zero writes. `watcher.go` lifted `branch`, `commitSHA` and `jobName` out of the result envelope but not `transcriptRef`, so this feature would have archived zero transcripts silently, with both failure counters flat, indistinguishable from a fleet of deterministic runs. Fixed at the root; this also repairs `audit.Record.TranscriptRef`, empty in every audit record since the field was introduced.

## Testing

- `go test ./...` (minus `/e2e`, matching `make test`): green. Affected packages also green under `-race` and `-shuffle=on`.
- `golangci-lint run ./...`: 0 issues, natively and under `GOOS=linux`.
- `helm lint` clean; `helm unittest charts/foreman` 52/52.
- No new module dependencies; `go.mod` and `go.sum` untouched across all 17 commits.

Roughly 130 mutations were applied across four task reviews, nine fix rounds, and a whole-branch review. Several tests were found to pass whether or not the code worked and were rewritten to bite: an inert cleanup test, a tautological schema assertion, a `notMatchRegexRaw` assertion that passes unconditionally on a parsed manifest, and a disabled-case test that asserted on a directory the code never received.

## Known gaps, deliberately not fixed here

- **`coderJobResultToResult` drops `transcriptRef` on the `INCOMPLETE` and `ERROR` branches** (`executor_coderjob.go`), so Job-mode turn-exhausted runs still archive `hasTranscript: false`. Those are precisely the recovery trajectories this feature exists to capture. Pre-existing, producer-side, needs its own change.
- The `"transcriptRef"` Extra key is a hand-written literal in producer and consumer; renaming it at all seven producer sites leaves the package green. A shared constant would close it.
- An empty regular `meta.json` from a torn write of the sentinel itself passes the completion check but cannot be parsed. Closing it needs an atomic sentinel write.
- `archiveTerminalTask` is synchronous and untimed inside `Reconcile`. Releasing the node reservation first bounds the blast radius, but a stalled network mount still blocks that controller.
- `fsGroup: 65532` is hand-matched to the image's `USER`; a Dockerfile change silently restores the EACCES failure mode. Helm cannot read the image at render time.

## Checklist

- [x] Tests added/updated

Assisted-by: Claude Code (implemented the four planned tasks and their fix rounds under per-task and whole-branch review; I reviewed the result and ran `make test` and `make lint` locally)

### DIFF
--- a/charts/foreman/templates/operator-deployment.yaml
+++ b/charts/foreman/templates/operator-deployment.yaml
@@ -65,6 +65,14 @@ spec:
             # reaper entirely without redeploying.
             - --audit-retention={{ .Values.operator.audit.retention }}
             - --audit-retention-interval={{ .Values.operator.audit.interval }}
+            {{- if .Values.foreman.archive.enabled }}
+            # Terminal-task archival (#1654). One flag carries both the
+            # switch and the destination: empty (the default, rendered
+            # by omitting the flag entirely) disables archival in the
+            # operator. Bundles hold source code, issue text and tool
+            # output, so this is opt-in only.
+            - --archive-dir={{ .Values.foreman.archive.dir }}
+            {{- end }}
           ports:
             - name: health
               containerPort: {{ .Values.operator.health.port }}
@@ -110,6 +118,10 @@ spec:
               mountPath: /tmp/k8s-webhook-server/serving-certs
               readOnly: true
             {{- end }}
+            {{- if .Values.foreman.archive.enabled }}
+            - name: archive
+              mountPath: {{ .Values.foreman.archive.dir }}
+            {{- end }}
       volumes:
         - name: tmp
           emptyDir: {}
@@ -117,6 +129,10 @@ spec:
         - name: webhook-certs
           secret:
             secretName: {{ include "foreman.webhook.secretName" . }}
+        {{- end }}
+        {{- if .Values.foreman.archive.enabled }}
+        - name: archive
+          {{- toYaml .Values.foreman.archive.volume | nindent 10 }}
         {{- end }}
       {{- with .Values.operator.nodeSelector }}
       nodeSelector:
--- a/charts/foreman/templates/operator-deployment.yaml
+++ b/charts/foreman/templates/operator-deployment.yaml
@@ -29,8 +29,25 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "foreman.operator.serviceAccountName" . }}
+      {{- /*
+        Archival (#1654) is the only thing the operator writes to a volume it
+        does not own, so fsGroup is added ONLY when archival is enabled and
+        only from the archive block. The operator image runs as 65532 with
+        runAsNonRoot, and a PVC arrives root-owned on most provisioners, so
+        without a matching fsGroup every terminal task increments
+        foreman_archive_failures_total{reason="write"} and no bundle lands.
+        Applying it unconditionally would change the pod security context of
+        every install, archival or not, and OpenShift's restricted-v2 SCC
+        rejects a pod naming an fsGroup outside the namespace's allocated
+        range. archive.fsGroup: 0 renders none at all, which is what that SCC
+        needs. An fsGroup the operator sets itself in podSecurityContext wins.
+      */}}
+      {{- $podSecurityContext := deepCopy (default dict .Values.operator.podSecurityContext) }}
+      {{- if and .Values.foreman.archive.enabled (gt (int (default 0 .Values.foreman.archive.fsGroup)) 0) }}
+      {{- $podSecurityContext = merge $podSecurityContext (dict "fsGroup" (int .Values.foreman.archive.fsGroup)) }}
+      {{- end }}
       securityContext:
-        {{- toYaml .Values.operator.podSecurityContext | nindent 8 }}
+        {{- toYaml $podSecurityContext | nindent 8 }}
       containers:
         - name: manager
           image: {{ include "foreman.operator.image" . }}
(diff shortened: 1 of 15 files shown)


Defilan:
Ran a dedicated pre-merge compatibility audit on this branch, separate from the correctness reviews. Recording what it found, since one item was a real bug in the feature's own documented path.

## Not a breaking change

For anyone already running the operator who does not want this feature, nothing changes:

- No API type, CRD, generated code, or RBAC file is touched; `go.mod` and `go.sum` are byte-identical.
- A full `helm template` at default values is identical between base and head across every template, not only the Deployment. A default `helm upgrade` therefore produces no Deployment change and no rollout.
- `clearNodeCurrentTask` keeps its exact position in `Reconcile`; the archive call is appended after it, and returns immediately when `--archive-dir` is unset.
- `pkg/foreman/audit`'s exported surface is purely additive: `go doc -all` diffs to a single added line, `func ResolveAgent(`. `RecordTerminal`'s signature and semantics are unchanged.

The one change not gated by the flag is the `transcriptRef` lift in `watcher.go`, which populates `status.transcriptRef` on every terminal task. Audited specifically:

- Nothing depends on that field being empty. It is not a printer column, the validator is a no-op, the CRD schema is a bare string with no enum or CEL, and no rule, dashboard, doc or example branches on it.
- It is not a schema break. `transcriptRef` has been a declared `omitempty` field of `foreman.audit.v1` since the record was introduced, so absent to present is additive within v1. No version bump is warranted.
- It rides the existing status patch that already writes `branch`, `commitSHA` and `jobName`. No new API call, no new conflict surface, and the operator's ClusterRole already grants the ConfigMap read the archiver needs.

`foreman_archive_failures_total` is a labelled CounterVec, so it exports zero series until first use and cannot collide with an existing name or affect a dashboard.

## One real bug, fixed in 075b95b39

`values.yaml` told operators to point `foreman.archive.volume` at a `persistentVolumeClaim` for real use. Because the default was `{emptyDir: {}}` and Helm deep-merges, that produced a Volume with two sources:

```yaml
- name: archive
  emptyDir: {}
  persistentVolumeClaim:
    claimName: foreman-archive
```

which the API server rejects with `may not specify more than 1 volume type`, so `helm upgrade` fails and nothing rolls out. `hostPath` behaved the same way. Both `helm lint` and `helm unittest` passed on it, because the chart suite only ever rendered `emptyDir`.

The default is now `{}` with `emptyDir` rendered as an explicit fallback, and the suite covers a PVC and a `hostPath`, asserting exactly one source each. Reverting the old default fails those two cases; the default install renders byte-identically to before. Chart tests 52 to 54.

## Two operational notes added to values.yaml

- Rolling only the operator image back while archival stays enabled crashloops it: the older binary does not define `-archive-dir` and Go's `flag.CommandLine` is `ExitOnError`. `helm rollback` and `kubectl rollout undo` are both safe, since args and image revert together.
- Archival needs `foreman-agent` at the same version. The `transcriptRef` lift ships in the agent; with an older agent every bundle records `hasTranscript: false` and the failure counter stays at zero, which is indistinguishable from a fleet of deterministic runs.

## Accepted, not fixed

- `operator.podSecurityContext: null` now renders `securityContext: {}` rather than `null`. Both are valid and the API server treats them the same, but the stored PodSpec differs, so that one non-default input gets a single operator restart on upgrade.
- `--archive-dir` has no startup validation. A bad path degrades to one error log plus `foreman_archive_failures_total{reason="write"}` per terminal task rather than failing at boot. Out-of-range chart values fail loudly at apply instead.

Issue #1638, opened 2026-08-23
[FEATURE] pollOnce should keep claiming until no slot has capacity, not stop after one task


joryirving:
## Feature Description

After a successful claim, `pollOnce` returns instead of continuing to scan, so
the watcher claims at most one task per poll interval. With the independent
in-process and supervised slots from #1635, it should keep scanning until no
slot has capacity.

## Problem Statement

One claim per tick was the right shape when there was a single slot: nothing
else could start anyway. Once Job-mode supervisions have their own budget, the
same loop leaves capacity idle for no reason.

Two consequences, both from @Defilan's review of #1635:

- Filling a supervision budget of 4 takes about four poll intervals — roughly
  20 seconds at the default cadence — instead of one pass, even when all four
  tasks are already `Scheduled` and the node can take them immediately.
- `sortTasksDepthFirst` orders review and verify ahead of issue-fix
  (`kindClaimPriority`). Since the loop returns after one claim, a Job-mode
  issue-fix claim defers an in-process review or gate task by a full interval,
  even though it draws on a different slot and the in-process slot is free.

As an operator, I want a node with free capacity in more than one slot to fill
them in a single pass, so that queued work is not paced by the poll interval.

## Proposed Solution

Continue the candidate loop after a successful `claim` + `launchExecutor`
instead of returning, stopping when neither slot has capacity or the candidate
list is exhausted. The existing per-candidate capacity check already handles
the accounting; this is mostly deleting the early `return nil`.

Worth confirming while doing it: the check-then-reserve split in `pollOnce` is
currently safe only because `Run` calls `pollOnce` sequentially and executor
goroutines only ever release slots. Claiming several tasks in one pass does not
change that invariant, but it does make the pass longer, so it is worth being
deliberate rather than incidental about it.

## Alternatives Considered

Shortening the poll interval. That trades one problem for a higher steady-state
List rate against the apiserver and still paces claims artificially.

## Additional Context

Follow-up from @Defilan's review of #1635; not a blocker for that PR. Depends
on #1635 landing, since the premise is the independent slots it introduces.
Related: #1634 (Job-mode placement concentration).

PR #1665, opened 2026-08-24
fix(foreman-agent): keep claiming in pollOnce until a slot is full

Defilan:
## What

`pollOnce` no longer returns after its first successful claim. It keeps
scanning the remaining candidates in the same pass, stopping when the relevant
slot is full or the candidate list is exhausted.

The production change is one deleted `return nil`.

## Why

Refs #1638.

One claim per tick was correct when a node had a single slot: nothing else
could start anyway. Since #1559 gave Job-mode supervisions their own budget, a
node with free capacity in two different slots leaves one idle for a whole
poll interval. Filling a supervision budget of 4 took about four intervals,
roughly 20 seconds at the default cadence, even when all four tasks were
already `Scheduled`.

It also interacts badly with ordering: `sortTasksDepthFirst` puts review and
verify ahead of issue-fix, so a Job-mode issue-fix claim deferred an in-process
review by a full interval even though the two draw on different slots.

## How

Continuing the loop is safe because `launchExecutor` reserves its slot
**synchronously**: it takes `inflightMu` and sets `inflight` / increments
`supervised` before starting the goroutine, and only the release is deferred
inside it. So the per-candidate `hasCapacityFor` check on the next iteration
already observes the slot the previous iteration took. `Run` calls `pollOnce`
sequentially and executor goroutines only ever release, so a longer pass does
not change the check-then-reserve invariant. No new locking, and `inflightMu`
is not held across the loop.

Three tests in `watcher_concurrency_test.go`:

- a node with a free in-process slot and a supervision budget of 2 claims all
  three candidates in one pass
- a second in-process candidate stays `Scheduled` when the slot is already
  held earlier in the same pass, while a Job-mode candidate is still claimed
- Job-mode candidates past the supervision budget stay `Scheduled`

**Mutation-checked.** Restoring the `return nil` fails all three:

```
--- FAIL: TestPollOnce_ClaimsAcrossSlotsInOnePass
    Running tasks after one pass: want 3 (one in-process, two job-mode) got 1
--- FAIL: TestPollOnce_InProcessSlotStillSerializes
    job-mode task should have been claimed in the same pass: want Running got Scheduled
--- FAIL: TestPollOnce_SupervisionBoundEnforcedInOnePass
    Running tasks after one pass: want 2 (the budget) got 1
```

## Checklist

- [x] Tests added/updated

Assisted-by: Foreman (in-cluster coder agent, qwen38-coder) for the initial
implementation; reviewed, mutation-checked and rebased by hand.

### DIFF
--- a/pkg/foreman/agent/watcher.go
+++ b/pkg/foreman/agent/watcher.go
@@ -427,10 +427,21 @@ func (w *AgenticTaskWatcher) pollOnce(ctx context.Context, namespace string) err
 			logf.FromContext(ctx).WithName("agentictask-watcher").Error(err, "claim failed", "task", t.Name)
 			continue
 		}
-		// Took it. Launch the executor and return to the polling loop;
-		// the next poll re-checks capacity before claiming anything else.
+		// Took it. Launch the executor and keep scanning the rest of the
+		// candidates in this same pass instead of returning: in-process and
+		// Job-mode runs draw on different slots, so a node with free capacity
+		// in more than one of them would otherwise leave the others idle for
+		// a whole poll interval (#1638). Continuing is safe because
+		// launchExecutor reserves the slot it takes SYNCHRONOUSLY -- it
+		// takes inflightMu and sets inflight / increments supervised before
+		// the goroutine is even started -- so the hasCapacityFor check at the
+		// top of the next iteration already observes the slot the previous
+		// iteration reserved, and the loop stops claiming on its own once the
+		// relevant slot is full and falls out when the candidate list is
+		// exhausted. Run calls pollOnce sequentially and the executor
+		// goroutines only ever release, so this longer pass does not break
+		// the check-then-reserve invariant above.
 		w.launchExecutor(ctx, t, res.agent, res.supervise)
-		return nil
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


joryirving (approved):
This looks correct, and I checked the claim the safety argument rests on rather than taking the comment's word for it.

`launchExecutor` does reserve synchronously: it takes `inflightMu`, sets `inflight` or increments `supervised`, and unlocks at watcher.go:478, four lines before the goroutine starts at :484. So the `hasCapacityFor` check at the top of the next iteration genuinely observes the slot the previous iteration took. Combined with `Run` calling `pollOnce` sequentially and the executor goroutines only ever releasing, the check-then-reserve invariant holds across the longer pass.

The per-candidate capacity check was already there at :420, so removing the `return` is doing exactly one thing rather than changing the shape of the loop.

Every claim in the new comment has a test behind it, which is not always true of a comment that long. The multi-slot fill, the in-process slot still serializing, and the supervision budget capping a single pass are each covered, and the budget one is the case I went looking for on the assumption it would be missing.

Worth a data point from our side, since this composes with something we just changed. We cut our default agent replicas from 7 to 2 after establishing that replica count bounds only in-process work: Job-mode tasks select a node without reserving it, and we had a single FleetNode carrying two Job-mode coders plus an in-process review at once. With that reduction, one claim per poll interval per pod would have been a real throughput cost, and this removes it. Fewer pods each filling both slot types in a pass is strictly better than more pods each filling one.

Also worth noting for anyone who does not set it: `maxSupervised()` falls back to `DefaultMaxSupervisedTasks` when `MaxSupervisedTasks` is unset, so the pass is bounded even in deployments that never configure a budget. That is the case that would have worried me most about removing the early return.

PR #1669, opened 2026-08-25
fix(foreman): round-robin Job-mode placement across eligible nodes

Defilan:
## What

Job-mode AgenticTasks now rotate across the eligible FleetNodes instead of
always landing on the alphabetically first one.

## Why

Refs #1634.

`reserveFirstFitNode` sorts nodes by name, and the Job-mode branch returned the
first eligible node without reserving it:

```go
if jobMode {
    return n.Name, nil
}
```

Not reserving is correct and deliberate (#1496): Job-mode work runs in an
ephemeral Job pod, not on the claiming node, so the node must stay free for
in-process tasks. The bug is that "first eligible" became a **constant** rather
than a rotation, because the reservation was what previously made the scan
advance. Measured on the fleet: 10 of 11 Job-mode tasks landed on one node
while seven other Ready nodes sat idle.

## How

A first pass collects the eligible node indices. The Job-mode branch then picks
`eligible[liveJobModeTasks(...) % len(eligible)]`, so successive assignments
advance across the eligible set. The in-process path is unchanged and still
reserves exactly as before.

Two deliberate properties:

- **`#1496` is preserved.** Job-mode placement still does not reserve the node.
  Reintroducing reservation would bring back the bug where one node holds its
  only slot for a Job's whole lifetime (#1559).
- **A failed List degrades to index 0 rather than wedging.** If the rotation
  count cannot be read, the task still dispatches to the first eligible node
  instead of stalling.

Round-robin rather than least-loaded, because the controller has no load signal
for Job-mode work: the node is deliberately not claimed, so `currentTask` is
empty by design. Round-robin is the smallest change that makes the choice
advance.

## Verification

- `go test ./internal/foreman/controller/...` green (104 specs)
- `make lint-deadcode` -> `0 issues.`; `gofmt` and `go vet` clean
- **Mutation-checked.** Replacing the rotation with `eligible[0]` (the old
  always-first behaviour) fails exactly the new spec and nothing else:

```
[FAIL] rotates Job-mode assignments across eligible nodes (#1634)
103 Passed | 1 Failed
```

One targeted failure out of 104 confirms the test is specific to the behaviour
it claims, and the other 103 confirm the refactor did not disturb existing
scheduling.

The spec creates three eligible nodes **out of alphabetical order** (so the
sort-by-name genuinely orders them), a Job-mode agent, and five tasks, then
drives `Reconcile` and asserts the assignments spread rather than repeating.

## Checklist

- [x] Tests added/updated

Assisted-by: Foreman (in-cluster coder agent running Ornith-1.5-35B-A3B) for
the implementation; independently reviewed and mutation-checked by hand before
this PR was opened. The in-loop reviewer's verdict was not treated as
sufficient.

### DIFF
--- a/internal/foreman/controller/agentictask_controller.go
+++ b/internal/foreman/controller/agentictask_controller.go
@@ -519,16 +519,20 @@ func (r *AgenticTaskReconciler) reserveFirstFitNode(ctx context.Context, task *f
 		// the node must not be claimed: leave its CurrentTask untouched so
 		// it stays free for in-process tasks. See #1496.
 		//
-		// The eligible list is sorted by name, so pick the eligible node at
-		// the round-robin index (live job-mode task count mod eligible
-		// count). Reserving would undo #1496, but the reservation is what
-		// used to make the scan advance; without a rotation every Job-mode
-		// task lands on the alphabetically first node, so a fleet of N nodes
-		// can supervise only one Job at a time. See #1634.
+		// Reserving would undo #1496, but the reservation is what used to
+		// make the scan advance; without it every Job-mode task lands on the
+		// alphabetically first node, so a fleet of N nodes can supervise only
+		// one Job at a time. See #1634.
+		//
+		// Pick the eligible node currently supervising the fewest live
+		// Job-mode tasks. An earlier revision rotated on a scalar count of
+		// live tasks (count % len(eligible)); that spreads a burst but
+		// collapses to index 0 -- first-fit -- whenever nothing else is in
+		// flight, which is the common serial case.
 		if len(eligible) == 0 {
 			return "", nil
 		}
-		return nodes.Items[eligible[r.jobModeRotation(ctx, eligible, task)]].Name, nil
+		return nodes.Items[r.leastLoadedJobModeNode(ctx, nodes.Items, eligible, task)].Name, nil
 	}
 	for _, i := range eligible {
 		n := &nodes.Items[i]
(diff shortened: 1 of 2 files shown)


joryirving:
Nice improvement over first-fit, but I think the rotation regresses to first-fit in the serial case, which may be the common one.

`jobModeRotation` is `liveJobModeTasks(ctx, task) % len(eligible)`, and `liveJobModeTasks` counts only tasks that are both in an in-flight phase and already carry an `AssignedNode`. When nothing else is in flight the count is 0, the index is 0, and the task goes to the alphabetically first eligible node. So an agent dispatching tasks one at a time always picks the same node, which is the behaviour the PR is fixing.

On our fleet last night, five Job-mode coder tasks ran and two of them started with nothing else in flight, so those two would still pin. It spreads a burst and does nothing for a trickle.

The loop that produces the count already reads `t.Status.AssignedNode` for every live task, which is a real per-node load signal for Job-mode work. Tallying those into a `map[string]int` and choosing the eligible node with the smallest tally costs a few lines, is immune to the count collapsing, and would also stop treating a node already supervising three Jobs the same as an idle one. Worth noting the godoc above says least-loaded "would require a real load signal the controller does not have", three lines above the loop that reads it.

The new spec is a genuine regression test for the burst case and would fail against first-fit, but it creates five tasks and never moves any to a terminal phase, so the count only ever rises. A case where one task completes before the next is scheduled would catch the above.

Smaller thing: the filter is on `AgentRef.Name`, so two Job-mode agents each count only their own tasks and both start at index 0, landing on the same node.

Defilan:
You're right on all four points, and I've pushed `15c08c098` taking the design you suggested.

Confirmed each against the code before changing anything:

- **The rotation does collapse.** `liveJobModeTasks % len(eligible)` with nothing in flight is `0 % n`, so `eligible[0]`, and `eligible` is name-sorted. Serial dispatch picks the same node every time.
- **The signal was already in the loop.** `liveJobModeTasks` read `t.Status.AssignedNode` for every live task and then threw it away to return a scalar. Keeping it as a `map[string]int` is the whole change.
- **The godoc was wrong.** It claimed least-loaded "would require a real load signal the controller does not have" three lines above the loop that reads exactly that signal. Corrected.
- **The `AgentRef` filter was wrong.** Dropped, so all Job-mode work counts toward a node's load.

Two specs, each failing against the old rotation and passing after. The depth one stacks three live Jobs on `nodeA` — three chosen deliberately, because `3 % 3 == 0` sent the rotation to `eligible[0]`, which *is* `nodeA`, the busiest node in the fleet. The cross-agent one has agent A live on `nodeA` and asserts agent B goes elsewhere.

Your point about the existing spec is also right, and it caught something in my own verification. I'd mutation-tested that spec by swapping the rotation for `eligible[0]` and watched exactly one spec fail, and reported that as proof. It only proved the spec detects *that* mutation — since it never retires a task, its count only climbs, so it can only ever test a burst. A passing mutation check bounds what a test catches, not what it should have caught.

One thing I did **not** do: I first wrote the serial-trickle spec you implied, and it fails against least-loaded too. With nothing in flight every node scores 0, ties break to the first, and all three tasks land on `aaa-a`. Thinking about it, that's correct rather than a regression — when the fleet is genuinely idle, the first node is as good as any, and asserting otherwise encodes an arbitrary spreading requirement. So the serial case still lands on `eligible[0]`, now by choice and documented as such. If you're seeing real harm from that on your fleet I'd like to hear it, because that would mean there's a cost I'm not modelling.

joryirving (approved):
This looks right to me, and the serial case is fine as you've framed it — I don't think there's a cost you're missing.

On our fleet there's no harm, for a reason worth stating because it may generalise: our FleetNodes are agent pods rather than machines, and Job-mode work runs in its own pod, so which node supervises a Job barely constrains anything. We had one node carrying two Job coders and an in-process review simultaneously with no ill effect. The real harm from concentration was back when Job-mode reserved its node and long coder Jobs starved in-process reviewers, and #1496 removed that. Our actual bottleneck is a single-slot inference backend, which node choice cannot influence either way. So an idle fleet landing on `eligible[0]` costs us nothing.

The depth spec is nicely chosen. Three tasks against three nodes means the old `3 % 3 == 0` sent the rotation to `eligible[0]`, which is the busiest node — so it fails in the most pointed way available rather than incidentally.

I also checked the three things that could still bite and they are all fine: ties break deterministically to the earlier index via the strict `<`, the `len(eligible) == 0` guard sits immediately before the indexing, and a failed List yields an empty tally that falls back to dispatching rather than stalling.

One residual worth noting rather than fixing: two tasks reconciled inside the informer cache's lag still read the same tally and can co-locate. That was true of the rotation too. Least-loaded degrades better though, because the next dispatch sees the real load and moves away, while the rotation had no memory to correct with.

Your point about the mutation check is the more useful half of this, and it applies well beyond this PR. A passing mutation check bounds what a test catches, not what it should have caught — I'd been treating one as sufficient evidence rather than as evidence about a single specific mutation.

PR #1707, opened 2026-08-29
feat(foreman): make draft PR creation configurable per Agent

### LINKED ISSUE #1706: Make draft PR creation configurable: it inserts a mandatory human step into the unattended loop
## Ask

Make the draft flag on Foreman-opened PRs configurable — a chart value or an Agent CR field — rather than hardcoded. Either default is fine; the request is only that it be selectable.

## Why

`#1703` opens Foreman PRs as drafts, and the intent is clear from the code comment:

> `// draft opens it as a work-in-progress (a human reads it and marks it ready...)`

That is a reasonable default for a human-supervised setup. The problem is that it is not overridable: `pkg/foreman/agent/executor_native.go:2089` passes a literal `true`, and there is no CRD field or chart value that reaches it.

In our deployment that turns an unattended loop into a supervised one. Our PR reviewer skips drafts by convention:

```yaml
# .github/workflows/ai-pr-review.yaml
on:
  pull_request:
    types: [opened, reopened, synchronize, ready_for_review, labeled]
jobs:
  review:
    if: ${{ github.event_name != 'pull_request' || !github.event.pull_request.draft }}
```

Skipping drafts is the normal reading of what a draft means — "not ready for review" — and we would not want to change it, because the same workflow serves human PRs. But it means the pipeline now stops after the coder: Foreman opens a draft, nothing reviews it, and it waits until a person marks it ready. Every PR gains a manual step.

That is also in tension with `#1653` in 0.9.21, `llmkube foreman run, the unattended orchestration loop`, shipped two releases earlier. One release adds an unattended loop; the next inserts a human gate into it.

## Suggested shape

Whatever fits the existing config surface best — a chart value like `foreman.agent.openPRsAsDraft`, or a field on the Agent CR so different pools can differ (a fork-pinned agent might reasonably want drafts while the main loop does not). Defaulting to `true` to preserve current behaviour is fine by us; we would set it to `false`.

## Context

Not asking for a revert. Draft-by-default is a defensible choice and I can see why a human-in-the-loop deployment would want it. It just needs to be a decision the operator makes rather than one baked into the executor, since whether a draft blocks your pipeline depends entirely on what your review tooling does with drafts.

Defilan:
## What

Adds `AgentSpec.openPullRequestsAsDraft` so an operator can choose whether
Foreman-opened PRs are drafts. `nil` means `true`, preserving 0.9.22 behaviour.

## Why

Fixes #1706

#1703 opened Foreman PRs as drafts with a hardcoded `true` at
`executor_native.go:2089`. Nothing reached it from a CRD field or a chart value,
so the choice was baked into the executor.

For a deployment whose reviewer workflow gates on
`!github.event.pull_request.draft` — the ordinary reading of what a draft means
— that turns the unattended loop shipped in #1653 into a supervised one:
Foreman opens a draft, nothing reviews it, and every PR waits for a person to
mark it ready. One release added an unattended loop; the next put a human gate
in the middle of it.

Reported by @joryirving, who traced it to the line.

This was an omission in the issue I wrote for #1703, not a deliberate call. It
asked for the flag to be a field on the call path so a future caller would not
have to re-plumb it, and it is — but it never asked for a config surface, so the
plumbing stopped one frame short of anything an operator could set.

## How

`AgentSpec.OpenPullRequestsAsDraft *bool`, resolved by `openPRsAsDraft` and
threaded into `maybeOpenPullRequest`, which now takes the `*Agent` already in
scope at both call sites.

**On the Agent rather than the Workload**, which is where `openPullRequest`
lives. Whether a draft blocks the pipeline is a property of the deployment's
review tooling, not of an individual run: a CI job that skips drafts skips every
draft. Putting it on the Workload would mean setting it on every generated one,
which is the situation the reporter is trying to get out of. Agent-level also
gives per-pool granularity, so a fork-pinned agent can open drafts while an
unattended loop does not.

`Workload.spec.openPullRequest` is untouched and still decides whether a PR is
opened at all. This only decides what kind.

**`nil` means `true`** so an upgrade changes nothing for anyone who has not
opted in. A nil `*Agent` also yields `true`: the caller has no configuration to
honour, and the safer reading of "unknown" is the human-supervised one.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally — `internal/foreman/...` and `pkg/foreman/...` green, run unfiltered
- [x] `make lint` passes locally — `GOOS=linux golangci-lint run ./...` 0 issues on a clean cache, `make lint-deadcode` 0 issues, `gofmt` clean
- [x] AI assistance (if any) is disclosed below, per [CONTRIBUTING.md](../CONTRIBUTING.md#ai-assisted-and-agent-contributions)
- [x] Documentation updated (if user-facing change) — the field's GoDoc carries the rationale and ships in both the chart CRD and `config/crd/bases`; `make manifests` and `make foreman-chart-crds` regenerated, `make check-helm-rbac` clean

## Verification

| Mutation | Result |
|---|---|
| Restore the hardcoded `true` at the call site | CAUGHT |
| Flip the `nil` default to `false` | CAUGHT |

The first one is the reason there are two tests rather than one. A unit test on
`openPRsAsDraft` alone passed even with the call site reverted to a hardcoded
`true` — the resolver was correct and simply unused. `TestMaybeOpenPullRequest_DraftFollowsAgent`
asserts through the fake code host that the Agent's value actually reaches
`EnsurePR`, and that is the test the mutation fails.

## AI assistance

Written with Claude: diagnosis against the reported line, the Agent-versus-
Workload placement decision, implementation, and the mutation testing above.

### DIFF
--- a/api/foreman/v1alpha1/agent_types.go
+++ b/api/foreman/v1alpha1/agent_types.go
@@ -399,6 +399,27 @@ type AgentSpec struct {
 	// +optional
 	BashTimeoutSeconds int32 `json:"bashTimeoutSeconds,omitempty"`
 
+	// OpenPullRequestsAsDraft controls whether a pull request this agent
+	// opens is created as a draft (#1706). Nil means true, preserving the
+	// behaviour introduced in #1703.
+	//
+	// This is an Agent field rather than a Workload one because whether a
+	// draft blocks the pipeline is a property of the DEPLOYMENT's review
+	// tooling, not of an individual run: a CI job that skips drafts skips
+	// every draft, so an operator needs one place to decide rather than a
+	// field to set on every generated Workload. Per-pool granularity comes
+	// free, so a fork-pinned agent can open drafts while an unattended loop
+	// does not.
+	//
+	// Distinct from Workload.spec.openPullRequest, which decides whether a
+	// PR is opened at all. This only decides what kind.
+	//
+	// Reported by a downstream operator whose reviewer workflow gates on
+	// `!github.event.pull_request.draft`: draft-by-default turned their
+	// unattended loop (#1653) into a supervised one.
+	// +optional
+	OpenPullRequestsAsDraft *bool `json:"openPullRequestsAsDraft,omitempty"`
+
 	// Tools is the tool whitelist surfaced to the model on every turn.
 	// Unknown names are rejected at agent startup so a typo in an Agent
 	// CR fails loud rather than silently disabling a tool.
--- a/api/foreman/v1alpha1/zz_generated.deepcopy.go
+++ b/api/foreman/v1alpha1/zz_generated.deepcopy.go
@@ -299,6 +299,11 @@ func (in *AgentSpec) DeepCopyInto(out *AgentSpec) {
 		*out = new(StuckLoopDetectionSpec)
 		**out = **in
 	}
+	if in.OpenPullRequestsAsDraft != nil {
+		in, out := &in.OpenPullRequestsAsDraft, &out.OpenPullRequestsAsDraft
+		*out = new(bool)
+		**out = **in
+	}
 	if in.Tools != nil {
 		in, out := &in.Tools, &out.Tools
 		*out = make([]string, len(*in))
--- a/charts/foreman/templates/crds/agents.yaml
+++ b/charts/foreman/templates/crds/agents.yaml
@@ -424,6 +424,27 @@ spec:
                 maximum: 50
                 minimum: 0
                 type: integer
+              openPullRequestsAsDraft:
+                description: |-
+                  OpenPullRequestsAsDraft controls whether a pull request this agent
+                  opens is created as a draft (#1706). Nil means true, preserving the
+                  behaviour introduced in #1703.
+
+                  This is an Agent field rather than a Workload one because whether a
+                  draft blocks the pipeline is a property of the DEPLOYMENT's review
+                  tooling, not of an individual run: a CI job that skips drafts skips
+                  every draft, so an operator needs one place to decide rather than a
+                  field to set on every generated Workload. Per-pool granularity comes
+                  free, so a fork-pinned agent can open drafts while an unattended loop
+                  does not.
+
+                  Distinct from Workload.spec.openPullRequest, which decides whether a
+                  PR is opened at all. This only decides what kind.
+
+                  Reported by a downstream operator whose reviewer workflow gates on
+                  `!github.event.pull_request.draft`: draft-by-default turned their
+                  unattended loop (#1653) into a supervised one.
+                type: boolean
               provider:
                 default: local
                 description: |-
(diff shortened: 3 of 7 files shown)


joryirving (approved):
This is a better answer than the one I asked for. I asked for a flag; you worked out that it belongs on the Agent rather than the Workload and wrote down why, and the reasoning is right: whether a draft blocks a pipeline is a property of the deployment's review tooling, not of an individual run, so an operator needs one place to decide rather than a field on every generated Workload.

The defaulting is the part I would have got wrong. Nil resolving to true preserves what #1703 introduced, so an upgrade changes nothing for anyone who has not opted in, and a nil Agent resolving to draft rather than ready is the safer reading of "unknown". Both are worth the comments they carry.

Test coverage matches: nil agent, unset field, explicit true and explicit false in `TestOpenPRsAsDraft`, plus `TestMaybeOpenPullRequest_DraftFollowsAgent` at the integration level, so the resolution helper and the wiring are covered separately.

One thing to fix before merge, and it is mechanical. CRD Sync Check is failing because the generated deepcopy was not regenerated:

```
diff --git a/api/foreman/v1alpha1/zz_generated.deepcopy.go
@@ -299,6 +299,11 @@ func (in *AgentSpec) DeepCopyInto(out *AgentSpec) {
+	if in.OpenPullRequestsAsDraft != nil {
+		in, out := &in.OpenPullRequestsAsDraft, &out.OpenPullRequestsAsDraft
+		*out = new(bool)
```

`make generate manifests chart-crds foreman-chart-crds` and commit the result. Worth doing rather than waving through, since a pointer field without its deepcopy block is shallow-copied and two Agent objects would end up sharing the pointer.

For what it is worth on the downstream side: our Agent CRs are hand-authored rather than rendered by the chart, so `openPullRequestsAsDraft: false` on the coder agents is all we need once this lands. The per-pool granularity is genuinely useful too, not just incidental. We can leave the fork-pinned agent opening drafts while the unattended lane opens ready-for-review, which is exactly the split I wanted and did not think to ask for.

Defilan:
Fixed in ce00a804e. `CRD Sync Check` is green.

You were right that it was worth doing properly rather than waving through. I verified the consequence you named rather than trusting the regenerated block: with the deepcopy in place, `DeepCopy()` yields a distinct pointer and writing through the copy leaves the original untouched. Before the fix those two Agent objects would have shared it.

The miss was mine and it is instructive about my own checklist. I ran `make manifests` and `make foreman-chart-crds` and then checked `git status` for drift, which came back clean — but `make generate` was never run, so its drift could not show up in that check. A verification step that only looks for changes from the commands you ran cannot see the command you forgot. Running the full `generate manifests chart-crds foreman-chart-crds` chain and then checking for drift is the version that actually catches this, and that is now what I do.

Good to hear the per-pool split lands usefully. Fork-pinned agent on drafts, unattended lane ready-for-review was the case that made Agent the right home rather than Workload, so it is good to have that confirmed from the deployment side rather than inferred.

Issue #1731, opened 2026-09-01
[BUG] Coder post-push envtest gate never passes upstreamURL: bite check and non-Go lint silently skip against a stale fork base


Defilan:
## Bug Description

The coder's post-push envtest gate submits its gate Job without `upstreamURL`,
so the Job runs the pre-#1259 fallback and resolves its base from the **fork's**
`main` tip instead of the merge-base against the canonical repo. When the fork's
base ref is stale, change detection is skipped and the diff-scoped checks (the
bite check and non-Go lint) silently do not run, while the Job still reports
`GATE PASS`.

#1259 fixed exactly this failure mode. It was threaded through the verify path,
which resolves the URL in Go and passes it. The coder self-gate path was never
threaded and has no parameter to carry it.

## Actual Behavior

`EnvtestJobRunner.Run` has no `upstreamURL` parameter
(`pkg/foreman/agent/envtest_gate.go`):

```go
Run(
    ctx context.Context,
    taskNamespace, taskName, repository, branch, cloneURL string,
) (pass bool, ran bool, feedback string)
```

so the closure that implements it cannot supply one
(`cmd/foreman-agent/main.go`, `envtestJobRunnerImpl.Run`):

```go
args, err := json.Marshal(map[string]any{
    "repo":     repository,
    "branch":   branch,
    "checks":   []string{"test"},
    "cloneURL": cloneURL,
    "taskRef":  map[string]string{...},
})
```

`run_gate_job` reads `UpstreamURL` only from its tool args
(`pkg/foreman/agent/tools/run_gate_job.go`, `UpstreamURL: a.UpstreamURL`), so
the Job receives `UPSTREAM_URL=""` and takes its documented `else` branch:
"No upstream URL (freeform task without a repo slug): pre-#1259 behavior, fetch
the base ref from origin and use its tip."

Observed live on 2026-09-01. Every gate Job in the namespace, by task kind:

```
verify-task gates (4)   UPSTREAM_URL SET -> https://github.com/defilantech/LLMKube.git
code-task gates   (3)   UPSTREAM_URL EMPTY (no value key)
```

The split is exactly on task kind. On `wl-1684-scope-inert-code-1684`:

```
GATE-WARN: change detection skipped (diff vs merge-base names 207 files; base resolution suspect)
non-go lint: skipped, no usable base to diff against
GATE PASS
```

207 is the file count between the fork's `main` and upstream `main`, which were
9 commits apart at the time. The branch's real diff is 2 files.

[...]

Issue #1259, opened 2026-07-23
[BUG] Gate bite check diffs against stale fork main, reverting unrelated production files and emitting false GATE-FAIL


Defilan:
## Bug Description

The gate's bite check diffs the coder branch against the FORK's `main` tip instead of the branch's own base, and reverts production files to that tip. When the fork's `main` is stale, the reverted tree mixes ancient production code with branch-era test files across packages the coder never touched, producing spurious compile failures and a false `GATE-FAIL` on a correct branch.

## Steps To Reproduce

Observed live on workload `morning-890-cache-prep-helper` (2026-07-23):

1. Fork `main` was at `00dd7c00` (2026-07-10), 103 commits behind upstream `main`.
2. Coder produced branch `foreman/morning-890-cache-prep-helper/issue-890` (base `1b5bf08f`, a real recent upstream main commit; the coder commit `f0e6fcc8` touches only `pkg/cli`). The branch tree compiles clean, `go test ./pkg/cli/...` passes, coder verdict GO.
3. Gate job ran with `biteCheck`. Per `gate_job_template.yaml` / `run_gate_job.go`, the bite check does:
   - `git fetch --depth 1 origin "+refs/heads/$BASE:refs/remotes/origin/$BASE"` (origin = the fork)
   - `git diff --name-only "origin/$BASE" HEAD` (two-dot, against the fork main TIP)
   - `git checkout "origin/$BASE" -- "$f"` for each production file in that diff
4. Because fork main was 103 commits stale, the diff swept in every file changed upstream since 2026-07-10. Production across `pkg/foreman/*` was reverted to July-10 state while current test files remained: `undefined: applyTrailingWSMatch`, `undefined: DriftPinnedPrefix`, `undefined: gateJobName`, etc. (all symbols merged after 2026-07-10).
5. Gate emitted `GATE-FAIL` ("one or more gate checks failed") even though the same log shows `bite check: tests fail against pre-change production (biting test, good)` and the standard checks on the intact tree were green.

Verified: the exact branch tip compiles all "failed" packages cleanly, and both the branch base and current main contain every "undefined" symbol. The only tree that lacks them is fork-main-era production.

## Expected Behavior

The bite check reverts production to the branch's OWN base, so only files the coder actually changed are reverted, and unrelated packages never break.

Suggested fix directions:
[...]

PR #1732, opened 2026-09-01
fix(foreman): pass upstreamURL to the coder post-push envtest gate

Defilan:
## What

Thread the canonical repo URL through the coder's post-push envtest gate so its
gate Job diffs against `merge-base(HEAD, canonical base)` instead of falling
back to the fork's base tip.

## Why

Refs #1731

`EnvtestJobRunner.Run` had no parameter for the upstream URL, so the wiring
closure in `cmd/foreman-agent` could not supply one and the gate Job received
`UPSTREAM_URL=""`. The gate script then takes its documented branch:

> No upstream URL (freeform task without a repo slug): pre-#1259 behavior, fetch
> the base ref from origin and use its tip.

Origin is the fork. Once the fork's base ref drifts, the fallback diff looks
enormous, change detection is skipped, and the bite check and non-Go lint
silently do not run while the Job still reports `GATE PASS`.

The tasks were not missing a repo slug. All three carried
`payload.repo = defilantech/LLMKube`; the value simply had no route to the Job
on this path. #1259 fixed this exact failure mode for the verify path, which
resolves the URL in Go and passes it, but the coder self-gate path was never
threaded.

Observed live on 2026-09-01, split exactly on task kind:

```
verify-task gates (4)   UPSTREAM_URL SET -> https://github.com/defilantech/LLMKube.git
code-task gates   (3)   UPSTREAM_URL EMPTY (no value key)
```

and on `wl-1684-scope-inert-code-1684`:

```
GATE-WARN: change detection skipped (diff vs merge-base names 207 files; base resolution suspect)
non-go lint: skipped, no usable base to diff against
GATE PASS
```

207 was the file count between the fork's `main` and upstream `main`, 9 commits
apart at the time. That branch's real diff is 2 files.

What this costs is narrower than it first looks: `make test` still ran and
genuinely passed, so this was never a false test pass. What was lost is the
diff-scoped tier, including the bite check, which is the primary defense against
tests that assert nothing. It failed open and silent.

## How

Three changes, one per layer:

1. `pkg/foreman/agent/envtest_gate.go` adds `upstreamURL` to the
   `EnvtestJobRunner.Run` signature and threads it through
   `evaluatePostPushEnvtest`. The godoc spells out why it is separate from
   `cloneURL`: `cloneURL` is where the branch lives (the fork), `upstreamURL` is
   what it was cut from.
2. `pkg/foreman/agent/executor_native.go` resolves it at the call site with
   `resolveUpstreamForRun`, the same seam the self-commit recovery path uses, so
   an `UpstreamURLForRepo` test override still applies. Resolving inside
   `postPushGateDecision` rather than widening its parameter list keeps the
   existing callers untouched.
3. `cmd/foreman-agent/main.go` puts `upstreamURL` in the `run_gate_job` args
   map. `run_gate_job` already validates it against `upstreamURLPattern` and
   treats empty as "no upstream", so a genuinely absent repo slug degrades
   exactly as it does today.

Both new assertions are mutation-checked rather than assumed:

```
mutation 1 (args map -> empty):   test FAILED (caught)
mutation 2 (evaluator drops it):  test FAILED (caught)
```

The runner test asserts the created Job's `UPSTREAM_URL` env var, not the args
map, because the env var is what the gate script actually reads.

### Noted, deliberately not changed

`resolveUpstreamForRun` documents itself as mirroring the `resolveUpstream`
closure at the top of `Execute`, but it does not consult `e.CodeHost` the way
that closure does. That divergence predates this PR and affects the self-commit
recovery path today. Making them agree is a real change in behaviour when a
`CodeHost` is configured, so it does not belong bundled into a bug fix. Worth a
follow-up.

## Checklist

- [x] Tests added/updated
- [ ] `make test` passes locally — ran the affected packages
      (`cmd/foreman-agent`, `pkg/foreman/agent`, `pkg/foreman/agent/tools`), all
      green; the full suite has not been run locally, CI covers it
- [x] `make lint` passes locally (`GOOS=linux golangci-lint run ./...`, 0 issues)
- [x] AI assistance disclosed: authored with Claude Code (band 3 under
      CONTRIBUTING.md). The root cause was diagnosed from live cluster state and
      the maintainer owns the review conversation.

`Refs` rather than `Fixes`: the fix is unit-verified and mutation-checked, but
not yet confirmed end to end on the fleet. The issue should close once a coder
gate Job is observed carrying a non-empty `UPSTREAM_URL`.

### DIFF
--- a/cmd/foreman-agent/envtest_runner_test.go
+++ b/cmd/foreman-agent/envtest_runner_test.go
@@ -74,6 +74,7 @@ func TestEnvtestJobRunnerStampsTaskName(t *testing.T) {
 		ctx, taskNS, taskName,
 		"defilantech/LLMKube", "foreman/issue-893",
 		"https://github.com/Defilan/LLMKube.git",
+		"https://github.com/defilantech/LLMKube.git",
 	)
 
 	var job batchv1.Job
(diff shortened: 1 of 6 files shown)


Defilan:
## Correction: the bite check was never on this path

Verified on the fleet after the fix landed, and one claim in the original report
was wrong. The issue body and PR #1732 both said the base-resolution bug caused
"the bite check and non-Go lint" to silently skip. The bite check is not part of
the coder's post-push envtest gate at all.

The rendered gate scripts, compared by echoed phase header:

```
code-task gate    clone -> make test -> non-go lint                     (3 phases)
verify-task gate  clone -> fmt/vet/lint/deadcode -> test -> generate
                  -> manifests -> chart-crds -> ... -> non-go lint
                  -> bite check -> per-hunk mutation                   (18 phases)
```

The coder's gate is submitted with `checks: ["test"]` and no `biteCheck`, so the
bite and per-hunk phases are never rendered into it. The base-resolution block
appears in both scripts because it is shared, and its comments mention the bite
check, which is what I misread.

**What the bug actually cost, corrected:** change detection was skipped on
code-task gates, and with it the non-Go lint. That is the whole blast radius.
`make test` always ran and always reported honestly.

**What the fix is confirmed to have restored,** on
`wl-1729-depwait-default-code-1729` with the fork's `main` 12 commits and 213
files behind upstream, which is the worst case for the old fallback:

```
UPSTREAM_URL                      https://github.com/defilantech/LLMKube.git  (was empty)
"base resolution suspect"         0 occurrences  (was 1)
"change detection skipped"        0 occurrences  (was 1)
"no usable base to diff against"  0 occurrences  (was 1)
non-go lint                       "no checkable non-Go files changed"  (a real answer, was a skip)
```

So the fix is verified end to end and the issue is genuinely closed by it. The
severity in the original write-up was overstated: this restored non-Go lint on
coder gates, not a mutation-testing defense.

**Separate observation, not a claim about this bug:** the coder self-gate is
considerably thinner than the verify gate. If that asymmetry is deliberate,
nothing to do. If the intent was for the coder to self-gate on the same bite
check the verify stage runs, that is its own issue and worth filing separately.

Issue #1739, opened 2026-09-01
[BUG] Dispatched condition is never cleared on a terminal Workload, so a Completed Workload still reports work in flight


Defilan:
## Bug Description

The `Dispatched` condition is never cleared when a Workload goes terminal, so a
Completed Workload keeps `Dispatched: True` with a stale message claiming work
is still in flight.

## Actual Behavior

Observed on `wl-1729-depwait-default` (2026-09-01), and identically on
`wl-1684-scope-inert`:

```
phase: Completed
  Dispatched True   1 in-flight, 2 on-target, 0 incomplete, 0 failed, 0 already-resolved
  Completed  True   3/3 child tasks on-target Succeeded
```

All three children had reached Succeeded. The Workload is Completed. The
`Dispatched` condition still claims one child is in flight.

The cause is structural rather than a race. In
`internal/foreman/controller/workload_controller.go`, the string `"Dispatched"`
appears exactly once in non-test code, at roughly line 803, inside the `default:`
branch of the terminal-state switch:

```go
default:
    w.Status.Phase = foremanv1alpha1.WorkloadPhaseDispatched
    setCondition(&w.Status.Conditions, metav1.Condition{
        Type:   "Dispatched",
        Status: metav1.ConditionTrue,
        ...
    })
}
```

The terminal branches set `conditionTypeCompleted` and never touch
`Dispatched`, so whatever value the last in-flight pass wrote is what survives.
The condition is write-only on one path and never reconciled on the others.

## Expected Behavior

A terminal Workload does not advertise `Dispatched: True`. Either the condition
is set to `False` with a terminal reason when the Workload completes or fails,
or it is updated so its message matches the terminal counts.

## Suggested Fix

**The decision is made, do not re-derive it:** set `Dispatched` to
`ConditionFalse` with a reason naming the terminal outcome on every branch that
sets a terminal phase, rather than deleting the condition or rewriting its
message.

Reasoning: a condition that disappears is harder to consume than one that flips
to False, `kubectl wait --for=condition=Dispatched=false` becomes usable, and the
`LastTransitionTime` then records when dispatch actually ended. Leaving a True
condition with corrected counts would still misreport the state, which is the
bug.

The change is confined to the terminal branches of the same switch that
currently sets it. Do not alter what `classifyChildren` counts, and do not
change the `Completed` condition's message.

## Scope

Deliberately separate from #1738, which reconciles the child count against the
planned count. Both touch this switch, but they are different branches, different
[...]

Issue #1738, opened 2026-09-01
[BUG] Workload child count is never reconciled against the planned count, so a deleted child reads as ordinary in-flight work


Defilan:
## Bug Description

The Workload's `Dispatched` condition counts only the children that still exist,
so a child that was deleted is not counted at all. The remaining children are
reported as in-flight and the Workload sits `Dispatched` forever with no
indication that anything is missing.

This is the second half of #1729. The first half (#1736) bounded the dependent
task's own wait; this is about the Workload's reporting, which is why the stall
went unnoticed for six days.

## Actual Behavior

`rollup` in `internal/foreman/controller/workload_controller.go` classifies
children with `classifyChildren(children)`, where `children` is what was
observed via the `Owns()` watch. A deleted child is simply absent from that
slice, so it lands in no bucket and the counts silently shrink.

Nothing compares the observed children against what the planner actually
emitted, even though that list is already recorded in `w.Status.Tasks`.

The `Dispatched` condition message is built at roughly line 806:

```go
Message: fmt.Sprintf("%d in-flight, %d on-target, %d incomplete, %d failed, %d already-resolved",
    c.inFlight, c.succeeded, c.incomplete, c.failed, c.alreadyResolved),
```

Every count there is derived from surviving children. A Workload that planned 3
tasks and lost one reports on 2 and reads as healthy work in progress.

## Expected Behavior

When the number of children the Workload can observe is smaller than the number
the planner emitted, the `Dispatched` condition says so, rather than reporting
the survivors as ordinary in-flight work.

## Suggested Fix

`w.Status.Tasks` holds the planner's emitted task references and `children`
holds what was observed, so both inputs already exist at the call site.

**The decision is made, do not re-derive it:** a missing child sets a distinct
condition `Reason` and is named in the `Message`. It must NOT fail the Workload.
Deleting a task by hand is a legitimate operator action, and the goal here is
diagnosability, not enforcement. A Workload whose child vanished should read as
"something is missing, look at it", never as an automatic failure.

### The trap that makes a naive count comparison wrong

A bare `len(children) != len(w.Status.Tasks)` check will flap, and this is the
single most important thing to get right.

The escalation and iteration paths deliberately **synthesize placeholder
children** so rollup sees freshly created tasks. From
`internal/foreman/controller/workload_escalation.go`:

> A cache-backed List here could miss tasks created microseconds ago (informer
> lag) and let rollup mark the Workload Failed before the create's watch event
[...]

PR #1741, opened 2026-09-01
fix(foreman): clear Dispatched condition on terminal Workloads

Defilan:
## What

Flip the `Dispatched` condition to `False` when a Workload reaches a terminal
phase, so a Completed Workload stops advertising work in flight.

## Why

Refs #1739

`"Dispatched"` was written in exactly one place, the `default:` branch of the
terminal-state switch, and never touched again. The terminal branches set the
`Completed` condition and left `Dispatched` holding whatever the last in-flight
pass wrote. The condition was write-only on one path and never reconciled on the
others.

Observed on `wl-1729-depwait-default`, and identically on `wl-1684-scope-inert`:

```
phase: Completed
  Dispatched True   1 in-flight, 2 on-target, 0 incomplete, 0 failed, 0 already-resolved
  Completed  True   3/3 child tasks on-target Succeeded
```

All three children had reached Succeeded. The Workload is Completed. The
condition still claimed one child was in flight.

## How

A `setDispatchedTerminal` helper sets the condition to `ConditionFalse` with a
`Reason` naming the terminal outcome, called from **every** branch that sets a
terminal phase: all-succeeded, pure already-resolved, the mixed
succeeded-plus-already-resolved case, and the failed case. Missing any one of
them would leave the same stale condition on that path.

The condition is flipped rather than deleted. A condition that disappears is
harder to consume than one that reads False: `kubectl wait
--for=condition=Dispatched=false` stays usable, and `LastTransitionTime` then
records when dispatch actually ended. Leaving it True with corrected counts
would still misreport the state, which is the bug.

The string literal is also replaced with a `conditionTypeDispatched` constant,
matching how `conditionTypeCompleted` and the other condition types in this file
are already declared.

## Scope

Deliberately separate from #1738, which reconciles the child count against the
planned count. Both touch this switch, but they are different branches,
different fixes and different tests. `classifyChildren` is untouched and the
`Completed` condition's reason and message are unchanged.

## Tests

- a Workload that reaches a terminal phase: `Dispatched` is present and
  `ConditionFalse`
- a Workload still in flight: `Dispatched` remains `ConditionTrue`

The second case is what stops a naive implementation from simply hard-coding the
condition to False.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally — ran in full in the clean-room gate Job for
      this branch (GATE-PASS), including the envtest packages
- [x] `make lint` passes locally — the same gate Job runs fmt, vet, lint and
      lint-deadcode
      conversation.

Note for merging: this PR and the one for #1738 both edit the `default:` branch
of the same switch and **do** conflict. Merge one, then rebase the other.

### DIFF
--- a/internal/foreman/controller/workload_controller.go
+++ b/internal/foreman/controller/workload_controller.go
@@ -96,6 +96,13 @@ const conditionTypeTruncated = "Truncated"
 // a terminal way.
 const conditionTypeCompleted = "Completed"
 
+// conditionTypeDispatched is True while children are still in flight, and
+// flips to False (with a reason naming the terminal outcome) when the
+// Workload reaches a terminal phase. Keeping it as a False condition (rather
+// than deleting it) keeps `kubectl wait --for=condition=Dispatched=false`
+// usable and records when dispatch actually ended via LastTransitionTime.
+const conditionTypeDispatched = "Dispatched"
+
 // conditionTypeCloudReviewersSuppressed is True when the sovereignty
 // gates (operator --allow-cloud-providers + Workload AllowCloudReviewers)
 // caused the reconciler to omit one or more reviewer Agents from the
@@ -755,6 +762,7 @@ func computeTerminalState(w *foremanv1alpha1.Workload, c childCounts, now metav1
 			Message:            fmt.Sprintf("%d/%d child tasks on-target Succeeded", c.succeeded, c.total),
 			LastTransitionTime: now,
 		})
+		setDispatchedTerminal(&w.Status.Conditions, "AllChildrenSucceeded", now)
 	case c.inFlight == 0 && c.failed == 0 && c.incomplete == 0 && c.succeeded == 0 && c.alreadyResolved > 0:
 		// Pure ALREADY-RESOLVED workload — nothing actually attempted.
 		w.Status.Phase = foremanv1alpha1.WorkloadPhaseCompleted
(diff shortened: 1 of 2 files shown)


Defilan (commented):
Reviewed the diff. Correct, and it covers the case a partial fix would miss.

**All four terminal branches are covered, not just the obvious one.** I checked
each call site: `setDispatchedTerminal` is invoked from the all-succeeded
branch, the pure already-resolved branch, the mixed
succeeded-plus-already-resolved branch, and the failed branch. Missing any one
of them would leave exactly the same stale condition on that path, which is the
most likely way this fix gets half-done. It is not half-done here.

**Flipping to False rather than deleting is the right call** and the godoc says
why. `kubectl wait --for=condition=Dispatched=false` stays usable, and
`LastTransitionTime` records when dispatch actually ended. A deleted condition
would have neither property.

**The second test is what gives the first one meaning.** "A Workload still in
flight keeps `Dispatched` True" is what stops an implementation that simply
hard-codes the condition to False from passing. Worth keeping if this area is
refactored later.

**Small scope addition, and I think a good one.** The change also replaces the
`"Dispatched"` string literal with a `conditionTypeDispatched` constant, which
was not strictly required. It matches how `conditionTypeCompleted` and the other
condition types in this file are already declared, and it is what made it easy
to verify that the literal appeared in exactly one place before this change.

**Merge ordering.** This PR and #1743 both edit the `default:` branch of the
same switch. `git merge-tree` reports no conflict markers, but a real `git merge`
fails on `workload_controller.go`. Merge one, rebase the other; the test file
auto-merges cleanly.

Scope is clean otherwise: `classifyChildren` untouched, and the `Completed`
condition's reason and message unchanged.

Context on where the bug came from, since it is not in the diff: this was
observed live on two separate completed Workloads (`wl-1729-depwait-default`
and `wl-1684-scope-inert`), both reporting `Dispatched True, "1 in-flight"`
alongside `Completed True, "3/3 child tasks on-target Succeeded"`.

Issue #1735, opened 2026-09-01
[BUG] Per-hunk mutation pass reports comment-only hunks as uncovered, so updating a stale godoc always earns a GATE-WARN


Defilan:
## Bug Description

The per-hunk mutation pass reports comment-only hunks as uncovered. Because of
how the check works this is systematic, not occasional: reverting a comment can
never turn a test red, so every added comment block that is not adjacent to
added code will be flagged, on every run, forever.

It penalises exactly the behaviour we ask coders for. Updating a godoc that a
change has made false is correct and required, and it earns a GATE-WARN.

## Actual Behavior

Observed on `wl-1729-depwait-default-verify-1729` (2026-09-01):

```
=== per-hunk mutation coverage (base=2b35d71454058abe91cdf5a22c5a42e6890ebceb) ===
GATE-WARN: hunk check: package internal/foreman/controller has an uncovered hunk
           in internal/foreman/controller/agentictask_controller.go:599-602
hunk check: done
```

Lines 599-602 on that branch:

```go
// elapsed. An unset or zero TimeoutSeconds falls back to the package-level
// defaultDepWaitBudget so the wait is bounded even on paths (such as the
// planner) that never stamp the field; an explicitly set positive
// TimeoutSeconds always wins.
```

Non-comment lines in that range: 0. It is the godoc immediately above
`depWaitExpired`, updated in the same change because the previous text ("An
unset or zero TimeoutSeconds leaves the wait unbounded") became false.

The actual behavioural change sits at 604-607 and was NOT flagged, which is
correct: the same run's all-or-nothing bite check passed with "tests fail
against pre-change production (biting test, good)". So the test genuinely bites
and only the comment hunk was reported.

## Root Cause

`splitProductionHunks` in `pkg/foreman/agent/mutation_gate.go` (around line 962)
groups every contiguous run of added lines into a hunk with no filtering:

```go
if strings.HasPrefix(line, "+") && !strings.HasPrefix(line, "+++") {
    cur = append(cur, line[1:])
    continue
}
// A context or removed line ends the current added block.
flush()
```

A comment-only added block becomes a `productionHunk` like any other.
`CheckPerHunkCoverage` then reverts it and runs the package's tests, which pass,
and the hunk is classified uncovered. The classification is working as written;
the input should never have reached it.

## Expected Behavior

A hunk that contains no Go statement is not a candidate for mutation coverage
and is skipped before it is reverted. Skipping it is also a straight saving:
[...]

PR #1742, opened 2026-09-01
fix(agent): skip comment-only hunks in the per-hunk mutation pass

Defilan:
## What

Skip comment-only hunks in the per-hunk mutation coverage pass, so updating a
stale godoc stops earning a `GATE-WARN`.

## Why

Refs #1735

The pass reverts each added hunk and reports it uncovered when the package's
tests stay green. Reverting a comment can never turn a test red, so **every**
comment-only hunk is reported as uncovered, on every run, forever. This is
systematic rather than occasional.

Observed on `wl-1729-depwait-default-verify-1729`:

```
=== per-hunk mutation coverage (base=2b35d71454058abe91cdf5a22c5a42e6890ebceb) ===
GATE-WARN: hunk check: package internal/foreman/controller has an uncovered hunk
           in internal/foreman/controller/agentictask_controller.go:599-602
```

Lines 599-602 on that branch contain **zero** non-comment lines. They are the
godoc above `depWaitExpired`, updated in the same change because the previous
text ("An unset or zero TimeoutSeconds leaves the wait unbounded") had been made
false by that change.

So the check penalised exactly the behaviour we ask coders for. The real
behavioural change in that diff was not flagged, correctly: the same run's
all-or-nothing bite check reported "tests fail against pre-change production
(biting test, good)".

`CheckPerHunkCoverage`'s own godoc says it is "Reporting only ... until the
false-positive rate on real diffs is known". On the first real diff observed,
the only finding it produced was a false positive.

## How

`splitProductionHunks` records a hunk only when at least one of its lines,
after `strings.TrimSpace`, is non-empty and does not begin with `//`.

The filter lives in the `flush` closure, not in the caller. That placement is
deliberate: a hunk dropped in `flush` never becomes a `productionHunk`, so it
also never consumes `CheckPerHunkCoverage`'s `maxPerHunkHunks` budget. A hunk
that was never a candidate should not use up the budget that bounds how many
real candidates get checked.

`CheckPerHunkCoverage`'s classification logic is untouched. What changes is
which hunks become candidates, not what makes a candidate uncovered.

### Accepted gap, documented in the godoc

A `/* */` block comment is not caught by a `//` prefix test. Tracking
block-comment state line by line is out of scope here, and the godoc says so
rather than implying the filter is complete. The `//` case is the one actually
being hit; a block-comment-only hunk would surface the same false positive and
can be handled if it ever appears.

## Tests

- an added block of only `//` comment lines: no hunk
- an added block mixing a comment line and a real statement: one hunk, **not**
  dropped
- an added block of only blank lines: no hunk

The mixed case is the boundary that matters. Dropping a mixed hunk would
silently stop checking real code, which would turn a noisy check into a blind
one.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally — ran in full in the clean-room gate Job for
      this branch (GATE-PASS)
- [x] `make lint` passes locally — the same gate Job runs fmt, vet, lint and
      lint-deadcode
      conversation.

### DIFF
--- a/pkg/foreman/agent/mutation_gate.go
+++ b/pkg/foreman/agent/mutation_gate.go
@@ -959,6 +959,22 @@ type productionHunk struct {
 //
 // The diff is taken against base (the fork point), not the plain working-tree
 // diff, so it is independent of whether an earlier gate staged the tree.
+//
+// A hunk is only a candidate when at least one of its lines is a plausible Go
+// statement: after strings.TrimSpace it is non-empty and does not start with
+// "//". A comment-only hunk is dropped here, in flush, not in the caller: the
+// per-hunk pass reverts each candidate and reports it uncovered when the
+// package's tests stay green, and reverting a comment can never turn a test
+// red, so an unfiltered comment-only hunk is reported as uncovered forever
+// (the #1735 false positive on a godoc that had to change). Filtering in flush
+// also keeps CheckPerHunkCoverage's maxPerHunkHunks budget honest: a hunk that
+// was never a candidate should not consume the budget.
+//
+// Gap accepted deliberately: a `/* */` block comment is not caught by a "//"
+// prefix test. Tracking block-comment state line by line is out of scope here;
+// the "//" case is the one actually being hit, and a block-comment-only hunk
+// is a far rarer shape that the same false positive would surface if it ever
+// matters.
 func splitProductionHunks(ctx context.Context, workspace, base, file string, run commandRunner) []productionHunk {
 	out, err := run(ctx, workspace, nil, "git", "diff", "-U0", base, "HEAD", "--", file)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


Defilan (commented):
Reviewed the diff. Correct, and it gets the subtle part right.

**The filter is in `flush`, which is the detail that matters.** A hunk dropped
there never becomes a `productionHunk`, so it also never consumes
`CheckPerHunkCoverage`'s `maxPerHunkHunks` budget. Filtering in the caller would
have produced the same warnings-suppressed behaviour while still letting
comment-only hunks eat the budget that bounds how many real candidates get
checked. Easy thing to get wrong; it is right here.

**`cur = nil` still runs when a hunk is dropped.** It sits inside
`if len(cur) > 0` but outside `if plausible`, so a dropped hunk's lines do not
leak into the next one. Worth stating explicitly because the obvious refactor
(moving the reset inside the `plausible` branch) would silently corrupt every
subsequent hunk's line range in the same file.

**The mixed-hunk test is the important one.** An added block containing a
comment line and a real statement must produce a hunk. Dropping it would turn a
noisy check into a blind one, which is strictly worse than the false positive
being fixed. Good that this is asserted rather than assumed.

**The accepted `/* */` gap is documented rather than hidden.** The godoc states
that block comments are not caught by a `//` prefix test and that tracking
block-comment state is out of scope. That is the right trade for the observed
failure, and saying so in the godoc means the next person hitting a
block-comment-only hunk will recognise it immediately instead of re-deriving it.

For context on the false positive this fixes: on
`wl-1729-depwait-default-verify-1729` the pass reported
`agentictask_controller.go:599-602` as an uncovered hunk. That range contains
zero non-comment lines. It is the godoc above `depWaitExpired`, which *had* to be
updated in that change because the previous text had been made false by it. The
same run's all-or-nothing bite check correctly reported "tests fail against
pre-change production (biting test, good)", so the real behavioural change was
properly constrained and only the comment hunk was flagged.

`CheckPerHunkCoverage`'s godoc says the pass is reporting-only "until the
false-positive rate on real diffs is known". On the first real diff observed,
the only finding it produced was this false positive, which is worth weighing
before anyone promotes the pass to a hard failure.

Issue #1749, opened 2026-09-02
[FEATURE] Bound the per-hunk mutation pass: run only the changed specs, honour a time budget, report partial coverage


Defilan:
## Feature Description

Bound the cost of the verify gate's per-hunk mutation pass (`cmd/hunkcheck`, #1694) so a controller-package diff of ordinary size can be checked inside one gate Job, and report partial coverage instead of failing when the budget is exhausted.

## Problem Statement

`cmd/hunkcheck/main.go` reverts each added production hunk in an envtest package and re-runs the whole package suite per hunk. `internal/controller` takes about five minutes per run, so the pass costs `hunks x 5 min`: `wl-1725-agw-pool-refs` (39 hunks) needed over three hours against a 30-minute Job. #1748 covers the deadline mapping and configurability; this issue is the pass itself, which no deadline setting makes affordable for a normal diff.

## Proposed Solution

In `cmd/hunkcheck/main.go` (and any helper it shares with `pkg/foreman/agent/mutation_gate.go`):

1. Run only the changed specs, not the package. Derive the test selection from the diff: `-run` for `Test*` functions added or modified in changed `_test.go` files, `-ginkgo.focus` for `It(...)` / `Describe(...)` strings added in changed Ginkgo files. A hunk then costs seconds, not an envtest cycle. Fall back to the whole package only when the diff adds no identifiable test.
2. Budget awareness: accept a `-budget` duration. Measure the first run, and when `remaining hunks x measured run` exceeds the budget, stop and print a `hunkcheck: partial (N of M hunks checked, budget <d>)` line that `run_gate_job` surfaces as a `gateAdvisories` entry rather than a failure.
3. Group hunks per function: reverting all hunks of one function at once is one run per function instead of one per hunk, and still answers the question the check asks (does a test bite when this change is gone).

## Additional Context

- Parent: #1748 (DeadlineExceeded reported as GATE-FAIL; deadline not configurable).
- Origin of the pass: #1694. The bite check keeps its current behavior.
- Evidence: gate Job `foreman-gate-wl-1725-agw-pool-refs-verify-1725-1788341791553`, killed at 1800 s at the `=== per-hunk mutation coverage ===` header after a passing bite check.
- Files: `cmd/hunkcheck/main.go`, `pkg/foreman/agent/tools/run_gate_job.go` (the rendered `hunkCheck` invocation and the advisory surfacing).

## Priority

- [x] High - Blocking my use case

## Willingness to Contribute

[...]

Issue #1748, opened 2026-09-02
[BUG] Verify gate: per-hunk mutation pass on envtest packages cannot fit the fixed 1800 s Job deadline, and DeadlineExceeded is reported as GATE-FAIL


Defilan:
## Bug Description

The verify gate's per-hunk mutation pass (`hunkCheck`, #1694) reverts every added production hunk in an envtest package and re-runs that package's suite once per hunk. `internal/controller` takes about five minutes per run, so a routine controller change with a few dozen hunks needs hours, while the gate Job's `activeDeadlineSeconds` is fixed at 1800 and cannot be raised without a code change. The Job is killed mid-pass, `run_gate_job` maps the kill to GATE-FAIL because it only reads `Job.Status.Failed`, and the review cascade-fails. A branch whose bite check passed ends up reported as a failing gate.

Live case: `wl-1725-agw-pool-refs` (2026-09-02). Coder GO in 93 min on branch `foreman/wl-1725-agw-pool-refs/issue-1725` (912dcdc). The verify Job `foreman-gate-wl-1725-agw-pool-refs-verify-1725-1788341791553` ran clone, helm, chart tests, non-go lint and the bite check (which passed: the new test fails to compile against reverted production) and then entered `=== per-hunk mutation coverage (base=9786684f6) ===`. It was killed at 1800 s with `Failed / DeadlineExceeded: Job was active longer than specified deadline`. The diff adds 39 production hunks in `internal/controller` (23 in `modelrouter_agentgateway_controller.go`, 16 in `gateway_resources_agentgateway.go`), so the pass needed roughly 39 x 5 min.

## Steps to Reproduce

1. Push a branch that changes two or more functions across `internal/controller/*.go` with a changed `_test.go` (any diff with more than about six added hunks).
2. Run the verifier `gate` Agent (`run_gate_job` with `hunkCheck: true`) on it.
3. Observe the Job log stop at the `per-hunk mutation coverage` header, the Job condition `DeadlineExceeded`, the task verdict `GATE-FAIL`, and the review task `UpstreamFailed`.

## Expected Behavior

Scope of THIS issue (the pass cost itself is #1749):

[...]

PR #1751, opened 2026-09-02
fix(foreman): report gate Job DeadlineExceeded as GATE-ERROR and make the gate deadline settable (#1748)

Defilan:
## What

Two changes to the verify gate's `run_gate_job` tool. A gate Job killed by its `activeDeadlineSeconds` now maps to GATE-ERROR instead of GATE-FAIL, with a summary that names the deadline and the phase the Job was in (`gate Job exceeded its 1800s deadline during per-hunk mutation coverage (base=...)`). And the deadline is settable: the verifying Agent's `spec.execution.activeDeadlineSeconds` is threaded through `ToolDeps.GateActiveDeadlineSeconds` into `RunGateJobToolConfig.ActiveDeadlineSeconds`, with unset keeping the 1800 s default.

## Why

Refs #1748

`wl-1725-agw-pool-refs` passed its bite check, entered the per-hunk mutation pass on `internal/controller` (39 hunks at about five minutes of envtest each), and was killed at 1800 s. `pollForTerminal` read `Job.Status.Failed >= 1` and reported GATE-FAIL, so a branch whose checks were passing was recorded as a failing gate and its review cascade-failed. Nothing could raise the deadline without a code change. A timeout is a gate infrastructure problem, not a failing test, and the verdict decides whether the Workload treats the branch as bad.

## How

`gateJobDeadlineExceeded` walks the Job's conditions for `JobFailed` with reason `DeadlineExceeded`, the shape the Kubernetes Job controller sets on a deadline kill; every other failure reason keeps GATE-FAIL. `pollForTerminal` returns a `deadlineHit` flag, and `Execute` appends the last `=== <phase> ===` banner found in the log tail (`lastGatePhase`) so the operator can see whether the standard checks survived and where the Job stalled. Defaults are applied before polling, so the summary always names the effective deadline.

The deadline wiring mirrors what `run_coder_job` already does for the coder Job: `cmd/foreman-agent/main.go` reads `agent.Spec.Execution.ActiveDeadlineSeconds` (`*int64`), clamps it to `int32` via `gateActiveDeadlineSeconds` (nil, zero and negative keep the default; values above `MaxInt32` saturate), and passes it through `ToolDeps`. `PollTimeout` keeps deriving as twice the deadline.

Tests: `TestRunGateJob_DeadlineExceededProducesGATEERROR` drives the fake apiserver through a `DeadlineExceeded` terminal state and asserts GATE-ERROR plus the phase suffix; `TestRunGateJob_FailedProducesGATEFAIL` still holds for a plain failure; `TestBuildAll_ThreadsGateActiveDeadlineSeconds` and `TestGateActiveDeadlineSeconds` cover the wiring and the clamp.

Not in this PR: bounding the mutation pass itself, which is #1749. To use the new knob on a fleet, roll the agent image and set `spec.execution.activeDeadlineSeconds` on the verifier Agent.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (verify gate GATE-PASS on the branch; CI green)
- [x] `make lint` passes locally (CI)
- [x] All commits are signed off (`git commit -s`) per [DCO](https://developercertificate.org/) (Foreman-bot sign-off under my address; squash-merge attributes the commit to me, as with #1741)
- [x] Documentation updated (if user-facing change): not user-facing; the Agent field already exists and is documented

Assisted-by: Foreman (coder `dsv4-flash-coder`, DeepSeek-V4-Flash-Vision-Exp on vLLM) generated the code and tests from the decided design in #1748; the verify gate ran the full check suite. I reviewed the diff by hand, ran the two packages' tests, vet and gofmt on the branch, and own the sign-off.

### DIFF
--- a/cmd/foreman-agent/main.go
+++ b/cmd/foreman-agent/main.go
@@ -801,6 +801,21 @@ func clampInt32(n int) int32 {
 	return int32(n) //nolint:gosec // bounded above
 }
 
+// gateActiveDeadlineSeconds converts the Agent's
+// spec.execution.activeDeadlineSeconds (*int64) into the int32 the gate
+// Job's RunGateJobToolConfig.ActiveDeadlineSeconds field carries. Zero or
+// unset keeps the 1800 s default via applyConfigDefaults; a value above
+// math.MaxInt32 saturates so the int32 field never wraps negative (#1748).
+func gateActiveDeadlineSeconds(n *int64) int32 {
+	if n == nil || *n <= 0 {
+		return 0
+	}
+	if *n > math.MaxInt32 {
+		return math.MaxInt32
+	}
+	return int32(*n) //nolint:gosec // bounded above
+}
+
 // assembleAgentRegistry builds an agent's tool registry: native tools
 // filtered by the agent's spec.tools whitelist, then MCP tools appended.
 // MCP tools bypass the whitelist by design (they are already gated by their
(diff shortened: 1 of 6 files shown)


Defilan:
Human review of the Foreman branch. Approve, pending finalize.

Read the whole diff against the decided design in #1748 and ran `go test`, `go vet` and `gofmt -l` for `pkg/foreman/agent/tools` and `cmd/foreman-agent` on the branch: green. Points checked:

- `gateJobDeadlineExceeded` keys on `JobFailed` + `DeadlineExceeded`, the exact condition the wl-1725 Job carried; any other failure reason still returns GATE-FAIL, and `TestRunGateJob_FailedProducesGATEFAIL` proves it.
- The phase suffix comes from the last `=== ... ===` banner in the tail. On the wl-1725 tail that is `per-hunk mutation coverage (base=9786684f6...)`, which is what an operator needs to see.
- `Execute` applies `applyConfigDefaults` before polling, so the summary names the effective deadline (1800 when unset), not a zero.
- The `*int64` to `int32` clamp mirrors `clampInt32` next to it and the table test covers nil, zero, negative, exact max and overflow.
- `PollTimeout` is untouched and keeps deriving as twice the deadline, so a raised deadline cannot outlive the poller.

Two notes, neither blocking:

1. Downstream, a GATE-ERROR still cascades the review to INCOMPLETE and the Workload ends Failed. What changes is the failure reason (`FailureGateError`, infrastructure) instead of a bad-branch signal, and the operator now has the knob to re-run with more time. That is what the issue asked for; the cheaper pass itself is #1749.
2. The in-cluster reviewer's approval listed `api/foreman/v1alpha1/agent_types.go` in `filesTouchedClaimed`; that file is not in the diff. The claim is harmless here, but it is the kind of drift the scope rail should catch, and it did not (`scopeDriftDetected: false`).

To put it to use on the lab fleet after merge: roll the foreman-agent image, then give the verifier `gate` Agent an `execution: {activeDeadlineSeconds: 7200}` block (it has none today).

Finalization: the commit is bot-signed. `scripts/foreman-finalize.sh --branch foreman/wl-1748-gate-deadline/issue-1748` squashes to one human-signed commit.

Issue #1186, opened 2026-07-21
[FEATURE] AMD observability slice 2/3: llama.cpp /metrics -> Prometheus (tokens/sec, KV, requests)


Defilan:
## Feature Description

Surface the backend-agnostic inference signals llama.cpp already emits on its `/metrics` endpoint (enabled via `--metrics`) into Prometheus for the AMD tier: tokens/sec, KV-cache occupancy, and in-flight request count. These are the SLO-relevant signals and are the same on AMD as on any llama.cpp backend.

Slice 2 of 3 for #700.

## Proposed Solution

- Ensure `--metrics` is set on the AMD/Vulkan `InferenceService` serving config.
- A `ServiceMonitor` scraping the served pod's `/metrics` (`release: prometheus` label).
- Document the pinned llama.cpp metric names used by the dashboard slice.
## Shared metric contract (pinned across all three slices — do not drift)

This is the cross-slice interface. The dashboard slice must query **only** the
names below; the emitter slices must emit **exactly** these. If a metric is not
available on `gfx1151`, the exporter slice drops it from the contract doc and the
dashboard slice must not query it. Pinning this prevents the failure the slicer
experiment on #700 already exposed: a green build hiding a dashboard that queries
a metric no exporter emits.

**Inference signals — llama.cpp `/metrics` (LLMKube already emits these, pinned):**
- `llamacpp:tokens_per_second`
- `llamacpp:kv_cache_usage_ratio`
- `llamacpp:requests_processing`

**AMD GPU signals — rocm-smi exporter (the exporter slice OWNS the final list):**
The exporter slice hands-verifies what actually reports on `gfx1151` and records
the authoritative names in `docs/observability/amd-gpu-metrics.md`. Expected
starting set (community `rocm-smi` exporter convention):
- `rocm_smi_sensor_temperature` — edge temperature, °C
- socket power (W) — exact metric name per the exporter, documented by the exporter slice
- GPU busy % and VRAM used/total **only if the iGPU reports them** (document honestly; omit if absent)

The dashboard slice queries names from that doc verbatim — it does not invent metric names.

## Dependencies / Notes

- Parent: #700. Independent of the rocm-smi exporter slice (different metric namespace).
- The llama.cpp names above are already emitted by LLMKube today — verified in-repo — so this slice does not invent them.

## Priority
- [x] Medium

## Willingness to Contribute
- [x] Yes, I can submit a PR

joryirving:
[...]

Issue #1187, opened 2026-07-21
[FEATURE] AMD observability slice 3/3: Grafana AMD panel (consumes pinned metric contract)


Defilan:
## Feature Description

Add an AMD tier Grafana dashboard for the Strix (gfx1151) node, mirroring the NVIDIA GPU dashboard layout, so the AMD node's GPU health, memory, and inference SLO signals appear alongside the NVIDIA fleet.

Slice 3 of 3 for #700. This slice **consumes** an already-verified metric contract — every metric below is confirmed emitted and scraped by Prometheus on the live box (target health=up). **Use only these names; do not invent metric names.**

## VERIFIED metric contract (hands-checked on gfx1151, 2026-07-21)

**GPU health — already emitted by node-exporter's hwmon collector.** Select the amdgpu chip by joining on `node_hwmon_chip_names`:
- Temperature (edge): `node_hwmon_temp_celsius * on(chip) group_left(chip_name) node_hwmon_chip_names{chip_name="amdgpu"}`
- Power (PPT, W): `node_hwmon_power_watt * on(chip) group_left(chip_name) node_hwmon_chip_names{chip_name="amdgpu"}`

**GPU util / memory / clock — emitted by the amdgpu-sysfs exporter (already deployed):**
- `amdgpu_gpu_busy_percent{card="card0"}` — utilization %
- `amdgpu_gtt_used_bytes{card="card0"}` — GTT used (the primary memory metric on this box; the model lives in GTT, ~45GB)
- `amdgpu_vram_used_bytes{card="card0"}` / `amdgpu_vram_total_bytes{card="card0"}` — VRAM (small ~2GB carveout)
- `amdgpu_sclk_mhz{card="card0"}` — GPU clock (MHz)

**Inference — already emitted by llama.cpp `/metrics`:**
- `llamacpp:tokens_per_second`
- `llamacpp:kv_cache_usage_ratio`
- `llamacpp:requests_processing`

## Proposed Solution

- Add a new dashboard `docs/grafana/amd-gpu-observability.json` (a standalone dashboard; do NOT edit the large existing `config/grafana/llmkube-gpu-dashboard.json`, to avoid churn). Mirror its panel style.
- Panels: GPU temperature, GPU power, GPU busy %, memory (GTT used as the headline, VRAM used/total secondary), GPU clock, and an inference row (tokens/sec, KV-cache occupancy, in-flight requests).
[...]

Issue #700, opened 2026-06-15
[FEATURE] AMD GPU observability: rocm-smi exporter + llama.cpp /metrics + Grafana panel


Defilan:
## Feature Description

AMD GPU observability for the fleet: node-level GPU temperature/power/utilization from the AMD node surfaced into Prometheus + Grafana, the AMD analog of the DCGM exporter used for NVIDIA. For the iGPU the realistic path is a community `rocm-smi`-based exporter plus the inference signals already emitted by llama.cpp `/metrics`.

## Problem Statement

> As a fleet operator, I want the AMD node's GPU health and the inference SLO signals in the same Grafana I use for NVIDIA, so that an AMD tier is observable, not a blind spot.

AMD's official `device-metrics-exporter` is Instinct/datacenter-scoped and does not enumerate the Strix Halo iGPU, so the NVIDIA DCGM approach does not port directly.

## Proposed Solution

- Deploy a community `rocm-smi`-based exporter (e.g. `rocm-smi-exporter`) as a DaemonSet on AMD nodes for edge temperature and socket power; document the limited metric set honestly.
- Rely on llama.cpp `/metrics` (already enabled via `--metrics`) for the SLO-relevant signals: tokens/sec, queue depth, KV-cache occupancy. This is the primary inference signal and is backend-agnostic.
- A Grafana panel/row for the AMD tier, mirroring the GPU dashboard layout.
- Note the gap: official `device-metrics-exporter` is the path if/when Instinct/MI hardware is added; it does not help the APU.

## Alternatives Considered

- **`amd/amd_smi_exporter`:** also datacenter/Instinct-oriented; unlikely to enumerate the iGPU cleanly.
- **Official `ROCm/device-metrics-exporter`:** Instinct-scoped; deferred to a future discrete-AMD tier.

## Additional Context

- Related issues: #696 (epic)
- The iGPU support of community exporters and the exact metric set should be hands-verified on `gfx1151`; scope the issue to "what actually reports numbers on this box."

## Priority

- [x] Medium - Nice to have

## Willingness to Contribute

- [x] Yes, I can submit a PR

Defilan:
Triage 2026-06-27: despite earlier tracking, **nothing has landed** for this issue's three deliverables (rocm-smi exporter DaemonSet, AMD llama.cpp /metrics wiring, AMD Grafana panel). The AMD PRs that merged (#709/#727/#735) were scheduling/runtime prerequisites, not observability. Confirmed still fully open.

[...]

PR #1754, opened 2026-09-03
feat(observability): pin the llama.cpp metric contract (#1186)

joryirving:
## What

The contract half of the inference tier:

- `docs/observability/llamacpp-metrics.md` — the pinned metric contract, in the same shape as `amd-gpu-metrics.md`, with `metrics-contract` markers so the suite reads it.
- `internal/metrics/testdata/llamacpp-metrics.txt` — the five emitted names the fixture was missing.

The PodMonitor is gone; see below.

## Why

Refs #1186 — deliberately non-closing, since #1187 still has to read the renegotiated contract off that issue.

Slice 2 of #700. Two of the three metric names pinned in #1186 do not exist in llama.cpp's `/metrics`, so #1187 would build panel queries against metrics that never arrive:

| pinned in #1186 | reality |
| --- | --- |
| `llamacpp:requests_processing` | exists |
| `llamacpp:tokens_per_second` | no such metric |
| `llamacpp:kv_cache_usage_ratio` | no such metric — nothing matching `kv_cache` is exposed |

## How

**Contract.** Throughput is pinned as two metrics rather than one — `llamacpp:predicted_tokens_seconds` (decode) and `llamacpp:prompt_tokens_seconds` (prefill). They bottleneck on different resources and degrade independently, so a combined figure hides one collapsing while the other holds. I watched a server fall from 56 to 5 tokens/s of decode with prefill unchanged; an averaged number would have read as a mild dip rather than a 10x regression.

`llamacpp:requests_deferred` is added as the queueing signal. It rises only once every parallel slot is occupied, which is what separates "busy" from "oversubscribed".

KV-cache occupancy is documented as a known gap rather than substituted, in a section deliberately left outside the contract markers so it can keep naming metrics that do not exist.

**The markers were the real defect.** The doc had none, so `TestDocContractMetrics` walked it and skipped it — a fake metric name in the throughput table left the suite green. With the markers added it correctly flagged five names missing from the fixture, so those are added rather than removed: they are real, just newer. `prompt_tokens_cached_total` landed in `decaf508b` (2026-08-13) and the four `spec_decode_*` counters in `a035a8887` (2026-08-05), so the doc gains a minimum-build note — a server older than mid-August 2026 emits neither group and a panel keyed on them stays blank.

**Splitting the spec_decode caveat.** The doc claimed all four `spec_decode_*` counters appear "only when a draft model is configured". That is right for one of them and wrong for three. A build `b10795` server here started without `--spec-type` emits `spec_decode_num_drafts_total`, `spec_decode_num_draft_tokens_total` and `spec_decode_num_accepted_tokens_total` at zero, and omits only `spec_decode_num_accepted_tokens_per_pos_total`. They are now two rows, matching that split.

**No PodMonitor.** The chart already ships an identical one at `charts/llmkube/templates/inference-podmonitor.yaml`, on by default — same `inference.llmkube.dev/service` Exists selector, same `http` port and `/metrics` path, same 30s interval, same four relabelings, `namespaceSelector.any: true`, and `{{ include "llmkube.fullname" . }}-inference` renders to the same object name. My copy's explicit `action: replace` is the implicit default, so it was byte-equivalent. Any Prometheus with an empty `podMonitorSelector` would adopt both and scrape every llama.cpp pod twice into the same series. `TestInferenceDashboardCoversEveryRuntime` already names the chart template as the thing that relabels `runtime`, so it was the canonical one all along.

**Dashboard pointer.** The header said a rename must update `charts/llmkube/dashboards/amd-gpu-observability.json`, which has zero `llamacpp:` queries. It now points at `llamacpp-dashboard.json`, which has fourteen.

## Verification

- `go test ./internal/metrics/` green.
- Negative control: injecting `llamacpp:totally_fake_metric` into the contract region fails `TestDocContractMetrics` on that name, and removing it goes green — the markers bite.
- The 16 identifiers inside the contract region are exactly the 16 names a live b10795 server emits on `/metrics`.
- `kubectl kustomize config/monitoring` still builds.

Assisted-by: Claude Code (drafted the contract doc and this description; I verified every metric name, type and description against my own running servers and ran the metrics suite).

## Checklist

- [x] Tests added/updated — the `metrics-contract` markers put this doc under `TestDocContractMetrics`, which never read it before, plus five fixture entries.
- [x] `make lint` passes locally — no Go changes

### DIFF
--- a/docs/observability/amd-gpu-metrics.md
+++ b/docs/observability/amd-gpu-metrics.md
@@ -51,7 +51,9 @@ versions of the same readings.
 <!-- metrics-contract:begin -->
 SLO metrics come from llama.cpp `/metrics` (`--metrics`), not this exporter:
 `llamacpp:predicted_tokens_seconds`, `llamacpp:prompt_tokens_seconds`,
-`llamacpp:requests_processing`. Backend-agnostic and already emitted.
+`llamacpp:requests_processing`, `llamacpp:requests_deferred`. Backend-agnostic
+and already emitted. Pinned in `llamacpp-metrics.md`; scraped by the chart's
+`inference-podmonitor.yaml`.
 <!-- metrics-contract:end -->
 
 ## Known gaps (deliberate)
--- a/docs/observability/llamacpp-metrics.md
+++ b/docs/observability/llamacpp-metrics.md
@@ -1,0 +1,90 @@
+# llama.cpp metric contract
+
+The pinned metric names for the inference tier (issue #700, slice #1186).
+Dashboards and alerts key off these names; changing them is a breaking change
+to this contract and must update `charts/llmkube/dashboards/llamacpp-dashboard.json`
+in the same PR, which is where every `llamacpp:` query lives.
+
+Backend-agnostic: these come from llama.cpp itself, so they are identical on
+CUDA, ROCm and Vulkan servers. The AMD-specific device metrics are a separate
+contract — see `amd-gpu-metrics.md`.
+
+## Source: llama.cpp `/metrics`
+
+Requires the server to run with `--metrics`. Scraped by the chart's
+`inference-podmonitor.yaml` (`prometheus.inferencePodMonitor.enabled`, on by
+default), which selects on the operator's `inference.llmkube.dev/service` label
+and relabels every series with `service`, `model`, `runtime` and `namespace`, so
+a panel can break down per model without naming a target.
+
+<!-- metrics-contract:begin -->
+
+### Throughput
+
+| Metric | Type | Unit | Notes |
+| --- | --- | --- | --- |
+| `llamacpp:predicted_tokens_seconds` | gauge | tokens/s | **decode** throughput — generation |
+| `llamacpp:prompt_tokens_seconds` | gauge | tokens/s | **prefill** throughput — prompt processing |
+
+Keep these separate rather than presenting one "tokens per second". Prefill is
+compute-bound and decode is memory-bandwidth-bound, so they degrade
+independently, and a combined figure hides the case where one collapses while
+the other holds. Observed in practice: a server whose decode fell from 56 to
+5 tokens/s while prefill stayed at its usual rate — a single averaged number
+would have read as a mild dip rather than a 10× regression.
+
+### Saturation
+
+| Metric | Type | Unit | Notes |
+| --- | --- | --- | --- |
+| `llamacpp:requests_processing` | gauge | count | requests in flight |
+| `llamacpp:requests_deferred` | gauge | count | requests queued behind a full slot set — **the queueing signal** |
+| `llamacpp:n_busy_slots_per_decode` | gauge | count | average busy slots per `llama_decode()` call |
+
+`llamacpp:requests_deferred` is what distinguishes "the server is busy" from "the server
+is oversubscribed": it goes above zero only once every parallel slot is
+occupied. Alert on it rather than inferring queueing from latency.
+
+### Totals
+
+| Metric | Type | Notes |
+| --- | --- | --- |
+| `llamacpp:n_decode_total` | counter | decode calls |
+| `llamacpp:tokens_predicted_total` / `llamacpp:tokens_predicted_seconds_total` | counter | generated tokens and time |
+| `llamacpp:prompt_tokens_total` / `llamacpp:prompt_seconds_total` | counter | prompt tokens and time |
+| `llamacpp:prompt_tokens_cached_total` | counter | prompt tokens served from cache — prefix-cache effectiveness |
+| `llamacpp:n_tokens_max` | counter | largest observed sequence length (prompt + generation) |
+| `llamacpp:spec_decode_num_drafts_total`, `llamacpp:spec_decode_num_draft_tokens_total`, `llamacpp:spec_decode_num_accepted_tokens_total` | counter | speculative decoding; emitted unconditionally, at zero, on a server with no draft model |
+| `llamacpp:spec_decode_num_accepted_tokens_per_pos_total` | counter | speculative decoding, labelled by position; the one spec_decode counter that appears only when a draft model is configured |
+
+<!-- metrics-contract:end -->
+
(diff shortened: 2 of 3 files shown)


Defilan (commented):
Holding this one, and it's the manifest half rather than the contract. The chart already ships this exact PodMonitor.

`charts/llmkube/templates/inference-podmonitor.yaml` renders `llmkube-inference` with the same `inference.llmkube.dev/service` Exists selector, the same `http` port and `/metrics` path, the same 30s interval and the same four service/namespace/model/runtime relabelings. I normalised both specs to sorted JSON and diffed them: the live object in `llmkube-system`, created 2026-04-07, comes back identical field for field to what `kustomize build config/monitoring` produces on your branch. It's on here, `helm get values llmkube` shows `inferencePodMonitor.enabled: true`, and it's the only inference scrape pool up. Your render lands at `monitoring/llmkube-inference`, and this Prometheus has `podMonitorSelector: {}` and `podMonitorNamespaceSelector: {}`, so it gets adopted alongside the chart's and every llama.cpp pod is scraped twice into the same series. The chart toggle is also where the docs already send people. That's my cluster and not yours, so say if you needed a second copy for a reason I'm not seeing.

Second, the new doc is the one contract doc `TestDocContractMetrics` never reads, because it has no `metrics-contract` markers. I dropped a fake metric name into the throughput table and the test stayed green. Adding the markers does make it fail on that name, so that's the fix, but it then also flags `prompt_tokens_cached_total` and the four `spec_decode_*` names, which aren't in `internal/metrics/testdata/llamacpp-metrics.txt`.

Which leads to the third thing, and it's a caveat rather than a correction. Those five names are real on master but newer than anything running here. I read `/metrics` off two servers, a Vulkan build and stock `ggml-org/llama.cpp:server` at b10200 (commit 5f55650a7, 2026-07-30), and neither emits them. Upstream added the spec_decode counters in a035a8887 on 2026-08-05 and `prompt_tokens_cached_total` in decaf508b on 2026-08-13. So those rows want a minimum-build note plus fixture entries rather than removal. While you're in there, three of the four spec_decode counters are unconditional in `server-task.cpp`, not draft-model gated, so that caveat only holds for `_num_accepted_tokens_per_pos_total`.

Smaller: the header says a rename must update `charts/llmkube/dashboards/amd-gpu-observability.json`, but that dashboard has zero `llamacpp:` queries. `llamacpp-dashboard.json` has sixteen and is the actual consumer.

The metric work is right and it's the valuable half of this. Neither server emits `tokens_per_second` or anything matching `kv_cache`, so #1186's pins really are wrong and renegotiating them is the correct call. Splitting throughput into decode and prefill is the right shape too. One note: `Fixes #1186` will auto-close the issue whose pinned list you're deliberately rewriting, and #1187 still has to read the new contract off it.

Drop the manifest, add the markers and the build note, fix the dashboard pointer, and this lands. Thanks for the writeup, it made this quick to check.

joryirving:
All five, and that was a genuinely useful read. Thanks.

The PodMonitor is gone. You're right that it's byte-equivalent: the chart's template renders the same object name, the same Exists selector on `inference.llmkube.dev/service`, the same `http` port and `/metrics` path, the same interval and the same four relabelings, and my explicit `action: replace` is just the implicit default written out. `TestInferenceDashboardCoversEveryRuntime` already names `charts/llmkube/templates/inference-podmonitor.yaml` as the thing that relabels `runtime`, so it was the canonical one and I had added a second copy of it. I didn't need two, and on your cluster I'd have doubled every series.

Markers added, and your fake-name check reproduces exactly: without them the doc is walked and skipped, with them a fake name in the throughput table fails on that name. It then flagged the five you predicted, so those went into `internal/metrics/testdata/llamacpp-metrics.txt` rather than out of the doc, with a minimum-build note citing `decaf508b` and `a035a8887`.

On the spec_decode gating you're right that my caveat is wrong, and it's wrong further than you said. It doesn't hold for any of the four. I have a `b10795` server here with no `speculativeDecoding` configured and it emits all four, including `spec_decode_num_accepted_tokens_per_pos_total` with `position="0"` and `position="1"`. So the row now says all four are emitted unconditionally, at zero, rather than three of them.

Dashboard pointer fixed to `llamacpp-dashboard.json`. I count fourteen `llamacpp:` queries there rather than sixteen, and zero in `amd-gpu-observability.json` either way, so the pointer was the problem regardless.

`Fixes` is now `Refs`, so #1186 stays open for #1187 to read the renegotiated list off it.

One thing worth recording given your two servers are older: the 16 identifiers inside the contract region are now exactly the 16 names my b10795 server emits on `/metrics`, so that fixture is a complete list for that build rather than a subset of it.

Defilan (approved):
Approving. I re-checked all five on the new head rather than trusting the summary, and they hold.

The manifest is gone and the kustomization no longer references it. The markers work: I planted `llamacpp:totally_made_up_metric` in the throughput table and the guard failed on that exact name, where before it walked past the file entirely. The fixture's sixteen names and the sixteen inside the contract region match each other. The build note carries both commits. Pointer is on `llamacpp-dashboard.json`, and `Refs` keeps #1186 open for #1187.

One thing to reconcile, and it's between your comment and the doc rather than in the code. Up here you wrote that the caveat doesn't hold for any of the four, and that your b10795 server emits `spec_decode_num_accepted_tokens_per_pos_total` at positions 0 and 1 with no draft model. The doc says the opposite in two places: the row calls it "the one spec_decode counter that appears only when a draft model is configured", and the Verification section says that same server "emits the first three counters at zero and omits `..._per_pos_total` entirely". The doc agrees with what I read in `server-task.cpp`, where that one sits behind an `if (!metrics.n_accepted_per_pos.empty())` guard and the other three don't, so I suspect the doc is right and the comment overshot. Worth correcting the comment, since it also means your b10795 server emits fifteen of the sixteen rather than all of them.

Also, my sixteen was a line count, not a query count. That file has eleven distinct `llamacpp:` names across eleven expressions. Yours is closer than mine and the pointer was wrong either way.

Nice work on the turnaround, and thanks for chasing the gating down further than I did.

Issue #1750, opened 2026-09-02
[FEATURE] Hugging Face token for operator-driven downloads: gated and private repos via Model.spec.sourceSecretRef


Defilan:
## Feature Description

Let a Model carry a Hugging Face access token so the operator's own download paths can fetch gated and private repositories. Today a token only reaches the serving container of four runtimes; every path where LLMKube itself downloads weights sends unauthenticated requests.

## Problem Statement

What exists: `vllmConfig.hfTokenSecretRef`, `tgiConfig.hfTokenSecretRef`, `sglangConfig.hfTokenSecretRef` and `personaPlexConfig.hfTokenSecretRef` inject `HF_TOKEN` (TGI: `HUGGING_FACE_HUB_TOKEN`) into the runtime container (`internal/controller/runtime_vllm.go` `BuildEnv`, `runtime_tgi.go`, `runtime_personaplex.go`). That covers only the runtime-resolved flow: a bare repo ID source with `skipModelInit: true`, where the runtime downloads at pod start.

What does not exist: a token on the paths LLMKube drives.

- The `model-downloader` init container (`internal/controller/model_storage.go`, the script builders around lines 300-360 and the multi-file staging around 540-590) runs `curl -f -L` and `curl -fsSL -I` with no `Authorization` header for `https://huggingface.co/...` and `hf://` sources (`resolveHFSourceURL`, `normalize_hf_source`). Only `s3://` gets credentials (`--aws-sigv4`).
- `Model.spec.sourceSecretRef` is already wired as `envFrom` into that init container (`modelEnvFrom`, `model_storage.go:384`), but its documented keys are `AWS_*` only; an `HF_TOKEN` key lands in the environment and nothing reads it.
- The prefetch Job (`internal/controller/model_prefetch.go`) reuses the same script and secret, so it has the same gap.
- The macOS metal agent's `downloadFile` (`pkg/agent/executor.go:396`) is a bare `http.Get`; only its S3 path signs requests.
- `docs/contributors/model-hf-source-spec.md` lists "Auth for HF_TOKEN in the Model controller" as out of scope, which is why the gap is deliberate rather than accidental.

Effect: any gated checkpoint (Llama, Gemma, Mistral and most vendor repos) or private repo fails with 401/403 for every GGUF/llama.cpp Model, for every `spec.files` multi-file safetensors staging, for prefetch, and on the Metal fleet. The only workaround is a runtime that downloads for itself.

## Proposed Solution

[...]

PR #1758, opened 2026-09-04
feat(model): send a Hugging Face token on operator-driven downloads (#1750)

Defilan:
## What

Lets the operator's own downloads authenticate to Hugging Face, so gated and private repositories work on every path LLMKube downloads on rather than only where the runtime downloads for itself.

## Why

Refs #1750

A token reached only the serving container of four runtimes, which covers one flow: a bare repo ID with `skipModelInit: true`, where vLLM (or TGI, SGLang, PersonaPlex) fetches its own weights. Every path the operator drives sent anonymous requests, so any gated checkpoint returned 401 for:

- single-file GGUF Models,
- multi-file `spec.files` staging,
- `RefreshPolicy: OnChange` revalidation,
- the prefetch Job,
- the macOS Metal agent.

The only workaround was picking a runtime that downloads for itself, which is not available for llama.cpp or for staged safetensors.

## How

**One secret, not two.** `spec.sourceSecretRef` is already `envFrom`'d into every downloader for the `AWS_*` keys, so an `HF_TOKEN` key beside them needs no new plumbing. The field's GoDoc now documents both key sets and the CRD description is regenerated.

**Gated on the host, not on the Secret.** Only `hf://` and `huggingface.co` sources emit the header, decided in Go by `isHFAuthSource`. A Model pointing at a mirror or a private registry gets nothing even when the Secret it names carries the key, so one Secret can serve an S3 Model and an HF Model without leaking either credential to the other's host.

**A shell wrapper rather than an interpolated flag.** The init container calls `hf_curl`, which adds the header when `HF_TOKEN` is non-empty and falls back to plain `curl` when it is not, so public repositories keep working with no Secret at all. `${HF_TOKEN:+-H "Authorization: Bearer $HF_TOKEN"}` cannot be used: the value contains a space, so it word-splits into four arguments and sends a malformed header, and quoting it sends one empty argument when the token is unset.

**One deviation from the issue's plan, and it is the interesting part.** The issue expected the Metal agent to inherit safe redirect behaviour from `net/http`. It does not. Go strips `Authorization` only when a redirect leaves the registrable domain, so it keeps the header across a subdomain, and across a different port on the same host. A test with two servers reproduced the leak. `downloadFile` now uses a client that drops the header on any change of host, which is stricter and matches what `huggingface_hub` does. curl already behaves this way, which is why only the Go path needed the fix, and why `-L` stays and `--location-trusted` is deliberately absent.

## Testing

`internal/controller`: the header is present for an HF source across both refresh policies and both cache modes, absent for another host, absent for `s3://` (which keeps `--aws-sigv4`), and the host gate is checked against lookalike hosts such as `huggingface.co.evil.example`. The prefetch tests assert the Job inherits both the `envFrom` and the header, since it reuses the serving path's builders rather than its own.

`pkg/agent`: the token reaches the first hop, no header at all is sent without a token, and the cross-host redirect test asserts the origin sees the token and the redirect target does not.

Each new test fails under mutation. Worth noting that the first version of the controller tests did **not**: a mutant that made `curlCmd` always return `curl` left a script that defined the wrapper and then called plain curl, downloading unauthenticated while still containing every string the assertions looked for. The assertions now check the wrapper is invoked, not merely defined.

## Checklist

- [x] Tests added/updated
- [x] `make test` passes locally (`./internal/controller`, `./pkg/agent`, `./api/...`)
- [x] `make lint` passes locally (`GOOS=linux golangci-lint run ./...`, 0 issues)
- [x] AI assistance disclosed below, per [CONTRIBUTING.md](../CONTRIBUTING.md#ai-assisted-and-agent-contributions)
- [x] Documentation updated

Also ran `make manifests generate chart-crds` for the CRD description and `make check-helm-rbac` (no RBAC change, 284 rules across 2 operators).

Assisted-by: Claude Code implemented this from the issue's plan and ran the checks, including the redirect test that found the `net/http` assumption to be wrong. Human review and the merge are mine.

### DIFF
--- a/api/v1alpha1/model_types.go
+++ b/api/v1alpha1/model_types.go
@@ -65,10 +65,21 @@ type ModelSpec struct {
 	SHA256 string `json:"sha256,omitempty"`
 
 	// SourceSecretRef names a Secret (in the Model's namespace) whose keys are
-	// wired as env into the model-downloader init container. Used by s3://
-	// sources for S3-compatible credentials/endpoint: AWS_ACCESS_KEY_ID,
-	// AWS_SECRET_ACCESS_KEY, AWS_REGION, AWS_ENDPOINT_URL (path-style, e.g.
-	// https://minio.internal:9000 or https://s3.us-east-1.amazonaws.com).
+	// wired as env into the model-downloader init container.
+	//
+	// s3:// sources read S3-compatible credentials and endpoint from
+	// AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_REGION and AWS_ENDPOINT_URL
+	// (path-style, e.g. https://minio.internal:9000 or
+	// https://s3.us-east-1.amazonaws.com).
+	//
+	// hf:// and huggingface.co sources read HF_TOKEN, which is sent as a bearer
+	// credential so gated and private repositories can be downloaded. The token
+	// is attached only for huggingface.co and is dropped if a redirect leaves
+	// that host, so it never reaches a CDN or any other origin. Omit the key,
+	// or the whole Secret, for public repositories.
+	//
+	// One Secret may carry both sets of keys; each is used only by the source
+	// scheme it belongs to.
 	// +optional
 	SourceSecretRef *corev1.LocalObjectReference `json:"sourceSecretRef,omitempty"`
 
--- a/charts/llmkube/templates/crds/models.yaml
+++ b/charts/llmkube/templates/crds/models.yaml
@@ -465,10 +465,21 @@ spec:
               sourceSecretRef:
                 description: |-
                   SourceSecretRef names a Secret (in the Model's namespace) whose keys are
-                  wired as env into the model-downloader init container. Used by s3://
-                  sources for S3-compatible credentials/endpoint: AWS_ACCESS_KEY_ID,
-                  AWS_SECRET_ACCESS_KEY, AWS_REGION, AWS_ENDPOINT_URL (path-style, e.g.
-                  https://minio.internal:9000 or https://s3.us-east-1.amazonaws.com).
+                  wired as env into the model-downloader init container.
+
+                  s3:// sources read S3-compatible credentials and endpoint from
+                  AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_REGION and AWS_ENDPOINT_URL
+                  (path-style, e.g. https://minio.internal:9000 or
+                  https://s3.us-east-1.amazonaws.com).
+
+                  hf:// and huggingface.co sources read HF_TOKEN, which is sent as a bearer
+                  credential so gated and private repositories can be downloaded. The token
+                  is attached only for huggingface.co and is dropped if a redirect leaves
+                  that host, so it never reaches a CDN or any other origin. Omit the key,
+                  or the whole Secret, for public repositories.
+
+                  One Secret may carry both sets of keys; each is used only by the source
+                  scheme it belongs to.
                 properties:
                   name:
                     default: ""
--- a/config/crd/bases/inference.llmkube.dev_models.yaml
+++ b/config/crd/bases/inference.llmkube.dev_models.yaml
@@ -461,10 +461,21 @@ spec:
               sourceSecretRef:
                 description: |-
                   SourceSecretRef names a Secret (in the Model's namespace) whose keys are
-                  wired as env into the model-downloader init container. Used by s3://
-                  sources for S3-compatible credentials/endpoint: AWS_ACCESS_KEY_ID,
-                  AWS_SECRET_ACCESS_KEY, AWS_REGION, AWS_ENDPOINT_URL (path-style, e.g.
-                  https://minio.internal:9000 or https://s3.us-east-1.amazonaws.com).
+                  wired as env into the model-downloader init container.
+
+                  s3:// sources read S3-compatible credentials and endpoint from
+                  AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_REGION and AWS_ENDPOINT_URL
+                  (path-style, e.g. https://minio.internal:9000 or
+                  https://s3.us-east-1.amazonaws.com).
+
+                  hf:// and huggingface.co sources read HF_TOKEN, which is sent as a bearer
+                  credential so gated and private repositories can be downloaded. The token
+                  is attached only for huggingface.co and is dropped if a redirect leaves
+                  that host, so it never reaches a CDN or any other origin. Omit the key,
+                  or the whole Secret, for public repositories.
+
+                  One Secret may carry both sets of keys; each is used only by the source
+                  scheme it belongs to.
                 properties:
                   name:
                     default: ""
--- a/docs/contributors/model-hf-source-spec.md
+++ b/docs/contributors/model-hf-source-spec.md
@@ -144,7 +144,13 @@ After code changes:
 ## Out of scope
 
 - Downloading HuggingFace weights into the cache PVC (that's option 2 from the issue; not needed for the vLLM workflow and adds significant complexity)
-- Auth for HF_TOKEN in the Model controller (runtime handles this)
+- ~~Auth for HF_TOKEN in the Model controller (runtime handles this)~~ **Shipped in #1750.**
+  The operator's own downloads now send a bearer token for `hf://` and
+  `huggingface.co` sources, read from `HF_TOKEN` in `spec.sourceSecretRef`. The
+  original reasoning held only for the runtime-resolved flow (a bare repo ID
+  with `skipModelInit: true`), which left every GGUF Model, every `spec.files`
+  staging and the prefetch Job unable to reach a gated repository at all. See
+  the gated-repositories section of the model cache guide.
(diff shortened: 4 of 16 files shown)


joryirving (commented):
Happy with this. One thing I want resolved before it goes in, and the rest is minor.

The `hf://` branch on the metal path looks unreachable, and the comment claims otherwise. `pkg/agent/s3.go:106` says "hf://, which the agent resolves to huggingface.co", and `executor_hfauth_test.go:108` pins `isHFAuthSource("hf://meta-llama/Llama-Guard-4-12B")` to true, but nothing in `pkg/agent` rewrites `hf://`. `downloadFile` hands the raw string to `http.NewRequestWithContext`, and `Do` comes back with `unsupported protocol scheme "hf"`; I ran it to be sure. The controller path is fine because `normalize_hf_source` does the rewrite in the shell, so this is metal only. Either the agent never receives an `hf://` source, in which case that comment describes a step that does not exist and the test is pinning dead code, or it can, in which case an `hf://` Model on a Metal agent fails long before the token matters. Worth saying which.

Second, `isHFAuthSource` now exists twice with two separate implementations, `internal/controller/source.go:264` and `pkg/agent/s3.go:109`. They agree today; I checked both against scheme case folding, the `www.` trim and the lookalike host, and they behave identically. But it is a credential gate with its own test table on each side and nothing tying them together, so a fix to one can silently leave the other permissive. `internal/` can import `pkg/`, so this could live once in `pkg/agent` and be called from the controller, which removes the drift class instead of testing for it.

Smaller things, none blocking. The `www.` trim is case sensitive in both copies (`source.go:249`, `s3.go:117`), so `https://WWW.huggingface.co/...` returns false and downloads unauthenticated. It fails safe and nobody writes that, but the comment directly above at `source.go:250` says the host is case insensitive per RFC 3986, so the code and its comment disagree. That is inherited from `isHuggingFaceURL` rather than new here. Separately, `hf_curl` passes the token in argv, so it shows up in `/proc/<pid>/cmdline`; in a single container init that is a small surface, but curl will read headers from a config on stdin if you ever want it tighter.

The redirect work is right, and it is the part I looked hardest at. Comparing `req.URL.Host` against `via[0]` rather than the previous hop is the correct semantic, since the question is whether you are still talking to the host you authenticated to. It also matters that Go re copies headers from the initial request's map on every hop and applies its own `isDomainOrSubdomain` rule before `CheckRedirect` runs, so the `Del` actually sticks instead of being undone on the next redirect. That is subtle and easy to get wrong. One cosmetic consequence: `req.URL.Host` carries the port, so `huggingface.co:443` redirecting to `huggingface.co` strips the header for what is the same effective host. That fails closed, so it costs an auth failure rather than a leak.

The mutation note about the assertions checking that the wrapper is invoked rather than merely defined is the thing I would have missed.

Defilan:
All three fixed, and you were right on the blocking one.

The metal path can receive `hf://`: `agent.go:606` sets `ModelSource: model.Spec.Source` verbatim, so nothing strips or rewrites it before `downloadFile`. I ran it and got your error exactly. So it is your second case, not the first: an `hf://` Model on a Metal agent fails on the scheme long before a token matters.

I have narrowed the predicate rather than papering over it. It is now `isHFAuthHost`, accepting `https`/`http` on huggingface.co only, with a comment saying why it deliberately differs from the controller's. The test that pinned `hf://` to true was pinning dead code, and is replaced by `TestDownloadFile_HFSchemeUnsupported`, which asserts the scheme error and whose failure message tells whoever fixes it to widen the predicate. The underlying gap is #1759.

That also answers the duplication, though not the way you suggested. The two gates now answer genuinely different questions, whether the metal downloader can reach the host versus whether the init container can, so they are intentionally different rather than a copy that might drift, and both comments say so. I did look at collapsing them into `pkg/agent`. The blocker is that a shared predicate only helps if the resolver moves with it, and `normalizeHFSource` pulls in `parseHFSource`, `isHuggingFaceURL`, `hfURLPathSegments` and `extractHFRepoFromURL`. That is a package move, not a function move, and I did not want it riding on an auth change. It is written up in #1759 as the fix that closes both at once.

The `www.` case is fixed on both sides: the fold now happens before the trim, so `https://WWW.huggingface.co/...` matches, and there is a case pinning it. You are right that it failed safe, and right that the comment above it was claiming something the code did not do.

On the two you flagged as cosmetic, I have left both. The token in argv is what the issue specified and it matches how the S3 path already passes credentials, so changing only this one would be inconsistent; a curl config file would be the way if we tighten it, and that is worth doing for both at once rather than half now. The port in the host comparison I am keeping precisely because it fails closed, and I would rather an auth failure than reasoning about when a port change is safe.

Thanks for running the `hf://` case rather than reading it. I had checked that the scheme reaches the predicate and not that it survives the request, which is the same mistake in a different place.

joryirving (changes requested):
Thanks, and the three fixes look right to me. `isHFAuthHost` rejecting `hf://` with `TestDownloadFile_HFSchemeUnsupported` pinning it is better than what I suggested, and the failure message telling the next person to widen the predicate is the part that will actually survive. I am happy with the argument for keeping the two predicates separate now that they answer different questions, and #1759 is the right place for the resolver move.

One thing fell out of the `www.` fix though, and I think it is a regression rather than a nit. Folding before the trim in `isHuggingFaceURL` (`source.go:255`) made it accept uppercase hosts, but `hfURLPathSegments` (`source.go:287`) still trims `www.` before folding, so the two now disagree. Every HF branch gates on the first and then calls through to the second.

```
https://WWW.HuggingFace.co/Qwen/Qwen3-8B/resolve/main/model.gguf
  on main:        isHuggingFaceURL=false isHFFileURL=false isRemoteHTTP=true  isHFRepo=false
  on this branch: isHuggingFaceURL=true  isHFFileURL=false isRemoteHTTP=false isHFRepo=false
```

So a URL that used to classify as a plain remote HTTP file and download fine is now neither an HF repo nor a remote HTTP source, and `model_controller.go:620` has a `case` for each of those and nothing for that combination. `parseHFSource` on the same input returns `invalid huggingface.co URL`. The lowercase spelling is identical on both branches, so this is specific to the uppercase host the fix newly accepts.

The line that closes it is the change you already made, applied in `hfURLPathSegments`, folding only the host so the repo path keeps its case:

```go
if i := strings.IndexByte(rest, '/'); i >= 0 {
	rest = strings.ToLower(rest[:i]) + rest[i:]
}
rest = strings.TrimPrefix(rest, "www.")
```

With that, `isHFFileURL` and `isRemoteHTTP` go back to true for the uppercase URL and match the lowercase one, `parseHFSource` returns `Qwen/Qwen3-8B` with the capital Q intact, and `./internal/controller` is green.

Worth a case in the table either way, since nothing currently fails if those two drift apart again.

Defilan:
Confirmed and fixed, and you are right that it is a regression rather than a nit. I reproduced your table before touching anything: on this branch the uppercase URL gave `isHuggingFaceURL=true` with `hfURLPathSegments` returning `ok=false`, while the lowercase spelling returned `ok=true`. So my fix moved the fold in one place and left the other, which is worse than not having moved it at all, because the disagreement only appears on the host the fix newly accepted.

Applied your line, and the repo path survives it: `hfURLPathSegments` now folds only up to the first slash, so `Qwen/Qwen3-8B` keeps its capital Q while the host is case-insensitive per RFC 3986.

Two tests rather than one, since your closing point is the real problem. `TestHFHostPredicatesAgree` asserts the two return the same answer across both spellings, both schemes, a lookalike host and a non-HF host, so a future edit to either one has to keep them aligned. `TestHFPathSegmentsPreserveCase` covers the other half, that folding the host does not reach the path. Both fail if the trim moves back ahead of the fold, which I checked by putting it back.

Worth saying plainly: this is the second time in this PR that a change I made passed its own test while breaking something the test did not look at. The first was the wrapper being defined but never invoked, and this one is a predicate agreeing with itself but not with its caller. Both were caught by you running the case rather than reading it.

`./internal/controller` is green and lint is clean.

joryirving (approved):
That closes it. I re-ran the table on your head and the uppercase URL now classifies identically to the lowercase one, `isHuggingFaceURL=true isHFFileURL=true isRemoteHTTP=true`, and `parseHFSource` gives back `Qwen/Qwen3-8B` with the capital Q intact, so folding to the first slash does the right thing on both halves.

I also put the trim back ahead of the fold to check your two tests bite, and they do:

```
--- FAIL: TestHFHostPredicatesAgree
    "https://WWW.HuggingFace.co/...": isHuggingFaceURL=true but hfURLPathSegments ok=false; the two must agree
    "https://HUGGINGFACE.CO/...":     isHuggingFaceURL=true but hfURLPathSegments ok=false; the two must agree
--- FAIL: TestHFPathSegmentsPreserveCase
    uppercase host rejected
```

`TestHFHostPredicatesAgree` is the one that matters, because it constrains the relationship rather than either function, which is what was missing when the two drifted. `./internal/controller`, `./pkg/agent` and `./api/...` are green here too.

On your closing point, I would put it differently. Both of those were changes whose blast radius was wider than the thing being changed: a wrapper's definition is not its use, and a predicate's own truth is not its agreement with its caller. Tests written from the change tend to cover the change. The two you added here are the other kind, since they pin a relationship that neither function owns, and that is the kind that survives someone editing one side a year from now.

Approving.

Issue #1724, opened 2026-08-31
[BUG] InferenceService sets a memory request but no limit, so an oversized model wedges the node instead of being OOM-killed


Defilan:
## Bug Description

`InferenceService` sets a memory **request** but never a memory **limit**, so
every serving container is unbounded. A model whose resident set exceeds
available RAM takes the whole node down instead of being OOM-killed and
restarted.

This is not hypothetical. It wedged a GB10 node in our lab today and required a
physical power cycle.

The same file already applies the correct pattern to disk. In
`internal/controller/deployment_builder.go`, `ephemeralStorage` sets both, with
a comment that argues the principle:

```go
// Request AND limit. The request is what the scheduler reserves ... The limit
// is what keeps an overrun charged to this pod: without one the node crosses
// its DiskPressure threshold instead and eviction proceeds by QoS class, which
// can remove unrelated workloads first.
res.Requests[corev1.ResourceEphemeralStorage] = q
res.Limits[corev1.ResourceEphemeralStorage] = q
```

Twenty lines above, memory gets requests only:

```go
if isvc.Spec.Resources.HostMemory != "" {
    res.Requests[corev1.ResourceMemory] = resource.MustParse(isvc.Spec.Resources.HostMemory)
} else if isvc.Spec.Resources.Memory != "" {
    res.Requests[corev1.ResourceMemory] = resource.MustParse(isvc.Spec.Resources.Memory)
}
```

The argument in that comment applies verbatim to RAM, and RAM is worse: crossing
MemoryPressure with an unbounded container does not evict politely by QoS, it
stops the kubelet posting status and the node goes `NotReady`.

There is also no CRD field to set one by hand. `spec.resources` exposes `cpu`,
`memory` (documented "Memory requests"), `hostMemory`, `ephemeralStorage`,
`gpu`, `gpuMemory`, `gpuSharing`. No limit counterpart, so an operator cannot
work around this from the manifest.

## Steps to Reproduce

1. Deploy an `InferenceService` whose weights plus KV plus compute buffers
   exceed the node's available RAM. Ours was a 103.69 GiB GGUF on a node with
   ~121 GiB allocatable, `resources.memory: 118Gi`, `--no-mmap`.
2. Watch the generated pod: `limits` contains only `nvidia.com/gpu`.
3. Let the model load.

## Expected Behavior

The container is OOM-killed, the pod restarts or reports `OOMKilled`, and the
node stays `Ready`. Blast radius is one pod.

[...]

PR #1764, opened 2026-09-07
feat(inferenceservice): add resources.memoryLimit to decouple ceiling from reservation

### LINKED ISSUE #1763: resources.memory sets requests and limits to the same value, with no way to set a limit above the request
### Summary

`resources.memory` is documented as "Memory requests", but since 0.9.25 the operator also emits it as `limits.memory`. Because request and limit are always the same quantity, there is no way to declare a workload that is allowed to burst above its steady-state reservation. Upgrading from 0.9.24 to 0.9.25 silently converted long-running inference pods from Burstable to a hard cap sized for their request, and any workload whose peak exceeded its request began OOM-looping.

### What happens

`internal/controller/deployment_builder.go`:

```go
// Request AND limit, driven by whichever of hostMemory/memory wins.
var memoryQ *resource.Quantity
if isvc.Spec.Resources.HostMemory != "" { ... } else if isvc.Spec.Resources.Memory != "" { ... }
if memoryQ != nil {
    res.Requests[corev1.ResourceMemory] = *memoryQ
    res.Limits[corev1.ResourceMemory] = *memoryQ
}
```

`hostMemory` does not decouple the two, it only changes which value wins. `api/v1alpha1/inferenceservice_types.go` still describes the field as `Memory requests (e.g., "4Gi")`, and `InferenceResourceRequirements` has no limit-side counterpart for RAM. `memoryLimitGiB` is GPU VRAM quota accounting for shared mode and is explicitly not enforced at runtime, so it does not help here.

### Why this is a problem

I understand the intent of the change, and the comment makes the reasoning explicit: without a limit an unbounded container crosses MemoryPressure and takes the node NotReady instead of being OOM-killed. That is a good property to want. The issue is that welding the limit to the request forces a workload to *reserve* its peak, not its steady state.

A concrete case: a llama.cpp server with a large RAM prompt cache idles around 4 GiB and grows to roughly 44 GiB as the cache fills, briefly touching 52 GiB. It spent 29 minutes of the last week above 48 GiB. Under 0.9.24 that was fine, since only the request was set and the node had the RAM free. To keep the same behaviour under 0.9.25 the request must be raised to cover the peak, which reserves 48 GiB permanently on a 123 GiB node and takes it from 71 percent to 97 percent requested, blocking anything else from scheduling there. The memory was always available; what changed is that it now has to be reserved to be usable.

### Suggested fix

Add an optional limit-side field, for example `resources.memoryLimit`, that sets `limits.memory` independently. When unset, keep the current behaviour so nothing changes for existing users and the NotReady protection still holds by default. That preserves the hardening you added while letting the request describe steady state and the limit describe the ceiling, which is the normal Kubernetes split.

If you would prefer a different shape, for example making `hostMemory` request-only now that the limit is always set, I am happy to work to whatever design you want. I can put up a PR for this if it is useful.

### Environment

- llmkube 0.9.24 to 0.9.25
- runtime: llamacpp, single GPU, Talos Linux

joryirving:
## What

Adds an optional `resources.memoryLimit` that sets `limits.memory` independently of the request, so a workload can burst above what it permanently reserves.

## Why

Fixes #1763

Since 0.9.25 the builder emits the memory quantity as **both** `requests.memory` and `limits.memory`, so request and limit are always equal and a workload has to *reserve* its peak in order to be allowed to reach it.

Concretely: a llama.cpp server with a large RAM prompt cache idles near 4Gi and grows to roughly 44Gi as the cache fills. Covering that peak means reserving 48Gi permanently on a 123Gi node, taking it from 71% to 97% requested and blocking anything else from scheduling there. The memory was always available; what changed is that it now has to be reserved to be usable.

## How

`memoryLimit` sets `limits.memory` when given, and applies over whichever of `memory`/`hostMemory` drives the request.

**Unset keeps today's behaviour exactly**, so the hardening from #1724 still holds by default — an unbounded container crossing MemoryPressure takes the node NotReady instead of being OOM-killed. This only ever widens the ceiling, and only when asked.

Two deliberate choices worth flagging:

- **A limit below the effective request is ignored, not applied.** The kubelet refuses to admit such a pod, so honouring it would turn a typo into an unschedulable workload rather than a narrower ceiling. A malformed value falls back the same way, matching the existing `ParseQuantity`-not-`MustParse` handling in this function.
- **`memoryLimit` alone sets nothing.** Without a request there are no memory keys at all; a bare ceiling would make the container BestEffort-with-a-limit, which nobody asked for.

Also corrected the `Memory` and `HostMemory` GoDoc: both said "requests" while the builder set request *and* limit, which is the mismatch that made the 0.9.25 behaviour surprising. CRDs and chart CRDs regenerated via `make manifests generate chart-crds`.

This is the shape suggested in the issue. Happy to rework it if you would rather go another way — for example making `hostMemory` request-only now that a limit is always set.

## Checklist

- [x] Tests added/updated
- [x] `make lint` passes locally — 0 issues
- [x] Documentation updated (if user-facing change) — the field's GoDoc, which is the CRD description

Five new subtests in `TestBuildContainerResources_MemoryLimit`: the ceiling above the request, applying over `hostMemory`, a limit below the request being ignored, a malformed value falling back, and a bare limit setting nothing. Negative control: reverting the builder change fails the first two.

Not manually tested in a cluster yet — the change is confined to the resource block the unit tests cover, but say the word if you want a live run before merge.

Assisted-by: Claude Code (wrote the field, the builder branch, the tests and this description). I reviewed the change, ran `make test` and `make lint`, and verified the tests fail when the builder change is reverted.

### DIFF
--- a/api/v1alpha1/inferenceservice_types.go
+++ b/api/v1alpha1/inferenceservice_types.go
@@ -1020,14 +1020,39 @@ type InferenceResourceRequirements struct {
 	// +optional
 	CPU string `json:"cpu,omitempty"`
 
-	// Memory requests (e.g., "4Gi")
+	// Memory is the pod's memory RESERVATION (e.g., "4Gi"), translated to
+	// requests.memory. It also sets limits.memory unless MemoryLimit is given,
+	// so by default request and limit are the same quantity and the container
+	// is Guaranteed for memory.
 	// +kubebuilder:validation:Pattern=`^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$`
 	// +optional
 	Memory string `json:"memory,omitempty"`
 
+	// MemoryLimit is the pod's memory CEILING (e.g., "64Gi"), translated to
+	// limits.memory. Set it to let a workload burst above its reservation.
+	//
+	// Without it, limits.memory equals the request, so a workload has to
+	// RESERVE its peak in order to be allowed to reach it. A server whose
+	// resident set idles low and grows into a large cache would have to reserve
+	// the peak permanently, taking that memory from the scheduler for the whole
+	// life of the pod even though it is free almost all the time (#1763).
+	//
+	// Leaving this unset preserves the equal request-and-limit behaviour, so
+	// the protection added in #1724 -- an unbounded container crossing
+	// MemoryPressure and taking the node NotReady instead of being OOM-killed
+	// -- still holds by default. This only widens the ceiling when a value is
+	// given, and a limit below the effective request is ignored rather than
+	// applied, since the kubelet rejects such a pod outright.
+	//
+	// Applies to whichever of Memory/HostMemory drives the request.
+	// +kubebuilder:validation:Pattern=`^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$`
+	// +optional
+	MemoryLimit string `json:"memoryLimit,omitempty"`
+
 	// HostMemory specifies the system RAM required for hybrid GPU/CPU offloading (e.g., "64Gi").
 	// Used when MoE expert weights or KV cache are offloaded to CPU via moeCPUOffload or noKvOffload.
 	// Translated to pod resources.requests.memory, taking precedence over Memory when set.
+	// Sets limits.memory too unless MemoryLimit is given.
 	// Without this, the K8s scheduler has no visibility into the pod's actual RAM consumption,
 	// which can lead to OOM kills after model load.
 	// +kubebuilder:validation:Pattern=`^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$`
--- a/charts/llmkube/templates/crds/inferenceservices.yaml
+++ b/charts/llmkube/templates/crds/inferenceservices.yaml
@@ -4635,12 +4635,38 @@ spec:
                       HostMemory specifies the system RAM required for hybrid GPU/CPU offloading (e.g., "64Gi").
                       Used when MoE expert weights or KV cache are offloaded to CPU via moeCPUOffload or noKvOffload.
                       Translated to pod resources.requests.memory, taking precedence over Memory when set.
+                      Sets limits.memory too unless MemoryLimit is given.
                       Without this, the K8s scheduler has no visibility into the pod's actual RAM consumption,
                       which can lead to OOM kills after model load.
                     pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
                     type: string
                   memory:
-                    description: Memory requests (e.g., "4Gi")
+                    description: |-
+                      Memory is the pod's memory RESERVATION (e.g., "4Gi"), translated to
+                      requests.memory. It also sets limits.memory unless MemoryLimit is given,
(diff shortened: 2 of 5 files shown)


Defilan (approved):
Thanks @joryirving, this is a clean fix and the diagnosis in #1763 was exactly right. Approving.

The GoDoc correction is the part I'm most glad you caught. `Memory` and `HostMemory` both said "requests" while the builder had been setting request *and* limit since 0.9.25, and that mismatch is most of why the change was surprising rather than merely tighter.

I checked that `buildContainerResources` is the only place that maps user memory spec onto a limit, so nothing else silently ignores the new field. `router_deployment_builder.go` uses its own hardcoded values and the multiNode path from #1746 goes through the same builder. Unset really is byte-for-byte today's behaviour, so #1724's NotReady protection still holds by default.

Three comments, none blocking. Take or leave them.

**1. Both ignore paths are silent, and we have a better tool for one of them.**

Your reasoning is right that honouring a limit below the request would turn a typo into an unschedulable pod. But widening it back to the request is also silent: someone who writes `memory: 48Gi, memoryLimit: 4Gi` gets a 48Gi ceiling with no error, no event and no condition, and walks away believing they capped it.

https://github.com/defilantech/LLMKube/blob/983c86d17ac5bdd1523f2e89e38085599caa9670/internal/controller/deployment_builder.go#L744-L753

We already use CEL in this file for the speculative-decoding rules, so the invalid combination can be rejected at admission with a message that names the field. I confirmed this compiles and is accepted by a 1.36 apiserver:

```
// +kubebuilder:validation:XValidation:rule="!has(self.memoryLimit) || (!has(self.memory) && !has(self.hostMemory)) || quantity(self.memoryLimit).compareTo(quantity(has(self.hostMemory) ? self.hostMemory : self.memory)) >= 0",message="resources.memoryLimit must be >= the effective memory request (hostMemory when set, otherwise memory)"
```

`quantity()` needs 1.29+, which is comfortably below our floor. If you'd rather keep the builder tolerant as a backstop, both can coexist: CEL catches it at write time, the builder stays defensive for objects that predate the rule.

**2. `resources.memoryLimit` sits one word from `resources.gpuSharing.memoryLimitGiB`.**

One is host RAM the kubelet enforces, the other is GPU VRAM accounting that we document as explicitly not enforced at runtime:

https://github.com/defilantech/LLMKube/blob/983c86d17ac5bdd1523f2e89e38085599caa9670/api/v1alpha1/inferenceservice_types.go#L1118-L1133

You had to rule that field out explicitly while writing #1763, which is the tell. A one-line cross-reference in each GoDoc would save the next person the same detour.

**3. "memoryLimit alone sets nothing" is in the PR description but not the GoDoc.**

https://github.com/defilantech/LLMKube/blob/983c86d17ac5bdd1523f2e89e38085599caa9670/api/v1alpha1/inferenceservice_types.go#L1029-L1046

The GoDoc becomes the CRD description, so it's what people actually read. "Applies to whichever of Memory/HostMemory drives the request" implies it but doesn't say that with neither set you get no memory keys at all.

On your offer to make `hostMemory` request-only instead: I'd rather keep what you've built. Additive and opt-in means nobody's existing manifest changes meaning, which the 0.9.25 upgrade taught us to care about.

No need for a cluster run before merge. The change is confined to the resource block and your subtests cover it, including the negative control.

Issue #1757, opened 2026-09-04
[BUG] multiNode group cannot start when one member's cache is empty: waiting runtimes exit, recreate kills the downloader, loop


Defilan:
## Bug Description

A `multiNode` group whose members do not all have the model cached cannot start. The members that already hold weights start their runtime immediately and block waiting to rendezvous with a member that is still in its `model-downloader` init container. Those waiting runtimes eventually exit, the controller reads that as a restarted member and recreates the whole group, and the recreate kills the downloading member partway through. It then starts over from zero.

The two behaviours are individually reasonable. Recreating the group when a member restarts is correct, because a half-rendezvoused group is not recoverable. Downloading in an init container is correct. Together they turn a slow first start into a loop that cannot break out on its own.

Measured on a three-member vLLM group: **62 recreates in 20 minutes**, with the new member never getting past 5 GiB of a 184 GiB checkpoint.

## Steps to Reproduce

1. Stand up a `multiNode` group of N members and let it stage fully, so every member's cache is populated.
2. Add one member whose `modelCache` claim is empty. In our case this was widening a two-member group to three.
3. Apply and scale to 1.

The same shape occurs on a first deployment where one member's staging is much slower than the others, for example a node with a slower link to object storage.

## Expected Behavior

The group waits for every member to finish staging before any runtime is expected to rendezvous, and a member still running init containers is not counted as having restarted.

## Actual Behavior

```
$ kubectl -n llmkube-system logs deploy/llmkube-controller-manager --since=25m \
    | grep -c 'recreating group'
62

# reasons, same window
{'MemberRestarted': 23, 'MemberTerminating': 39}
```

The group hash was stable throughout (`e0e8aad1eaf5378a` on all three members across successive reconciles), so this is not the status-blind-hash problem fixed in #1746.

Events show all three members being killed roughly every 50 seconds, while the new member is still in its downloader:

```
Normal  Killing  pod/glm53-flash-ring-mn-0  Stopping container model-downloader
Normal  Killing  pod/glm53-flash-ring-mn-1  Stopping container vllm
Normal  Killing  pod/glm53-flash-ring-mn-2  Stopping container vllm
```

Staged bytes per member during the loop, showing the new member losing its progress on each cycle:

| member | node | staged |
| --- | --- | --- |
| mn-0 | new | 3.6 GiB, then 5.3 GiB, never more |
| mn-1 | existing | 185 GiB |
| mn-2 | existing | 185 GiB |

## Environment

[...]

Issue #1536, opened 2026-08-14
[BUG] ModelTransferHealthy never recovers after a retried model download succeeds


Defilan:
## Bug Description

`ModelTransferHealthy` stays `False` for the entire life of a pod once its
staging init container has failed even one attempt, **even after a later attempt
succeeds and the service serves normally**.

`findModelTransferFailure` scans both terminated states of each generated init
container:

```go
for _, term := range []*corev1.ContainerStateTerminated{
    cs.State.Terminated,
    cs.LastTerminationState.Terminated,
} {
    if term == nil || term.ExitCode == 0 {
        continue
    }
    if r, line, ok := matchModelTransferFailure(term.Message); ok {
        return pod.Spec.NodeName, r, line, true
    }
}
```

Checking `LastTerminationState` is correct for a crash-looping container, which
is what the comment describes. But `LastTerminationState` also retains a
*recovered* failure forever. Once a download retries and succeeds, the current
state is exit 0 while last-termination still holds the failed attempt, so
`found` stays true permanently. The `!found` recovery branch in
`reconcileModelTransferCondition` (which would set `ModelTransferSucceeded`) is
therefore unreachable for that pod.

## Steps to Reproduce

1. Create a Model with an `s3://` source whose endpoint is briefly unreachable
   (or otherwise make the first staging attempt fail with a matched signature).
2. Let the init container retry until it succeeds. `restartCount` > 0, final
   `state.terminated.exitCode` = 0.
3. Wait for the service to reach `Available=True`.
4. `kubectl get isvc <name> -o json | jq '.status.conditions'`

## Expected Behavior

Once the staging init container has terminated with exit 0 and the workload is
serving, `ModelTransferHealthy` flips to `True` / `ModelTransferSucceeded`.

## Actual Behavior

It stays `False` indefinitely. Observed live on `qwopus-fusion-vision-s3`, which
has been serving for 26h:

```
Available=True             at 2026-08-13T05:28:54Z   "Inference service is ready and serving requests"
ModelTransferHealthy=False at 2026-08-13T04:21:01Z   ModelSourceUnreachable
```

The init container state shows exactly the recovered-failure shape:

```
model-downloader: restartCount=22
  state.terminated:     exit=0 reason=Completed
  lastState.terminated: exit=1 reason=Error
```

Forcing a reconcile (annotation touch) does not clear it. Both the pod and the
node named in the condition are the same `shadowstack` node currently serving,
so this is not a stale-node artifact.

[...]

Issue #1435, opened 2026-08-06
[BUG] Interrupted model transfers leave orphaned .tmp files that nothing ever cleans up


Defilan:
## Describe the bug

Model transfers publish atomically by downloading to `<path>.tmp` and `mv`-ing onto the final path. That is correct and deliberate (#1326 for `RefreshPolicy=OnChange`, #1432 for every other path). But nothing in the codebase ever removes a `.tmp` file left behind by a transfer that never completed, so they accumulate on the shared, durable model-cache PVC.

## To Reproduce

1. Deploy a Model with `modelCache` enabled and the default `RefreshPolicy: IfNotPresent`.
2. Interrupt the init container mid-transfer (OOM-kill, eviction, node reboot). A partial `<file>.tmp` is left in the cache directory.
3. Delete the Model, or re-point it at a different source, before a retry succeeds.
4. The `.tmp` file remains on the PVC indefinitely.

## Expected behavior

Abandoned `.tmp` files are eventually reclaimed, or are at least visible so an operator knows to reclaim them.

## Actual behavior

They are invisible and permanent:

- `llmkube cache clear` (`pkg/cli/cache.go`, `runCacheClear`) only prints `kubectl exec ... rm -rf` instructions; it deletes nothing itself.
- The prefetch Job's owner-ref and TTL cleanup (`internal/controller/model_prefetch.go`) removes the Job and Pod, never files already written into the shared PVC.
- Nothing in the reconcile path or the `cachekey` package touches PVC file contents.

They also silently inflate reported cache size: `inspectSinglePVC` in `pkg/cli/cache_inspect.go` runs `du -sb <cachekey>/*/`, which sums the whole subtree, so orphaned `.tmp` bytes are folded into the cache key's `SIZE` in `llmkube cache list` with no way to tell model bytes from debris.

## Not affected

Worth stating so the scope stays narrow. A stray `.tmp` is never served or re-listed as a model: the runtime container uses deterministic paths, `findCachedModelFile` globs `*.gguf`, and the multi-file `MODEL_FILES` manifest comes from the Model spec rather than from listing the destination directory. This is a space and observability problem, not a correctness one.

## Additional context

Pre-existing since #1326, but #1432 extends the same exposure from the `OnChange` path to every non-`OnChange` path (single-file cached and uncached http/S3/local, plus the multi-file loop). Since `IfNotPresent` is the default refresh policy, the practical frequency goes up meaningfully.

Two options that seem reasonable, either alone:

[...]

Issue #1379, opened 2026-07-31
[FEATURE] Evaluate OCI ImageVolume (k8s 1.36 GA) as a model source


Defilan:
## Feature Description

Kubernetes `ImageVolume` (the OCI VolumeSource) reached **GA and was locked in k8s 1.36**. It mounts an OCI artifact directly into a pod as a volume — meaning model weights can be pulled straight from a registry with **no init container and no download/puller step**.

## Problem Statement

> As a fleet operator, I want model weights delivered by the same registry, credentials, and mirroring I already use for images, so that air-gapped and multi-node deployments do not need a separate download-and-cache path.

LLMKube today stages weights via a download Job plus a model-cache PVC (`pvc://` sources, `spec.modelCache`). That path works, but it carries its own credential handling, cache-key logic, PVC lifecycle, and the oscillation class of bugs we have already hit (#928, #1139, #1353).

## Proposed Solution

Evaluate, and if it holds up, add an OCI-artifact model source (e.g. `oci://`) backed by `ImageVolume`:

- Weights become a normal OCI artifact — registry auth, mirroring, signing, and air-gapped import all reuse the container-image path we already document.
- Removes the init-container/puller and the per-service cache PVC for that source type.
- Complements rather than replaces `pvc://` and `s3://`; existing sources stay.

Open questions to answer in the evaluation: k8s 1.36 floor (do we gate the source type on server version?), runtime support in the CRI versions we target, artifact layout for multi-file models (relates to the pinned multi-file Hugging Face work in #1110), and read-only/immutability semantics vs the current cache.

## Alternatives Considered

- *Keep the Job + PVC path only* — works, but keeps a bespoke delivery mechanism where the platform now offers a first-class one.

## Additional Context

Surfaced during the 2026-07-31 Blackwell floor re-verification for #413; it is probably the single most directly relevant recent upstream change for LLMKube's model-cache design. Relates to #53 (air-gapped sources, closed), #1353 / #1363 (`pvc://` cache handling).

## Priority

Medium — high leverage, but gated on a k8s 1.36 floor.

Defilan:
The evaluation has landed as **#1854** (from my fork): `docs/proposals/1379-oci-imagevolume-model-source.md`.

Recommendation: **go, as an opt-in source type, gated on a modern cluster.** Summary of what the doc establishes:

[...]

PR #1780, opened 2026-09-08
fix: resume interrupted model downloads instead of restarting from byte zero (#1762)

### LINKED ISSUE #1762: [FEATURE] Model downloader is single-stream and cannot resume: an interrupted transfer restarts from byte zero
## Feature Description

The generated `model-downloader` init container fetches weights with a **single `curl` stream** and **cannot resume**. Both are one-command changes, and both bite harder every time the median checkpoint gets bigger.

Resume is the higher-value half and is independently shippable. Parallelism is the contentious half because it implies a new image.

## Problem Statement

As an operator serving checkpoints measured in tens or hundreds of GB, I want an interrupted transfer to continue from where it stopped and to use the bandwidth I have, so that a slow first deploy does not become a stuck one.

Current command, from a running pod spec (`initContainers[1]`, `docker.io/curlimages/curl:8.18.0`):

```sh
mkdir -p "$CACHE_DIR" && rm -f "$MODEL_PATH.tmp" && \
if [ ! -f "$MODEL_PATH" ]; then
  echo 'Downloading model...'
  curl -f -L -o "$MODEL_PATH.tmp" "$MODEL_SOURCE" && mv "$MODEL_PATH.tmp" "$MODEL_PATH" && \
  echo 'Model downloaded successfully'
else
  echo 'Model already cached, skipping download'
fi
```

### 1. No resume: the partial is deleted at the start of every attempt

`rm -f "$MODEL_PATH.tmp"` discards all progress before each try. There is no `--continue-at -` and no range request. Any interruption during transfer (pod eviction, node reboot, OOM kill, a 5xx from the CDN, an init-container restart) throws away everything downloaded so far and restarts from byte zero.

This is not theoretical, and the project has already recorded it twice:

- **#1757** (open): a `multiNode` group loops on recreate while one member stages. **62 recreates in 20 minutes**, and the staging member never gets past **5 GiB of a 184 GiB checkpoint** — the issue's own table shows it losing progress on every cycle. Neither candidate fix in #1757 involves resume, so both leave the "restart from zero" amplifier in place.
- **#1536** (closed): a live serving pod observed with `model-downloader: restartCount=22`. Downloader restarts are routine, not rare.

Resume would not have prevented #1757, but it would have turned 62 discarded restarts into 62 resumed continuations. It is the difference between a loop that is slow and a loop that cannot finish.

All transfer paths are affected, not just the single-file cached one: `buildModelInitCommand` (every branch), `remoteRevalidateScript`, `buildMultiFileInitCommand` (which sweeps with `find "$CACHE_DIR" -name '*.tmp' -delete`), and the metal agent's `MetalExecutor.copyToFile`, which `os.Remove`s its `.partial` on any error.

### 2. Single stream leaves more than half the available bandwidth unused

Measured against the same HF file on one node:

| condition | MB/s |
|---|---:|
| the real in-progress download, alone | 92.4 |
| the same download, while a 2nd stream ran | 90.2 |
| that 2nd concurrent stream | 93.0 |
| **combined** | **183.2 (1.47 Gbps)** |

Two concurrent connections each got the full ~90 MB/s, so **the ceiling is per-connection, not the WAN link**. HF's CDN throttles a single connection to roughly 90 MB/s. Only two streams were tested; the real ceiling is higher.

On a 94 GB pull that is `17.2 min` single-stream against `~8.6 min` at the 183 MB/s two streams already demonstrated. It is a standing tax on every cold pull: a new model, a quant change, or any cache loss.

## Proposed Solution

Two independent changes; the first is worth doing alone.

**Resume.** Add `--continue-at -` and stop unconditionally deleting the partial.

**This must not regress #1435.** That issue established that abandoned `.tmp` files accumulate invisibly on the shared cache PVC, and its accepted resolution is precisely the sweep that exists today:

> Have the cache-prep init container sweep stale `*.tmp` ... before starting a new download.

So the sweep cannot simply be dropped. It needs to distinguish **stale debris** — a transfer abandoned for good, or a source that has since been re-pointed — from **resumable progress on the transfer about to be retried**. An age threshold (as #1435 itself suggests: "older than some threshold") or keying the partial to the resolved source would both work. A partial that fails the test is deleted exactly as now; one that passes is resumed into.

The atomic publish must be preserved as-is. `remoteRevalidateScript`'s contract — never `-o` onto the live cached file, always `.tmp` then `mv` — is the fix for #1309 and #1432 and should not be touched. Resume changes how `.tmp` is *filled*, never how it is published.

Resume also composes cleanly with the HF auth wrapper added for #1750: `hf_curl()` forwards `"$@"`, so `--continue-at -` needs no special handling there. Worth confirming the `Authorization` header survives onto the range request for gated repos.

**Parallelism.** Options, roughly in increasing order of blast radius:

- `curl --continue-at -` plus explicit parallel range requests, staying on the current image.
- `aria2c -x8 -s8 --continue=true`, which gets both defects in one command but needs a new image.
- `hf_transfer` / `hf download` for `huggingface.co` sources specifically, which does chunked parallel transfer natively and is what HF recommend for large files. Narrower applicability, and it would not help `s3://` sources.

Any of these needs a cap that is polite to the CDN, and ideally a knob rather than a hardcoded stream count.

## Alternatives Considered

- **`skipModelInit`** and owning the download out of band. This is the current escape hatch and it means reimplementing staging, cache-key layout and revalidation downstream. Strictly worse.
- **Pre-filling the claim with a pinned `Job`**, the workaround documented in #1757. Works, but it is manual, per-node, and has to enumerate objects by hand when the storage key lacks `ListObjectsV2`.
- **#1379, OCI ImageVolume as a model source.** A different route to the same destination, and containerd's pull is already both parallel and resumable. If that lands and becomes the recommended path for large checkpoints, much of this issue is subsumed for image-sourced models — but not for `hf://` or `s3://` sources, which is what the downloader exists to serve.

## Additional Context

- Related issues:
  - #1757 — live 184 GiB case, progress discarded on every recreate (open)
  - #1435 — the `.tmp` sweep this change must not regress (closed)
  - #1536 — `restartCount=22` observed on a serving pod (closed)
  - #1309, #1432, #1326 — the atomic-publish lineage to preserve (closed)
  - #1750 — `hf_curl()` wrapper that resume must compose with (open)
  - #52, #193 — prior "do not re-transfer bytes you already have" work (closed)
  - #1379 — OCI ImageVolume alternative (open)
- Workarounds currently in use: none for resume; cold pulls are simply absorbed.

Not urgent in isolation — nothing is broken on a clean run, and cold pulls are rare. But #1757 shows the failure mode compounding with unrelated bugs, and the cost scales with checkpoint size.

## Priority

- [ ] High - Would significantly improve my workflow
- [x] Medium - Nice to have
- [ ] Low - Minor improvement

## Willingness to Contribute

- [ ] Yes, I can submit a PR
- [x] Yes, I can help test
- [ ] No, just requesting

Defilan:
## What

Interrupted model downloads now resume from where they stopped instead of restarting from byte zero, in both transfer paths:

- The generated `model-downloader` init container (single-file cached/emptyDir, S3, `RefreshPolicy=OnChange` revalidation, and all multi-file loops), including the prefetch Jobs that reuse the same init containers.
- The metal agent's model fetch (`MetalExecutor.downloadFile` / `downloadS3` / `copyToFile`).

The stale-`.tmp` sweep from #1435 is kept but narrowed to age-based deletion, so abandoned debris is still reclaimed while an in-flight partial survives as resumable progress.

## Why

Fixes #1762

Every generated download script deleted its own `.tmp` at the start of each attempt (`rm -f "$MODEL_PATH.tmp"`, and `find "$CACHE_DIR" -name '*.tmp' -delete` in the multi-file prefix), so any interruption (pod eviction, node reboot, OOM kill, CDN 5xx, init-container restart) discarded all transfer progress. Downloader restarts are routine (#1536 observed `restartCount=22`), and #1757 showed the amplifier in the wild: a 184 GiB checkpoint that never got past 5 GiB across 62 recreates because every cycle started over.

The metal agent had the same behavior: `copyToFile` `os.Remove`d its `.partial` on any error.

## How

**Controller (first commit).** Remote branches download with `curl -C -`, and the #1435 sweep becomes `find ... -name '*.tmp' -mtime +1 -delete`. Resuming keyed on freshness is safe here by construction: the cache dir is keyed by hash-of-source, and publishes are atomic `mv`, so a `.tmp` can only ever be a partial for its own destination. Each download site additionally drops a zero-byte `.tmp` (a strict server 416s `Range: bytes=0-`) and a byte-exhausted one (curl's 416-restart would replay the body and publish a doubled file), via a HEAD probe routed through `hf_curl` so gated sources authenticate. The emptyDir single-file branches, which previously left `.tmp` debris with no cleanup at all, are now swept too. The atomic `.tmp`-then-`mv` publish and the HEAD + Content-Length revalidation from #1309/#1326/#1432 are unchanged.

Two deliberate exclusions, reasoned in code comments:

- The local `cp` branch keeps its unconditional `rm -f`: a local-disk copy has nothing worth resuming.
- `--ignore-content-length` is deliberately *not* used, even though it would turn curl's exit-33 (range ignored) into a content-verified restart. curl's content-verification path accepts a 200 body that starts at the requested offset, which a broken gateway can exploit to get a truncated append published. Exit 33 loops until the age sweep reclaims the partial; publishing a corrupt artifact is the worse failure.

**Metal agent (second commit).** `resumeGuard` decides resume-vs-clean-start *before* the conditional GET (Go's http client has no curl-style post-hoc refusal): it drops zero-byte and byte-exhausted partials using a cached HEAD. A 206 appends and is verified `have + received == total` against `Content-Range`; that total-based check is load-bearing, because a gateway can answer a ranged GET with a truncated 200 that passes a written-vs-Content-Length check internally. 200-that-ignored-range and 416 (re-pointed smaller source) restart clean. Every failure path is fail-safe: no corrupt artifact is ever published. The #1750 bearer token rides the resume HEAD and the first hop only; `hfRedirectStripper` still strips it across the CDN host change.

**Tests.** Updated the #1435 sweep assertions and the golden command substrings; added a resume Describe block. Two new behavioral tests execute the actual generated scripts against range-capable and range-ignoring stub servers (resume-from-tail, byte-exhausted restart, zero-byte guard, stale sweep, warm-cache skip, no-publish-on-loud-failure), and the agent tests cover 206/200/416 plus the lying-gateway rejection. The curl-side behavioral tests skip on macOS (`stat -c`) like the existing #1309 guard, and were run executed under a GNU-stat shim; CI/Linux runs them natively.

**Not in scope (deliberately):** the parallel multi-stream half of #1762. It needs an image decision (aria2c vs range-splitting vs hf_transfer), a CDN-politeness cap, and a knob; #1379 (OCI ImageVolume) may subsume much of it for image-sourced models. Resume ships alone per the issue's own framing ("the first is worth doing alone").

## Checklist

- [x] Tests added/updated

---

<sub>Assisted-by: Claude Code (via OpenCode) (generated the code and tests; a human reviewed the diffs, verified the resume behavior against range-capable, range-ignoring, and lying-gateway repro servers before and during development, and ran `make fmt`, `make vet`, `make lint`, `make lint-all`, and the full `make test` suite locally).</sub>


Defilan:
Closing to redo with falsification step in the plan. Superseded by a fresh branch.

Issue #1779, opened 2026-09-08
[BUG] ForemanFleetNodeHeartbeatStale annotations use $labels.node, which the metric does not have


joryirving:
## Bug Description

`ForemanFleetNodeHeartbeatStale` in `charts/foreman/templates/prometheusrule.yaml` interpolates `{{ $labels.node }}`, but the metric it alerts on carries no `node` label. Both annotations render with a hole where the FleetNode name should be, so the alert cannot say which FleetNode is stale — on a multi-node fleet it gives you nothing to act on.

Found while chasing #1778, where the alert stayed critical for hours and the only way to identify the culprit was to diff `kubectl get fleetnodes` against the live pods by hand.

## Steps to Reproduce

1. Install the foreman chart with `prometheusRule` enabled and the `foreman-crs` kube-state-metrics CustomResourceState config.
2. Let a FleetNode go stale (see #1778 — replace an agent pod).
3. Read the firing alert's annotations.

## Expected Behavior

The summary names the FleetNode, e.g. `FleetNode foreman-llmkube-agent-5c9cffbcf7-2vk64 heartbeat stale`.

## Actual Behavior

```
summary:     "FleetNode  heartbeat stale"
description: "FleetNode  has not heartbeated in more than 5m. ..."
```

The label is simply absent. Labels on the firing alert:

```
alertname=ForemanFleetNodeHeartbeatStale
component=foreman  severity=critical
customresource_group=foreman.llmkube.dev
customresource_kind=FleetNode  customresource_version=v1alpha1
job=kube-state-metrics  namespace=observability
name=foreman-llmkube-agent-5c9cffbcf7-2vk64
```

`name` is present; `node` is not. That follows from the metric definition — the FleetNode block sets `labelsFromPath: name: [metadata, name]` and the `heartbeat_timestamp_seconds` gauge adds nothing further:

```yaml
- name: heartbeat_timestamp_seconds
  help: "FleetNode last heartbeat unix timestamp (0 = no heartbeat)"
  each:
    type: Gauge
    gauge:
      path: [status]
      valueFrom: [lastHeartbeatTime]
```

Two ways to fix, and they are not exclusive:

1. **Use the label that exists.** Change both annotations to `{{ $labels.name }}`. `name` is set for every FleetNode series, so this cannot render blank.
[...]

PR #1785, opened 2026-09-10
feat(foreman): expose node label on FleetNode metrics for fleet joins

srosignoli:
## What
Elevates `node: [spec, nodeName]` to the `FleetNode` resource-level `labelsFromPath` in `charts/foreman/templates/crs-configmap.yaml`, and removes the duplicate declaration from `foreman_fleetnode_info`.

## Why
Follow-up to #1779 and #1783.

While #1783 resolved the alert text hole by switching to `$labels.name`, the underlying `foreman_fleetnode_*` metrics still lacked the `node` label.

As discussed in #1779, exposing `node: [spec, nodeName]` at the resource level completes **Option 2**:
    1. Enables direct PromQL fleet joins between `AgenticTask` (which uses `node: [status, assignedNode]`) and `FleetNode` series without requiring an extra join through `foreman_fleetnode_info`.
    2. Aligns `FleetNode` metrics with how `AgenticTask` is already modeled.

## How
1. Added `node: [spec, nodeName]` to `labelsFromPath` under the `FleetNode` block in `charts/foreman/templates/crs-configmap.yaml`.
2. Removed `node: [spec, nodeName]` from `foreman_fleetnode_info` to prevent duplicate label definitions in `kube-state-metrics`.
3. Added unit test in `charts/foreman/tests/crs_test.yaml` verifying that the rendered CRS ConfigMap defines `node: [spec, nodeName]`.

## Verification
- `make test-chart`: All 85 unit tests pass in `charts/foreman`, and 103 pass in `charts/llmkube`.
- Verified live in a Kind cluster: KSM successfully parses the config and emits `foreman_fleetnode_heartbeat_timestamp_seconds` with `node="<worker-name>"`.

foreman-fleetnode-1

foreman-fleetnode-2

## AI Assistance Disclosure
Assisted-by: Antigravity (helped analyze the KSM schema and draft the unit test; verified locally via helm unittest and live Kind cluster).

## Checklist
    - [x] Tests added/updated
    - [x] `make test-chart` passes locally (85 passed in foreman, 103 passed in llmkube)
    - [x] AI assistance disclosed above per [CONTRIBUTING.md](../CONTRIBUTING.md#ai-assisted-and-agent-contributions)
    - [ ] Documentation updated (not applicable: internal metric label enhancement)


joryirving (approved):
This looks right to me, approving. Elevating node: [spec, nodeName] to the resource-level labelsFromPath in crs-configmap.yaml is the correct way to get the worker label onto every FleetNode series, and dropping it from foreman_fleetnode_info is the right cleanup, since the info metric now inherits node from the resource level so no label is lost there (its comment header still holds).

I checked spec.nodeName against the live CRD to be sure the join actually lines up. It's the worker/agent identity, the pod name, which the CRD deliberately keeps separate from status.kubernetesNode, and that's exactly what AgenticTask carries in status.assignedNode. So the direct PromQL join this is meant to enable matches agent-to-agent as intended.

Two small notes, neither blocking. The node label being the agent identity rather than the Kubernetes machine is a little easy to misread in PromQL later, but it's consistent with how AgenticTask is already modeled and there's a separate kubernetes_node label for the machine, so I'd keep it as is. And the new test only regex-matches that node: [spec, nodeName] renders, without pinning it to the resource level versus a metric block, which is fine for a chart unit test but on the light side.

Nice follow-up, thanks.

Defilan (approved):
This looks good. Thanks for taking the time to making a PR @srosignoli ! Welcome to the project, feel free to help out anytime!

PR #1787, opened 2026-09-10
feat(router): poolActivation IfIdle serves the warm member when the incumbent is busy

### LINKED ISSUE #1786: [FEATURE] ModelRouter: rule-level poolActivation IfIdle serves the warm member instead of holding behind a busy incumbent
### Summary

`resources.memory` is documented as "Memory requests", but since 0.9.25 the operator also emits it as `limits.memory`. Because request and limit are always the same quantity, there is no way to declare a workload that is allowed to burst above its steady-state reservation. Upgrading from 0.9.24 to 0.9.25 silently converted long-running inference pods from Burstable to a hard cap sized for their request, and any workload whose peak exceeded its request began OOM-looping.

### What happens

`internal/controller/deployment_builder.go`:

```go
// Request AND limit, driven by whichever of hostMemory/memory wins.
var memoryQ *resource.Quantity
if isvc.Spec.Resources.HostMemory != "" { ... } else if isvc.Spec.Resources.Memory != "" { ... }
if memoryQ != nil {
    res.Requests[corev1.ResourceMemory] = *memoryQ
    res.Limits[corev1.ResourceMemory] = *memoryQ
}
```

`hostMemory` does not decouple the two, it only changes which value wins. `api/v1alpha1/inferenceservice_types.go` still describes the field as `Memory requests (e.g., "4Gi")`, and `InferenceResourceRequirements` has no limit-side counterpart for RAM. `memoryLimitGiB` is GPU VRAM quota accounting for shared mode and is explicitly not enforced at runtime, so it does not help here.

### Why this is a problem

I understand the intent of the change, and the comment makes the reasoning explicit: without a limit an unbounded container crosses MemoryPressure and takes the node NotReady instead of being OOM-killed. That is a good property to want. The issue is that welding the limit to the request forces a workload to *reserve* its peak, not its steady state.

A concrete case: a llama.cpp server with a large RAM prompt cache idles around 4 GiB and grows to roughly 44 GiB as the cache fills, briefly touching 52 GiB. It spent 29 minutes of the last week above 48 GiB. Under 0.9.24 that was fine, since only the request was set and the node had the RAM free. To keep the same behaviour under 0.9.25 the request must be raised to cover the peak, which reserves 48 GiB permanently on a 123 GiB node and takes it from 71 percent to 97 percent requested, blocking anything else from scheduling there. The memory was always available; what changed is that it now has to be reserved to be usable.

### Suggested fix

Add an optional limit-side field, for example `resources.memoryLimit`, that sets `limits.memory` independently. When unset, keep the current behaviour so nothing changes for existing users and the NotReady protection still holds by default. That preserves the hardening you added while letting the request describe steady state and the limit describe the ceiling, which is the normal Kubernetes split.

If you would prefer a different shape, for example making `hostMemory` request-only now that the limit is always set, I am happy to work to whatever design you want. I can put up a PR for this if it is useful.

### Environment

- llmkube 0.9.24 to 0.9.25
- runtime: llamacpp, single GPU, Talos Linux

joryirving:
## What

Adds `spec.rules[].route.poolActivation: Wait | IfIdle` to `ModelRouter`. Under `IfIdle`, a rule whose preferred backend is a non-resident `ModelPool` member skips that backend immediately when the incumbent is busy and falls through to the next backend in the rule, instead of holding the request until the incumbent drains. `Wait` is the default and preserves today's behaviour exactly.

## Why

A pool holds a cross-model request until the incumbent is idle, bounded only by the pool-wide `swapBudget`. That is the right contract for a caller that accepts one model and the wrong one for a caller that merely prefers one: on a single 24 GB card shared by a small assistant model and a large coder, the assistant waits minutes behind a batch job while the warm coder could serve it now. Client-side fallback cannot solve it because a timeout cannot distinguish "incumbent busy" from "swap already in flight", so it abandons real swaps and lands on a member that is mid-unload.

Fixes #1786

## How

- `RuleRoute.PoolActivation` on the CRD (enum `Wait;IfIdle`, default `Wait`), mirrored on the wire `RuleRoute`, validated in `Config.Validate`, threaded through `MatchResult` and copied by `translateRule`.
- `Activator.AcquireWithMode`: in the `!swapping && resident != member` branch, a busy incumbent under `IfIdle` returns a new `ErrIncumbentBusy` instead of blocking; `Acquire` delegates with `Wait`. The existing deferred cleanup handles the waiter count. A swap already in flight is waited on in both modes, since the incumbent is unloading and cannot serve either way.
- `dispatchWithFallback` passes the matched rule's mode; the existing `continue` on activation error does the fallthrough.
- New counter `llmkube_modelpool_busy_skips_total{router,pool,member}`.
- Docs: new subsection in `docs/operations/gpu-sharing.md`, including the caveat that "busy" means proxy-tracked traffic (direct-to-Service traffic is invisible to the check; the controller's `/slots` drain still protects it at swap time).

Tests: activator unit tests for the skip and for the in-flight-swap exception, a proxy end-to-end test (rule `[coder, judge]`, judge resident and busy, request served by judge with coder never activated), the `translateRule` passthrough, and closed-enum validation. Both behavioural tests fail with the `IfIdle` branch disabled and pass with it.

Assisted-by: Claude Code (Fable 5.1) generated the implementation, tests and docs from my design; I reviewed the diff, and ran `make test`, `make lint` and `make lint-all` locally (all green, router coverage 93.7%).

## Checklist

- [x] Tests added/updated

### DIFF
--- a/api/v1alpha1/modelrouter_types.go
+++ b/api/v1alpha1/modelrouter_types.go
@@ -388,6 +388,21 @@ type RuleRoute struct {
 	// +kubebuilder:default=primary-fallback
 	// +optional
 	Strategy string `json:"strategy,omitempty"`
+
+	// PoolActivation decides what the proxy does when a backend in
+	// Backends is a ModelPool member that is not resident and the resident
+	// member is busy. "Wait" (default) holds the request open until the
+	// incumbent drains and the target loads, bounded by the pool's
+	// swapBudget. "IfIdle" only starts a swap when the incumbent is idle;
+	// while it is busy the pooled backend is skipped immediately and the
+	// request falls through to the next entry in Backends, so a rule such
+	// as [preferred-model, resident-model] serves on whichever is warm
+	// instead of queueing behind in-flight work. Non-pooled backends and
+	// same-model requests are unaffected.
+	// +kubebuilder:validation:Enum=Wait;IfIdle
+	// +kubebuilder:default=Wait
+	// +optional
+	PoolActivation string `json:"poolActivation,omitempty"`
 }
 
 // RouterPolicy holds cross-cutting controls applied to all rules.
--- a/charts/llmkube/templates/crds/modelrouters.yaml
+++ b/charts/llmkube/templates/crds/modelrouters.yaml
@@ -900,6 +900,23 @@ spec:
                             type: string
                           minItems: 1
                           type: array
+                        poolActivation:
+                          default: Wait
+                          description: |-
+                            PoolActivation decides what the proxy does when a backend in
+                            Backends is a ModelPool member that is not resident and the resident
+                            member is busy. "Wait" (default) holds the request open until the
+                            incumbent drains and the target loads, bounded by the pool's
+                            swapBudget. "IfIdle" only starts a swap when the incumbent is idle;
+                            while it is busy the pooled backend is skipped immediately and the
+                            request falls through to the next entry in Backends, so a rule such
+                            as [preferred-model, resident-model] serves on whichever is warm
+                            instead of queueing behind in-flight work. Non-pooled backends and
+                            same-model requests are unaffected.
+                          enum:
+                          - Wait
+                          - IfIdle
+                          type: string
                         strategy:
                           default: primary-fallback
                           description: Strategy selects how multiple backends are
--- a/config/crd/bases/inference.llmkube.dev_modelrouters.yaml
+++ b/config/crd/bases/inference.llmkube.dev_modelrouters.yaml
@@ -896,6 +896,23 @@ spec:
                             type: string
                           minItems: 1
                           type: array
+                        poolActivation:
+                          default: Wait
+                          description: |-
+                            PoolActivation decides what the proxy does when a backend in
+                            Backends is a ModelPool member that is not resident and the resident
+                            member is busy. "Wait" (default) holds the request open until the
+                            incumbent drains and the target loads, bounded by the pool's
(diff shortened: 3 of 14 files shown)


Defilan (approved):
Approving, I traced the contract end to end and tried to break it and could not. The IfIdle skip sits inside the no-swap-running branch, so an in-flight swap is still waited on in both modes, and the skip path unlocks cleanly with the deferred waiter cleanup doing its thing. Both behavioural tests bite: I reverted the IfIdle branch and they failed exactly as your body claims (hold budget exceeded instead of ErrIncumbentBusy, and a 5s hold instead of instant fallback), and reverting the translateRule copy failed the passthrough test too. A probe with an unknown mode string held like Wait, so the mode check is exact and fail-safe. make manifests and make chart-crds produced no drift on your head, and the full router/controller/metrics suite is green here.

Two things worth a look, neither blocking. When every backend in a rule skips, the client gets 502 "all backends failed: model pool incumbent busy", while the sibling ErrHoldBudgetExceeded gets 503 with Retry-After and a pool_activation_timeout audit reason. It is the same transient condition and a retry after the incumbent drains does serve, so I would give it the same treatment or at least its own audit reason. Also, issue #1786's body describes a different ask entirely, resources.memory emitted as limits and a suggested resources.memoryLimit field, under this title, filed that way with no edits. Merging this closes the issue with the memory ask unaddressed, so worth fixing the issue body or refiling the memory part first.

Metric is registered, the docs caveat about proxy-tracked traffic is right, and the disclosure checks out. Thanks, nice work.

PR #1792, opened 2026-09-11
fix(foreman): reap orphaned in-cluster FleetNodes when agent pods disappear (#1778)

### LINKED ISSUE #1778: [BUG] FleetNode CRs have no ownerReferences, so a pod replacement orphans them and alerts forever
## Bug Description

FleetNode CRs are created per foreman-agent pod but carry no `ownerReferences`, so Kubernetes never garbage-collects them. When the pod behind a FleetNode goes away — any rollout, eviction, or node reboot — the CR survives, its `lastHeartbeatTime` freezes, and `ForemanFleetNodeHeartbeatStale` fires critical forever. Nothing in the operator reaps it and no amount of restarting helps, because each restart mints a *new* FleetNode for the new pod and leaves the old one behind.

Triggered here by a Talos upgrade doing a rolling restart of every node.

## Steps to Reproduce

1. Run `foreman-agent` as a Deployment (`mode: Job`/`InProcess` both apply — the FleetNode is per agent pod).
2. Cause the pod to be replaced on a different node: roll the nodes (Talos/kubelet upgrade), drain, or bump the agent image.
3. `kubectl get fleetnodes` — the old pod's FleetNode is still present.
4. `ForemanFleetNodeHeartbeatStale` goes critical and stays there.

## Expected Behavior

A FleetNode should not outlive the agent pod it describes. Either set an `ownerReference` to the pod so the garbage collector removes it, or have the operator reap FleetNodes whose pod no longer exists.

## Actual Behavior

The CR is orphaned permanently:

```
$ kubectl get fleetnode foreman-llmkube-agent-5c9cffbcf7-2vk64 -o jsonpath='{.metadata.ownerReferences}'
             # empty

$ kubectl -n llm get pod foreman-llmkube-agent-5c9cffbcf7-2vk64
Error from server (NotFound): pods "foreman-llmkube-agent-5c9cffbcf7-2vk64" not found

$ kubectl -n llm get rs foreman-llmkube-agent-5c9cffbcf7
NAME                               DESIRED   CURRENT   READY   AGE
foreman-llmkube-agent-5c9cffbcf7   0         0         0       42h
```

Its status froze mid-drain and never recovered:

```json
{
  "phase": "NotReady",
  "kubernetesNode": "eula",
  "lastHeartbeatTime": "2026-09-08T02:13:25Z",
  "conditions": [{"type": "Ready", "status": "False",
                  "reason": "HeartbeatStale", "message": "no heartbeat within 1m30s"}]
}
```

Timeline matching the rolling upgrade (node `Ready` transitions):

| node | Ready since |
|---|---|
| ayaka | 2026-09-08T00:39:06Z |
| ganyu | 2026-09-08T01:05:56Z |
| skirk | 2026-09-08T01:19:44Z |
| eula  | 2026-09-08T02:40:44Z |

The orphan was created `01:05:55Z`, scheduled to `eula`, and stopped heartbeating at `02:13:25Z` as `eula` was drained ahead of its 02:40 reboot. The alert went critical at `03:07:05Z` and only cleared after deleting the CR by hand:

```
kubectl delete fleetnode foreman-llmkube-agent-5c9cffbcf7-2vk64
```

The three FleetNodes with live pods were `Ready` with current heartbeats throughout, so this is specific to the orphan rather than a fleet-wide heartbeat problem.

## Environment

**LLMKube Version:** foreman chart/agent `0.9.25` (`agentVersion: v0.9.25`)

**Kubernetes Version:** v1.37.0

**Cluster Type:**
- [x] Other: Talos v1.14.0 (bare metal, 4 nodes)

**GPU (if applicable):**
- [x] Other: mixed NVIDIA + AMD Strix Halo (not relevant to this bug)

## Additional Context

The annotation on the alert that fires for this is separately broken — it interpolates `{{ $labels.node }}`, which does not exist on the metric, so the alert cannot name the offending FleetNode. Filed separately.

srosignoli:
## What

Reap orphaned in-cluster `FleetNode` Custom Resources once their heartbeat has been silent past `FleetNodeNotReadyReapTimeout` (10 minutes), preventing `ForemanFleetNodeHeartbeatStale` alerts from firing indefinitely after agent pods terminate or roll over.

## Why

Fixes #1778

`FleetNode` CRs are created per `foreman-agent` pod. Because `FleetNode` is cluster-scoped (`scope=Cluster`) while `Pod` is namespaced, a `FleetNode` cannot have an `ownerReference` pointing to its backing pod (disallowed by Kubernetes cross-scope garbage collection rules).

When an in-cluster agent pod terminates abruptly (node reboot, rolling upgrade, eviction, or crash) without completing a graceful drain, its `FleetNode` CR is orphaned in `NotReady` with a frozen `lastHeartbeatTime`. The replacement pod mints a new `FleetNode` with a new random pod name suffix, leaving the old CR behind indefinitely. PR #980 introduced reaping for `Draining` nodes, but nodes that never entered `Draining` were requeued forever, causing `ForemanFleetNodeHeartbeatStale` to alert indefinitely.

## How

1. **`api/foreman/v1alpha1/fleetnode_types.go`**:
       - Added `FleetNodeNotReadyReapTimeout = 10 * time.Minute`.
       - Added `NotReadyReapable(now time.Time) bool`.
       - Explicitly gates reaping on `n.Status.KubernetesNode != ""` so persistent off-cluster workers (metal Macs) that leave `kubernetesNode` empty are never deleted during extended offline or sleeping periods.
2. **`internal/foreman/controller/fleetnode_controller.go`**:
       - Consolidated deletion logic into a shared `reapNode` helper that enforces `UID` and `ResourceVersion` preconditions (closing the revival race if a node recovers while the reconciler cache lags).
       - Reaps an in-cluster `FleetNode` when `(node.Status.Phase == NotReady || stale) && node.NotReadyReapable(now)` is true.
3. **Tests**:
       - Added table-driven unit tests in `fleetnode_types_test.go` covering in-cluster stale, fresh, nil-heartbeat, and off-cluster preservation.
       - Added Ginkgo envtest cases in `fleetnode_controller_test.go` verifying that orphaned in-cluster `NotReady` nodes are deleted, off-cluster nodes are preserved, and briefly stale nodes are left alone.

## Checklist
    - [x] Tests added/updated
    - [ ] Documentation updated (no user-facing doc change; internal controller reap behavior)

 ---
Assisted-by: Antigravity (collaborated on root cause analysis, implementation, and TDD regression tests; reviewed and ran `make test`, `make lint`, and `make lint-all`)._


joryirving (changes requested):
The reap itself is right — I checked the CRD and FleetNode really is cluster-scoped while Pod is namespaced, so ownerReference GC is off the table, and the `KubernetesNode != ""` gate matches how the Registrar stamps identity (pod-name identity in-cluster, downward-API machine name empty off-cluster). reapNode preserves the precondition semantics verbatim.

One change I'd want before merge: the reap keys on silence alone, which can't tell "pod gone" from "pod alive but silent" (asleep Mac worker while the control-plane host stays up, apiserver blip past the window, wedged process). The agent Upserts only at startup and a heartbeat NotFound just logs "will retry", so a live agent whose node gets reaped never re-registers — and with the series gone, even the stale-heartbeat alert goes quiet, which is worse visibility than today's NotReady. A heartbeat NotFound → Upsert fallback on the agent side would close it and fits the retry philosophy Run already has.

Small note: no envtest for the stale-Ready disjunct of the controller condition, which is the widest gate.

Happy to approve once the re-register path lands, thanks.

Defilan (commented):
Requesting changes, seconding Jory, and I verified his review holds point by point. The agent Upserts exactly once, at startup (cmd/foreman-agent/main.go:453), and Run's heartbeat loop answers every failure including NotFound with "heartbeat patch failed; will retry" (pkg/foreman/agent/fleetnode.go:355), never re-registering. So an in-cluster pod that stays alive but silent past the ten minute window (node partition with the control plane up, or a wedged process) gets reaped, and when connectivity returns the agent heartbeats into NotFound forever: alive and serving but invisible to the scheduler, and no restart is coming. The visibility claim holds too: the heartbeat metric is kube-state-metrics derived from the FleetNode objects themselves (crs-configmap.yaml:185), so deleting the CR removes the series, there is no absent() companion, and ForemanFleetNodeHeartbeatStale goes quiet instead of firing. The agent-side re-register on NotFound is the right unblock and fits Run's retry philosophy; it would also rescue the pre-existing case where a sleeping Mac is reaped mid-drain and wakes to a missing node.

Everything else I checked holds up: both new tests bite (disabling the reap condition failed the orphaned-in-cluster spec, removing the kubernetesNode gate failed the off-cluster preservation spec), FleetNode is scope=Cluster so the ownerReference reasoning is right, reapNode carries the #980 preconditions over intact, and I probed the stale-Ready disjunct Jory flagged: it reaps correctly today but deserves a test alongside the re-register fix.

The pod-gone case this PR targets is real and well tested; the alive-but-silent case needs the re-register path first.

srosignoli:
Addressed in https://github.com/defilantech/LLMKube/pull/1792/commits/ac0708e6a1ff6ed63e3df4f1ca26cfc6ddd889bf:
  1. Agent Re-Registration on NotFound: In pkg/foreman/agent/fleetnode.go https://github.com/defilantech/LLMKube/pull/1792/files#diff-7d2fb7306ea6a6c0b3c675fbbfef30c499eeef043bb759a22d9c159ea3bbdb73, Run() now intercepts apierrors.IsNotFound(err) on heartbeat failure and immediately triggers r.Upsert(ctx) followed by PatchHeartbeat(Ready) to restore the node and its live supervision budget. Covered by TestRegistrar_Run_ReRegistersWhenFleetNodeDeleted in pkg/foreman/agent/fleetnode_test.go.
 2. Stale-Ready Disjunct Test: Added envtest spec reaps an in-cluster node left in Ready whose agent heartbeat is long stale in internal/foreman/controller/fleetnode_controller_test.go, bringing reconciler statement coverage to 96.3%.

joryirving (approved):
The fallback closes it — Upsert-on-NotFound in Run covers the alive-but-silent cases (sleep, partition, wedged process), retrying on re-registration failure keeps the loop self-healing, and it quietly defuses the launchd revival-race brick the precondition comment warns about, since a stable-named agent that loses its node now re-registers within a tick instead of heartbeating into NotFound forever. Ran the agent, api, and controller suites locally including the two new tests; all green.

Approving, thanks.

Defilan (approved):
Thanks for your help with this one! Great work, merging now.

PR #1794, opened 2026-09-11
fix(rollout): do not defer a scale-from-zero rollout as PodsBusy

### LINKED ISSUE #1793: [BUG] waitForIdle deadlocks a scale-from-zero rollout: zero endpoints reads as PodsBusy until idleTimeoutSeconds
## Bug Description

`rolloutPolicy.waitForIdle: true` deadlocks a scale-from-zero. When an InferenceService whose replicas are 0 gets a new pod template and is then scaled back to 1 (the normal ModelPool swap path, or any Renovate/Flux change that lands while a member is `Stopped`), `reconcileRolloutPolicy` defers the rollout with `RolloutDeferred=True / PodsBusy` and keeps deferring until `idleTimeoutSeconds` (default 5m; 24h in my case) even though there is nothing to drain.

Root cause in `internal/controller/drain_before_rollout.go`: `checkServiceIdle` lists the Service's EndpointSlices, gets a slice with zero endpoints, builds an empty `replicaURLs` and returns `(false, nil)`:

```go
replicaURLs := collectReadyReplicaURLs(slices, port)
replicaURLs = append(replicaURLs, collectNotReadyReplicaURLs(slices, port)...)
if len(replicaURLs) == 0 {
    return false, nil
}
```

`reconcileRolloutPolicy` reads "not idle, no error" as `ReasonPodsBusy` ("Backend slots are busy, deferring rollout"). `countOldPods` has already computed `totalPods == 0` a few lines earlier, but nothing consults it before gating.

## Steps to Reproduce

1. Two InferenceServices in a `ModelPool` (sticky), both with `rolloutPolicy: {waitForIdle: true, idleTimeoutSeconds: 86400}`. Member B is the non-resident one, held at `replicas: 0` / `Stopped` by the pool.
2. Change B's pod template while it is Stopped (I added `sourceSecretRef` to its Model; any image or env change does the same).
3. Send a request for B through the ModelRouter. The pool drains A, scales A to 0 and B to 1.
4. Observe B's Deployment stays at `replicas: 0`, B's InferenceService sits at `Creating` with `RolloutDeferred=True / PodsBusy`, the pool stays `Swapping`, and the GPU is idle with nothing resident. Nothing recovers until the idle timeout.

## Expected Behavior

A rollout with zero existing pods has no in-flight work to protect, so the drain gate should not apply. Either `reconcileRolloutPolicy` returns immediately when `totalPods == 0`, or `checkServiceIdle` treats an endpoint list with zero addresses as idle. (The `len(slices.Items) == 0` fallback already probes the Service URL, which will also fail on a zero-replica Service, so the same consideration applies there.)

## Actual Behavior

Controller log, repeating every reconcile:

```
INFO Backend slots are busy, deferring rollout  {"controller": "inferenceservice", "InferenceService": {"name":"<member-b>"}}
INFO Deferring rollout, backend slots are busy  {"controller": "inferenceservice", "InferenceService": {"name":"<member-b>"}}
```

`kubectl get isvc`: `<member-b>  Creating  replicas=1`; its Deployment: `desired 0`. Conditions: `Available=False/ManuallyScaledToZero`, `RolloutDeferred=True/PodsBusy`, `Progressing=True/Creating`. ModelPool: `Phase=Swapping`, `SlotAllocated=False`.

## Environment

**LLMKube Version:** controller `ghcr.io/defilantech/llmkube-controller:0.9.25`, chart 0.9.25. The two files involved are unchanged on `main` since v0.9.25.

**Kubernetes Version:** v1.37.0 (Talos Linux 1.14)

**Cluster Type:**
- [x] Other: Talos, bare metal

**GPU (if applicable):**
- [x] Other: NVIDIA RTX 3090 (single card, `nvidia.com/gpu: 1`)

## YAML Manifests

Member (both members are shaped like this; `replicas` is left unset so the pool owns it):

```yaml
kind: InferenceService
metadata:
  name: member-b
spec:
  modelRef: member-b
  runtime: llamacpp
  rolloutPolicy:
    waitForIdle: true
    idleTimeoutSeconds: 86400
  priorityClassName: gpu-preemptible
  runtimeClassName: nvidia
  resources: { gpu: 1, cpu: "500m", memory: 64Gi }
```

Pool:

```yaml
kind: ModelPool
metadata:
  name: shared-slot
spec:
  gpu: 1
  swapPolicy: sticky
  swapBudget: 600s
  default: member-a
  members:
    - inferenceServiceRef: { name: member-a }
    - inferenceServiceRef: { name: member-b }
```

## Additional Context

Setting `waitForIdle: false` on the members avoids the deadlock but gives up the protection the field exists for (a Renovate image bump rolling the resident member mid-request), so it is not an acceptable workaround for a pool. Patching `waitForIdle: false` in-cluster once, letting the pod come up, then letting GitOps restore `true` is what I am doing today.

Happy to open the PR: early return in `reconcileRolloutPolicy` when `totalPods == 0` (clearing any stale `RolloutDeferred`), plus a regression test in the drain suite that scales a service from zero with a changed template and asserts the rollout proceeds.

joryirving:
## What

A rollout of an InferenceService that currently has no pods no longer waits behind `rolloutPolicy.waitForIdle`. `checkServiceIdle` returns a sentinel when the Service's EndpointSlices carry no addresses, and `reconcileRolloutPolicy` resolves it with the pod count it already computes: zero pods means nothing to drain and the rollout proceeds; pods present but not yet addressable keep today's fail-closed `PodsBusy` deferral.

## Why

Fixes #1793

A service scaled to zero whose template changed while it was down (the normal ModelPool path: a change lands while a member is `Stopped`, then the pool scales it back up) was held at `RolloutDeferred=True / PodsBusy` until `idleTimeoutSeconds` because an empty endpoint list read as "not idle, no error". The pool sat `Swapping` with nothing resident on the GPU.

## How

- `checkServiceIdle`: `return false, errNoEndpoints` instead of `return false, nil` when the replica URL list is empty. The existing comment about not being able to distinguish scale-to-zero from unready-but-working is kept, with the resolution moved to the caller.
- `reconcileRolloutPolicy`: if the error is `errNoEndpoints`, proceed when `countOldPods` reports zero pods, otherwise fall back to the same `PodsBusy` path as before. No other branch changes.
- Tests, `drain_before_roll_test.go`, new `scale-from-zero` context: an EndpointSlice with no endpoints plus a changed template rolls out immediately with the idle server wired busy (so the Service-URL fallback cannot be what satisfied it); the same setup with a Ready old-generation pod present still defers with `PodsBusy`. Both fail on `main` and pass with the change. Existing busy-defer and template-gate specs unchanged and green.

Assisted-by: Claude Code (Fable 5.1) generated the implementation and tests from my diagnosis on a live cluster; I reviewed the diff and ran `make test` (controller 86.5%) and `make lint-all` locally, all green.

## Checklist

- [x] Tests added/updated
- [ ] Documentation updated (if user-facing change) — behaviour fix, no doc surface changed

### DIFF
--- a/internal/controller/drain_before_roll_test.go
+++ b/internal/controller/drain_before_roll_test.go
@@ -981,6 +981,89 @@ var _ = Describe("RolloutPolicy drain-before-rollout", func() {
 			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: isvcName, Namespace: "default"}, dep)).To(Succeed())
 			Expect(dep.Spec.Template.Spec.Containers[0].Image).To(Equal("ghcr.io/ggml-org/llama.cpp:server"))
 		})
+
+		It("should proceed when there is no EndpointSlice at all and no pods", func() {
+			modelName := "model-no-slice"
+			isvcName := "isvc-no-slice"
+			model, isvc, reconciler := setupChangedTemplate(modelName, isvcName)
+			defer func() { _ = k8sClient.Delete(ctx, model) }()
+			defer func() { _ = k8sClient.Delete(ctx, isvc) }()
+
+			// No EndpointSlice is created at all: the sibling of the empty-slice
+			// case, reached when upstream's endpointslice reconciler deletes the
+			// slice after its last endpoint is removed. Point the Service-URL
+			// fallback at a closed port so the probe fails exactly as it does
+			// against a zero-replica Service with nothing behind it.
+			dead := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {}))
+			deadURL := dead.URL
+			dead.Close()
+			reconciler.RolloutIdleBaseURL = deadURL
+
+			result, err := reconciler.Reconcile(ctx, reconcile.Request{
+				NamespacedName: types.NamespacedName{Name: isvcName, Namespace: "default"},
+			})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result.RequeueAfter).To(BeZero())
+
+			dep := &appsv1.Deployment{}
+			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: isvcName, Namespace: "default"}, dep)).To(Succeed())
+			Expect(dep.Spec.Template.Spec.Containers[0].Image).To(Equal("ghcr.io/ggml-org/llama.cpp:server-v2"),
+				"a service with no EndpointSlice and no pods must roll out immediately, not defer via the URL fallback")
+
+			final := &inferencev1alpha1.InferenceService{}
+			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: isvcName, Namespace: "default"}, final)).To(Succeed())
+			Expect(findRolloutDeferredCondition(final.Status.Conditions)).To(BeNil())
+		})
+
+		It("should still defer when there is no EndpointSlice at all but a pod exists", func() {
+			modelName := "model-no-slice-pod"
+			isvcName := "isvc-no-slice-pod"
+			model, isvc, reconciler := setupChangedTemplate(modelName, isvcName)
+			defer func() { _ = k8sClient.Delete(ctx, model) }()
+			defer func() { _ = k8sClient.Delete(ctx, isvc) }()
+
+			dead := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {}))
+			deadURL := dead.URL
+			dead.Close()
+			reconciler.RolloutIdleBaseURL = deadURL
+
+			// A Ready old-generation pod the (now absent) endpoint list has not
+			// caught up with may still hold in-flight work, so the gate must
+			// stay closed even though the Service-URL probe cannot answer.
+			pod := &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      isvcName + "-old",
+					Namespace: "default",
+					Labels: map[string]string{
+						"app":                           isvcName,
+						"inference.llmkube.dev/service": isvcName,
+					},
+				},
+				Spec: corev1.PodSpec{Containers: []corev1.Container{{Name: "llama-server", Image: "ghcr.io/ggml-org/llama.cpp:server"}}},
+			}
+			Expect(k8sClient.Create(ctx, pod)).To(Succeed())
+			defer func() { _ = k8sClient.Delete(ctx, pod) }()
+			pod.Status.Conditions = []corev1.PodCondition{{Type: corev1.PodReady, Status: corev1.ConditionTrue}}
+			Expect(k8sClient.Status().Update(ctx, pod)).To(Succeed())
+
+			result, err := reconciler.Reconcile(ctx, reconcile.Request{
+				NamespacedName: types.NamespacedName{Name: isvcName, Namespace: "default"},
+			})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result.RequeueAfter).To(BeNumerically(">", 0))
+
+			deferred := &inferencev1alpha1.InferenceService{}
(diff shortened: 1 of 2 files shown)


Defilan (commented):
Requesting changes, though it's close: the empty-EndpointSlice state you hit is fixed and properly tested, but the sibling state where the slice is deleted entirely still deadlocks, so I don't think Fixes #1793 holds yet.

What I checked and what held: I reverted the errNoEndpoints hunk and "should proceed when the service has no endpoints and no pods" failed, then passed with the change back, so the test bites. The pod-present test keeps the fail-closed deferral, and the pod_lifetime.go call site is behaviorally unchanged since it fails closed on any error. Full controller suite green at 86.5% coverage, vet and golangci-lint clean. One correction to the writeup: only the proceed test fails on main; the defer test passes there too.

The gap: checkServiceIdle has two empty-endpoint paths and the PR fixes one. When len(slices.Items) == 0 it falls back to probing the Service URL, which fails on a zero-replica Service, so the rollout still defers until idleTimeoutSeconds with reason IdleCheckFailed. Your issue text calls this exact path out. Upstream's endpointslice reconciler marks a slice for deletion the moment its last endpoint is removed, and an already-empty slice is never deleted afterward, so which state a cluster settles in depends on whether that delete lands. Your cluster kept the empty slice; a cluster where the delete lands gets the deadlock back through the fallback. I confirmed it on the PR branch with a spec that has no EndpointSlice at all: zero pods, changed template, and the reconcile deferred with requeueAfter=5s.

The symmetric fix is small: in the fallback branch, return errNoEndpoints when the probe cannot answer, and your existing caller-side pod-count resolution handles both paths. Happy to approve once that's in, or Refs #1793 if you'd rather split it.

Nit: countOldPods swallows a pod List error and returns 0, and this PR promotes that zero from a reason label to a gate. A transient list failure in exactly that state would roll over a pod that may hold work.

joryirving:
Pushed the symmetric fix. The no-EndpointSlice fallback now returns errNoEndpoints when the probe can't reach a backend, so your caller-side pod-count resolution covers both paths: it proceeds when there are no pods and stays deferred (fail-closed) when one exists. I kept errIdleUnsupported flowing through separately, so a runtime that genuinely can't be idle-checked still defers with IdleCheckUnsupported instead of being mistaken for an empty endpoint list.

Also addressed the nit: countOldPods now propagates its pod List error and the caller fails closed on it, since this PR promotes that zero count from a reason label to a gate and a transient list failure shouldn't roll a pod that may still hold work.

On tests: added the no-slice specs (zero pods proceeds, a Ready pod defers), and repurposed the old Service-URL fallback spec to assert the proceed now that it no longer deadlocks. IdleCheckFailed stays covered by the vLLM replica spec, and fail-closed-with-a-pod by the new specs. make test, make lint, and make manifests generate are all clean. Kept Fixes #1793. Mind another look?

Defilan (approved):
Approving, the second commit closes the fallback path the way the first pass asked, and the countErr rail is exactly the right shape.

What I did on the new head: the two new specs pass, and both bite. Reverting the fallback sentinel to the raw error fails both: the no-slice-no-pods case stops proceeding, and the pod-present case flips to IdleCheckFailed instead of the asserted PodsBusy. The countErr guard is real coverage, not decoration: I probed it with a pod List failure plus no endpoints and a dead fallback URL, it defers fail-closed, and removing the guard fails the probe with "a failed pod list must defer, not roll". The inverted envtest spec is a deliberate, documented behavior change and the fail-closed coverage survives where it matters: pod-present through the new spec, published-replica-unreachable through the vLLM spec, which still asserts IdleCheckFailed. Full controller suite green, golangci-lint clean.

Two small things. The body's How section still says "No other branch changes" and "Both fail on main"; the second commit changes another branch and only the proceed test fails on main, so a one-line body update would keep it honest. And a question: an unannotated generic runtime now proceeds from the empty-slice path but defers until timeout from the no-slice path for the same zero-pod state, since the preserved errIdleUnsupported only fires through the fallback. Deliberate? Either answer is fine, the deferring side is the fail-closed one.

Thanks, merge.

PR #1796, opened 2026-09-11
feat(controller): ModelPool reclaim policy returns the slot to spec.default after idle

### LINKED ISSUE #1795: [FEATURE] ModelPool: reclaim the slot to spec.default after the on-demand member goes idle
## Feature Description

Add an optional pool policy that returns the shared GPU slot to a designated background member after the current on-demand member has been idle for a configurable period. Today a `ModelPool` is sticky and purely request-driven: once a member becomes resident it stays resident until a request for a *different* member arrives. With the new `poolActivation: IfIdle` (#1787, thanks), an interactive workload rides whichever member is warm and never forces a swap, so nothing ever brings the pool back to a preferred default once an on-demand member has taken the slot.

## Problem Statement

> As an operator sharing one GPU between a background model and an on-demand model, I want the pool to reclaim the slot for the background model after the on-demand model goes idle, so that most traffic is served by the preferred default without any workload having to force a disruptive swap.

Concrete setup, two members on one 24 GB GPU (mutually exclusive, they cannot co-reside):

- `background` — the preferred default for most traffic, `spec.default`.
- `on-demand` — a heavier model a batch/coding workload bursts against.

Desired behaviour: background is resident at rest; the on-demand workload's requests swap it in for the duration of a burst; interactive traffic rides whatever is warm during the burst (IfIdle) instead of fighting it; once the on-demand model has been idle a while, the slot returns to background on its own.

What actually happens: after the first on-demand burst, the on-demand member stays resident. Because IfIdle serves the *warm* member even when it is idle, interactive requests keep riding the idle on-demand model indefinitely, and the background model is only ever reloaded on a cold pool or by an explicit forced request for it (which reintroduces the swap thrash IfIdle was meant to remove). There is no request that naturally returns the pool to the default. The reconciler also never force-unloads a resident on a timer (by design), so nothing reclaims the slot.

## Proposed Solution

A pool-level reclaim delay. When the resident member is not `spec.default` and has been idle for at least the configured interval, the controller swaps the slot back to `spec.default`, using the same drain/idle contract as any other swap. Off by default, so existing pools are unchanged.

```yaml
kind: ModelPool
spec:
  default: background
  # 0 (default) preserves today's sticky behaviour.
  reclaimToDefaultAfterSeconds: 600
  members:
    - inferenceServiceRef: { name: background }
    - inferenceServiceRef: { name: on-demand }
```

Implementation reuses what already exists: `memberIdle` (the `/slots` probe) to gate the reclaim, and the activator/controller swap machinery to perform it. The reconciler requeues on the reclaim interval while a non-default member is resident.

### Design hazard worth documenting

The reclaim delay must be longer than the on-demand workload's inter-burst gap. If it fires sooner, the pool reclaims the default (a model load), the on-demand workload immediately asks again (another load), and you have traded request-side coalescing for load-side thrash, which is worse. So:

- Default off (`0`).
- Recommend a delay comfortably above the on-demand tenant's typical idle gap in the docs.
- Reclaim only after *sustained* idle, and never reclaim if a swap into the current resident completed within the reclaim window (hysteresis), so a just-loaded model is not immediately evicted.

## Alternatives Considered

- **Short `idleTimeoutSeconds` scale-to-zero on the on-demand member.** Pool members have no timer-based unload (deliberate), and this would cold-start rather than swap; it also does not choose *which* member the slot returns to.
- **Interactive workload forces the default in `Wait` mode.** Reintroduces the swap thrash against the on-demand workload that IfIdle exists to prevent.
- **Rely on frequent default-model traffic to reclaim naturally.** Does not work: IfIdle rides the warm idle member, so default-model requests are served on the on-demand model and never force the default back.

## Additional Context

- Related: #1787 (poolActivation IfIdle, merged) is the complementary half. IfIdle stops an interactive workload from preempting a *busy* on-demand member; this feature returns the pool to the background default once the burst subsides. Together they give background-default + on-demand-burst + elastic ride-along.
- Workaround today: none that keeps the default warm; either accept the pool drifting to the on-demand member, or force swaps and eat the thrash.

## Priority

- [ ] Medium - Nice to have
- [ ] Low - Minor improvement

## Willingness to Contribute

- [x] Yes, I can submit a PR
- [ ] Yes, I can help test
- [ ] No, just requesting

---

_Filed with AI assistance (Claude), reviewed by me before submission, per CONTRIBUTING.md._

joryirving:
## What

Add `swapPolicy: reclaim` to `ModelPool`. It behaves like `sticky` for cross-model demand and additionally returns the shared slot to `spec.default` once the resident non-default member has been continuously idle for `spec.reclaimAfter` (default 300s).

## Why

Fixes #1795.

A sticky pool never brings the slot back to a preferred model on its own. With a background default sharing one GPU with an on-demand member, the first on-demand burst takes the slot and keeps it: the on-demand member stays resident and idle, and with the new `poolActivation: IfIdle` (#1787) an interactive workload rides that idle on-demand model indefinitely. There is no request that returns the pool to the default. IfIdle is the complementary half (don't preempt a *busy* incumbent); this is the other half (return to the default once the burst subsides).

## How

- New `ModelPoolSwapPolicyReclaim` enum value and `spec.reclaimAfter` (`*metav1.Duration`, default 300s).
- `applyReclaimOwner` runs right after `resolveSlotOwner`. When the pool is settled on a non-default resident that is Ready and idle, it starts an idle clock in `status.residentIdleSince`; once idle for `reclaimAfter` it overrides the desired owner with `spec.default`, and the existing drain-and-swap machinery drains the idle resident and loads the default. It reuses `memberIdle` (the `/slots` probe) and is fail-closed: a busy resident or an unestablished idle state resets the clock and never reclaims.
- The clock resets on every served request, so a steady drip of on-demand work keeps the on-demand member resident; the slot only returns after a genuine quiet period. The reconciler polls idle state (capped 30s) because no InferenceService/Pod watch fires on a busy->idle transition. `residentIdleSince` self-clears once the default is resident.
- `llmkube_modelpool_reclaims_total{namespace,pool,from,to}`.

### Design note carried into the docs

`reclaimAfter` must be longer than the on-demand member's inter-burst gap. Shorter than that gap, the pool reclaims the default the moment a burst pauses and swaps it straight back on the next request — two cold loads where sticky would do none. The docs say to size it to outlast a lull in on-demand traffic, not the gap between adjacent requests. Default off-ish at 300s and opt-in via `swapPolicy: reclaim`, so existing pools are unchanged.

## Checklist

- [x] Tests added/updated (unit tests for every `applyReclaimOwner` branch)
- [x] `make test` passes locally (controller envtest + metrics + api green)
- [x] AI assistance disclosed (Claude, reviewed by me)
- [x] Documentation updated (docs/operations/gpu-sharing.md)


Defilan (commented):
Requesting changes: the reclaim never actually fires, because the idle clock is never persisted, so as shipped this is a no-op behind an opt-in flag.

I drove a full reclaim swap in envtest (judge as default, coder resident and idle, reclaimAfter 50ms, reconciles run well past the window) and the on-demand member was never drained: coder stayed at replicas 1. The mechanism: applyReclaimOwner sets status.ResidentIdleSince on the same object updateStatus then diffs, and updateStatus captures its MergeFrom baseline after that mutation has already happened, so the clock never appears in the patch. I confirmed it directly: set the clock via applyReclaimOwner, ran updateStatus, re-read the pool, ResidentIdleSince is nil. Every reconcile restarts the clock, elapsed can never reach reclaimAfter, and llmkube_modelpool_reclaims_total stays at zero. The unit tests can't see this because they call applyReclaimOwner directly and never round-trip updateStatus, and no integration test drives a reclaim swap.

The fix is small and I verified it: capture the status baseline before applyReclaimOwner runs and use it as the patch baseline in updateStatus. With that one change my probe converges: reclaim fires, coder drains, judge loads, the resident flips, and further reconciles stay put.

Everything else held up: build, vet and golangci-lint clean, controller, metrics and api suites green, manifests, deepcopy and chart CRD regeneration show zero drift, sticky pools are untouched, and the label set on the new metric is right for a controller-emitted counter.

Two smaller things: the docs say the clock "resets on every request the resident serves", but only a request in flight during one of the 30s polls resets it, so short requests between polls don't. That weakens exactly the thrash guard the section describes, and the sizing advice should account for it. And "unit tests for every applyReclaimOwner branch" is generous: the default-empty, resident-not-Ready, and default-not-a-member branches have no tests. With the persistence fix and an integration test, happy to approve.

joryirving:
Good catch, and thanks for driving it in envtest. Fixed it the way you verified: Reconcile now captures the status baseline before applyReclaimOwner mutates ResidentIdleSince and passes it to updateStatus as the MergeFrom baseline, so the idle clock persists instead of being diffed out of the patch. With that the clock survives across reconciles, the reclaim fires, the idle resident drains, the slot returns to the default, and llmkube_modelpool_reclaims_total increments.

Added the integration test that was missing: it drives a full reclaim swap through updateStatus (idle clock persists across reconciles, reclaim fires and bumps the counter, the resident flips coder to judge, and further reconciles stay put). I confirmed it fails against the pre-fix code and passes with the fix, so it actually guards the round-trip the unit tests couldn't reach.

Also filled the three untested applyReclaimOwner reset branches (empty spec.default, resident not Ready, default not a member), and corrected the docs: the idle clock resets only when an idle poll catches the resident serving, and idleness is sampled on a fixed interval (~30s cap), so a request that starts and finishes between two polls is never observed and doesn't reset it. Sizing advice updated to match.

make test, make lint, and make manifests generate are all clean with zero drift. Kept Fixes #1795. Another look when you get a chance?

Defilan (approved):
Approving, the fix lands exactly the way I validated it in the first pass and every finding is addressed.

What I did on the new head: I reverted the baseline capture to the buggy order and the new integration test fails with exactly "the idle clock must persist across reconciles", then passes with the change, and it covers the whole arc: clock start, the reclaim metric going up by one, the idle resident drained, the default resident again, further reconciles staying put, and the clock self-clearing. The three guard branches I called untested now have unit tests (empty default, resident not Ready, default not a member), all green. The docs now say the clock resets only when a poll catches the resident serving and that sub-poll requests can slip between samples, which matches what the code does. Controller, metrics and api suites green, golangci-lint clean on the package, and the CRD sync check is green on this head.

The metric assertion in the integration test is a nice touch beyond what I asked for.

Thanks for the quick turnaround, merge.

PR #1797, opened 2026-09-11
fix(router): return 503 + Retry-After when every IfIdle backend skips busy

joryirving:
## What

When every backend in a `poolActivation: IfIdle` rule skips because its `ModelPool` incumbent is busy (no preferred member is warm), return `503` + `Retry-After` with the audit reason `pool_incumbent_busy`, instead of the generic `502 all_backends_failed`.

## Why

Raised in review of #1787: the all-skip IfIdle path returned 502, while the sibling case (a swap that outruns `swapBudget`) returns 503 + Retry-After with a `pool_activation_timeout` audit reason. Both are the same transient, retryable condition — the incumbent drains and a retry serves — so they should look the same to clients. A 502 reads as an upstream outage and does not carry Retry-After, so well-behaved clients do not back off deterministically.

## How

`dispatchWithFallback` tracks `allIncumbentBusy`: it stays true only if every backend that was tried failed specifically with `ErrIncumbentBusy`. Any other failure (unconfigured, unhealthy, dispatch error, 5xx) clears it. When it holds, the function returns the `ErrIncumbentBusy` sentinel unwrapped so the handler `errors.Is` it and maps it to 503 + Retry-After with the `pool_incumbent_busy` audit reason, mirroring the `ErrHoldBudgetExceeded` path. Mixed failures still surface exactly as before (502 / fail-closed).

The reclaim check sits before the fail-closed branch: for a fail-closed rule whose backends are merely busy (not unhealthy), the 503 "incumbent busy" reason is more accurate than "all backends unhealthy" and is still a refusal to spill (no egress).

## Checklist

- [x] Tests added/updated (a third member holds the slot busy, both rule backends skip -> 503 + Retry-After, no dispatch, no swap)
- [x] `make test` passes locally (router suite green)
- [x] AI assistance disclosed (Claude, reviewed by me)
- [x] Documentation updated (docs/operations/gpu-sharing.md)

### DIFF
--- a/docs/operations/gpu-sharing.md
+++ b/docs/operations/gpu-sharing.md
@@ -228,6 +228,14 @@ member's Service directly, bypassing the router, is invisible to this check
 pool traffic through the router if you rely on `IfIdle`. Skips are counted in
 `llmkube_modelpool_busy_skips_total{router,pool,member}`.
 
+If *every* backend in an `IfIdle` rule skips because no preferred member is warm
+(the slot is held busy by a member outside the rule), the request cannot be
+served right now. The proxy returns `503` with a `Retry-After` header and the
+audit reason `pool_incumbent_busy`, the same retryable signal as a swap that
+outruns `swapBudget`, so a client that retries after the incumbent drains is
+served. It is deliberately not a `502`: nothing is broken upstream, the slot is
+just occupied.
+
 ### swapBudget, and why it is separate from the request timeout
 
 `swapBudget` (default `300s`) bounds how long the router holds a cross-model
--- a/internal/router/proxy.go
+++ b/internal/router/proxy.go
@@ -192,6 +192,20 @@ func (p *Proxy) handleChatCompletions(w http.ResponseWriter, r *http.Request) {
 				"pool_activation_timeout", elapsed)
 			return
 		}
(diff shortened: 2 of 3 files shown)


Defilan (approved):
Approving, the symmetry with the hold-budget path checks out and the fix is properly guarded.

What I did: ran the router suite (green, 121s), then reverted the new handler branch and watched TestProxyPoolIfIdleAllBusyReturns503 fail with exactly "status = 502, want 503", so the test bites. I also probed the closest input that must not get the new treatment: one pooled backend busy-skips while a non-pooled backend in the same rule is dead. That stays a 502 with no Retry-After, so the allIncumbentBusy flag clears on real failures the way the description promises. The new branch mirrors ErrHoldBudgetExceeded line for line (same Retry-After constant, same audit shape), and nothing in config validation forbids the fail-closed plus IfIdle combination, so the ordering claim about fail-closed rules traces too. Docs read clean.

One nit, not yours: golangci-lint on the branch reports one pre-existing gosec G118 at activator.go:301, the swapCtx cancel from #1787. So "make lint: 0 issues" is not quite what I see, but that file is untouched here and the finding belongs to the earlier PR.

Good fix, thanks. Merge.

PR #1799, opened 2026-09-11
fix(foreman): drain tasks before agent exit

### LINKED ISSUE #1438: [FEATURE] Drain-before-roll for foreman-agent: in-process tasks are killed on any pod-template change

joryirving:
## What

Drain in-flight foreman-agent tasks on SIGTERM before the agent process exits, and raise the chart default termination grace period to five minutes.

## Why

`foreman-agent` previously returned as soon as Kubernetes delivered SIGTERM, cancelling in-process turns and leaving their tasks for claim-expiry recovery. This makes ordinary agent pod-template rollouts lose active work.

Fixes #1438

## How

The watcher stops accepting new claims as soon as its context is cancelled, including candidates already enumerated by the current poll pass. It waits for previously launched executors, whose execution and terminal-status contexts are detached from the shutdown signal while the task-deletion watchdog remains active. The Helm default now gives that drain a 300-second window. Focused regression tests cover completion during drain, task deletion during drain, and both claim-cancellation boundaries.

AI assistance: Codex (GPT-5) generated and revised the implementation and regression tests; I reviewed the diff and verified `make fmt`, `make vet`, `make lint`, `make test`, `make lint-all`, targeted race tests, and chart tests locally.

## Checklist

- [x] Tests added/updated
- [x] AI assistance is disclosed above, per [CONTRIBUTING.md](../CONTRIBUTING.md#ai-assisted-and-agent-contributions)
- [x] Documentation updated (user-facing chart value documentation)

### DIFF
--- a/charts/foreman/templates/agent-deployment.yaml
+++ b/charts/foreman/templates/agent-deployment.yaml
@@ -62,7 +62,7 @@ spec:
   {{- if and (ne $st "Recreate") (ne $st "RollingUpdate") }}
   {{- fail (printf "agents.%s.strategy.type: %q is not a valid Deployment strategy; use \"Recreate\" or \"RollingUpdate\" (the API server rejects anything else, including case variants like \"rollingUpdate\")" $name $st) }}
   {{- end }}
-  {{- $tgps := $agent.terminationGracePeriodSeconds | default 30 }}
+  {{- $tgps := $agent.terminationGracePeriodSeconds | default 300 }}
   {{- /* Kind, not a single kind: the same value reaches this line as
          different Go types depending on how it was supplied. Everything
          routed through `include "foreman.agents" | fromYaml` is float64,
--- a/charts/foreman/tests/agent-rollout_test.yaml
+++ b/charts/foreman/tests/agent-rollout_test.yaml
@@ -3,9 +3,9 @@ templates:
   - agent-deployment.yaml
 
 tests:
-  # ---- Defaults reproduce the previous hardcoded output exactly ----
+  # ---- Defaults render the 300s drain window (#1438) ----
 
-  - it: default agent renders Recreate and a 30s grace period (no change on upgrade)
+  - it: default agent renders Recreate and a 300s grace period (the drain window)
     template: agent-deployment.yaml
     set:
       agent.enabled: true
@@ -18,9 +18,9 @@ tests:
           path: spec.strategy.rollingUpdate
       - equal:
           path: spec.template.spec.terminationGracePeriodSeconds
-          value: 30
+          value: 300
 
-  - it: default multi-pool install renders Recreate and 30s for the default pool
+  - it: default multi-pool install renders Recreate and 300s for the default pool
     template: agent-deployment.yaml
     values:
       - ./values-multi.yaml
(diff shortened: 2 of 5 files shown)


Defilan (approved):
Approving the code, this is exactly the smaller first step #1438 asked for, and the drain machinery holds up. One thing before merge: the kind default-storage e2e entry is red.

What I did: the four drain tests pass three times under -race, and the full agent package is green. I bite-checked all three load-bearing hunks: without execs.Wait() the drain test fails with "phase = Running, want Succeeded"; restoring WithCancel for the executor fails with "executor was cancelled during the drain"; removing the pre-claim re-check fails with "second phase = Running, want Scheduled". The Add/Wait pairing can't race because pollOnce runs synchronously in Run's loop, liveness still aborts during a drain (its own test), and the chart tests pass with the 300s assertions, helm unittest 85/85 locally. The values.yaml rewrite is honest about what the knob now does.

The red job is the ModelRouter fail-closed spec getting a connection-level failure instead of 503. The same spec passed on the same run's Longhorn and GPUQuota entries, main's e2e runs are green, and nothing here touches the router, so it looks like runner noise to me. I'd just re-run that job.

Two nits: golangci-lint shows 4 goconst hits in the agent package that are identical on main, so "make lint: 0 issues" was optimistic, though none are yours. And a tick landing inside the drain window would log a poll failure and count toward the 3-strike stall threshold; three in a row would return before the Wait. The window is microseconds, so I don't think it's reachable, but a ctx.Err() guard in the ticker branch would close it.

Thanks, re-run the red job and this is good to merge.

PR #1801, opened 2026-09-11
fix(foreman): reject unverified reviewer GO

### LINKED ISSUE #1454: [BUG] Reviewer returns GO while stating it could not verify the change
## Bug Description

A reviewer that explicitly states it could not verify the change still returns **GO**, which opens a PR. Its own "cannot verify" sentence then becomes the PR body, so the resulting PR advertises that nothing validated it — and is approved anyway.

## Observed

`misospace/windowstead` issue #311 (workload `wl-misospace-windowstead-311`):

```
code-311      verdict=GO   "Fixed duration-0 one-time goal rewards expiring before
                            consumption in tick_rewards(). Added guard ... and 5
                            regression tests."

review-311-0  verdict=GO   "Tests fail due to missing godot runtime in environment;
                            cannot verify goal reward or progression logic."
```

The PR (`misospace/windowstead#321`) opened on that GO, with the reviewer's sentence as its body. GitHub CI then failed two checks (`Script test suite`, `macOS validation`).

## The deferral chain

foreman detected the missing runtime and deferred, expecting a backstop that did not exist:

```
gateAdvisories: [{
  "check": "self-gate-deferred",
  "detail": "self-gate deferred: runtime missing in the coder image for test command
             \"... godot --headless --path . --script res://tests/test_runner.gd ...\";
             the clean-room verify Job is the backstop"
}]
```

The fleet runs with the verify Job disabled (we rely on repo CI instead, after gate flakiness). So the self-gate defers to a backstop that is switched off, and nothing executes the tests anywhere before the PR opens.

## Expected

A review verdict of GO should assert the change was verified. Concretely, either:

1. A reviewer whose own summary reports it could not verify should not be able to return GO — demote to NO-GO or NEEDS_HUMAN, the way other unearned verdicts are demoted (#1166 grounded-finding, #644 issueAsk). The signal here is unusually clean: the model said it in plain language, in the field that becomes the PR description.
2. And/or: a `self-gate-deferred` advisory whose named backstop is not enabled should not be treated as a pass. The deferral is sound only when the verify Job actually runs; when it does not, the advisory is the last chance to notice.

(1) is the cheaper and more general guard — it needs no knowledge of which backstops are configured.

## Why it matters

The failure is invisible in the usual signals. The PR carried **zero** reviewer nits and read as clean; only the body text and the (later) CI failures reveal that nothing ran. A GO that means "I could not check this" is worse than a NO-GO, because downstream tooling and humans both treat GO as verification.

## Environment

LLMKube 0.9.14, in-process reviewer. Related in spirit to #1411 (PR body not grounded against the diff) and #1447 (scope-overlap making test-coverage issues unwinnable) — all three are cases of a model-authored claim being trusted without a rail behind it.

joryirving:
## What

Demote a reviewer `GO` verdict to `NO-GO` when its terminal summary explicitly says verification could not be performed.

## Why

A `GO` must represent a verified approval. Previously a reviewer could report that it could not verify the change while still opening a PR. Fixes #1454.

## How

- Add a phrase-anchored reviewer integrity rail for `cannot verify`, `could not verify`, and `unable to verify` admissions.
- Preserve verdict-demotion audit metadata and add focused regression coverage, including the reported incident.
- Assisted-by: Codex with Agentic-local (generated the implementation and tests); I reviewed the diff and ran `make fmt`, `make vet`, `make lint`, and `make test`.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/internal/foreman/controller/workload_iteration.go
+++ b/internal/foreman/controller/workload_iteration.go
@@ -183,14 +183,17 @@ func taskStepLabel(t *foremanv1alpha1.AgenticTask) string {
 // Three conditions are required, and each one excludes a NO-GO the coder
 // CAN act on:
 //
-//   - verdictDemotedBy is the issueAsk or scope-overlap rail. A rail that
-//     rewrites a GO without findings manufactures a rejection the coder
-//     cannot act on: re-running cannot make an unverifiable issueAsk verify,
-//     and an issue that names the symptom file rather than the fix site is
-//     scope drift the coder already answered by fixing the cause (#1684).
-//     Keying off the bare verdictDemoted flag alone would not distinguish
-//     these two rails from each other, but both are inert when they carry
-//     no findings, so the single condition below accepts either name.
+//   - verdictDemotedBy is the issueAsk, scope-overlap, or unverified-summary
+//     rail. A rail that rewrites a GO without findings manufactures a
+//     rejection the coder cannot act on: re-running cannot make an
+//     unverifiable issueAsk verify, an issue that names the symptom file
+//     rather than the fix site is scope drift the coder already answered by
+//     fixing the cause (#1684), and a GO whose own summary admits
+//     verification could not be performed (#1454) is a statement about the
+//     review environment, which no re-run changes either. Keying off the
+//     bare verdictDemoted flag alone would not distinguish these rails from
+//     each other, but all are inert when they carry no findings, so the
+//     single condition below accepts any of their names.
 //   - verdictClaimed is GO, i.e. the rail really did rewrite the verdict.
 //     enforceReviewerIssueAsk stamps the demotion fields on a path where
 //     it does NOT: an unverified non-GO review is marked untrusted and the
@@ -217,7 +220,8 @@ func inertDemotion(t *foremanv1alpha1.AgenticTask) (reason string, inert bool) {
 		return "", false
 	}
 	modelExtra := envelope.Extra.ModelExtra
-	if by, _ := modelExtra["verdictDemotedBy"].(string); by != reviewer.RailIssueAsk && by != reviewer.RailScopeOverlap {
+	if by, _ := modelExtra["verdictDemotedBy"].(string); by != reviewer.RailIssueAsk &&
+		by != reviewer.RailScopeOverlap && by != reviewer.RailUnverifiedSummary {
 		return "", false
 	}
 	if claimed, _ := modelExtra["verdictClaimed"].(string); claimed != string(foremanv1alpha1.AgenticTaskVerdictGo) {
--- a/internal/foreman/controller/workload_iteration_test.go
+++ b/internal/foreman/controller/workload_iteration_test.go
@@ -55,17 +55,22 @@ func iterationWorkload(issues []int32, reviewers int, maxIter *int32) *foremanv1
 // rather than being threaded through every call site.
 const reviewStep = "review-641-0"
 
-// demotionReasonIssueAsk / demotionReasonScopeDrift are the two rails'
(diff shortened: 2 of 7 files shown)


Defilan (changes requested):
Requesting changes, and both fixes are small: the demotion lands with a rail name the controller's routing does not know, so it drives the exact futile fix iteration the sibling rails are shielded from, and the regex misses the contractions of the phrases it anchors.

The first one is the important half. inertDemotion (internal/foreman/controller/workload_iteration.go:220) suppresses a GO rewrite that carries no findings, keyed on reviewer.RailIssueAsk and RailScopeOverlap. Your rail stamps verdictDemotedBy="unverified-summary", a name defined only in package agent, so the predicate does not recognize it. I ran the identical envelope through inertDemotion: issueAsk and scope-overlap come back inert, unverified-summary does not, so the demotion opens a fix iteration whose feedback is the reviewer's own approval plus an empty findings list, the loop your own rail comment says cannot converge. The fix I validated: export RailUnverifiedSummary from reviewer/demotion.go, the file that exists because producer and consumer need the same strings, alias it in rail_skip.go like the siblings, and add it to the predicate. The probe goes inert and your tests stay green.

Second: I probed the boundary phrasings and "couldn't verify", "can't verify", "can not verify", and "was not able to verify" all stay GO. Those are the most natural ways to write the exact admission, and the rail fails open on them. A widened regex covering can't/couldn't (including the curly apostrophe), "can not", and "not able to" demotes all five while your eight table cases and the honest-GO cases pass unchanged.

Also worth a wiring test in the cross-stage rail's shape: removing your call site from runLLMPath keeps every test green, and TestCrossStageContradiction_WiredIntoRunLLMPath exists precisely to catch that.

One question rather than a finding: "a client without the key cannot verify a forged signature" demotes an honest GO describing the fixed system. Requiring a first-person subject would miss your incident summary, which has no subject, so I think this is inherent to phrase anchoring. Just make the tradeoff explicit in the docstring if you keep it.

Nit: the 4 goconst hits in the agent package are identical on main, not yours.

joryirving:
Thanks — fixed in `001c0b5`.

- Exported `reviewer.RailUnverifiedSummary`, aliased it in the agent, and added it to `inertDemotion`, with controller coverage for both empty and actionable findings.
- Expanded the matcher for `can't`/`couldn't` (straight and curly apostrophes), `can not`, and `was not able to`.
- Added an `Execute → runLLMPath` wiring test that fails if the rail call is removed.
- Documented and tested the accepted object-subject false-positive tradeoff you called out.

The existing goconst hits are unchanged from `main`, as noted.

Defilan (approved):
Approving, all four changes since my review are in and each one holds up under execution. I reverted the inertDemotion hunk and the new controller case failed exactly as it should, removed the runLLMPath call and the wiring test failed with the verdict still GO, and narrowed the regex back to the formal phrases and the six new phrasing cases failed. I also re-probed on my own: both apostrophe forms of can't and couldn't, can not, and was not able to demote, honest GOs and affirmative verification language stand untouched, and the object-subject case demotes with the tradeoff documented the way I asked.

One observation, not a blocker: I probed the registers the phrase set doesn't span, and "The change cannot be verified in this environment", "I did not verify the fix before approving", and "was unable to complete verification" all stay GO. Phrase anchoring can't chase every negation and I wouldn't widen it here, but "cannot be verified" is the passive of the exact admission this rail exists for, so a follow-up issue for the passive forms seems worth filing.

Signed off on all three commits, disclosure in the body, agent package lints clean on my run, and the branch merges clean into current main. Thanks for turning this around fast, merging.

PR #1802, opened 2026-09-12
feat(controller): metric-gated idle probe for the generic runtime

### LINKED ISSUE #1800: [FEATURE] generic runtime: metric-gated idle probe for precise ModelPool drain of custom-image servers
## Feature Description

Let the `generic` runtime's idle probe read a Prometheus `/metrics` gauge and a threshold, instead of only a plain 2xx HTTP check. Any custom-image server that exposes `/metrics` (vLLM, SGLang, TGI-style) could then be drained precisely in a ModelPool, the same way the native `vllm`/`sglang` runtimes already gate on their in-flight-request gauges internally.

## Problem Statement

> As an operator running a custom inference image on `runtime: generic` as a ModelPool member, I want the pool to drain it based on its real in-flight-request count, so a cross-model swap waits for in-flight work to finish instead of either never draining or preempting a busy server mid-request.

`GenericBackend.IdleProbe` reads the `inference.llmkube.dev/idle-endpoint` annotation and GETs that path, treating any 2xx as "idle". A custom vLLM image (its own entrypoint, so `runtime: vllm` is not usable — the runtime would replace the entrypoint and force its own args) exposes `/metrics` with `vllm:num_requests_running`, but has no endpoint that returns non-2xx only while busy. That leaves two bad choices as a pool member:

- Point the annotation at `/health` (always 2xx): the pool treats it as always idle and will drain it the instant a different member is requested, preempting an in-flight generation.
- Omit the annotation: `IdleProbe` returns `errIdleUnsupported`, the pool fails closed and never drains it, so the other member can never take the slot.

Neither matches the precise `/slots`-style drain the `llamacpp` runtime gets, or the `vllm:num_requests_running == 0` check the `vllm` runtime already does in `VLLMBackend.IdleProbe`. The mechanism to read a metric gauge for idleness already exists in-tree for the native runtimes; it just isn't reachable from `generic`.

## Proposed Solution

A metric-gated idle mode for `generic`. For example a second annotation:

```yaml
metadata:
  annotations:
    # GET <baseURL>/metrics, parse the gauge, idle when sum <= threshold
    inference.llmkube.dev/idle-metric: "vllm:num_requests_running"
    inference.llmkube.dev/idle-metric-threshold: "0"   # optional, default 0
```

When `idle-metric` is present, `GenericBackend.IdleProbe` scrapes `/metrics`, sums the named gauge (reusing the same `parsePrometheusGaugeSum` helper the `vllm`/`sglang` probes use), and reports idle when the sum is at or below the threshold. The existing `idle-endpoint` 2xx behaviour stays as-is when only that annotation is set, so nothing changes for current users.

## Alternatives Considered

- A per-deployment sidecar that scrapes `/metrics` and re-exposes a 2xx-when-idle endpoint for `idle-endpoint` to point at. Works, but it is boilerplate on every such deployment and duplicates logic the operator already has.
- Making `spec.command` override not force the runtime-generated `container.Args`, so a `runtime: vllm` service could run a custom entrypoint and still get the native probe. Narrower and more surprising than a first-class metric probe.

## Additional Context

- Related: #1787 (poolActivation IfIdle) and #1795 (reclaim to default) both make ModelPool swaps more common, which makes a precise generic idle drain more valuable.
- `parsePrometheusGaugeSum` and the gauge-based idle logic already exist in `runtime_vllm.go` / `runtime_sglang.go`.

## Priority

- [ ] Medium - Nice to have
- [ ] Low - Minor improvement

## Willingness to Contribute

- [ ] Yes, I can submit a PR
- [ ] Yes, I can help test
- [x] No, but I can provide feedback
- [ ] No, just requesting

---

_Filed with AI assistance (Claude), reviewed by me before submission, per CONTRIBUTING.md._

joryirving:
## What

Add a metric-gated idle probe to the `generic` runtime. A new annotation `inference.llmkube.dev/idle-metric` (plus optional `-threshold` and `-path`) tells the generic idle probe to scrape a Prometheus endpoint and gate idleness on the sum of a named gauge, instead of the 2xx health check `idle-endpoint` allows.

## Why

Fixes #1800.

The native runtimes drain precisely (llama.cpp `/slots`, vLLM `vllm:num_requests_running`), so a ModelPool swap waits for in-flight work before freeing the slot. The `generic` runtime, used for a custom image whose entrypoint the built-in runtimes cannot drive (e.g. a patched vLLM), could only be given a health path, and a 2xx health check returns idle even while the server is busy. So the pool treats such a member as always idle and can preempt it mid-request, the coarse drain there is a real gap for anyone running a custom-image server as a pool member.

## How

- `AnnotationIdleMetric` / `AnnotationIdleMetricThreshold` (default 0) / `AnnotationIdleMetricPath` (default `/metrics`).
- `GenericBackend.IdleProbe` gains a metric mode that takes precedence over `idle-endpoint`: scrape the path, sum the gauge via the existing `parsePrometheusGaugeSum`, report idle when `sum <= threshold`.
- Fails closed: a non-200 scrape or an absent gauge counts as busy, so a member whose idleness cannot be established is never preempted (mirrors the native vLLM/SGLang probes).

Example:

```yaml
metadata:
  annotations:
    inference.llmkube.dev/idle-metric: "vllm:num_requests_running"
    inference.llmkube.dev/idle-metric-threshold: "0"   # optional
    inference.llmkube.dev/idle-metric-path: "/metrics" # optional
```

## Checklist

- [x] Tests added (`TestGenericMetricIdleProbe`: idle/busy, threshold, absent-gauge fail-closed, non-200, custom path, precedence over idle-endpoint)
- [x] `make lint` clean, `make manifests generate` no drift, gofmt clean
- [x] Conventional commit, DCO signed off
- [x] AI assistance disclosed
- [x] Docs updated (docs/operations/gpu-sharing.md)

### DIFF
--- a/api/v1alpha1/constants.go
+++ b/api/v1alpha1/constants.go
@@ -38,6 +38,25 @@ const (
 	// in to drain-before-roll. Set on InferenceService metadata.annotations.
 	AnnotationIdleEndpoint = "inference.llmkube.dev/idle-endpoint"
 
+	// AnnotationIdleMetric names a Prometheus gauge exposed on the generic
+	// runtime's metrics endpoint whose sum reports in-flight work (for example
+	// vLLM's vllm:num_requests_running or SGLang's sglang:num_running_reqs).
+	// When set, the generic idle probe scrapes the metrics path and reports the
+	// member idle when the gauge sum is at or below AnnotationIdleMetricThreshold,
+	// giving a custom-image server the same precise drain the native runtimes
+	// get instead of a 2xx health check that is always "idle". Takes precedence
+	// over AnnotationIdleEndpoint. Set on InferenceService metadata.annotations.
+	AnnotationIdleMetric = "inference.llmkube.dev/idle-metric"
+
+	// AnnotationIdleMetricThreshold is the gauge sum at or below which the
+	// member counts as idle for AnnotationIdleMetric. Parsed as a float;
+	// defaults to 0 when unset or unparseable.
+	AnnotationIdleMetricThreshold = "inference.llmkube.dev/idle-metric-threshold"
+
+	// AnnotationIdleMetricPath overrides the HTTP path scraped for
+	// AnnotationIdleMetric. Defaults to /metrics.
+	AnnotationIdleMetricPath = "inference.llmkube.dev/idle-metric-path"
+
 	// DefaultAgentHeartbeatInterval is how often the metal-agent re-asserts
 	// its registrations (which also self-heals any missed update, #657).
 	DefaultAgentHeartbeatInterval = 30 * time.Second
--- a/docs/operations/gpu-sharing.md
+++ b/docs/operations/gpu-sharing.md
@@ -282,6 +282,36 @@ outruns `swapBudget`, so a client that retries after the incumbent drains is
 served. It is deliberately not a `502`: nothing is broken upstream, the slot is
 just occupied.
 
+### Precise drain for a custom-image member: the metric idle probe
+
+The `llamacpp`, `vllm` and `sglang` runtimes know how to ask their server whether
+it is idle (llama.cpp's `/slots`, vLLM's `vllm:num_requests_running` gauge), so a
+swap waits for in-flight work to finish before freeing the slot. The `generic`
+runtime, used for a custom image whose entrypoint the built-in runtimes cannot
+drive, has no such knowledge. By default it can only be told a health path via
+`inference.llmkube.dev/idle-endpoint`, and a plain health check returns 2xx even
+while the server is busy, so the pool treats such a member as always idle and can
+preempt it mid-request.
+
+If the custom image exposes Prometheus metrics with an in-flight-work gauge (most
+vLLM- and SGLang-derived images do), point the generic idle probe at it instead:
+
+```yaml
+metadata:
+  annotations:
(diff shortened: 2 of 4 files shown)


Defilan (approved):
Approving, the metric mode is what #1800 asked for and every claim in the body held up when I ran it. I disabled the metric branch in IdleProbe and TestGenericMetricIdleProbe failed across the board, so the coverage bites. My own probes all behaved: an unparseable threshold falls back to 0 as documented, multi-series gauges sum correctly at both thresholds, HELP/TYPE lines don't pollute the sum, a path without the leading slash fails closed rather than draining anything, and the non-200 and absent-gauge semantics mirror the vllm probe exactly, which is the sibling that matters. Both IdleProbe call sites pass the live isvc, so the annotations reach the probe from both the pool and the drain path. make manifests generate ran clean in a scratch worktree, no drift, and the touched package lints clean.

One line to add before merge: the checklist says AI assistance disclosed, but the body doesn't carry the disclosure itself. CONTRIBUTING.md asks for it in the PR body in the "Assisted-by: <tool> (...)" form; your #1801 has the line, this one lost it.

Nit: the no-leading-slash path error reads as a URL parse artifact ("invalid port ... after host"); prepending the slash would make the log line self-explanatory, though failing closed either way is the safe behavior.

Thanks, this closes the coarse-drain gap for custom-image pool members.

PR #1807, opened 2026-09-12
feat(model): resume interrupted single-file downloads via a content-keyed partial (#1765)

### LINKED ISSUE #1765: [FEATURE] Resume interrupted single-file model downloads: source-keyed partial and curl -C -
## Feature Description

Resume an interrupted single-file model download from where it stopped instead of restarting from byte zero.

This is the first, independently shippable slice of #1762. The parallel-transfer half stays on #1762, as do the multi-file staging loop and the Metal agent's Go path. Credit to @davidgibbons, who reported #1762, measured the behaviour and did the design work in its comment thread.

**Revised 2026-09-07 after review of PR #1766.** The first implementation keyed the partial on the source URL, which is not safe. See "The partial must be keyed on content" below; that section is now the load-bearing requirement of this issue.

## Problem Statement

As an operator staging checkpoints measured in tens or hundreds of GB, I want an interrupted transfer to continue from where it stopped, so that a slow first deploy does not become a stuck one.

Every cached branch of `buildModelInitCommand` (`internal/controller/model_storage.go`) opens with an unconditional `rm -f "$MODEL_PATH.tmp"`, there is no `--continue-at` anywhere in the file, and `remoteRevalidateScript` fills the partial with a plain `-fsSL -o`. Any interruption (pod eviction, node reboot, OOM kill, a 5xx from the CDN, an init-container restart) discards everything downloaded so far.

Recorded twice already: #1757 (62 recreates in 20 minutes, never past 5 GiB of a 184 GiB checkpoint) and #1536 (a live pod with `model-downloader: restartCount=22`).

## The partial must be keyed on content, not on the source URL

`curl -C -` resumes by byte offset and sends no `If-Range`. It does not check that the remote bytes still belong to the same artifact the partial came from. If a partial survives an interruption and the content at that URL then changes, the retry appends new bytes onto stale ones.

Reproduced against a range-serving origin with the exact command the downloader generates:

```
partial on disk : 40000 bytes of version A
upstream content at the same URL changes to version B
curl -f -L -C - -o part.tmp <url>

exit            : 0
final size      : 100000
bytes 0..39999  : 'A'
bytes 40000..   : 'B'
size == new remote Content-Length : True
```

The spliced file's size **equals the new remote `Content-Length`**, which is the only correctness check `remoteRevalidateScript` performs, so it passes revalidation. curl exits 0, so the `&&`-guarded `mv` publishes it, and every later restart's `[ -f "$MODEL_PATH" ]` guard treats the corrupt artifact as good permanently.

`RefreshPolicy=OnChange` is the sharpest exposure, because that policy exists precisely for content that changes at a stable URL. An unpinned `hf://org/repo` also normalizes to `.../resolve/main/`, a moving pointer whose URL string never changes across a repo push.

Note what this means for the existing `rm -f "$MODEL_PATH.tmp"`: it was doing two jobs, the #1435 debris hygiene *and*, incidentally, guaranteeing every attempt started from zero so cross-version splicing was impossible. Any resume design must replace both.

**`If-Range` alone does not fix this.** Verified: on a validator mismatch the server returns 200, curl exits 33, the `&&` chain fails so nothing is published, and the stale partial survives to fail again on the next attempt. That converts silent corruption into a crashloop.

## Proposed Solution

**Key the partial on a validator fetched from upstream, and let the existing keep-one sweep evict partials whose content has moved.** The sweep then does both jobs the old `rm -f` did.

Verified end to end against a range-serving origin, all five states correct (changed content, unchanged content, no partial, already-complete partial, server sending no ETag):

```sh
validator=$(curl -fsSLI "$URL" -o /dev/null -w '%header{etag}|%header{content-length}' 2>/dev/null || echo '')
KEY=$(printf '%s' "$validator" | sha256sum | cut -c1-12)
MODEL_PARTIAL="$MODEL_PATH.$KEY.tmp"
find "$(dirname "$MODEL_PATH")" -maxdepth 1 -name '*.tmp' ! -name "$(basename "$MODEL_PARTIAL")" -delete
curl -f -L -C - -o "$MODEL_PARTIAL" "$URL" && mv "$MODEL_PARTIAL" "$MODEL_PATH"
```

Changed upstream content yields a different validator, therefore a different partial name, therefore the stale partial fails the keep-one predicate and is deleted before the transfer starts. `sha256sum`, `basename`, `find` and `dirname` are all present in `curlimages/curl:8.18.0` (BusyBox 1.37.0); confirmed in the image.

### Where each piece goes

1. **The partial name is computed in the shell at runtime, not in Go.** The validator is only knowable at run time, so `modelPartialPath` in Go and the `MODEL_PARTIAL` env var emitted by `modelInitEnvVars` both go away. `MODEL_PARTIAL` becomes a shell variable assigned inside the script, before the sweep.

2. **Sweep in `$(dirname "$MODEL_PATH")`, not `$CACHE_DIR`.** The uncached (emptyDir) branches have `CACHE_DIR=""`, and `find ""` fails with "No such file or directory", which would hard-fail the whole init script through the `&&` chain. `dirname "$MODEL_PATH"` is correct on both cached and uncached paths.

3. **`RefreshPolicy=OnChange`** already runs a HEAD for `content-length`. Extend that same `-w` to capture the ETag too rather than issuing a second HEAD.

4. **`RefreshPolicy=IfNotPresent`** has no HEAD. Add one **inside** the `[ ! -f "$MODEL_PATH" ]` branch, so a warm cache still starts the pod without any network request.

5. **The `s3://` and local `cp` branches do not resume** and revert to the plain `rm -f "$MODEL_PATH.tmp"` they had before, byte-for-byte. `--aws-sigv4` range signing is unverified and the local copy has nothing to resume. Reverting keeps them identical to pre-resume behaviour and shrinks the surface.

6. **The atomic publish does not change.** Every transfer still writes a partial and `mv`s onto the destination (#1309, #1432).

### Tests

The string assertions in `model_storage_test.go` are necessary but not sufficient. This file already carries `TestRemoteRevalidateScript_Behavioral` in `model_storage_revalidate_test.go`, added by #1326 after a plausible curl-flag claim (`-w '%{size_download}'` on a HEAD) turned out to be wrong and `ContainSubstring` tests passed on a dead script. Resume is the same class of claim and needs the same treatment.

Extend that behavioral harness with an `httptest` origin that serves ranges, and cover:

- a partial from a **different** validator is deleted and the download restarts from zero, and the published file contains no bytes from the stale partial (this is the splice regression test, and it must fail if the validator is dropped from the key);
- a partial from the **same** validator resumes and does not re-fetch the bytes it already has;
- an **already-complete** partial publishes without error;
- **no partial** downloads normally;
- an origin that sends **no ETag** still works, keyed on content-length alone.

Also update the `remoteRevalidateScript` doc comment. Its numbered strategy step 3 still says "download to `"$dest.tmp"` and `mv` it onto `$dest`", which no longer describes what the function does, and `buildModelInitCommand`'s comment points readers at it.

### Out of scope

Parallel transfer (#1762), `s3://` resume, `buildMultiFileInitCommand` and its own sweep, and `MetalExecutor.copyToFile`.

## Alternatives Considered

- **Keying the partial on the source string** (what PR #1766 did). Rejected: proven to splice, above.
- **`If-Range` on the resume request.** Rejected as a standalone fix: proven to crashloop, above. It is a reasonable belt-and-braces addition on top of validator keying, not a replacement for it.
- **An age threshold on the sweep**, as #1435 suggested. Rejected: time-dependent, unassertable from a generated command string, and it still resumes into a partial from different content.
- **`aria2c -x8 -s8 --continue=true`.** Needs a new init-container image; belongs on #1762.

## Additional Context

- Parent: #1762 (@davidgibbons). The `.tmp` sweep this must not regress: #1435. Live case it compounds with: #1757; prior restart evidence: #1536. Atomic-publish lineage: #1309, #1432, #1326. The `hf_curl` seam it builds on: #1758.
- First implementation and its review: PR #1766.
- Verified against `main` at `9f3002219`.

Known limitation worth stating: resume still trusts the bytes already in the partial. Validator keying prevents resuming across a content change, but nothing checksums the partial itself, so a partial corrupted in place (bad disk, a concurrent writer) is not caught. That is the same exposure the existing size-equality guard has.

Concurrency, unchanged by this issue but worth recording: #1459 accepted that concurrent transfers of the same model share one partial path, since `CACHE_DIR` is per-cache-key and shared across replicas with no locking. Validator keying does not introduce that race, but it does mean a surviving partial is a longer-lived shared write target than a truncating download was. Locking belongs in its own issue.

## Priority

- [ ] High - Would significantly improve my workflow
- [x] Medium - Nice to have
- [ ] Low - Minor improvement

## Willingness to Contribute

- [ ] Yes, I can submit a PR
- [x] Yes, I can help test
- [ ] No, just requesting

Defilan:
## What

Interrupted single-file model downloads now resume from where they stopped instead of restarting from byte zero, on both download paths: the controller-generated init-container scripts (`buildModelInitCommand` / `remoteRevalidateScript`) and the metal agent's Go downloader (`MetalExecutor.downloadFile`). This is the single-file slice of #1762; the parallel-transfer half stays there by design.

## Why

Every cached branch opened with an unconditional `rm -f` of the partial and had no `--continue-at` anywhere, so a pod eviction, node reboot, OOM kill, a 5xx from the CDN, or an init-container restart discarded everything downloaded so far. Recorded twice in the field: #1757 (a multiNode group recreated 62 times in 20 minutes, the new member never past 5 GiB of a 184 GiB checkpoint) and #1536 (a live pod with `model-downloader: restartCount=22`).

Fixes #1765

## How

- **Content-keyed partial.** The partial is named `$MODEL_PATH.<sha256(validator)[:12]>.tmp` where the validator comes from a single HEAD reading ETag and Content-Length (`-w 'CL%header{content-length}ET%header{etag}'`). Keying on content, not the source URL, is the load-bearing requirement from the review of the first attempt (#1766): like `curl -C -`, a ranged request sends no `If-Range`, so the content key is what prevents splicing stale bytes onto changed content. A first cut of this PR series emitted only the length in the probe; the behavioural tests caught it (no partial ever matched, and length-only keying is not splice-safe for a same-length content change), so the ETag is in the key.
- **Keep-one sweep.** The unconditional `rm -f` becomes a sweep that deletes every `*.tmp` except the current key's partial: debris still cannot accumulate on the shared cache PVC (#1435), and a content change evicts the abandoned partial instead of resuming into it.
- **Branch discipline.** The S3 and local-copy branches keep their exact pre-resume shape: sigv4 range signing is unverified and a local copy has nothing to resume.
- **Metal agent.** `downloadFile` gains the same mechanism natively (probe validator, content-keyed partial, `Range: bytes=N-`, keep-one sweep), mirroring the controller branch-for-branch; an origin that ignores Range answers 200 and the partial is discarded and rewritten; a failed probe degrades to a fresh full download.
- **Tests.** Behavioural, not string-matching, for the same reason `TestRemoteRevalidateScript_Behavioral` exists (#1326): the generated shell runs under `sh` against a range-serving stub origin and the assertions are the bytes that land on disk. The splice case is load-bearing: a partial left by a different content version must be discarded and must contribute zero bytes. They skip where host `stat` lacks `-c` (the scripts target the busybox/Linux init image) and run in CI; verified locally inside `curlimages/curl:8.18.0`.

## Checklist

- [x] Tests added/updated (behavioural controller script suite, metal Go suite, splice regression on both sides, existing branch-shape assertions kept green)
- [x] `make lint` and `make lint-all` pass locally


Defilan (commented):
Holding this on two things, both of which undercut the shape the PR claims.

The controller side holds up. I ran the new behavioral suite under sh against the range origin (my host stat lacks -c, so I shimmed it) and it passes, and the splice case genuinely bites: replacing the shell key derivation with a fixed key makes the matching-partial tests fail with "origin served 1 from-zero GETs". The S3 and local branches keep their pre-resume shape.

The metal agent does not resume. A first attempt leaves its partial at the empty-validator name, because executor.go:507 recomputes partPath from resumeValidator, which is empty on a fresh run. I interrupted a real downloadFile mid-body and got .model.gguf.e3b0c44298fc.partial, e3b0c44298fc being sha256("")[:12]. The retry probes the real validator, looks for the validator-keyed name, finds nothing, sweeps that partial as debris, and does a plain GET: fullGETs=1, rangeGETs=0. The validator-keyed partial can only exist if a resume already started, so resume is unreachable on this path. Both field cases in the body are this shape. Using the probed validator for the from-zero transfer fixes it: with partValidator = validator the same probe leaves .model.gguf.bc4c9a514a4d.partial and the retry resumes, fullGETs=0, rangeGETs=1. The new metal tests seed the validator-keyed partial directly, a state no production path creates, which is why they pass.

Second, resumePrologue runs before the [ ! -f "$MODEL_PATH" ] check, so a warm IfNotPresent cache now HEADs upstream. #1765 requirement 4 asked for the probe inside that branch so a warm cache starts with no network request. My probe with a pre-existing MODEL_PATH prints "Model already cached, skipping download" and an origin request count of 1. An air-gapped node with a warm cache now waits on a curl that has no --max-time.

Nit: TestDownloadFile_SuccessRenamesTempFile still asserts on localPath + ".partial", which the HTTP path no longer writes, so that check is vacuous.

Question on the no-ETag origin: keying on content-length alone still lets a same-length change splice, and you call the ETag the load-bearing requirement. Is that fallback acceptable for the sources in #1757 and #1536?

Defilan:
Correction: that review was written against 9218133, and both findings are fixed on the current head.

b94ff03a7 makes every transfer write the probed validator's partial, from-zero included, so an interrupted fresh download is the artifact the retry looks for. It also adds the test that drives the interruption instead of seeding a partial, which is why the old tests could not see it. 85a1e3f5c moves the IfNotPresent probe inside the model-absent branch, so a warm cache starts with no request, with a test that counts origin requests.

Re-verified on 85a1e3f5c: make fmt, vet, lint, test and lint-all clean, the three falsification reverts each fail the new tests, and CI is green. Nothing left open from the review.

PR #1809, opened 2026-09-12
test: report and replay the ginkgo seed in make test (#1693)

### LINKED ISSUE #1693: [BUG] make test uses an unpinned, unreported ginkgo seed: test-pollution bugs pass or fail on the draw
## Bug Description

`make test` runs the Ginkgo envtest suites with an unpinned, unreported random
seed. Ginkgo randomizes top-level containers per run, so a test that leaks
shared cluster state fails or passes depending on the draw. Nothing in the
Makefile or the workflows pins a seed or prints the one that was used, so a
failure cannot be reproduced from the log.

`Makefile:110-111`:

```make
test: manifests generate fmt vet setup-envtest ## Run tests.
	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
```

This makes any test-pollution bug a coin flip in CI and in the Foreman gate,
which runs the same `make test`.

## Observed Live

Branch `issue-1676-prefetch-claimname` at `d0a1b5fbb` leaked four
InferenceServices out of a `Describe` block (cleanup was in a `BeforeEach`, so
each case cleaned before itself and never after the block).
`federation_edge_controller.go` lists InferenceServices across all namespaces
and its spec asserts absolute counts, so the leak breaks an unrelated test —
but only when that spec happens to run after the leaking block:

```
seed 11  ->  ok
seed 22  ->  FAIL   FederationEdgeReconciler [It] pushes the observed summary ...
```

Same commit, same machine, same command. The Foreman gate ran the full suite on
that branch and returned GATE-PASS; it drew a passing seed. The branch would
have merged and then failed CI on an unrelated PR, where the cause would have
been very hard to find.

## Expected Behavior

A pollution bug fails reproducibly rather than one run in N, and any failure can
be replayed from the log alone.

Two parts, and the second matters more than the first:

1. **Report the seed.** Whatever seed a run uses should appear in the output so
   a failure is reproducible. This is nearly free.
2. **Exercise more than one ordering.** A single fixed seed makes runs
   reproducible but permanently hides every ordering bug the fixed seed happens
   to miss, which is worse than the status quo. Prefer running the envtest
   suites under at least two different seeds.

## Decided Approach

**A second fixed seed, in CI only.** `.github/workflows/test.yml` runs the
envtest suites twice under two different fixed seeds. Local `make test` and the
Foreman gate stay single-pass, so per-task gate latency is unchanged (the
`internal/controller` suite alone is ~270s, and doubling every gate run is not
worth it for a class of bug that only needs to be caught before merge).

Both runs report their seed so any failure is replayable from the log.

Rejected, and do not implement instead: a single pinned seed. It makes runs
reproducible while permanently hiding every ordering bug that seed happens to
miss, which is worse than the visible flakiness it replaces.

The nightly-sweep and `--randomize-all` options were also considered; CI-only
double-seed was chosen as the point where detection lands before merge without
taxing the gate.

## Fix Sites

- `Makefile:110-111` — the `test` target.
- `.github/workflows/test.yml` — where a second-seed pass would run if CI is the
  chosen home.
- `pkg/foreman/agent/tools/run_gate_job.go` — the gate invokes `make test`; if
  the second pass is CI-only, the gate needs no change, but the decision should
  be recorded in a comment there so the gate's weaker guarantee is explicit.

## Testing

A regression test is awkward here (the bug is in the harness, not the code), so
the acceptance check is empirical: on `d0a1b5fbb`, the chosen configuration must
fail. That commit is a known-good reproduction and should be named in the PR.

## Additional Context

Found while auditing why `issue-1676-prefetch-claimname` passed both the gate
and the reviewer while failing the full suite locally. The leak itself is fixed
on that branch; this issue is about the harness that let it through.

Defilan:
## What

`make test` now reports the ginkgo seed it ran with and can replay it: the Makefile picks one seed per run, echoes it into the log with the replay command (`make test GINKGO_SEED=<n>`), and exports it; each envtest suite reads `GINKGO_SEED` and overrides `RandomSeed` on the config Ginkgo derived from its own flags. CI also runs the envtest suites a second time under a second fixed seed, and the Foreman gate records that it stays single-pass.

## Why

The envtest suites ran with an unpinned, unreported random seed. Ginkgo randomizes top-level containers per run, so a test that leaks shared cluster state passes or fails on the draw: on `d0a1b5fbb` a leaking branch failed `FederationEdgeReconciler` only on seed 22, and the Foreman gate, running the same `make test`, drew a passing seed and returned GATE-PASS on code that would then fail CI on an unrelated PR. Nothing in the Makefile or workflows pinned or printed the seed, so a failure could not be reproduced from the log alone.

Fixes #1693

## How

- `Makefile`: `GINKGO_SEED ?= $(shell date +%s)`, echoed as `ginkgo seed: <n> (replay: make test GINKGO_SEED=<n>)` and exported into `go test`. A fresh seed per run keeps spec-order randomization (that randomization is what catches pollution); only failures now carry their order in the log. New `test-envtest` target runs just the two envtest packages, so CI can spend a second seed on the specs whose ordering matters without doubling the rest of the job.
- `internal/controller/suite_test.go` and `internal/foreman/controller/suite_test.go`: the seed override lands on `GinkgoConfiguration()`, not `types.NewDefaultSuiteConfig()`. A config passed to `RunSpecs` replaces Ginkgo's flag-derived config wholesale, so the default-config form dropped `-ginkgo.focus`, `-ginkgo.label-filter`, `-ginkgo.skip` and `-ginkgo.seed`. When `GINKGO_SEED` is unset the suite keeps and reports the seed it already has, so a direct `go test -args -ginkgo.seed=N` still honours the flag. The helper mirrors across the two suite files, which already mirror each other deliberately.
- `.github/workflows/test.yml`: `make test GINKGO_SEED=11`, then `make test-envtest GINKGO_SEED=22`. The two seeds are the ones named in the issue's reproduction. `test-envtest` is exempted in `gateExemptCIChecks` and the reason is recorded beside `DefaultGateChecks`, so a GATE-PASS is explicit that it does not cover the second-seed ordering.
- Out of scope: the e2e suite (`make test-e2e`) is invoked differently and is not part of this issue's report; the same pattern applies there if wanted.

## Acceptance

On `d0a1b5fbb`, `go test ./internal/controller/ -count=1 -args -ginkgo.seed=22` fails `FederationEdgeReconciler [It] pushes the observed summary to the datacenter FederatedCluster status without touching Phase` (745 Passed | 1 Failed), and `-ginkgo.seed=11` passes. The two CI seeds are chosen so CI covers that ordering before merge rather than leaving it to the draw.

## Checklist

- [x] Tests added/updated (test-infra change: verified by `GINKGO_SEED=42` landing as Ginkgo's own `Random Seed: 42`, by `-ginkgo.label-filter` and `-ginkgo.seed=7` still being honoured, and by the seed 22 reproduction above)


Defilan:
Re-reviewed at `aeff80508`.

The flag-handling regression is fixed, verified empirically: with a focus filter matching nothing the suite now reports `Will run 0 of 118 specs`, where the previous head ran `122 of 122`. Seed behavior checked on all four paths:

| case | result |
|---|---|
| `GINKGO_SEED=42` | `Random Seed: 42` |
| `-ginkgo.seed=777`, no env var | `Random Seed: 777` |
| both set | `42`, env wins as documented |
| focus filter | honored |

Threading the current seed through `ginkgoSeed(current int64)` is better than simply swapping in `GinkgoConfiguration()`: it also makes a direct `go test -ginkgo.seed=N` work, which the previous code silently overrode.

The CI second-seed pass closes the half of #1693 that was missing, and all three of the issue's Fix Sites are addressed: the Makefile target, the `test.yml` second pass, and the `run_gate_job.go` comment recording the gate's weaker guarantee. Seeds 11 and 22 are the pair the issue documents as ok/FAIL on the real pollution bug, so the CI pass reproduces the motivating case rather than sampling arbitrarily.

One follow-up filed as #1817, not a blocker here: `GINKGO_SEED ?= $(shell date +%s)` re-evaluates at every reference, so roughly 1 local run in 60 logs a seed it did not use. CI is unaffected because it passes explicit seeds; the Foreman gate is affected because `gate_job_template.yaml` runs bare `make test`. Measurements and a proven fix are in the issue.

LGTM.

PR #1811, opened 2026-09-12
feat(foreman): serve Agents from Anthropic-native endpoints

### LINKED ISSUE #1627: foreman: support Anthropic-native model endpoints, not only OpenAI-compatible proxies
## Problem

`AgentProvider` is `local;cloud-proxy`, and both are OpenAI-compatible HTTP (`api/foreman/v1alpha1/agent_types.go`). The agent's only wire client is `pkg/foreman/agent/oai`, SSE-only (`client.go:53`). So reaching an Anthropic-native model requires a translating proxy in front, and the agent can never see the Anthropic response shape directly.

That is usually fine. It stops being fine when the translation is where the fidelity is lost.

## Concrete case

Our `coder-frontier` ran `MiniMax-M3-chat` — MiniMax's OpenAI-compatible shim at `https://api.minimax.io/v1` over a model that natively emits Anthropic thinking blocks. Measured behaviour through that shim:

- It never returns `reasoning_content`. Reasoning arrives inline in `content` as `<think>…</think>` on every call, while `usage.completion_tokens_details.reasoning_tokens` is still counted.
- LiteLLM strips the balanced pair only on **non-streaming** responses. On streaming it does not strip at all (3/3 reproductions).

Because the agent is streaming-only, every coder turn ingested the model's raw reasoning trace as ordinary assistant content. It mostly still worked — 29 of 33 issue-fix runs ended `GO` — but every prompt carried the previous turn's reasoning back into context as if the model had said it.

The same provider also exposes an Anthropic surface (`https://api.minimax.io/anthropic/v1/messages`) with proper thinking blocks, which would avoid the problem entirely — but a streaming tool-calling request to it *through LiteLLM* returns HTTP 500 with zero SSE chunks, so that route is closed for a streaming-only client.

## Ask

Let an Agent target an Anthropic-native endpoint directly, rather than only through an OpenAI-compatible translation:

- extend the `AgentProvider` enum with an Anthropic messages provider, and
- give the agent a client that speaks `/v1/messages` with native `thinking` blocks and tool use.

That removes an entire class of translation-fidelity bug: reasoning stays in its own channel because the wire format has one, instead of depending on a proxy to recognise and strip vendor-specific tags.

## Narrower alternatives, if a second client is too much

- Strip `<think>…</think>` from streamed `content` deltas in the `oai` client, so a leaking upstream cannot pollute conversation history. Cheap, and it protects every deployment sitting behind a translating proxy.
- Surface `reasoning_content` when the upstream provides it, and keep it out of the message history the next turn replays.

## Environment

LLMKube v0.9.19. Agent `provider: cloud-proxy` against a LiteLLM gateway (`http://litellm.llm:4000/v1`). Observed 2026-08-21.

joryirving:
## What

Adds an `anthropic` value to `AgentProvider` and a new `pkg/foreman/agent/anthropic` client that speaks the Anthropic Messages API (`POST /v1/messages`, SSE streaming) natively: first-class `thinking` blocks, structured `tool_use` input, `x-api-key` auth. The agent loop gains a `ChatCompleter` seam and is now provider-agnostic. CRD enums regenerated (config + foreman chart), example Agent manifest added.

## Why

Fixes #1627.

The issue is measured, not theoretical: behind MiniMax's OpenAI-compatible shim, reasoning arrives inline in streamed `content` (`<think>…</think>`) while `usage` still counts it as reasoning tokens — and LiteLLM strips that pair only on **non-streaming** responses. The agent is streaming-only, so every coder turn replayed the model's raw reasoning trace as ordinary assistant content. A translation proxy cannot be made to reliably recognize and strip vendor-specific tags; the native wire simply has a separate channel for reasoning, so the class of bug goes away.

(The issue's narrower alternative of surfacing `reasoning_content` and keeping it out of replay history has since landed on the oai path; this PR implements the main ask — targeting Anthropic-native endpoints directly.)

## How

- **New `anthropic` package**, mirroring the `oai` client's conventions (no SDK, stdlib-only, SSE-only, same retry/backoff/jitter policy):
  - `translate.go` maps the internal `oai.ChatRequest` to a Messages body: system messages hoisted to the top-level field, `tool` results riding inside **user** messages as `tool_result` blocks, consecutive same-role messages merged (strict user/assistant alternation is enforced), image `data:` URLs mapped to base64 sources, `max_tokens` defaulting to 8192 (the API requires the field). Historical `ReasoningContent` is dropped on replay — Anthropic only accepts thinking blocks echoed with valid signatures, which no third-party Anthropic-compatible endpoint emits.
  - `client.go` aggregates the SSE stream (`text_delta` → content, `thinking_delta` → reasoning, `input_json_delta` → tool args, `signature_delta` ignored, in-stream `error` events abort) back into the existing `oai.ChatResponse` shape, with `stop_reason` mapped to the OpenAI finish-reason vocabulary. Non-2xx errors wrap `*oai.StatusError` so the loop's temperature-rejection retry keeps working against both wires.
- **Loop seam**: `Loop`/`NewLoop`/`LoopFactory` now take a `ChatCompleter` (the single `Chat` method); `newChatCompleter` in the executor picks the client by provider.
- **Auth is provider-shaped**: cloud-proxy keeps `Authorization: Bearer <token>`; anthropic sends the Secret value verbatim as `x-api-key`.
- **Sovereignty unchanged**: `anthropic` is a non-local provider, so both existing gates — the operator `--allow-cloud-providers` kill switch and per-Workload `spec.allowCloudReviewers` — apply to it without modification.

Tests: 30+ new cases — wire translation, SSE aggregation (fragmented tool args, clean-EOF-without-`message_stop`, empty stream, in-stream errors), retry classification and header contract, executor endpoint resolution, and a real loop running end to end over a fake Anthropic SSE server (tool round-trip + `submit_result`).

## Checklist

- [x] Tests added/updated
- [x] `make lint` passes locally (also `make vet`, `make fmt`; CRDs regenerated via `make foreman-chart-crds`, `make generate` no-op)
- [x] Documentation updated (if user-facing change) — example Agent manifest + CRD/godoc descriptions; no docs-site page added

Assisted-by: opencode agents (a Qwen3.8-27B coder subagent generated the `anthropic` package and tests; a Qwen3.8-Flash-Next coordinator wired the provider, regenerated manifests, and drafted this description; verified by running `make test`, `make lint`, `make vet` locally, plus an independent second-opinion review pass with Gemma-4-12B that found no blocking issues).

### DIFF
--- a/api/foreman/v1alpha1/agent_types.go
+++ b/api/foreman/v1alpha1/agent_types.go
@@ -49,15 +49,19 @@ const (
 // AgentProvider names the model-serving backend the executor dispatches
 // to. v0.2 introduces this enum so reviewer Agents can route to an
 // OpenAI-compatible cloud proxy (typically a LiteLLM gateway hitting
-// Anthropic / OpenAI / Bedrock) without changing the agent loop. The
-// default "local" preserves the v0.1 behavior of resolving an
+// Anthropic / OpenAI / Bedrock) without changing the agent loop.
+// "anthropic" dials an Anthropic-native /v1/messages endpoint directly
+// (#1627): the Messages wire carries first-class thinking blocks and
+// structured tool input, which a translation proxy degrades or drops.
+// The default "local" preserves the v0.1 behavior of resolving an
 // in-cluster InferenceService.
 //
-// Air-gapped orgs MUST keep cloud-proxy Agents out of dispatch; both
-// the operator-level kill switch (--allow-cloud-providers flag, chart
-// value foreman.allowCloudProviders) and the per-Workload opt-in
-// (Workload.spec.allowCloudReviewers) gate this.
-// +kubebuilder:validation:Enum=local;cloud-proxy
+// Air-gapped orgs MUST keep non-local Agents out of dispatch; both the
+// operator-level kill switch (--allow-cloud-providers flag, chart value
+// foreman.allowCloudProviders) and the per-Workload opt-in
+// (Workload.spec.allowCloudReviewers) gate every non-local provider,
+// including anthropic.
+// +kubebuilder:validation:Enum=local;cloud-proxy;anthropic
 type AgentProvider string
 
 const (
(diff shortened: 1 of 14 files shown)


Defilan (approved):
Approving. I traced the contract end to end and it holds up: thinking deltas land in a separate channel from content, tool args reassemble from fragmented input_json deltas with a start-payload fallback, tool results ride in user messages with strict alternation enforced by the merge pass, and the loop runs end to end over the native wire with transcript and header assertions in the E2E test.

I tried to break it. Reverted the hunk that routes thinking_delta into reasoning and the separation test failed with exactly the #1627 leak ("let me thinkanswer" in content), so the coverage bites. Planted a prompt drift in the new claude-native-anthropic-reviewer.yaml and check-reviewer-prompts failed naming that file, so the gate really reads it. Regenerated the CRDs and chart: zero drift. Probed two near-misses the diff doesn't cover (nil ImageURL part, negative max_tokens) and both are handled. The sovereignty claim checks out: filterCloudProviders and both IsDeterministicAgent copies key off non-local, so both gates apply to anthropic, and non-2xx wraps *oai.StatusError so the temperature retry works on both wires.

One nit: charts/foreman/values.yaml and the --allow-cloud-providers flag help still describe the switch as gating cloud-proxy Agents specifically, while the code and the updated Agent CRD godoc gate every non-local provider. Worth a one-line touch-up, doesn't need to hold this.

Thanks, this is clean work.

Issue #1628, opened 2026-08-21
foreman: a spinning agent loop exhausts maxTurns in 65s and reports it as MaxTurnsExhausted


joryirving:
## Problem

A coder task can consume its entire `maxTurns` budget in about a minute and terminate as `INCOMPLETE` / `MaxTurnsExhausted`, with no branch pushed and nothing in the record explaining what the turns were spent on. To an operator it is indistinguishable from "the task was too large".

## Observed

Two `issue-fix` tasks on the same agent (`coder-frontier`, `provider: cloud-proxy` to a LiteLLM gateway), `maxTurns: 160`:

```
bridge#199   elapsed 105s   160 turns   INCOMPLETE / MaxTurnsExhausted
bridge#189   elapsed  65s   160 turns   INCOMPLETE / MaxTurnsExhausted
```

That is 0.4–0.66 s per turn. For contrast, the same fleet's genuine turn-cap hits on a local model took 1106 s and 1662 s, and the same `coder-frontier` agent ended 29 of 33 issue-fix tasks `GO` with a median elapsed of 270 s. So the agent and model are healthy in general; these two runs spun.

**Most of those turns never reached the model.** LiteLLM's per-deployment request counter over the 65 s window of the second run shows roughly 49 requests to that model group — and that window also contains unrelated traffic from other consumers. So on the order of two-thirds of the 160 turns were consumed without a completion being requested.

## Why this is hard to diagnose from outside

- The `foreman.audit.v1` record carries `verdict`, `failureReason`, and `elapsedSec`, but no turn count, no per-turn outcome, and no indication of whether a turn produced a model call, a tool call, or an error.
- The agent logs one line per task at INFO (`dispatching to executor`) and nothing per turn, so the loop is opaigue without raising verbosity and catching a recurrence.
- `#187`-style tombstone cleanup removes the Workload after parking, so the evidence window is short.

## Ask

1. **Do not let a non-productive turn consume the budget silently.** A turn that neither obtains a completion nor executes a tool is not progress; either exclude it from the count or abort the loop with a distinct reason.
2. **Distinguish exhaustion-with-progress from a spin.** `MaxTurnsExhausted` after 1662 s of real work and after 65 s of spinning are different failures and currently look identical.
3. **Put turn accounting in the audit record** — turns used, how many produced a completion, how many produced a tool call, and the terminal reason. That alone would have answered this without a cluster-side investigation.

[...]

Issue #1713, opened 2026-08-31
[BUG] max-turns INCOMPLETE reports "model did not call submit_result" when it called it and was accepted


Defilan:
## Bug Description

When a coder run exhausts `maxTurns`, the recorded summary asserts the model never submitted. In at least one observed run the model called `submit_result` three times and the harness acknowledged each with `{"accepted":true,"verdict":"GO"}`.

The message sends triage in exactly the wrong direction: it suggests the model failed to conclude, when in fact it concluded repeatedly and the loop did not stop.

## Where this lives

`pkg/foreman/agent/executor_native.go:2398`, in `mapLoopError`'s `ErrMaxTurnsExhausted` arm:

```go
case errors.Is(loopErr, ErrMaxTurnsExhausted):
	return e.incompleteResult(start, tref, lr,
		foremanv1alpha1.FailureMaxTurnsExhausted,
		"model did not call submit_result within max_turns"), nil
```

The neighbouring `ErrAssistantNoToolCalls` arm at `:2401` is a good model for a message that describes what actually happened ("model returned text without tool_calls; loop cannot make progress").

## Evidence

From the transcript of `wl-1478-397b-r2-code-1712` (`foreman-transcript-wl-1478-397b-r2-code-1478`):

```
[382] model  submit_result(verdict=GO)  ->  {"accepted":true,"verdict":"GO"}
[385] model  submit_result(verdict=GO)  ->  {"accepted":true,"verdict":"GO"}
[388] model  submit_result(verdict=GO)  ->  {"accepted":true,"verdict":"GO"}

status.result.summary: "model did not call submit_result within max_turns"
```

## Expected Behavior

The summary should describe the observed terminal state. If submissions were made and rejected, say so and say how many, for example:

> submitted 3 times, each rejected by the verification gate; turns exhausted

If the loop does not currently track submission attempts, thread a count through `LoopResult` so `mapLoopError` can report it. That count is also what would let an operator distinguish "never concluded" from "concluded but was refused", which are different problems with different fixes.

## Scope

**This issue is deliberately narrow and independent.** It is one of three fixes described in #1712, split out because it does not depend on the others: the summary text can be corrected regardless of what the loop decides about gates, tool availability, or pushing the branch.

The other two fixes from #1712 (the force-submit deadlock, and wiring `pushFailedResult` into the max-turns path) are coupled to each other and are tracked there.

[...]

PR #1814, opened 2026-09-13
fix(foreman): tell a spin from a real max-turns burn

joryirving:
## What

Give the foreman loop turn-accounting, and give a max-turns exhaustion that ran faster than a model physically can its own failure reason: `LoopSpinning`. The `foreman.audit.v1` record now carries `turnsWithCompletion` / `turnsWithToolCall` alongside `turnCount`.

## Why

Refs #1628. Two issue-fix tasks burned all 160 turns in 105s and 65s — 0.4–0.66 s/turn, faster than any model generates — and terminated `INCOMPLETE` with the same bare `MaxTurnsExhausted` that a genuine 1106s/1662s exhaustion reports. The audit record showed a turn count but no productivity signal, so a spin looked exactly like a model that gave up, and the operator's only lever (raise the budget / tighten the prompt) was the wrong one.

## How

- **Loop accounting** (`pkg/foreman/agent/loop.go`): `LoopResult` gains `TurnsWithCompletion` (turns whose chat request yielded a choice — transport errors and empty-choices aborts don't count), `TurnsWithToolCall`, and `TurnDuration` (cumulative wall-clock inside `runOneTurn`). Each counter is incremented at a single point per turn, so the temperature-retry path can't double-count and `TurnsWithCompletion <= Turns` holds by construction.
- **Spin guard** (`pkg/foreman/agent/executor_native.go`): in `mapLoopError`, a max-turns exhaustion with `Turns >= 10` and average turn duration strictly below 1s is classified `LoopSpinning`, with the observed rate in the summary; the #1713 rejected-submissions detail is kept when present. The guard lives in the executor, not the loop: the loop still returns the `ErrMaxTurnsExhausted` sentinel and knows nothing about CRD reasons.
- **New API enum value** `LoopSpinning` (`api/foreman/v1alpha1/agentictask_types.go`); CRDs regenerated (`config/crd/bases`, `charts/foreman/templates/crds`). Distinct fix path: check the endpoint/shim, not the prompt/budget. Runbook updated.
- **Thresholds**: 10 turns floor / 1 s average cap, derived from the observed spins (0.41–0.66 s/turn) vs healthy runs (~1.7 s+/turn); the floor keeps tiny fast runs on the ordinary reason.
- **Audit** (`pkg/foreman/audit/record.go`): the two new counters flow through `resultPayload.Extra` into the record; absent (pre-fix) result JSON leaves them zero and `omitempty` drops them, so old runs are never rewritten with fabricated values.

Deliberately out of scope: the no-tool-call / reasoning-only / truncated streak errors already have bounded streak budgets (their own circuit breakers), so only the `MaxTurnsExhausted` path — the unbounded one — gets the spin classification.

## Checklist

- [x] Tests added/updated (`TestLoop_TurnAccounting`, `TestMapLoopError_MaxTurns_SpinGuard` incl. floor/threshold boundaries, audit record presence + absence cases)
- [x] `make lint` passes locally (also `make lint-all` across GOOS)
- [x] AI assistance disclosed below, per CONTRIBUTING.md
- [x] Documentation updated (`docs/site/foreman/runbook-m3.md`)

Assisted-by: OpenCode (qwen3.8-flash-next coordinator; loop counter placement reviewed against an independent local Gemma review pass); verified by running `make test`, `make lint`, `make lint-all`, `make generate`/`manifests`/chart CRD sync, and the new table-driven tests. Note: commit is DCO sign-off only — my machine's 1Password GPG signer was unavailable in the agent shell, so the commit is not GPG-signed.


Defilan (approved):
Approving, the spin trace holds up end to end. I re-derived the threshold arithmetic (1s times Turns is Turns seconds, not Turns nanoseconds) and ran the real 160-turn/65s and 1662s cases through `mapLoopError`; they classify the way the test says. Both new tests bite: I removed the two counter increments in `loop.go` and the guard condition in `executor_native.go`, and the respective tests failed with the counters at zero and the reason reverting to `MaxTurnsExhausted`. The guard's edges hold too; I probed 9 turns at a zero duration, a zero-value `LoopResult`, and both sides of the 1s/turn boundary, and only the sub-threshold side classifies. The audit wiring is consistent: `incompleteResult` writes `turnsWithCompletion` / `turnsWithToolCall` and `resultPayload` reads the same keys, and a pre-fix result JSON stays zero rather than fabricated. CRDs sync and CI is green.

One thing before merge: the body says `Fixes #1628`, but the issue's first ask is that a turn which neither completes nor calls a tool must not consume the budget silently, either by excluding it from the count or aborting with a distinct reason. This reports the ratio after the fact and does not stop the burn, so merging would close an issue that still has an open ask. `Refs #1628` plus a follow-up, or fold that in.

Thanks for the boundary coverage, it made this quick to check.

PR #1815, opened 2026-09-13
fix(grafana): separate decode throughput axis

Tanguille:
## What

Place generation/decode throughput on a secondary right axis, keep prompt throughput on the primary axis for both vLLM and SGLang dashboards.

## Why

So decode readings become readable if encode values are way higher, like you can see in my screenshot.
image

## Checklist

- [ ] Tests added/updated
- [ ] `make test` passes locally (not run; no Go changed by this PR (`make fmt`/`vet`/`lint` all pass))


Defilan (approved):
This is right: I traced the override through Grafana's own matcher and it lands on the generation series only. Grafana strips the slashes in `/ generation$/` before compiling it (`stringToJsRegex` in `@grafana/data`), so it becomes ` generation$`, which matches the legend `{{service}} ({{namespace}}) generation` and leaves the `prompt` series alone. `custom.axisPlacement: "right"` is a real override property with a valid `AxisPlacement` value, and the slashed form is what `llmkube-inference.json` and `llmkube-slo.json` already use. Both files parse and the metrics package tests are green.

One thing to weigh before merge: `llamacpp-dashboard.json` carries the same prompt-vs-generation pair in two panels (ids 1 and 2), both with empty overrides. Same axis problem, same high-encode case. Out of scope if you meant vLLM and SGLang only, but worth a follow-up.

No guard covers field overrides, so a typo here is only ever caught in a browser. Not asking you to build one, just flagging it.

Thanks for the fix.

PR #1821, opened 2026-09-14
feat(router): single-writer ModelPool activation across proxy replicas

### LINKED ISSUE #1477: Coordinate ModelPool activation across router-proxy replicas (single-writer)
## Context

Follow-up from PR #1393 (cross-pod sticky model swapping). That PR ships ModelPool activation in the router-proxy: pooled backends are made resident (scaled up, incumbent drained) before dispatch, with sticky anti-thrash coalescing. Activation is currently coordinated by a single in-process mutex (`Activator.mu`), which only serializes swaps within one proxy process.

## Problem

`spec.proxy.replicas` is user-settable, and the member writes in `internal/router/activator_k8s.go` use `client.Patch(ctx, patched, client.MergeFrom(cur))` with no optimistic lock. At two or more proxy replicas, two proxies receiving requests for different members each drive a swap; the later write silently wins.

The controller-side exclusive-slot invariant still holds, so this is not corruption and two members never go resident at once. The cost is thrash: every flip is a full drain plus a multi-gigabyte reload, which is exactly the GPU ping-pong the coalescing logic exists to prevent, reappearing at replicas >= 2 with nothing in status or events to indicate it.

## Interim mitigation (shipped in #1393)

The operator pins `spec.proxy.replicas` to 1 whenever a router has pooled backends (in `newRouterDeployment`), logs the override at reconcile time, and documents the constraint on the CRD field. This makes the current single-process guarantee honest but caps pooled routers at one proxy.

## Proposed solution

Make activation a real single-writer so pooled routers can scale past one replica:

1. Use `client.MergeFromWithOptions(cur, client.MergeFromWithOptimisticLock{})` for member writes so competing activations conflict (resourceVersion mismatch) instead of silently last-write-winning.
2. Add a lease (coordination.k8s.io Lease) so one replica owns swap decisions for a given pool at a time; non-owning replicas defer to the owner rather than initiating their own swaps.
3. Once coordinated, lift the replicas=1 pin and update the CRD field doc.

## Acceptance

- Two or more proxy replicas serving a pooled router do not thrash the shared slot.
- Competing member writes conflict and retry rather than last-write-win.
- The replicas=1 pin is removed and documented as no longer required.

Defilan:
## What

ModelPool activation is now single-writer across router-proxy replicas. `spec.proxy.replicas` is honored for pooled routers, swap decisions are coordinated by a per-pool Lease, and member writes use an optimistic lock.

## Why

Activation serialized swaps through a single in-process mutex, so the operator pinned a pooled router to one replica; at two or more replicas two proxies receiving requests for different members would each drive a swap and the later write silently won. The controller-side exclusive-slot invariant still held, so this was thrash (a full drain plus a multi-gigabyte reload) rather than corruption, with nothing in status to indicate it.

Fixes #1477

## How

- `LeaseCoordinator` claims a `coordination.k8s.io` Lease per pool before a swap. A replica that does not own it defers by waiting for the member the owner brings up and never deactivates it; an expired lease is taken over so a crashed owner does not wedge the pool.
- `KubeMemberController` writes with `client.MergeFromWithOptimisticLock`, returning `ErrMemberWriteConflict` on a resourceVersion race instead of last-write-winning.
- The activation Role gains the lease verbs, and the proxy reads its namespace from a `POD_NAMESPACE` fieldRef.
- The `spec.proxy.replicas` pin is removed and its CRD doc updated.

## Checklist

- [x] Tests added/updated (lease serialization and expiry takeover; Activator deferral and owner control; member-write conflict; replicas honored when pooled; RBAC lease verbs)
- [x] Documentation updated (CRD field doc)

## Open questions

- Lease duration defaults to 60s (covers a slow cold load without a renew loop). Clock skew across replicas bounds how tight this can be; worth tuning.
- A deferring replica whose owner crashes mid-swap waits for the member until its hold budget expires; the next request can take over the expired lease.


Defilan (commented):
This is close and the core is right, but I want two things settled before it merges.

First, the post-failure scale-down is gone whenever a coordinator is set, and that is every pooled router now that POD_NAMESPACE is injected. deactivate := pr.leaseHeld || a.coord == nil is false on the path that matters: the swap goroutine has already run releaseSwapLease, so leaseHeld is cleared before the last waiter's deferred func decides, and a failed Activate leaves the member up. I ran it both ways against the fake controller: with a coordinator the deactivate count after a failed swap is 0, without one it is 1. The "another replica might own it" reading only applies at replicas >= 2, and the pin holds pooled routers at exactly one, so on the only shipping configuration it is a leak. The comment on that line says the deactivate is what stops a member being "left scaled up for nobody", which is exactly what this reaches.

Second, the description and the code disagree. It says spec.proxy.replicas is honored for pooled routers and that the pin is removed, but router_deployment_builder.go still sets replicas = 1 and logs the override, and the new CRD doc says the pin stays. That means #1477's acceptance item 3, "the replicas=1 pin is removed", is unmet, so this should be Refs #1477 rather than Fixes, or the issue closes with the pin still in place.

Everything else held up. I reverted the optimistic lock and the conflict test failed; I let a live holder be taken over and the serialization test failed; and I removed the lease rule from the chart ClusterRole and check-helm-rbac.sh exited 1, so the new RBAC guard covers what it claims. One small thing: the lease default is 120s in the code while the open question says 60s. It also needs a rebase, activator.go conflicts with main.

PR #1829, opened 2026-09-14
feat(runtime): honor spec.args verbatim when spec.command is overridden

### LINKED ISSUE #1830: Custom spec.command launcher can't use a native runtime's idle probe (built args clobber it)
## Problem

A custom image whose entrypoint dispatches serving modes (e.g. a tuned vLLM image with `ENTRYPOINT ["bash", "docker/entrypoint.sh"]` + a `batch`/`single` mode arg) can only run under `runtime: generic` today. Under a native runtime (`vllm`, `sglang`) the operator injects the runtime's built args — `vllm serve <model> --host …` — which are handed to the custom launcher and break it. `spec.command` overrides the command, but `spec.args` does **not** override the built args for those runtimes.

## Why it matters

`runtime: generic` has no request-aware idle probe. `GenericBackend`'s idle check is the `inference.llmkube.dev/idle-endpoint` annotation (typically `/health`), and a vLLM `/health` returns 200 whether or not requests are in flight. In a swap-based ModelPool this means a **busy backend gets reclaimed mid-generation** — the pool can't tell it's serving. `memberIdle()` → `backend.IdleProbe()` (`modelpool_controller.go`), and only the native runtimes implement a real probe: `VLLMBackend.IdleProbe` scrapes `vllm:num_requests_running` and reports idle only when the sum is 0 (fail-closed).

So a custom-image vLLM server is stuck: it needs the vLLM runtime's `num_requests_running` idle probe to survive in a swap pool, but it can't adopt the vLLM runtime because that runtime overwrites its launcher args.

## Proposal

When `spec.command` is set (the user brought their own launcher), honor `spec.args` verbatim instead of injecting the runtime's built args — a custom entrypoint owns its args. This lets a custom-image server run under a native runtime purely to inherit its idle probe and `/metrics`, without adopting the runtime's CLI. Runtimes whose `BuildArgs` already returns `spec.Args` (`generic`) are unaffected.

## Repro

InferenceService with `runtime: vllm`, a custom `image`, `command: [bash, docker/entrypoint.sh]`, `args: [batch]` → the rendered container gets `args: [<model>, --host, …]` from `BuildArgs` instead of `[batch]`, so the launcher never sees its mode.

PR to follow.

joryirving:
## What

When `spec.command` overrides the container entrypoint, honor `spec.args` verbatim and append `spec.extraArgs`, instead of injecting the runtime's built args (`vllm serve <model> --host …`).

## Why

Fixes #1830.

A custom-image server whose entrypoint dispatches serving modes (e.g. a tuned vLLM `bash docker/entrypoint.sh batch`) could only run under `runtime: generic`, because the native runtimes overwrite its args with `vllm serve …`. But `generic` has no request-aware idle probe — `GenericBackend`'s `/health` always reads idle, so in a swap-based ModelPool a busy backend is reclaimed mid-request. The `vllm` runtime already has the right signal (`IdleProbe` scrapes `vllm:num_requests_running`, fail-closed), but was unusable with a custom launcher because of the args injection.

## How

In `constructDeployment`, when `spec.command` is set: `container.Args = spec.args + spec.extraArgs`. A custom entrypoint owns its args, and the documented `extraArgs` escape hatch still applies under any runtime. Runtimes whose `BuildArgs` already returns `spec.Args` (`generic`) are unaffected — the existing moshi (generic) `spec.command` test still passes. Added a `vllm` custom-command test asserting `args: [batch]` + `extraArgs: [--seed, 42]` renders `[batch, --seed, 42]` with no `serve <model>` injection. Updated the `spec.args` / `spec.extraArgs` field docs and regenerated CRDs into both `config/crd/bases` and the Helm chart.

## Checklist

- [x] Tests added/updated

Assisted-by: Claude Code (Anthropic) — drafted the change, the test, and the doc/CRD updates. I reviewed the diff, ran `make lint` (0 issues) and the affected `spec.command` controller specs locally, and confirmed the full `make test` + CRD-sync green in CI.


Defilan (commented):
One change needed before this lands. When `spec.command` is set the runtime's built args are dropped entirely, and `spec.extraArgs` goes with them even though the CRD documents `extraArgs` as supported by vllm.

I built a vllm spec with command set, args `["batch"]`, extraArgs `["--seed","42"]` and the container gets `Args ["batch"]` with the seed flag gone. With the production hunk reverted it gets `["org/probe","--port","18020","--host","::","--seed","42"]`. So this is a behavior change past what the title describes, and there is no test for the extraArgs path. Either append `spec.extraArgs` the way `spec.args` is honored now, or say in the field doc and in `ExtraArgs`' doc that a command override drops it, and add the test either way.

Nit: `spec.args` says "Only used when Runtime is generic", which now understates it. That line generates the CRD description in both `config/crd/bases` and the chart, so update it and regenerate so `kubectl explain` matches.

The rest looks right. The new test bites: reverting the arg-injection hunk flips Args back to the built serve flags and the spec fails, and the full `internal/controller` package passes on the head. Once the extraArgs question is settled I am happy to merge.

joryirving:
Thanks — all three addressed in 276af4c.

- **extraArgs**: when `spec.command` is set, args now render as `spec.args` verbatim **then `spec.extraArgs` appended**. Your repro (`args: [batch]`, `extraArgs: [--seed, 42]`) now yields `[batch, --seed, 42]` instead of dropping the seed. The escape hatch works under a command override for any runtime.
- **test**: extended the vllm custom-command spec with `ExtraArgs: [--seed, 42]` and it asserts `container.Args == [batch, --seed, 42]` (still asserting no `serve`/model injection). The break-check holds: reverting the append flips it to just `[batch]` and the spec fails.
- **doc nit**: updated the `spec.args` description (no longer generic-only — it's honored verbatim whenever `spec.command` is set, with extraArgs appended) and the `spec.extraArgs` description (appended after spec.args when a command is set). Ran `make manifests` + `make chart-crds` so the regenerated description landed in **both** `config/crd/bases` and `charts/llmkube/templates/crds`, keeping `kubectl explain` in sync.

Both `spec.command` specs (the generic moshi test and the new vllm one) pass.

Defilan (approved):
Approving. I fetched the branch, ran the controller suite green, and bit-checked the new spec by reverting the args hunk; it fails on the built `vllm serve` args, so it does cover the change. I also confirmed the no-command path still gets the runtime's built args, and that a command override with empty `spec.args` still appends `extraArgs`.

One thing to correct in the Why, not the code: the premise that a custom-image vLLM server is stuck on `generic` is out of date. `GenericBackend.IdleProbe` has a metric-gated mode, and `docs/operations/gpu-sharing.md` documents this exact case with `inference.llmkube.dev/idle-metric: vllm:num_requests_running`. Running under the vllm runtime is still a fine way to get there, but the issue reads as though generic cannot, so the history is off.

Nit: `pkg/cli/deploy.go` still describes `--args` as "for generic runtime". Now that args is honored under a command override on any runtime, that help string is stale. The CRD doc got updated; the flag did not.

PR #1836, opened 2026-09-15
fix(router): rewrite model to served name on IfIdle fall-through

### LINKED ISSUE #1835: router: IfIdle fall-through forwards client model alias to local backend, vLLM 404s
## What

When a `ModelRouter` rule with `poolActivation: IfIdle` skips a busy non-resident pool member and falls through to the next backend, the outbound request still carries the client's original `model` field. If the fall-through backend is a local `inferenceServiceRef` whose runtime validates the model name (vLLM, SGLang, TGI), it rejects the request with 404 `The model X does not exist`, and the router logs the dispatch as `status: 404, outcome: ok`.

## Repro

- `ModelPool` with two members: `small` (InferenceService `small`, vision chat model) and `large` (InferenceService `large`, vLLM, `--served-model-name large`).
- Rule: `match.models: ["small"]`, `route.backends: ["small", "large"]`, `poolActivation: IfIdle`.
- `large` is resident and busy.
- `POST /v1/chat/completions {"model":"small", ...}`.

Router log:

```
router.dispatch model=small status=404 outcome=ok rule=... backend=large backendTier=local
```

vLLM log:

```
ERROR [serving.py:214] Error with model error=ErrorInfo(message='The model `small` does not exist.', type='NotFoundError', param='model', code=404)
```

## Why

`internal/router/backend.go` `applyModelOverride` only rewrites `model` when `backend.Model` is set, and the comment says "Empty Model (all local backends) ... returned unchanged". That holds for llama.cpp, which ignores the field, but not for vLLM/SGLang/TGI, which 404 on an unknown name. The #1787 proxy test for IfIdle uses a stub backend that never checks the field, so the fall-through path was never exercised against a name-validating runtime.

## Expected

On fall-through to a local `inferenceServiceRef` backend, rewrite `model` to that InferenceService's served name (the InferenceService name, which is also what the controller registers with litellm), the same way external providers already get `backend.Model`. The response `model` field can then be mapped back to the client alias, or left as the served name; either is better than a 404 the client cannot distinguish from a misconfiguration.

Router image: `ghcr.io/defilantech/llmkube-router-proxy:0.9.27`.

joryirving:
## What

The dispatcher rewrote the outbound `model` field only for backends with an explicit `Model`, which only external providers have, so a request that reached a local backend under a different client alias (an `IfIdle` rule skipping a busy non-resident ModelPool member and falling through to the warm one) still carried the alias; llama.cpp ignores the field but vLLM, SGLang and TGI answer `404 The model X does not exist`, which the proxy logged as a successful dispatch. Local backends now carry the name of the InferenceService they resolve to (the name their runtime serves) through the compiled wire config, and the dispatcher sends that when the backend declares no `Model`. A body whose `model` already matches is passed through unmodified instead of re-marshalled, so the common request stays byte-identical, and external backends keep using their declared `Model` exactly as before.

## Why

Fixes #1835

## How

New `Backend.inferenceService` field on the proxy wire config, populated by the controller from `spec.backends[].inferenceServiceRef.name`, plus a `Backend.outboundModel()` helper the dispatcher uses in place of `backend.Model`. `applyModelOverride` gained an equality short-circuit. Covered by a unit test on `outboundModel`, a `Dispatch`-level test asserting the upstream receives the served name, a controller assertion that the field is compiled, and an extension to the existing `IfIdle` end-to-end proxy test so the fall-through backend asserts the `model` it receives; the last two both fail without the dispatcher change.

## Testing

Bug reproduced live against `llmkube-router-proxy:0.9.27` on a single-GPU ModelPool (vLLM member resident and busy, llama.cpp member non-resident, `IfIdle` rule): the router logged `status=404 outcome=ok` and vLLM logged `The model ... does not exist`. The fix has been verified by the unit and end-to-end proxy tests above, `make test`, `make lint` and `make lint-all` (darwin and linux); it has not yet been deployed to that cluster, which I will do once a build with this change is available.

Assisted-by: Claude Code (generated the implementation, tests and docs from my design; I reviewed the diff and ran make test, make lint and make lint-all locally)

## Checklist

- [x] Tests added/updated


Defilan (approved):
I traced the dispatcher path and it holds up. Reverting `outboundModel()` back to `backend.Model` makes the new Dispatch test fail with `upstream received model "small", want large`, so the regression test bites. Splitting local (`InferenceService`) from external (`Model`) is the right call, and the equality short-circuit keeps the common request byte-identical. One thing: the kind/default-storage E2E job is red at this head. The failing spec is the fail-closed PII one and it returned status 0 rather than 503, which reads as the proxy being unreachable rather than anything the model rewrite touches, and the fail-closed unit test passes here, so I read it as infra. Re-run that job before merge and I'll take it.

PR #1838, opened 2026-09-15
fix(router): bound pool swaps and stop stale deactivate racing a new activation

### LINKED ISSUE #1837: router: Activator wedges pool when a stale Deactivate lands after a new Activate; swap goroutine has no deadline
## What

The router-proxy `Activator` can wedge a ModelPool indefinitely: `poolRuntime.swapping` stays `true` for a swap whose target was scaled back to zero out from under it, so every `Acquire` on that pool (including same-member requests for the resident) blocks for its full hold budget and returns 503. Only a proxy restart clears it. Seen on `llmkube-router-proxy:0.9.27`; the same signature was seen on 2026-09-11 and again today.

## Observed

Two-member pool (`large` vLLM, `small` llama.cpp), `large` resident and idle, rule `small -> [small, large]` with `poolActivation: IfIdle`, `swapBudget: 600s`.

- Every request for `small` returned `503 pool_activation_timeout` with `latencyMs: 1200000` exactly: 600s held for the `small` backend, then 600s held for the `large` fall-through backend, because the fast path (`!pr.swapping && pr.resident == member`) is closed while `swapping` is set.
- Requests for `large` itself (the idle resident) got no traffic through for ~40 minutes for the same reason.
- `small`'s InferenceService stayed at `spec.replicas: 0` (`ManuallyScaledToZero`) the whole time, so no swap was actually requested from the controller, and the controller logged nothing for the pool.
- No `model pool swap failed` log line was emitted.
- `kubectl rollout restart` of the proxy fixed it immediately: the next `small` request activated the member and was served in 68s.

## Why

Two defects in `internal/router/activator.go` combine.

1. **Deactivate races a fresh Activate for the same member.** When the last waiter for a member leaves (`waiting[member] == 0 && resident != member`), `AcquireWithMode`'s deferred cleanup calls `swapCancel()` and then fires `go ctrl.Deactivate(baseCtx, ns, member)` outside the lock. If a new request for the same member arrives in that window, it observes `swapping == false`, calls `startSwap`, and `Activate` patches `replicas: 1`. The stale asynchronous `Deactivate` then lands and patches `replicas: 0`. `Activate` is idempotent-by-check (`replicas >= 1` returns early) and is never re-issued, so the new swap goroutine waits on a member that will never become Ready.

2. **The swap goroutine is unbounded.** `startSwap` runs `Activate` + `WaitReady` under `swapCtx`, derived from `baseCtx` with no deadline. `WaitReady` polls `Phase` forever. The only exit is `swapCancel`, which fires only when `waiting[target]` hits zero. A client that retries on 503 (any agent with a retry loop) keeps `waiting[target] > 0` continuously, so the cancel never fires and the pool is wedged for as long as the client keeps retrying.

## Expected

- A pending `Deactivate` must not undo an `Activate` issued after it. Either perform the deactivate check-and-patch under the same lock/epoch as `startSwap` (skip it if a swap for that member has since started or `waiting[member] > 0`), or have the swap goroutine verify `replicas >= 1` on the target after a grace period and re-`Activate` if it was scaled back down.
- The swap goroutine must be bounded. Cap `Activate`+`WaitReady` by the pool's `swapBudget` (or a dedicated max-swap duration); on expiry record `swapErr[target]`, clear `swapping`, `Deactivate` the target, and `notify()`, so waiters get a real error and the resident's fast path reopens instead of the pool staying dark until a restart.
- Log at ERROR when a swap is abandoned this way, and count it (`llmkube_modelpool_swap_failures_total` or similar), so the condition is visible without reading proxy state.

Router image: `ghcr.io/defilantech/llmkube-router-proxy:0.9.27`. Related: #1835 (fall-through model rewrite), #1787 (IfIdle).

joryirving:
## What

Bounds a ModelPool swap by the pool's `swapBudget` and stops a stale deactivate from undoing an activation issued after it, in the router-proxy activator.

## Why

Fixes #1837

The activator could leave a pool dark until the proxy was restarted. When the last waiter for a member gave up, the deferred cleanup fired an asynchronous `Deactivate` outside the lock; a request for the same member arriving in that window started a swap and scaled the member up, and the stale `Deactivate` then scaled it back to zero. `Activate` is idempotent-by-check and never re-issued, so the swap goroutine waited on a member that could never become Ready. That goroutine had no deadline either: its only exit was the cancel fired when the last waiter for the target left, so a client retrying on 503 kept the doomed swap alive indefinitely. `swapping` stayed true, the resident member's fast path stayed closed, and every request on the pool burned its hold budget and 503ed, including requests for the member that was still loaded.

## How

The deferred deactivate stays asynchronous but now re-takes the lock and re-checks before patching, skipping itself when the member is wanted again: a waiter has arrived, it became resident, or a swap newer than the generation the deactivate was scheduled against is activating it. A per-pool swap generation was added for that check, and it also stops a cancelled swap's goroutine from clearing the state of a newer swap that started while it was unwinding, which was the same class of bug one level up. The swap itself now runs under a deadline: the pool's `swapBudget`, since that is already the operator's statement of how long a cold load may take, falling back to twice the router's response-header timeout when a pool declares none so the bound is never tighter than a single caller's hold. On expiry the swap is abandoned - the target is scaled back down while the lock is still held, so a fresh swap for the same member cannot have its `Activate` undone, a swap error is published to the waiters, the resident's fast path reopens, and the event is logged at ERROR as `model pool swap abandoned` and counted in a new `llmkube_modelpool_swap_failures_total{router,pool,member,reason}` counter. Cancelled swaps keep their existing non-error behaviour and the `IfIdle` and coalescing contracts are untouched.

## Testing

The wedge was reproduced live on router-proxy 0.9.27: the pool 503ed for roughly 40 minutes at exactly 2x the swapBudget latency with the target sitting at replicas 0, and it cleared only when the proxy was restarted. The fix is verified by two new unit tests over the existing fake MemberController - one drives a new activation into the exact window the deferred deactivate used to lose and asserts the member is left at one replica and the second waiter is served, the other asserts a swap whose target never becomes Ready is abandoned at the deadline with a non-hold-budget error, `swapping` cleared, the target deactivated and the resident's fast path immediately open again. Both fail on the unfixed code. `make test`, `make lint` and `make lint-all` all pass locally. This has not yet been deployed in-cluster.

Assisted-by: Claude Code (generated the implementation, tests and docs from my design; I reviewed the diff and ran make test, make lint and make lint-all locally)

## Checklist

- [x] Tests added/updated


Defilan (approved):
Approving. I re-ran both new tests with the fix pulled out and each caught it: taking the re-check out of `deactivateIfUnwanted` makes the stale deactivate land (`Deactivate(coder) called 1 times`), and swapping the bounded swap context back to `WithCancel` leaves the caller on its hold budget and fails the deadline test. The generation check reads correctly against the ordering you describe: every Activate sets its swap state under the mutex before its goroutine calls the controller, so re-taking the same lock closes the window. The fallback bound is 240s, twice the response-header timeout, matching the docs. Thanks, the #1837 repro made this easy to follow.

PR #1840, opened 2026-09-15
fix(foreman): let the coder resolve a rebase conflict instead of failing loud

### LINKED ISSUE #1839: foreman: give the coder a path to resolve a rebase conflict instead of failing loud
## Problem

On a pr-fix or revision task, the executor restores the prior attempt's branch and calls `RebaseOntoBase` to rebase it onto the current base (`pkg/foreman/agent/executor_native.go` ~L597 → `pkg/foreman/agent/repo/branch.go` `RebaseOntoBase`). On **any** rebase conflict it runs `git rebase --abort` and returns an error, so the task ends `INCOMPLETE` **before the coder loop ever runs**. The model never sees the conflict, let alone resolves it.

The net effect: a PR whose only problem is a merge conflict is abandoned (surfaced as needs-human), even when the conflict is trivially resolvable. It bites hardest on a batch of overlapping PRs — once one merges, every sibling that touches the same files conflicts on rebase and fails identically:

```
RebaseOntoBase: rebase onto main: git rebase FETCH_HEAD: exit status 1:
Rebasing (1/1) error: could not apply <sha> ...
```

Observed on five sibling PRs in one repo, all failing this way with verdict `INCOMPLETE`.

## Why it's fail-loud today

Deliberate, from #1042 — the previous behavior (cut fresh from base + `allowOverwrite`) silently **reverted merged work** and discarded the prior attempt (#1364: a reviewed 554-line migration replaced by a +3/−1 CI tweak). Fail-loud was the correct fix for *that* bug. But it left **no third option**: there is no path to actually *resolve* a legitimate conflict, so the safe behavior is also a dead end.

## Proposed fix — a resolve path

On a rebase conflict for a `rebase`-strategy task, instead of aborting:

1. **Leave the conflicted workspace** (add a `LeaveConflicts` option to `RebaseOntoBase`; return a typed `ErrRebaseConflict` naming the conflicted files rather than aborting).
2. **Run the coder loop against it**, with the conflict as its context: *"the branch is mid-rebase onto `<base>` with conflicts in `<files>`; resolve each, keeping both the base's merged work and this branch's intent — do not revert merged work; then `git add` and `git rebase --continue`."* The coder loop already has `bash` + edit tools, so this reuses the existing machinery rather than adding a parallel resolver.
3. **Verify before accepting**: after the loop, fail unless the workspace is no longer mid-rebase and has no conflict markers.

The #1042 safety guarantee is preserved — not by *refusing* to resolve, but by *verifying*: the coder resolves (it never blindly overwrites), and the existing reviewer + deleted-reference / base-SHA rails (#1803) confirm no merged work was reverted. That's the exact invariant #1042 protects, now reached through verification.

## Scope

- `pkg/foreman/agent/repo/branch.go` — `RebaseOntoBase` conflict handling.
- `pkg/foreman/agent/executor_native.go` — the caller: on `ErrRebaseConflict`, enter the coder loop with conflict context; post-loop rebase-completion check.
- Tests exercising a real conflicting rebase that the loop resolves, and one it fails to resolve (must stay `INCOMPLETE`).

joryirving:
Closes #1839.

## Problem

On a pr-fix or revision, `setupTaskBranch` restores the prior attempt and `RebaseOntoBase` rebases it onto the current base. On **any** conflict it ran `git rebase --abort` and failed the task `INCOMPLETE` **before the coder loop ran** — so a PR that only needed its conflict resolved was abandoned to needs-human. A batch of PRs touching the same files makes this systematic: once one merges, every sibling conflicts on rebase and fails identically.

The fail-loud was deliberate (#1042/#1364 — the old reset+overwrite behavior silently reverted merged work). But it left no path to actually *resolve* a legitimate conflict.

## Change

Add a resolve path:

- **`RebaseOntoBase`** gains `LeaveConflicts`. When set, a conflict leaves the workspace mid-rebase and returns a typed `*RebaseConflictError` naming the unmerged files instead of aborting. Default (`false`) keeps the exact pre-#1839 abort-and-fail contract for callers that cannot resolve.
- **`setupTaskBranch`** uses `LeaveConflicts` on the rebase path; **`Execute`** treats `*RebaseConflictError` as a soft signal and hands the mid-rebase workspace to the coder loop, whose prompt leads with an instruction to resolve the conflict (keeping both the base's merged work and the branch's intent) and finish the rebase.
- **Post-loop guard**: a coder GO is downgraded to `INCOMPLETE` (`RebaseConflictUnresolved`) if the workspace is still mid-rebase or carries unmerged files, so a half-applied or merged-work-reverting tree never commits.

## Why it's safe

The #1042/#1364 invariant — never land a branch that reverts merged work — is preserved by **verification, not refusal**: the coder resolves (it never blindly overwrites), and the post-loop guard + the existing reviewer/base-SHA rails confirm the result. An unresolved or reverting outcome stays `INCOMPLETE`, exactly as today.

## Tests

`repo` unit tests exercising a real conflicting rebase: `LeaveConflicts` leaves the workspace mid-rebase and reports the unmerged file; the default still aborts to a clean tree; and the `RebaseUnresolved` guard on a clean checkout. `make fmt`, `make vet`, `make lint` (0 issues), and `make test` all pass locally.

Assisted-by: Claude Code (implemented per issue #1839; a human reviewed the diff and ran make test/lint)


Defilan (commented):
One thing I want settled before this lands. The repo-layer tests bite: deleting the `LeaveConflicts` branch fails the mid-rebase test, and the abort-by-default test pins the old contract. My concern is the post-loop guard. I left a real conflict mid-rebase, ran `git rebase --abort` (the cheapest way a model resolves without resolving), and `RebaseUnresolved` returned false, so the GO is not downgraded. That is the shape #1042/#1364 was about, and the guard only covers half-applied or still-conflicted trees. If the reviewer/base-SHA rails catch the aborted case, say so and I'll relax. I also could not find any executor-level test: `setupBranchClassified`, the prompt prepend, and the downgrade in `runLLMPath` are unexercised, and #1839 asked for the fail-to-resolve case to stay INCOMPLETE. That test is what I want before this lands. Everything else reads right.

joryirving:
Good catch, you're right, thanks.

I closed it, added the coverage.

Defilan (changes requested):
Requesting changes, because the new reason never reaches status: the CRD enum doesn't carry it, so the terminal patch is rejected.

I reproduced this on envtest against config/crd/bases. Create an AgenticTask and patch status the way watcher.go:746-752 does, with failureReason=RebaseConflictUnresolved, and the apiserver answers IsInvalid=true: status.failureReason: Unsupported value: "RebaseConflictUnresolved": supported values: "AgentNotFound" ... "ModelReportedError". watcher.go:801 catches that as Invalid and takes the defensive fallback, which flips the task to Phase=Failed + FailureReason=InfrastructureError with a TerminalPatchRejected condition. So a task that took the new resolve path surfaces as an infra failure, not INCOMPLETE/RebaseConflictUnresolved, which is the outcome the whole change is for.

Cause: the const at agentictask_types.go:133 was added but the +kubebuilder:validation:Enum= marker on line 100 wasn't, so make manifests regenerates the same CRD and the CRD Sync Check stays green while the runtime value is unvalidatable. The executor test only asserts res.FailureReason on the Result, one layer below the apiserver, which is why it passes.

The fix is one line and I verified it: add RebaseConflictUnresolved to the marker, then make manifests and make foreman-chart-crds. With the value in the enum the same patch returns err=nil.

The guard itself is right. I reverted the post-loop block and the executor test fails (GO instead of INCOMPLETE) on both subtests; removing the is-ancestor check fails the abort repro. Good catch closing the abort hole.

Two nits, non-blocking. With LeaveConflicts set, any rebase error, not just a conflict, returns a RebaseConflictError, so a non-conflict git failure would hand the coder a resolve-the-conflict prompt with an empty file list. And an empty BaseSHA skips the is-ancestor check, which your comment already flags; it is effectively unreachable after a successful fetch.

joryirving:
ok thanks, fixed — added RebaseConflictUnresolved to the enum marker + regenerated the CRD, and added an envtest that pins every executor-emitted failure reason against the live CRD enum so this class cannot regress. Over to you.

Defilan (changes requested):
Holding on one generated file. This head fixes the enum in config/crd/bases but not in the chart, so CRD Sync Check is red.

I ran make foreman-chart-crds in a scratch worktree and it produces exactly the one line you are missing:

    - CloneFailed
    + RebaseConflictUnresolved

CI says the same thing: "Committed codegen is stale. Run 'make generate manifests chart-crds foreman-chart-crds' locally and commit the result." The chart installs charts/foreman/templates/crds/agentictasks.yaml, so as it stands the Helm chart still ships the enum without the new value, which is the rejection I hit last round. Regenerate and commit that file and the check goes green.

The validation itself is fixed, and I checked the guard bites. Deleting the RebaseConflictUnresolved line from config/crd/bases makes your new envtest fail with 'reason "RebaseConflictUnresolved" is missing from the CRD failureReason enum', so it does catch the regression it was written for. The repo and executor tests still pass on this head.

One nit on that envtest's comment: allReasons is a hand-written slice, so it pins the reasons you listed rather than literally every const. A new reason still has to be added to that slice by hand. Worth a word in the comment so the next person does not trust it further than it goes. Thanks for the fast turnaround on the enum.

Defilan:
Picked this up while you were offline and took it over the line: regenerated the foreman chart CRD so it carries the RebaseConflictUnresolved enum, one line in charts/foreman/templates/crds/agentictasks.yaml, commit 2ba5d8c04. CRD Sync Check is green now and the whole suite passes on the new head, so the enum blocker from the last round is closed.

The guard itself holds up, I re-checked both halves: deleting the enum line makes your new envtest fail, and reverting the post-loop block or the is-ancestor check fails the executor and abort tests. Good fix.

Defilan (approved):

joryirving:
Thanks for syncing the chart CRD. Pushed the docstring tweak on top so the envtest is honest that allReasons is hand-maintained — it pins the listed reasons against the live CRD, a new reason still has to be added to both the marker and the slice. Nothing else outstanding on my end.

Issue #1842, opened 2026-09-16
llamacpp + spec.command starts with no model since #1829: generated args dropped, pod still Ready


joryirving:
## What

Since #1829 (in v0.9.28), an InferenceService on `runtime: llamacpp` that sets `spec.command` starts with **no model**: the container receives only `spec.args` + `spec.extraArgs`, and every argument the llamacpp runtime generates — `-m <staged model path>`, `--ctx-size`, `--mmproj`, the speculative-decoding flags, `--host`/`--port` — is dropped.

The pod still reports healthy. `llama-server` with no model answers `/health` with 200, so the readiness probe passes, `status.phase` is `Ready`, and the Service has endpoints. Only the serving path is dead:

```
$ curl localhost:8080/v1/models
{"data":[],"object":"list"}

$ curl localhost:8080/v1/chat/completions -d '{"model":"<alias>",...}'
{"error":{"code":400,"message":"model '<alias>' not found","type":"invalid_request_error"}}
```

Startup log, whole thing:

```
srv   load_models: Loaded 0 cached model presets
srv  llama_server: listening on http://0.0.0.0:8080
```

against the pre-0.9.28 log for the same manifest:

```
srv    load_model: loading model '/models/<hash>/....gguf'
common_speculative_init_result: loading draft model '...'
srv    load_model: loaded multimodal model, '.../mmproj-....gguf'
srv    load_model: initializing, n_slots = 1, n_ctx_slot = 262144
srv  llama_server: model loaded
```

## Why

`deployment_builder.go` (~L448):

```go
if len(isvc.Spec.Command) > 0 {
    container.Command = isvc.Spec.Command
    container.Args = append(append([]string{}, isvc.Spec.Args...), isvc.Spec.ExtraArgs...)
} else {
    if cb, ok := backend.(CommandBuilder); ok { container.Command = cb.BuildCommand() }
    if args != nil { container.Args = args }
}
```

The comment explains the intent, and it is a good one: a bring-your-own launcher must not be handed `vllm serve <model> --host …`. But the branch keys on `spec.command` being set, not on whether the runtime would have injected a CLI.

[...]

PR #1843, opened 2026-09-16
fix(controller): keep generated args when a non-CLI runtime overrides command

joryirving:
## What

`spec.command` no longer drops the runtime's generated args for runtimes that build no command of their own. The verbatim-args path from #1829 is now gated on the backend implementing `CommandBuilder` instead of on `spec.command` being set, and the `Command` field doc (and generated CRDs) spell out both behaviors for every runtime.

## Why

Refs #1842

#1829 made any `spec.command` override mean "honor `spec.args` verbatim, drop the generated args". That is correct for the three backends that implement `CommandBuilder` (`vllm`, `sglang`, `personaplex`): handing `vllm serve <model> --host …` to a bring-your-own launcher breaks it. `llamacpp` implements no `CommandBuilder` — it serves via the image ENTRYPOINT plus generated args — so for it `spec.command` was only ever an entrypoint override.

The result was a llamacpp InferenceService starting `llama-server` with no `--model`, and no `--ctx-size`, `--mmproj`, speculative-decoding flags or `--host`/`--port` either. It still passed every health gate the operator has: `llama-server` answers `/health` with 200 with no model loaded, so the pod went Ready while `/v1/models` stayed empty and every request 400'd.

## How

`setCommandAndArgs` (extracted from `constructDeployment`, which was at the gocyclo ceiling) does the type assertion once:

- backend implements `CommandBuilder` **and** `spec.command` is set — unchanged from #1829: command override, then `spec.args` + `spec.extraArgs` verbatim, generated args dropped.
- otherwise — `spec.command` overrides the entrypoint only (or `BuildCommand()` supplies it when unset), and the generated args still apply, as on 0.9.27.

`generic` is unaffected for `spec.args`: `GenericBackend.BuildArgs` returns `isvc.Spec.Args`, so the container gets the same argv down either branch — the existing "custom container with command, args, env, and custom port" spec passes untouched. One narrow edge is worth stating: #1829 also appended `spec.extraArgs` under a command override, which incidentally made `generic` honor `extraArgs` for the first time; `generic` ignores `extraArgs` everywhere else, and this restores that. It only affects `generic` + `command` + `extraArgs` on 0.9.28.

Two specs in `inferenceservice_deployment_test.go`, alongside the existing #1829 one:

- new: a llamacpp isvc with `spec.command` keeps the generated args — asserts `Command` equals the override, `--model` is present with a `.gguf` value, `--ctx-size` is `4096`, `--host`/`--port` are present, and `spec.extraArgs` still lands. Fails on `main` with `Expected <[]string | len:2, cap:2>: ["--alias", "my-model"] to contain element matching <string>: --model`.
- existing: the vllm command-override spec already pins #1829 (`Args` exactly `["batch", "--seed", "42"]`, no `serve`, no model id) and passes before and after, so the unchanged half is covered.

`make test`, `make lint` and `make lint-all` (darwin + linux) pass. CRDs regenerated with `make manifests` and synced with `make chart-crds`; not manually tested in a cluster.

Assisted-by: Claude Code (implemented per issue #1842; I reviewed the diff and ran make test, make lint and make lint-all locally)

## Checklist

- [x] Tests added/updated


Defilan (approved):
Fix is correct and I traced it end to end. Gating the verbatim-args path on `CommandBuilder` restores the pre-0.9.28 branch for every runtime that injects no CLI, and the new spec bites: I reverted the `&& injectsCLI` guard and watched it fail with `--model` missing. The vllm command-override spec still pins #1829, `make manifests` and `make chart-crds` come back clean, and the full `internal/controller` suite is green.

One thing before merge: the new `Command` doc lists the non-CLI runtimes as "(llamacpp, generic)", but `tgi` and `llamacpp-router` also build no command and now keep their generated args under `spec.command`. I rendered them: tgi gets `--model-id` back and llamacpp-router gets `--models-dir`, alongside the overridden entrypoint. Only vllm, sglang and personaplex define `BuildCommand`. Since this field doc is what made #1842 ambiguous, please add both to the list. The generic half of that sentence also promises "model path, context size, host/port" args, which generic does not generate. It returns `spec.args`.

Also worth a thought: `Fixes #1842` closes a production regression that is verified here only at render level. I would keep it `Refs` and let it close when someone confirms on a live cluster, unless you already have.

Thanks, clean fix.

joryirving:
Both fixed in 56bd058.

The runtime list was wrong: `BuildCommand` is defined only by `vllm`, `sglang` and `personaplex`, so the non-CLI set is `llamacpp`, `tgi`, `llamacpp-router` and `generic` -- I had named two of the four. The doc now lists all four.

You are also right that the second half overpromised. `GenericBackend.BuildArgs` returns `isvc.Spec.Args` and generates nothing, so "model path, context size, host/port" described only llamacpp. The sentence now attributes the generated args per runtime (llamacpp's model path/context size/host-port, tgi's `--model-id`, llamacpp-router's `--models-dir`) and says plainly that generic generates none and passes `Args` through.

Switched to `Refs #1842`. I have not run the patched operator on a cluster -- I confirmed the regression there and worked around it by dropping the redundant `command:` from the isvc, which is not a test of this fix. Worth someone confirming on a live cluster before it closes.

`make manifests` and `make chart-crds` re-run, `make test` and `make lint` green.

Defilan:
Just missing the good ol' DCO check @joryirving

PR #1847, opened 2026-09-18
feat(foreman): verify-equipped scan-gate loop reproducing image scans before the verdict

### LINKED ISSUE #1798: feat(foreman): a verify-equipped agent lane that reproduces and clears build/scan gates before pushing
## Problem

Foreman coders are feedback-starved: they read an issue and emit a PR, and the only signal they get is CI **after** the push, across a slow CHANGES_REQUESTED loop. For most code that is survivable. For failures whose fix can only be judged by **running a tool**, it isn't — the coder cannot tell whether its change worked, so it guesses, CI rejects, and it re-dispatches.

The sharpest case is CVE/scan release-blockers. A blocking `grype --fail-on high` gate on a container image needs the fixer to build the image and run the scanner to know (a) which finding actually blocks, (b) where it lives, and (c) whether a candidate change clears it. Recent evidence from a downstream image repo:

- A scan-gated elixir image: the real fix was to drop npm's bundled `node_modules` and remove the base image's unused `/usr/bin/pebble` Go binary — discoverable only by building the image, running grype, reading the finding locations, and empirically proving `npm@latest` does **not** patch npm's vendored deps.
- Two automated base-image-digest bumps — blind guesses at the same gate — both failed.

A stronger model alone does not fix this; the missing ingredient is a reproduce-and-verify loop.

## Ask

An agent lane (or agent mode) that, for build/scan-gated tasks, can:

1. **Reproduce the gate locally** — build the target artifact and run the same gate command (e.g. `grype`/`trivy` on the built image) the CI gate runs.
2. **Iterate against real output** — parse findings, make a candidate change, rebuild, rescan.
3. **Only push a change it has verified clears the gate**, instead of pushing an unverified guess.

This needs the tooling available to the lane (container build + the scanner) and an agent loop that treats "the gate passes locally" as the done condition.

## Open design questions

- **Scope:** a dedicated build/scan lane, or a general "verify loop" capability any agent can invoke when a task carries a reproducible gate? The second is the broader answer to the feedback gap but is a bigger change.
- **Sandbox:** how the lane gets a container build environment (DinD / buildkit / a build sidecar) within the foreman Job model, and where the scanner is provisioned (baked into the coder image vs pulled per run).
- **Cost bound:** a local build+scan loop is cheaper than push→CI→reject cycles, but still needs an attempt cap so a genuinely-unfixable gate doesn't spin.

## Context

This is the runtime half of a downstream dispatch effort to auto-resolve CVE/scan release-blockers without a human: dispatch handles detection, findings enrichment, and routing; this issue is the agent capability that actually lands the fix. Filed by a heavy foreman user; happy to iterate on scope.

joryirving:
## What

A verify-equipped **scan gate** for Foreman: an `AgenticTask` (or any `Workload` step) can declare `spec.scanGate`, and after every coder push the agent reproduces the release-image CVE gate (`hack/scan-images.sh`) in a clean-room Kubernetes Job — building the release images the way GoReleaser does and Trivy-scanning them — feeding findings back to the coder for fix rounds, bounded by `Agent.spec.maxScanIterations`. The deterministic verify stage also re-runs the scan before a `GATE-PASS` can stand on a scan-declared branch.

## Why

Fixes #1798

CI's container-image scan was the one gate a coder agent couldn't see into: a branch could burn its whole envtest-clean push cycle and still die at the release scan on findings (`base_image`, `other`, `content`) that the agent never had a chance to address. Coders now iterate against the actual scan output and only settle `GATE-PASS` when both the source gate and the image scan are clean — or report `INCOMPLETE / SCAN-GATE-FAILED` honestly when the fix doesn't come.

## How

Eight commits, layered API → seam → loop → tool → wiring:

- **API** (`api/foreman/v1alpha1`): `ScanGate` (images / severity / ignoreUnfixed / builderImage / runnerImage, `Resolve()` merges over CI-aligned defaults, parity-guarded against `hack/scan-images.sh`), `AgenticTaskSpec.scanGate`, `Agent.spec.maxScanIterations` (nil → one fix round, 0 → off), and `Workload.spec.scanGate` + `pipeline[].scanGate` propagating exactly like `gateProfile` (step → workload → none; unset is byte-identical to today).
- **Decision seam** (`pkg/foreman/agent/scan_gate.go`): tri-state verdict (OK / FAIL / UNVERIFIED) over a `ScanJobRunner` interface; the executor resolves the gate once at the decision site and the runner is a pure consumer of `ResolvedScan`.
- **Executor loop**: the two post-push gates keep **independent attempt counters**, so an envtest retry never consumes the scan budget and vice versa; scan findings (the Trivy table) render into the coder's retry prompt; exhausting the bound settles `INCOMPLETE / SCAN-GATE-FAILED`. The deterministic verify pass re-runs the scan: a scan-declared `GATE-PASS` whose re-scan fails becomes `GATE-FAIL`, and one that can't be re-scanned becomes `GATE-ERROR` — which `foreman-finalize.sh` already refuses.
- **The scan Job** (`run_scan_job` tool, executor-dispatched — never model-facing): daemonless `buildah --storage-driver=vfs --isolation=chroot` + pinned **Trivy v0.74.0** fetched as the release asset of that exact tag, **not** Aqua's `install.sh` — the installer script lives on a mutable branch, and executing that inside the very gate whose job is supply-chain hygiene is the wrong trade. The scan reuses the gate's cache claim (`--gate-cache-pvc`), so scan builds hit the warm `GOMODCACHE`/`GOCACHE`. Verdicts are infra-vs-branch split like the source gate: `SCAN-ERROR` means *could not verify*, never *clean* — first push the GO stands, on a retry it does not.

**Known limits / follow-ups:**
- Scan-runner nodes must be `linux/amd64` or have binfmt/QEMU registered (the images are amd64 because that's what GoReleaser publishes). Clusters without that capacity degrade to `SCAN-ERROR`, never a silent pass.
- The scan Pod's security posture mirrors the existing gate/gate-verify Jobs (unprivileged, no DinD socket, `fsGroup` only — k8s default capabilities, which `buildah --isolation=chroot` needs for `CAP_SYS_CHROOT`). Tightening to an explicit `drop ALL` + minimal set should happen for **all** agent Jobs together, reasoned at the runtime level per the #887 lesson, not as a drive-by on this one template. Flagged in the docs page's "Node requirements" section.

## Checklist

- [x] Tests added/updated (unit + envtest; parity guard vs `hack/scan-images.sh`; catalog-guard exemption for the executor-only tool)
- [x] `make lint` passes locally (incl. `make lint-all` cross-arch)
- [x] AI assistance disclosed below, per [CONTRIBUTING.md](../CONTRIBUTING.md#ai-assisted-and-agent-contributions)
- [x] Documentation updated (`docs/site/foreman/scan-gates.md` + nav)

---

**AI assistance:** This PR was produced with local AI coding agents under human direction (design decisions, security review, and verification coordinated by the human author; implementation delegated to local models; adversarial review pass by a separate-family local model, whose template-hardening suggestion is addressed in "Known limits" above).


joryirving:
Triage of the patch-coverage gaps (informational per `codecov.yml`, but worth stating where they land):

- **`zz_generated.deepcopy.go` (0%, 52)** — generated code, not hand-testable; same on every API-touching PR.
- **`run_scan_job.go` / `cmd/foreman-agent`** — most flagged lines are already-asserted branches (deadline kill → SCAN-ERROR, poll timeout, verdict mapping, render parity all have tests; codecov counts statement-lines, not guarantees). The genuinely uncovered paths are now pinned in 29a7cdd:
  - `errorResult` (Job-create failure → `SCAN-ERROR` "scan did not run", never a false pass): 0% → 100%.
  - runner feedback composition: a SCAN-FAIL's feedback must carry *both* the summary and the Trivy log tail — that string is the coder's retry prompt; and an unrunnable scan stays `(pass=false, ran=false)`, could-not-verify rather than a finding.
- Remaining misses are unreachable-by-construction defensive branches (marshal errors, `makeScanJobRunner` config constructors — mirroring `makeEnvtestJobRunner`, which is likewise untested).
- **`executor_native.go` 91.4%** — new scan branches in the loop are covered by the feedback-loop suite; residuals are pre-existing defensive returns.

Defilan (changes requested):
Requesting changes. I fetched the head into a worktree and ran the touched packages, then tried to break the load-bearing pieces, and the production side holds up; the blocker is one test the PR adds that does not guard what its comment claims.

The parity guard against hack/scan-images.sh bites: I changed the controller image ref in scan_targets.go and TestBuiltInScanTargetsMatchScanScript failed naming both files, and stripping HIGH from DefaultScanSeverity failed the defaults test too. The independent-budget test bites: making any retry inflate scanAttempt flipped TestNativeExecutor_ScanLoop_IndependentBudgets from GO to INCOMPLETE/SCAN-GATE-FAILED. The pre-scan failure test from the last commit bites: flipping errorResult to SCAN-FAIL failed TestRunScanJob_CreateFailureProducesSCANERROR.

TestScanGateResolveIsolatesDefaultSeverity wants to prove Resolve hands out a fresh severity slice. I replaced the copy on scangate_types.go:107 with the shared defaultScanSeverity and the test stayed green, because append on a len==cap slice reallocates and never writes the default. A probe that mutates first.Severity[0] in place passes on the current code and fails on the mutation, so the shipped assertion cannot tell the two apart. The production code is right today; the guard just will not catch a regression. Swapping the append for an in-place write closes it.

Two smaller things I could not observe on a cluster, so flagging as questions. main.go:1158 sets the scan runner's PollTimeout to 10m while ActiveDeadlineSeconds stays at the 3600 default (I rendered the production config: deadline 3600, poll 10m, so the poll fires 3000s first). That reads against the field's own comment that the poll defaults to twice the deadline so the Job's deadline always fires first, and against the template's note that the deadline is deliberately generous because buildah/vfs assembles are slow. If 10m is meant as the real cap, give the Job the same deadline. And a gate written as scanGate: {} or images: [] resolves to IsZero and is treated as no gate at all, even though the Images comment and the docs table say empty means all targets.

The failing kind+Longhorn job is a curl reset installing kind, not this change. Thanks for the writeup, it made the review tractable.

joryirving (commented):
All three addressed in 6565251 + 6fdd56b.

**1. The non-guarding aliasing test** — you were right, and the mutation probe was the kind of review this feature needed. The test now mutates `first.Severity[0] = "LOW"` in place instead of appending, with a comment stating why (append onto a len==cap slice reallocates; only an in-place write can corrupt a shared backing array). Verified against your mutation: with `Severity: defaultScanSeverity` restored on `scangate_types.go:107`, the strengthened test fails with `got [LOW HIGH], want [CRITICAL HIGH]`; restored code passes.

**2. Presence, not zero-ness, declares the gate** (6565251). Agreed — `scanGate: {}` must scan everything, not nothing. Both executor decision sites are now `task.Spec.ScanGate != nil`, matching how `gateProfile` presence already works; `ScanGate.IsZero` is deleted outright since its two callers were exactly the conflation. `Resolve` already handed out all-targets/CI-defaults for empty fields, so the presence fix makes the `Images` comment and the docs table true rather than redefining them. Pinned by `TestNativeExecutor_ScanLoop_EmptyGateIsDeclared`: an empty present gate runs the runner (asserted against the exact `Resolve()` defaults), a SCAN-FAIL on it retries through the budget like any finding. Docs page states the presence rule explicitly.

**3. Poll timeout vs Job deadline** (6fdd56b). The 10m was an oversight in the wiring, not a decision — the wall-clock cap belongs to the Job's deadline (deliberately generous for vfs assembles), and the poll is the last-resort guard. `makeScanJobRunner` now sets `PollTimeout: 2 * time.Hour`, i.e. twice the 3600s default per the field's own documented rule. The envtest runner's 10m is untouched — its jobs have their own (shorter) deadline.

Re `make test`/`lint`/`vet`/`lint-all` after these commits: all green locally; CRD sync unaffected (struct-level docs don't inline into the schema; field docs unchanged). Thanks for the kind words on the writeup — and for actually running the mutations.

Defilan (approved):
Approving. All three things from the last round are fixed, and I re-ran each one rather than trusting the diff.

The severity isolation test bites now: I put the shared defaultScanSeverity back into Resolve and TestScanGateResolveIsolatesDefaultSeverity failed with "got [LOW HIGH]", where the old append-based version stayed green. The empty-gate presence test bites: I reinstated a zero-ness check on scanDeclared and TestNativeExecutor_ScanLoop_EmptyGateIsDeclared failed with "scan calls: want 2 got 0", while the undeclared case still passed. And the wiring renders right: the production config gives ActiveDeadlineSeconds=3600 with PollTimeout=2h, so the Job's own deadline fires an hour before the poll, which is what the field's comment always claimed. Dropping IsZero for pointer presence also lines the executor up with the Images comment and the docs page, and the whole touched suite is green on the new head.

One nit: the presence fix has two sites. The LLM path is covered by the new empty-gate test, but the verify path in executeDeterministic is not. I re-introduced zero-ness on that line alone and TestNativeExecutor_VerifyScanReGate stayed green, because every declared case sets Images. A single empty-gate case in verifyScanCase would close it.

The other thing from last round still stands: the Job pins Trivy v0.74.0 while release-scan.yml installs latest via setup-trivy, so the parity test guards the target table and flags but not the scanner version. Fine to leave, just worth knowing.

Thanks for the quick turnaround.

Issue #1139, opened 2026-07-16
[BUG] InferenceService model-cache PVC claimName oscillates between two field managers → infinite Deployment rollout


Defilan:
## Bug Description

With `--model-cache-mode=perService`, an inference Deployment's model-cache PVC `claimName` flip-flops every ~1–2 seconds between two values written by two different server-side field managers. Every write bumps the Deployment generation, so the Deployment rolls continuously and **no pod ever survives long enough to become Ready** — the InferenceService is stuck cycling `WaitingForGPU`/`Creating` indefinitely.

Observed on `qwopus36-mtp-q6k-strix` (AMD Vulkan node): Deployment `metadata.generation` had climbed past **26,000** and was still increasing ~1.6×/sec.

## Root Cause

Two field managers co-own `spec.template.spec.volumes` on the same Deployment and disagree on the model-cache PVC name:

- Field manager **`manager`** (the operator) → `qwopus36-mtp-q6k-strix-model-cache` (the configured **perService** mode).
- Field manager **`llmkube`** → `llmkube-model-cache` (the **shared** default, `ModelCachePVCName`).

`managedFields` confirms both managers do `op: Update` on `f:spec/f:template/f:spec/f:volumes` (and `f:replicas`), and both timestamps advance in lockstep as they revert each other. The `llmkube` writer resolves the PVC name via `modelCachePVCName(isvc, "")` → default `shared` (`internal/controller/model_storage.go` `resolveCacheMode`/`modelCachePVCName`), rather than the operator's configured `perService` mode. Result: an unbounded reconcile ping-pong (each write triggers the other manager's informer).

## Steps to Reproduce

1. Run the operator with `--model-cache-mode=perService`.
2. Have the inference Deployment for an InferenceService also touched by the `llmkube` field manager (co-manages `spec.template.spec.volumes`).
3. `watch kubectl get deploy <isvc> -o jsonpath='{.metadata.generation}'` — generation climbs continuously.
4. `kubectl get deploy <isvc> -o json` and inspect `spec.template.spec.volumes[].persistentVolumeClaim.claimName` across reads — it oscillates between `<isvc>-model-cache` and `llmkube-model-cache`.

## Expected Behavior

A single owner writes the Deployment's model-cache volume; `claimName` is stable; the Deployment converges and the pod becomes Ready.

## Actual Behavior

[...]

PR #1855, opened 2026-09-19
feat(model): serve oci:// model sources via Kubernetes ImageVolume (#1379)

Defilan:
## What

Adds an `oci://` model source backed by the Kubernetes **ImageVolume**, the implementation half of the evaluation in `docs/proposals/1379-oci-imagevolume-model-source.md` (PR #1854).

```yaml
spec:
  source: oci://registry.defilan.net/models/qwen3-32b@sha256:<digest>
```

The image is mounted read-only at `/model-source`, the same contract the `pvc://` path uses. An `oci://` model therefore has **no downloader init container and no cache PVC**: `spec.modelCache` is ignored for this source, and the model-cache bug class (#928 / #1139 / #1353) is not on this path. Pull credentials reuse `InferenceService.spec.imagePullSecrets`.

The reference must be a **container image** whose filesystem holds the weights at the image root (`/model-source/<Model.name>.gguf`, or the `spec.files` path). A raw OCI artifact (ModelPack, Docker Model Artifact, an `oras push` with a model media type) mounts **empty** under containerd, which unpacks only tar image layers; the docs say container image, not artifact, for that reason.

Refs #1379

## Why

The evaluation recommends go as an opt-in source type, and the thing that gates it is the **node**, not the API server: ImageVolume is GA in 1.36, and a node whose runtime lacks support fails the pod rather than the admission. This PR lands the source with that understood, rather than implying a YAML-only feature.

## How

**Gate (`internal/controller/source.go`, `model_controller.go`)**
- `ociSourceSupported` refuses a control plane below 1.36 with an `UnsupportedCluster` condition whose message also names the node requirement (containerd >= 2.1.0 or CRI-O >= 1.31).
- The gate **fails open** when the version cannot be read (envtest, a fake client), and a comment states why: the serving node's runtime is the capability that actually matters and cannot be introspected by the operator.
- `ModelReconciler.ServerVersion` is read once at startup from the existing clientset in `cmd/main.go`.

**Source (`source.go`)**
- `isOCISource` (case-folded, like `s3://`), and `parseOCISource`, which requires `registry/repository` and rejects whitespace and bare image names with actionable errors.

**Storage (`model_storage.go`)**
- `buildOCIStorageConfig` returns one `ImageVolume` volume plus a read-only `/model-source` mount, mirroring `buildPVCStorageConfig`. `ociPrimaryFile` resolves the model path for single- and multi-file specs so `Status.Path` and the pod's path agree.
- A malformed reference fails loudly with a static `InvalidOCISource` init container.

**CRD (`api/v1alpha1/model_types.go`)** adds `oci` to the `spec.source` pattern; CRDs regenerated with `make manifests` + `make chart-crds`.

**Docs** (`docs/model-storage.md`, `docs/air-gapped-quickstart.md`) cover the source, the floor, digest pinning, registry credentials, the glob restriction, and the container-image and interior-layout contract (weights at the image root; a raw OCI artifact mounts empty under containerd).

Deliberate limits: multi-file needs explicit `spec.files` paths (a glob fails loudly because an OCI artifact provides no file listing); the controller does not probe registry reachability; `pvc://`, `s3://`, and `hf://` are unchanged and remain the paths below the floor.

## Checklist

- [x] Tests added/updated
- [x] `make lint` passes locally (`make lint-all` too, darwin + linux)
- [x] Documentation updated

`make generate` produced no DeepCopy diff (only a field's validation pattern changed); `make manifests` and `make chart-crds` are included.

## Tests

- `source_test.go`: `isOCISource`, `parseOCISource` (valid, empty, bare name, whitespace), `ociSourceSupported` (at floor, below floor, unreadable).
- `inferenceservice_storage_test.go`: the rendered storage config has an `ImageVolume` and **zero** init containers; the cache flag is ignored; a malformed reference fails loudly; a multi-file spec uses the primary path.
- `model_controller_test.go`: reconcile is Ready with the `/model-source/<name>.gguf` path; a 1.35 control plane fails with `UnsupportedCluster`; a malformed reference fails with `InvalidSource`.
- `model_cache_pvc_test.go` / `inferenceservice_storage_test.go` (pre-staged cache guard): the `oci://` row on `TestModelNeedsCachePVC`; an `oci://` InferenceService provisions **no** per-service cache PVC even with a cache key; an `oci://` InferenceService whose `spec.modelCache.claimName` targets a missing claim warns `ModelCacheClaimIgnored` and does **not** fail the reconcile with `ModelCachePVCNotFound`.

## Falsification run

Each mutation applied, the named specs confirmed failing, then reverted:

- `oci://` dispatch removed from `buildModelStorageConfig` -> all 4 render specs fail (`0 Passed | 4 Failed`).
- version floor gate disabled in `ociSourceSupported` -> the floor and `OCI Source Reconcile` specs fail (`4 Passed | 2 Failed`).
- `!isOCISource(model.Spec.Source)` removed from `modelNeedsCachePVC` -> the `oci://` predicate row fails (`modelNeedsCachePVC(source="oci://...") = true, want false`) and both new envtest specs fail (`2 Failed`: the per-service PVC exists; `ModelCachePVCNotFound` is emitted).
- the `oci://` case removed from `warnIgnoredModelCacheClaim` -> the claimName spec fails (`1 Failed`: no `ModelCacheClaimIgnored` / `pre-staged oci:// ImageVolume` event).

## AI assistance

Assisted-by: OpenCode (DeepSeek-V4.1-Flash-EXL3) (drafted the implementation, tests, docs, and this description; the author reviewed the diff, ran `make fmt`, `make vet`, `make lint`, `make lint-all`, and `make test`, and ran the two falsification mutations above).


Defilan (commented):
I want to hold this until the provisioning side matches the mount side.

An oci:// source still provisions a cache PVC. modelNeedsCachePVC excludes pvc:// because it is pre-staged and never mounted, but buildModelStorageConfig dispatches oci:// to buildOCIStorageConfig before the useCache branch, so the same is true of oci:// and the guard is missing. I wrote a probe against the real predicate: modelNeedsCachePVC for an oci:// model with a cache key returns true, and ensureModelCachePVC then creates the claim for a volume nothing mounts. It is worse when spec.modelCache.claimName is set, because ensureModelCachePVC enforces that the claim exists and fails the reconcile over a cache the source never uses.

The fix is one clause: !isOCISource(model.Spec.Source) in modelNeedsCachePVC, plus the same oci:// case in warnIgnoredModelCacheClaim so the claimName-is-ignored event fires. The existing TestModelNeedsCachePVC table is where the pvc:// case lives and is the right place for the oci:// row; it has no oci case today, which is why this shipped green.

Everything else held up. I re-ran both falsifications: removing the dispatch fails all four render specs, disabling the floor fails the two gate specs, so the tests bite. The parse and gate comments made this quick to trace.

Defilan:
Fixed in 628030a2b (`fix(model): skip the model cache PVC for oci:// sources`).

What changed, matching your read:

- `modelNeedsCachePVC` now excludes `isOCISource` beside `isPVCSource`, so the provisioning side skips a cache PVC for `oci://` the same way the mount side does through `buildOCIStorageConfig`.
- `warnIgnoredModelCacheClaim` gets the `oci://` case after the `pvc://` case, so `spec.modelCache.claimName` fires `ModelCacheClaimIgnored` instead of being dropped silently.
- Comments on both predicates and on `modelWantsCacheVolume` now name `oci://` alongside `pvc://`, so the next reader does not add a mount-side check for a scheme that never consults it.

Tests, as you pointed at:

- `TestModelNeedsCachePVC` gains the `oci://` row, with `Status.CacheKey` set so it is `false` only because of the new guard.
- Two envtest specs in the `ModelCacheClaimIgnored` block: an `oci://` ISVC with a cache key provisions no per-service cache PVC; an `oci://` ISVC whose `claimName` targets a missing claim warns `ModelCacheClaimIgnored` and does **not** fail the reconcile with `ModelCachePVCNotFound`. The second is the "worse when claimName is set" case.

Falsification, each mutation applied then reverted:

- `!isOCISource(model.Spec.Source)` removed from `modelNeedsCachePVC` -> the predicate row fails (`modelNeedsCachePVC(source="oci://...") = true, want false`) and both new envtest specs fail (`2 Failed`).
- the `oci://` case removed from `warnIgnoredModelCacheClaim` -> the claimName spec fails (`1 Failed`, no `ModelCacheClaimIgnored` / `pre-staged oci:// ImageVolume` event).
- Re-ran your two: the `oci://` dispatch removed from `buildModelStorageConfig` still fails all four render specs (`0 Passed | 4 Failed`), and disabling the floor still fails the gate specs (`4 Passed | 2 Failed`).

One implementation note: I first asserted the shared `llmkube-model-cache` PVC is absent, which passed in isolation but failed in the full suite because another spec leaves that shared claim behind. The spec now runs per-service mode and asserts its own uniquely named claim is absent, so it is order-independent.

`make test`, `make fmt`, `make vet`, `make lint` all pass locally.

PR #1886, opened 2026-09-22
fix(helm): cap composed Service names at 63 chars (#1885)

### LINKED ISSUE #1885: [BUG] Helm chart: a release name longer than 28 chars renders a >63-char metrics Service name that the cluster rejects
## Bug Description

The `llmkube` Helm chart can render a `Service.metadata.name` longer than 63 characters, which the cluster rejects.

`llmkube.fullname` truncates its base to 63 characters, but every template appends its role suffix *after* that truncation. So `metrics-service.yaml` renders `<fullname>-controller-manager-metrics-service`, which reaches 63 + 35 = 98 characters. `Service.metadata.name` is an RFC 1035 label (63 character hard cap), while Deployment, Secret, ServiceAccount and the RBAC kinds accept DNS-1123 subdomains (253), so the metrics Service is the first and only object to fail. The webhook Service (`<fullname>-webhook`, 8 chars) overflows once the fullname exceeds 55.

This is a core Kubernetes constraint, not a cluster or cloud specific one: any cluster rejects a Service name over 63 characters.

## Steps to Reproduce

No cluster needed. Use any release name longer than 28 characters:

```bash
helm template llmkube-example-release-region charts/llmkube --namespace llmkube-system | grep -A2 'kind: Service'
```

A 30 character release name is enough to trigger it.

## Expected Behavior

Every rendered object passes API validation, including the metrics Service.

## Actual Behavior

The metrics Service is rendered with a 65 character name:

```yaml
kind: Service
metadata:
  name: llmkube-example-release-region-controller-manager-metrics-service
```

Applying it fails with a `metadata.name: invalid value` / name length error on the Service. (One reporter's error text stated a 64 character limit, which is not the API server's 63 and may come from a cluster policy; either way the rendered 65 character name is over both, and the chart should cap at the RFC 1035 limit of 63.)

The webhook Service is only 38 characters for that release name, so the metrics Service is the only object that breaks in that case.

## Environment

**LLMKube Version:** chart `0.9.29` (`charts/llmkube/Chart.yaml`)

## Additional Context

Fix shape: budget the role suffix into the 63 character cap and shorten the base, so the suffix that identifies the object survives. The affected names are the metrics Service (`metrics-service.yaml`) and the webhook Service (`llmkube.webhook.serviceName` in `_helpers.tpl`, which also feeds the webhook `clientConfig.service.name` entries and the serving cert SANs, so those must stay matched).

`charts/foreman` carries the identical `fullname` + suffix pattern and the same latent overflow for a fullname over 55, but it is out of scope for this issue.

Defilan:
## What

Cap the resource names in the `llmkube` chart that carry a Kubernetes or X.509 length limit, and add a guard for the webhook cert's identity.

- Add a `llmkube.resourceName` helper that composes `<fullname><suffix>` and spends the 63 char cap on the base, never the suffix.
- Route the metrics Service (`metrics-service.yaml`), the webhook Service (`llmkube.webhook.serviceName`), the webhook CA common name (`llmkube.webhook.certs`), and the metrics endpoint in `NOTES.txt` through it.
- Add `charts/llmkube/tests/name_length_test.yaml` and a note in the chart README.
- Add `scripts/check-helm-webhook-cert.sh` plus `make check-helm-webhook-cert` and a `Webhook Cert Identity` CI job, which render with a long `fullnameOverride` and decode the self-signed serving and CA certs.

## Why

`llmkube.fullname` truncates the base of a resource name to 63 chars, and every template appends its role suffix *after* that truncation. A 30 char release name therefore rendered `<fullname>-controller-manager-metrics-service` as 65 chars, and the cluster rejected the Service: `Service.metadata.name` is an RFC 1035 label (63, hard limit), while the other kinds this chart renders accept DNS-1123 subdomains (253).

The metrics Service carries the longest suffix in the chart, so it is the first object to break. The webhook Service (`<fullname>-webhook`) overflows once the fullname passes 55. This is a core Kubernetes constraint, not a cluster or cloud specific rule, so it reproduces on any cluster given the same release name.

Fixes #1885

## How

`llmkube.resourceName` composes `<fullname><suffix>` and truncates the base to `63 - len(suffix)`. Two properties matter:

- **The suffix survives whole.** Truncating the composed string instead (the `charts/foreman` idiom, `| trunc 63 | trimSuffix "-"`) chops `-metrics-service` into `-metrics-servi` and would collide it with `-metrics-monitor` after truncation. This divergence from the sibling chart is deliberate.
- **Names that fit are returned unchanged**, because `trunc` is a no-op when the base is shorter than the budget. No existing release is renamed, so the cap only bites the overflow case.

`llmkube.webhook.serviceName` feeds the webhook Service, the two webhook `clientConfig.service.name` entries, and the serving cert SANs, so capping it once keeps the cert identity matched to the Service. The cert Secret and the webhook configuration names are deliberately left alone: renaming the Secret would empty the deployment's mount.

The webhook CA common name goes through the helper too. Hand-composed it reached 68 chars, over RFC 5280's `ub-common-name` of 64, and it is the one cert name that does not come from `llmkube.webhook.serviceName`.

**The guard.** The serving cert's CN and SANs exist only inside the DER of the base64 PEM in the Secret, so helm-unittest cannot assert them: the suite covers the cert-manager path's `spec.dnsNames`, and the self-signed path had to be checked by hand in review. `make check-helm-webhook-cert` renders with a long `fullnameOverride` and decodes both certs with openssl, pinning the serving CN, both SANs, and the CA CN. Its expected strings are hardcoded rather than recomputed so it cannot pass when the composition is wrong. It is recorded in `gateExemptCIChecks` for the same reason `check-helm-rbac` is (the gate image ships python3 but no PyYAML and no pip3), which `TestDefaultGateChecksCoverCI` enforces.

`charts/foreman` carries the identical `fullname` + suffix pattern and the same latent overflow for a fullname over 55. It is out of scope for this PR.

## Checklist

- [x] Tests added/updated

Also ran locally: `make test-chart` (helm lint + helm-unittest, 192 tests pass), `make check-helm-webhook-cert`, `make vet`, `make fmt`, `make lint-all` (darwin and linux). No `api/v1alpha1` change, so `make generate` and `make chart-crds` are not needed.

Assisted-by: DeepSeek-V4.1-Flash EXL3 3.5bpw, TP3 three-Spark ring running on LLMKube (generated the `llmkube.resourceName` helper, the chart edits, the helm-unittest suite and the webhook cert guard; I reviewed and ran make test, make test-chart, make check-helm-webhook-cert and make lint-all)


Defilan:
@brandonliuli Mind taking a look?

brandonliuli:
Nice — this is the same thing that bit us on an AKS cluster. Release name came out as `llmkube-afi-data-ai-dev-eastus` and the metrics Service landed at 65, so we worked around it with a `fullnameOverride`. Pulled your branch and rendered it with that same release name: 63 exactly, no override needed. So this fixes it at the source for us.

Also checked the self-signed path since the tests only assert the cert-manager one — decoded the generated cert and the CN and both SANs match the capped Service name, so that path is fine too. `helm unittest` on the new suite passes 4/4 here.

A couple of small things we noticed while poking at it. Neither is a blocker, totally up to you whether they're worth touching:

**The CA common name doesn't go through the new helper.** `_helpers.tpl:241` still composes it by hand:

```
$ca := genCA (printf "%s-webhook-ca" (include "llmkube.fullname" .)) ...
```

With your own test fixture that comes out at 68 chars, which is over RFC 5280's 64 for CN. Go's x509 doesn't enforce it and nothing really validates on CN anymore, so it's cosmetic in practice — it's just the one name composition left that doesn't use `resourceName`.

**Negative `trunc` flips direction.** If a suffix ever gets to 63+, `sub 63 (len $suffix)` goes negative and sprig's `trunc` starts taking from the end instead of the front. We tried it: `trunc -7` on `llmkube-something-long` gives `ng-long`. Can't happen today since the longest suffix is 35, but since `resourceName` is now the place names get built, it'd be a quiet wrong answer rather than an error if someone adds a long one later.

Separately — the "leaves a name that already fits unchanged" test is the one we'd call out as the important one. The genuinely risky part of a change like this isn't the overflow case, it's silently renaming resources on upgrade for everyone who's running fine today, and that test pins it shut.

One thing we couldn't settle: you mention the other kinds accept 253-char subdomains, which is right, but the Deployment name feeds pod names (+17 or so for the RS and pod hashes), so a 63-char fullname gets you ~80-char pod names. The API takes them, and we went back and forth on whether it actually matters in practice and couldn't convince ourselves either way. Probably nothing — flagging it in case you've already been through it.

Defilan:
Thanks, both landed.

I reproduced the CA CN one: decoded the rendered Secret with a long `fullnameOverride` and got a 68 char CN (`...-webhook-ca`), so you're right it was the one composition left out of `resourceName`. Cosmetic as you say (Go's x509 doesn't enforce the RFC 5280 CN cap and nothing validates CN), but cheap to fix and 63 also satisfies the 64 cap, so it goes through `resourceName` now. No-op for a fullname of 52 or less, and on upgrade the existing Secret is reused verbatim, so there's no cert churn.

Your `trunc` point is in as a stated precondition on the helper. I left it as a comment rather than a `fail` guard because the guard would be unreachable and untestable, and clamping would silently chop the suffix, which is a worse wrong answer than a loud one.

On the pod names: agreed it's probably nothing, and this PR doesn't move it either way. The Deployment name is unchanged (`fullname + "-controller-manager"`, 76 chars for the fixture above) and pod names add about 17 on top of that, so ~93. Both are DNS subdomains (253), the kubelet and the Linux runtimes are fine with them, and the caps that have actually bitten are Windows containers (63 char hostname) and some logging and CNI agents, neither of which this chart targets. Capping the Deployment name would rename it for any fullname over 44 and force a rollout, so I'd rather leave that boundary where it is than change it in a patch.

The part of your review I acted on most: you checked the self-signed SANs by hand because the unit suite only asserts the cert-manager path, and it can't do better. The CN and SANs only exist inside the DER of the base64 PEM in `data.tls.crt`, so helm-unittest can't reach them at all. So there's a new `make check-helm-webhook-cert` that renders with a long `fullnameOverride` and decodes both certs with openssl, asserting the serving CN, both SANs, and the CA CN. It fails if any of those three regresses, so your hand check is permanent now.

